PR #7, opened 2018-07-30
Fix typo in cryptodomex req check

HuzefaMandvi:
Was printing out the m2crypto version


jetwhiz:
Good catch!  Thanks @HuzefaMandvi

Issue #10, opened 2018-12-19
PEP8 compliance


lukehinds:
# Environment
 * OS / version:  N/A
 * Processor architecture: N/A
 * TPM Manufacturer: N/A
 * Keylime version: N/A

# Description 

Code linted and refactored to comply with the [PEP8 standard](https://www.python.org/dev/peps/pep-0008/). 

## Expected behaviour vs. actual behaviour 

PEP8 tests fail

Pass PEP8 linting

Once this issue is resolved, PEP8 can be automatically called as part of CI for #9 

## Steps to reproduce problem

1.  Run a PEP8 tool (for example `pylint`)

lukehinds:
I should be able to do this one file at a time as separate pull requests to make it easier to review

PR #14, opened 2018-12-19
PEP8 Compliance Changes for ca_impl_cfssl.py

lukehinds:
Changes made to ca_impl_cfssl.py to allow for PEP8 Compliance as per issue #10

Added two pylint ignores for false postives on constant naming case and
Redefining built-in 'open'.  E1300 for seeing ':' as a bad character
E0401 for false postive import errors & W0603, R0914 for global
statements and too many locals (acceptable as its an array).

### DIFF
--- a/keylime/ca_impl_cfssl.py
+++ b/keylime/ca_impl_cfssl.py
@@ -1,34 +1,36 @@
 '''
 DISTRIBUTION STATEMENT A. Approved for public release: distribution unlimited.
 
-This material is based upon work supported by the Assistant Secretary of Defense for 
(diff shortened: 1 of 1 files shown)


lukehinds:
I will hold off for now on pushing more PEP8 changes to see how you folks view the recommendation to go to PEP8.

The thinking is, that as more people start to contribute , having automatic PEP8 checks helps keep on top of keeping the code consistent for style. If any seem to strict, we can flag them out with `# pylint: disable-msg=<NUMBER>` in each file.

Issue #11, opened 2018-12-19
Demo scripts to support multiple linux dists


lukehinds:
# Environment
 * OS / version: Fedora / CentOS
 * Processor architecture: N/A
 * TPM Manufacturer: N/A
 * Keylime version: *

# Description 

Currently the demo scripts do not support other distributions outside of Debian, yet the project is listed as supporting CentOS / Fedora

## Expected behavior vs. actual behavior 

Running the `demo_setup.sh` will fail due to calls being made with only the `apt-get` package manager. 

A conditional / case statement should be added to first establish the distribution and then call the relevant package manager and package naming.

## Steps to reproduce problem
1.  Run `demo_setup.sh` on a non Debian based Linux Distribution.

Issue #12, opened 2018-12-19
create_whitelist.sh support for RHEL family


lukehinds:
# Environment
 * OS / version: Fedora / CentOS
 * Processor architecture: N/A
 * TPM Manufacturer:  N/A
 * Keylime version: *

# Description 

`create_whitelist.sh` seeks to [generate a sha1 file hash list](https://github.com/mit-ll/python-keylime/blob/master/keylime/create_whitelist.sh#L61) by un-archiving with the [initrd](https://github.com/mit-ll/python-keylime/blob/master/keylime/create_whitelist.sh#L55) naming convention. In RHEL family distributions, the naming convention is initramfs

The initramfs is also not a gzip file type. 

Since fedora 19, initramfs format has been changed from gz to xz. So to un-archive it requires `xz` tool:

    xz -dc /boot/initramfs*

## Expected behavior vs. actual behavior 

Will fail on any distribution that is not debian based.

Should work on other distributions, such as the listed Fedora and CentOS

This can be achieved using a simple conditional check on what the underlying distribution is running the script.

## Steps to reproduce problem
1. run `create_whitelist.sh` on a non debian dist.

PR #15, opened 2018-12-19
Add RHEL Family support to demo scripts

lukehinds:
This change adds support for all of the RHEL family for the `demo_setup.sh` and `create_whitelist.sh` scripts.

It closes issues #11 and #12


lukehinds:
Just realised I need to change from gzip for the rhel conditional in `create_whitelist.sh`

*Please don't merge this yet*

nabilschear:
Thanks!  @jetwhiz should be back working on this by early next week.  We'll make an effort to get this merged soon.  It's been crazy with the holidays.

lukehinds:
> Thanks! @jetwhiz should be back working on this by early next week. We'll make an effort to get this merged soon. It's been crazy with the holidays.

No big rush, its not holding me up.

jetwhiz:
It might be better just to use your version of the create_whitelist.sh, since the version I hacked up a while back (which is conflicting) isn't as thorough as the approach you've taken.

jetwhiz:
@lukehinds go ahead and clobber the existing create_whitelist.sh with your version to resolve the conflict, since I like your solution better.

PR #44, opened 2019-02-07
Allows tpm2-abrmd restart for test run

lukehinds:
This change introduces a slight change to the package manager
selection and in turn installs python-dbus

Further changes are made to `test_restful.py` whereby if the
system is running within systemd mode with an active dbus, restart
`tpm2-abrmd.service`. If not, do not try to restart the tpm2-abrmnd.

The logic being, a non systemd based service will be running a
tpm1.2 system.

The service restart is required as when using mssim, the resource
manager cannot reconnect.

This allows continued use of one set of test scripts for both
TPM versions.

### DIFF
--- a/test/run_tests.sh
+++ b/test/run_tests.sh
@@ -1,3 +1,6 @@
+#!/bin/bash
+##########################################################################################
+#
 # DISTRIBUTION STATEMENT A. Approved for public release: distribution unlimited.
 #
 # This material is based upon work supported by the Assistant Secretary of Defense for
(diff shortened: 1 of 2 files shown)


jetwhiz:
This looks really good, @lukehinds!  Just a few small changes so we can run this from the command line and on non-dnf systems and then I think it is ready to merge.

lukehinds:
Thanks @jetwhiz for reviewing , well spotted about the hacked off shebang. I have seen this before and I think it might be a wayward editor extension I have. Change incoming.

lukehinds:
@jetwhiz , we should be good now, give me a quick review and I will merge this and we have CI coverage.

p.s.. I was thinking last night, I will refine this over time. It might be more tidy to move the tpm2_wrapper into run_tests.sh. Let's see though, if upstream see the rmd disconnect as an issue (btw, I put a comment in above the dbus python code).

jetwhiz:
Hi @lukehinds -- this looks better, I think we just need to put in spaces around the "=" sign for the if-else statements (starting at line 142) and then this is ready to merge.  

As-is, the run_tests.sh script doesn't run properly on Ubuntu (and non-dnf systems) since `if [ $PACKAGE_MGR="dnf" ]; then` is always true (the equals is not evaluated when there are no spaces around the `=`).

lukehinds:
Side note. Will look at running a CI job in Ubuntu (which will capture what you found above 👍)

PR #57, opened 2019-02-12
Don't pass -I option to bootstrap

tstruk:
Recent version of tss doesn't need -I param to
be passed to bootstrap.


tstruk:
@lukehinds it was done in this PR https://github.com/tpm2-software/tpm2-tss/pull/1254

PR #65, opened 2019-03-04
Script to replicate CI build

lukehinds:
Can be useful a local test environment for unstaged code testing


lukehinds:
Some more details

Running this pulls down the same images used in CI. It then mounts a local repo as a volume and i turn run the test scripts (based on choice of TPM1.2 or 2.0).

The idea is a developer can make local changes and quickly test them in a local replication of the CI build process we use on travis. They can then capture any failures before they make a PR.

Issue #81, opened 2019-03-13
pyyaml dependency handling


nabilschear:
# Environment
 * OS / version: mac osx
 * Processor architecture: 
 * TPM Manufacturer: 
 * Keylime version: 

# Description 

pyyaml dependency wasn't handled the same way in setup.py and installer.sh as our other dependencies.  So, when i work with keylime on my mac with the emulator the yaml dependency wasn't pulled in by setup.py

## Expected behavior vs. actual behavior 
 expected: keylime runs
actual: import yaml not found error

## Steps to reproduce problem
1. create virtual environment
2. activate virtual environment
3. python setup.py install
4. attempt to run any of the keylime applications

# Relevant logs

Attach any relevant log files that can help to debug your issue.

nabilschear:
this is also related to #80

lukehinds:
I think we are good to close this one too @nabilschear ?

nabilschear:
agreed

Issue #80, opened 2019-03-13
update pyyaml to 5.1


nabilschear:
# Environment
 * OS / version: 
 * Processor architecture: 
 * TPM Manufacturer: 
 * Keylime version: 

# Description 

pyyaml had a nasty vulnerability reported last year that causes remote code execution https://nvd.nist.gov/vuln/detail/CVE-2017-18342.  We use yaml to parse the output from tpm2-tools.    this has finally been updated in pyyaml version 5.1, which was just released.  Once this version propagates to distro packages, we should update keylime to require it.

## Steps to reproduce problem
1. 

# Relevant logs

Attach any relevant log files that can help to debug your issue.

PR #82, opened 2019-03-13
handle pyyaml just like other dependencies

nabilschear:
first step towards fixing pyyaml dependency stuff.  we'll stick with the distro versions for now.

#81  and eventually will relate to #80

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,12 +86,10 @@ The following python packages are required:
 * pyzmq>=14.4
 * setuptools>=0.7
 * python-dev
+* pyyaml
 
-The latter of these are usually available as distro packages.
-
-On Ubuntu: `apt-get install -y python-dev python-setuptools python-tornado python-m2crypto python-zmq`
-
-On CentOS: `yum install -y python-devel python-setuptools python-tornado python-m2crypto python-zmq`
+The latter of these are usually available as distro packages. See [installer.sh](https://github.com/keylime/python-keylime/blob/master/installer.sh) for more information if you want to install them this way.
+You can also let keylime's `setup.py` install them via PyPI.
 
 #### TPM utility prerequisites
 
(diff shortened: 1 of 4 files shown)


jetwhiz:
Yeah it looks like they're keeping safe_load around as an alias in 5.1, so it should be safe to use it here.  

It might be good to both lower the min version to 3.12 or 3.11 (ubuntu xenial) and change `yaml.load` to `yaml.safe_load`.  In the future we can plan to integrate OWASP or similar into the CI process so we don't need to worry about checking against min versions anymore.

nabilschear:
So, the installer will already warn on an old version and ask the user for an override.  So, they can proceed.  Do you think we should still have the version check and warning? @lukehinds

lukehinds:
> So, the installer will already warn on an old version and ask the user for an override. So, they can proceed. Do you think we should still have the version check and warning? @lukehinds

Hmm, I personally would not if we use `safe_load`. I say this on basis that a lot of folks might see the warning and out of concern start then trying to use pip or another means of getting a later version of `pyyaml` outside of ubuntu / debians main package repositories and this in turn might bite them later when upgrading etc. I think moving to OS packages over pypi is a good move though and we should definitely keep that code!

lukehinds:
I also wanted to add its highly likely that ubuntu have backported the security fixes into the version in the package repos. 

Red Hat do something similar, they maintain a long term stable version and back port in security fixes. So a version number that is insecure upstream, may well likely be patched for the same release number downstream.

lukehinds:
Here we go: https://launchpad.net/ubuntu/+source/pyyaml/3.11-1ubuntu0.1

nabilschear:
ok, i've left 3.11 in setup.py to make sure that it will just work.  i've removed the check and added the safe_load call.

PR #86, opened 2019-03-19
Add TPM Emulator warning text

lukehinds:


nabilschear:
i think i've got more warnings about the emulator strewn about enough that people would notice them.

PR #88, opened 2019-03-20
Update demo_setup.sh and create_whitelist.sh for Debian systems

jetwhiz:
Noticed that there are some issues grabbing the initrd files in modern Debian-based systems using the traditional `gzip`-`cpio` method.  Newer Debian-based systems now provide `unmkinitramfs` to unpack initramfs images, which we can leverage.  In the worst case, fall back to using `gzip`-`cpio`. 

This method also allows `create_whitelist.sh` to be more flexible with parsing init ram images (no need to pass in "initrd" or "initrams" on the command line.

### DIFF
--- a/demo/demo_setup.sh
+++ b/demo/demo_setup.sh
@@ -172,7 +172,7 @@ esac
 
 echo 
 echo "=================================================================================="
-echo $'\t\t\t\tUpdating $NAME packages'
+echo $'\t\t\t\tUpdating packages'
 echo "=================================================================================="
 ${PACKAGE_MANAGER} update -y 
 
@@ -284,17 +284,5 @@ if [[ "$IMA_ENABLE" -eq "1" ]] ; then
     echo $'\t\t\t\tGenerating IMA whitelist'
     echo "=================================================================================="
     cd $KEYLIME_DIR/keylime/
-    case "$ID" in
-    debian | ubuntu)
-        initdisk="initrd"
-    ;;
-
-    redhat | centos | fedora)
-        initdisk="initramfs"
-    ;;
-    *)
-        echo "${ID} is not currently supported."
-        exit 1
-    esac
-    ./create_whitelist.sh whitelist.sh ${initdisk}
+    ./create_whitelist.sh whitelist.txt
 fi
--- a/keylime/create_whitelist.sh
+++ b/keylime/create_whitelist.sh
@@ -21,6 +21,29 @@
 #
 ##########################################################################################
 
+# Configure the installer here
+INITRAMFS_TOOLS_GIT=https://salsa.debian.org/kernel-team/initramfs-tools.git
+INITRAMFS_TOOLS_VER="master"
+
+
+# Grabs Debian's initramfs_tools from Git repo if no other options exist
+if [[ ! `command -v unmkinitramfs` && ! -x "/usr/lib/dracut/skipcpio" ]] ; then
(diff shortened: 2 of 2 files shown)


jetwhiz:
Good catch!  It seems the raw `cpio` method is fraught with issues in general, since we would have to: 
1. see if there are any uncompressed, prepended archives and skip them (it sounds like `-id 2` is not always sufficient)
2. determine if the archives contained are compressed, and see which algorithm is used for compression
3. the cpio could contain many archives in addition to the main archive, incl. kernel drivers, etc.

Do you think it would be better just to pull in `unmkinitramfs` if it does not exist as a last-ditch effort (if the dracut `skipcpio` option is not there): https://salsa.debian.org/kernel-team/initramfs-tools

Would that introduce an issue for older, pre-skipcpio versions of Fedora/RHEL/CentOS that we might still want to support?  I imagine this would affect RHEL/CentOS 6 and prior?

lukehinds:
I would keep the OS detection, we can then at least nail down to a specific distribution and its release. This gives us some future room to work with more OSes as well (arch, opensuse etc).

Every systemd based dist has a `/etc/os-release` which is very easy to work with:

check this out:

```
$ source /etc/os-release

$ echo $ID
fedora

$ echo $VERSION
29 (Server Edition)

$ echo $VERSION_ID
29
```

```
$ source /etc/os-release 

$ echo $ID
ubuntu

$ echo $VERSION_ID
18.04

$ echo $VERSION
18.04.2 LTS (Bionic Beaver)
```

You can see the setup here:

https://github.com/keylime/python-keylime/blob/b05d5d7cfbabe163d4bca936cd2f6efdfa929914/demo/demo_setup.sh#L93-L99

jetwhiz:
That's a good point ... that might be a good option instead of command detection.  

I was looking at the different distributions and it seems they all package initramfs the same way though, it's just that some compress (with various algorithms) and some don't compress.  Basically the workflow is `compress? -> cpio`. We might not need to have a bunch of different methods to extract initramfs at all if we just use `unmkinitramfs`: https://salsa.debian.org/kernel-team/initramfs-tools/blob/master/unmkinitramfs

This shell script should work on all platforms, since all they're doing is decompressing using the detected algorithm and then skipping any of the extra prepended archives.  I've tested this on Fedora 29 and it works perfectly. At the very least, that means having a universal fallback where we download and run `unmkinitramfs` should work for suse, Fedora/RHEL/CentOS, and older Ubuntus/Debians.  

In this case, we either replace `gzip -dc $i | cpio -id 2> /dev/null` with the `unmkinitramfs` shell script download+execute, or just always use `unmkinitramfs` and don't bother with dracut's `skipcpio` at all.

jetwhiz:
Do you have preferences on this, @lukehinds ?  I think the global fallback option of (download+run) unmkinitramfs.sh should be a good solution to cover all remaining bases (pre-skipcpio Fedora/RHEL, pre-unmkinitramfs Debian/Ubuntu, suse, etc.)

lukehinds:
hey @jetwhiz if `unmkinitramfs` captures strays and works on the supported releases for debian / rhel and family, that's acceptable to me.

jetwhiz:
@lukehinds Tested this update out on Ubuntu 18 LTS and 16 LTS, Fedora 29 and CentOS 7 and everything works well.  Let me know what you think!

PR #91, opened 2019-03-21
Replace and remove `raise Exception` calls

### LINKED ISSUE #87: use of `raise Exception`
Within `tenant.py` we use `raise Exception` and I wonder if this is needed in all cases.

For example we may see the following:

```
1552982503.81 - keylime.cloudverifier - WARNING - Node of uuid D432FBB3-D2F1-4A97-9EF7-75BD81C00000 already exists
  File "/usr/lib/python2.7/site-packages/keylime-1.2-py2.7.egg/keylime/tenant.py", line 744, in main
    mytenant.do_cv()
  File "/usr/lib/python2.7/site-packages/keylime-1.2-py2.7.egg/keylime/tenant.py", line 481, in do_cv
    raise Exception("Node %s already existed at CV.  Please use delete or update."%self.node_uuid)
Exception: Node D432FBB3-D2F1-4A97-9EF7-75BD81C00000 already existed at CV.  Please use delete or update.
```

The use of `raise Exception` here makes it look like the code is broken as we get a traceback as CLI output, but in fact the code has captured the condition correctly and could instead report with the logger (most likely `logger.info`)

If we have agreement here on using the logger instead of `raise Exception` . I am happy to work on a PR that we can then review.

lukehinds:
Please see GH issue #87 for details

Fixes #87

### DIFF
--- a/demo/node_monitor/tenant_node_monitor.py
+++ b/demo/node_monitor/tenant_node_monitor.py
@@ -152,7 +152,7 @@ class InstancesHandler(BaseHandler):
         except Exception as e:
             common.echo_json_response(self, 400, "Exception error: %s"%e)
             logger.warning("POST returning 400 response. Exception error: %s"%e)
-            logger.warning(traceback.format_exc())
+            logger.exception(e)
                
     def put(self):
         """This method handles the PUT requests to add instances to the Node Monitor. 
--- a/keylime.conf
+++ b/keylime.conf
@@ -437,8 +437,8 @@ keys = consoleHandler
 keys = formatter
 
 [formatter_formatter]
-format = %(created)s - %(name)s - %(levelname)s - %(message)s
-datefmt = 
+format = %(asctime)s.%(msecs)03d - %(name)s - %(levelname)s - %(message)s
+datefmt = %Y-%m-%d %H:%M:%S
 
 [logger_root]
 level = INFO
@@ -446,11 +446,11 @@ handlers = consoleHandler
 
 [handler_consoleHandler]
 class = StreamHandler
-level = DEBUG
+level = INFO
 formatter = formatter
 args = (sys.stdout,)
 
 [logger_keylime]
-level = DEBUG
+level = INFO
 qualname = keylime
 handlers = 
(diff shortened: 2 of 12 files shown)


nabilschear:
so the exceptions were nice because they would stop the flow of execution and head out to an error handler.  If we just let it keep going, then more errors (and less easy to understand) are going to happen.  What if instead, we drop the duplicate logger.error lines put raise Exception back in and instead just remove this line of code or change it to logger.debug? https://github.com/keylime/python-keylime/blob/ba0a323ceaffab9c82d5a94ccbfb698d49e136e6/keylime/tenant.py#L768  i think that will clean it up a bit.  in that spirit, we should also change this https://github.com/keylime/python-keylime/blob/ba0a323ceaffab9c82d5a94ccbfb698d49e136e6/keylime/tenant.py#L765-L766 and this https://github.com/keylime/python-keylime/blob/ba0a323ceaffab9c82d5a94ccbfb698d49e136e6/keylime/tenant.py#L702-L703 to use exceptions rather than directly calling sys.exit.  Not sure if this should also be converted to a raise Exception? https://github.com/keylime/python-keylime/blob/ba0a323ceaffab9c82d5a94ccbfb698d49e136e6/keylime/tenant.py#L721  

also the tenant webapp deals with things this way: https://github.com/keylime/python-keylime/blob/ba0a323ceaffab9c82d5a94ccbfb698d49e136e6/keylime/tenant_webapp.py#L527-L531  i'm not sure if that will print a stack trace or not.

nabilschear:
ha it already is logger.debug.  maybe we just need to change the default loglevel in keylime.conf to not be debug =)

lukehinds:
> ha it already is logger.debug. maybe we just need to change the default loglevel in keylime.conf to not be debug =)

That's not a bad idea. We could also set a more human friendly format.

```
[formatter_formatter]
format = %(asctime)s.%(msecs)03d - %(name)s - %(levelname)s - %(message)s
datefmt = %Y-%m-%d %H:%M:%S
```

My main motivation was to not make correctly captured events look like a bug (when it spews out a trackback to the various code objects). Report everything gracefully.

nabilschear:
@lukehinds i went ahead and tried out the special exception solution, updated the log formatting/level, and cleaned up other instances of stack traces being printed.  what do you think?

lukehinds:
> @lukehinds i went ahead and tried out the special exception solution, updated the log formatting/level, and cleaned up other instances of stack traces being printed. what do you think?

Thanks, will take a look (not out machine just now).

PR #94, opened 2019-04-01
Rename of node / agent

lukehinds:
As per mailing list discussion:

https://keylime.groups.io/g/main/topic/node_agent/30833147

### DIFF
--- a/keylime.conf
+++ b/keylime.conf
@@ -437,8 +437,8 @@ keys = consoleHandler
 keys = formatter
 
 [formatter_formatter]
-format = %(created)s - %(name)s - %(levelname)s - %(message)s
-datefmt = 
+format = %(asctime)s.%(msecs)03d - %(name)s - %(levelname)s - %(message)s
+datefmt = %Y-%m-%d %H:%M:%S
 
 [logger_root]
 level = INFO
@@ -446,11 +446,11 @@ handlers = consoleHandler
 
 [handler_consoleHandler]
 class = StreamHandler
-level = DEBUG
+level = INFO
 formatter = formatter
 args = (sys.stdout,)
 
 [logger_keylime]
-level = DEBUG
+level = INFO
 qualname = keylime
 handlers = 
--- a/keylime/tenant_webapp.py
+++ b/keylime/tenant_webapp.py
@@ -290,8 +290,7 @@ class InstancesHandler(BaseHandler):
                                         "http://%s:%s/instances/%s"%(tenant_templ.cloudverifier_ip,tenant_templ.cloudverifier_port,instance_id),context=tenant_templ.context)
         except Exception as e:
             logger.error("Status command response: %s:%s Unexpected response from Cloud Verifier."%(tenant_templ.cloudverifier_ip,tenant_templ.cloudverifier_port))
-            logger.error(traceback.print_exc())
-            logger.error("Error: %s "%str(e))
+            logger.exception(e)
             common.echo_json_response(self, 500, "Unexpected response from Cloud Verifier", str(e))
             return
         
@@ -356,8 +355,7 @@ class InstancesHandler(BaseHandler):
                                         "http://%s:%s/instances/"%(tenant_templ.registrar_ip,tenant_templ.registrar_port),context=tenant_templ.context)
         except Exception as e:
             logger.error("Status command response: %s:%s Unexpected response from Registrar."%(tenant_templ.registrar_ip,tenant_templ.registrar_port))
-            logger.error(traceback.print_exc())
-            logger.error("Error: %s "%str(e))
+            logger.exception(e)
             common.echo_json_response(self, 500, "Unexpected response from Registrar", str(e))
             return
         
--- a/demo/node_monitor/tenant_node_monitor.py
+++ b/demo/node_monitor/tenant_node_monitor.py
@@ -152,7 +152,7 @@ class InstancesHandler(BaseHandler):
         except Exception as e:
             common.echo_json_response(self, 400, "Exception error: %s"%e)
             logger.warning("POST returning 400 response. Exception error: %s"%e)
-            logger.warning(traceback.format_exc())
+            logger.exception(e)
                
     def put(self):
         """This method handles the PUT requests to add instances to the Node Monitor. 
--- a/keylime/cloud_verifier_tornado.py
+++ b/keylime/cloud_verifier_tornado.py
@@ -221,7 +221,7 @@ class InstancesHandler(BaseHandler):
         except Exception as e:
             common.echo_json_response(self, 400, "Exception error: %s"%e)
             logger.warning("POST returning 400 response. Exception error: %s"%e)
-            logger.warning(traceback.format_exc())
+            logger.exception(e)
         
         self.finish()
         
@@ -272,7 +272,7 @@ class InstancesHandler(BaseHandler):
         except Exception as e:
             common.echo_json_response(self, 400, "Exception error: %s"%e)
             logger.warning("PUT returning 400 response. Exception error: %s"%e)
-            logger.warning(traceback.format_exc())
+            logger.exception(e)
             
         
         self.finish()
@@ -332,8 +332,7 @@ class InstancesHandler(BaseHandler):
 #                     self.get_q_log_file.flush()  
                  
             except Exception as e:
-                logger.debug(traceback.print_exc())
-                logger.critical("Unexpected exception occurred in worker_get_quote.  Error: %s"%e )            
+                logger.exception(e)            
 
 
 
@@ -363,9 +362,6 @@ class InstancesHandler(BaseHandler):
  
     def process_instance(self, instance, new_operational_state):
         try:
-            if instance is None:
-                #import traceback
-                traceback.print_stack()
             main_instance_operational_state = instance['operational_state']
             stored_instance = self.db.get_instance(instance['instance_id'])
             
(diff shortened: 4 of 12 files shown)


lukehinds:
@jetwhiz  / @nabilschear - might look like a very heavy review, but github makes it very easy to speed read through and should only take 2-3 mins

nabilschear:
what should we do with the binary content (e.g., presentations, papers, etc.) in the repo?

nabilschear:
one other potential issue: this will break if you have an old keylime.conf in /etc because of the rename of the cloud_node section

lukehinds:
hmm, not to sure on the binary objects..I could change the .doc stuff.

> one other potential issue: this will break if you have an old keylime.conf in /etc because of the rename of the cloud_node section

What a code check be acceptable for this (I could put something into config parser)? the other option is we release this as a non backwards compatible release?

lukehinds:
I have been playing around with this and I don't like adding checks in, its starting to get messy.

We could do something like this:

```
if config.has_section('cloud_node'):
    cert_path=config.get('cloud_node','revocation_cert')
else:
    cert_path=config.get('cloud_agent','revocation_cert')
```

But it means collecting some technical debt, as every time someone needs to use that config call, they need to make sure they check for both values.

The other option is to run something to update the `cloud_node` section to `cloud_agent`, but configparser deals with `_sections` as a dict, so the code becomes over burdened - also there are many entry points (`keylime_verifier`, `keylime_registrar` etc).

I honestly think we are best making the change and highlighting it in the release notes. I don't believe we (keylime) are used widely just yet, so we should not cause breakage for any users?

nabilschear:
it's not like there's a lot of users yet.  so no backwards compatibility is not a big deal.

nabilschear:
i was looking at the rest api doc.  it seems like we should also get rid of the word *instance* from our keylime vocabulary.  and replace it with agent too.

nabilschear:
@lukehinds and you thought this PR was big. now it's even bigger!  i've banished instance from keylime too.  One question for @jetwhiz and @lukehinds , i looked at https://github.com/keylime/python-keylime/blob/master/doc/keylime-elevator-slides.pptx?raw=true and i wasn't sure if we should take the word node out of there too?

lukehinds:
@nabilschear , the more, the merrier as far as I am concerned.

for the elevator slides, I am not of any strong opinion, you guys can do as you please there. Once this is decided, we should be good to merge then?

nabilschear:
i went and actually looked at all the instance changes.  Hopefully tests pass now.  i think i caught all the perl -pie errors.  One more person to look over the whole PR would be good.  @jetwhiz

nabilschear:
woohoo it passed the build!

keylime/static/css/webapp.css
```diff
@@ -37,14 +37,14 @@ html, body {margin:0px;padding:0px;height:100%;width:100%;font-family:verdana,sa
...
-#node_container {list-style-type: none;}
+/* agent boxes and containers */
+#agent_container {list-style-type: none;}
 #terminal-frame {margin-top:25px;margin-left:10px;background-color:#000088;border:2px solid #dddddd;width:50%;min-width:500px;max-width:750px;}
 #terminal-header {padding:3px;background-color:#dddddd;font-weight:bold;cursor:pointer;}
 #terminal {width:100%;min-width:500px;max-width:750px;height:400px;background-color:#000088;color:#ffff00;overflow-x:auto;font-family:monospace;font-size:medium;}
 
-/* Each node instance for 'instances' table */
-#instance_body {padding:10px;}
+/* Each agent agent for 'agents' table */
```
jetwhiz: This now reads: `Each agent agent for 'agents' table`

keylime/static/js/webapp.js
```diff
@@ -197,10 +197,10 @@ function dragoverCallback(event) {
 function fileUploadCallback(event) {
     let target = event.target;
     
-    // Bubble up to find the file_drop node 
+    // Bubble up to find the file_drop agent 
     if (target.classList.contains) {
         while (!target.classList.contains("file_drop")) {
-            target = target.parentNode;
+            target = target.parentAgent;
```
jetwhiz: This needs to stay 'parentNode' ... it is a JavaScript property

keylime/static/js/webapp.js
```diff
@@ -351,71 +351,71 @@ function updateNodesInfo() {
...
+        // Find removed agents (in old, not in new)
+        let removedAgents = arrayDiff(existingAgentIds, agentIds);
+        //console.log(removedAgents);
         
-        // Add node
-        for (let i = 0; i < newNodes.length; i++) {
-            let ele = document.getElementById('node_template').firstElementChild.cloneNode(true);
+        // Add agent
+        for (let i = 0; i < newAgents.length; i++) {
+            let ele = document.getElementById('agent_template').firstElementChild.cloneagent(true);
```
jetwhiz: This needs to stay 'cloneNode', it's a JavaScript function.

keylime/static/js/webapp.js
```diff
@@ -351,71 +351,71 @@ function updateNodesInfo() {
...
         
-        // Remove node
-        for (let i = 0; i < removedNodes.length; i++) {
-            let ele = document.getElementById(removedNodes[i]);
+        // Remove agent
+        for (let i = 0; i < removedAgents.length; i++) {
+            let ele = document.getElementById(removedAgents[i]);
             //console.log(ele);
-            ele.parentNode.removeChild(ele);
+            ele.parentAgent.removeChild(ele);
```
jetwhiz: This needs to stay 'parentNode' (JavaScript)

keylime/tenant_webapp.py
```diff
@@ -106,25 +106,25 @@ def get(self):
...
-                                <h3>Add Node</h3>
+                                <h3>Add agent</h3>
                                 <h4 id='uuid_str'></h4>
                             </center>
-                            <form id='add_node' name='add_node' onsubmit='submitAddNodeForm(this); return false;'>
+                            <form id='add_agent' name='add_agent' onsubmit='submitAddAgentForm(this); return false;'>
                                 <div class="form_block">
-                                    <label for='node_ip'>Node IP: </label>
-                                    <input type='text' id='node_ip' name='node_ip' value='127.0.0.1' required onfocus='this.select()'>
+                                    <label for='agent_ip'>agent IP: </label>
```
jetwhiz: Can we keep the case here, `Agent IP:`?  It is a GUI field visible to users.

keylime/tenant_webapp.py
```diff
@@ -335,135 +335,135 @@ def get(self):
...
-        instance_id = rest_params["nodes"]
-        
-        # let Tenant do dirty work of deleting node 
+
+        agent_id = rest_params["agents"]
+
+        # let Tenant do dirty work of deleting agent
         mytenant = tenant.Tenant()
-        mytenant.node_uuid = instance_id
+        mytenant.agent_uuid = agent_id
```
jetwhiz: Does this still work?  I thought this variable was changed to `AGENT_UUID` in tenant.py (uppercase now)?

keylime/tenant_webapp.py
```diff
@@ -515,11 +515,11 @@ def post(self):
...
             'ima_exclude': ima_exclude,
         }
-        
-        # let Tenant do dirty work of adding node 
+
+        # let Tenant do dirty work of adding agent
         try:
             mytenant = tenant.Tenant()
-            mytenant.node_uuid = instance_id
+            mytenant.agent_uuid = agent_id
```
jetwhiz: Changed to `AGENT_UUID` in tenant.py (uppercase)?

keylime/tenant_webapp.py
```diff
@@ -529,105 +529,105 @@ def post(self):
...
-        instance_id = rest_params["nodes"]
-        
-        # let Tenant do dirty work of reactivating node 
+
+        agent_id = rest_params["agents"]
+
+        # let Tenant do dirty work of reactivating agent
         mytenant = tenant.Tenant()
-        mytenant.node_uuid = instance_id
+        mytenant.agent_uuid = agent_id
```
jetwhiz: Changed to `AGENT_UUID` in tenant.py (uppercase)?

scripts/ima/ascii_runtime_measurements_ima
```diff
@@ -206,7 +206,7 @@
 10 d09349a60461558c35a7d54a5e5fc46e78ddd621 ima 0633935bea94ffa21495127c859c4744338dd796 /lib/systemd/system/sys-kernel-config.mount
 10 010ce5b9e6e47fbba9570bb94c4d02b4cfee1534 ima c5e2ba84c06e1340aa2cee4917f034fbc52432cb /lib/systemd/system/systemd-hwdb-update.service
 10 02889950415fd92d9efde3b73be6786155b5ca2d ima ab99878d8d7a7831665f9afda53c26a06d5d308a /lib/systemd/system/systemd-tmpfiles-setup-dev.service
-10 e348f50c7298d33e2e99ea882e4666e70f7f45dc ima 4ede83c2caf9602514a436189756db673a570b14 /lib/systemd/system/kmod-static-nodes.service
+10 e348f50c7298d33e2e99ea882e4666e70f7f45dc ima 4ede83c2caf9602514a436189756db673a570b14 /lib/systemd/system/kmod-static-agents.service
```
jetwhiz: This shouldn't be changed (this will change the SHA1 hash written to PCR10).

keylime.conf
```diff
@@ -51,39 +51,39 @@ dec_payload_file = decrypted_payload
...
 extract_payload_zip = True
 
-# set the node's uuid to the given value.  
+# set the agent's uuid to the given value.
 # set to 'openstack', it will try to get the uuid from the metadata service
 # if you set this to 'generate', keylime will create a random uuid
 # if you set this to 'hash_ek', keylime will set the UUID to the result
 # of SHA256(public EK in PEM format)
-node_uuid = D432FBB3-D2F1-4A97-9EF7-75BD81C00000
+AGENT_UUID = D432FBB3-D2F1-4A97-9EF7-75BD81C00000
```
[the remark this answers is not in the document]
lukehinds: done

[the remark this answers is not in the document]
lukehinds: done

lukehinds:
sorry guys, messed up here fixing a merge request, working on getting this open again.

PR #118, opened 2019-04-18
Add support for legacy tpm2-tools v3

jetwhiz:
Currently Keylime only supports master branch of tpm2-tools (v4+).  This PR adds support for the legacy (v3.X) version of the tpm2-tools (currently used on RHEL, etc.)

Tools have all been upstreamed to 3.X.  Need to update the CI infrastructure to support both current (tpm2-tools v4 fork) and upstream 3.X (https://github.com/tpm2-software/tpm2-tools/tree/3.X) TPM tool code bases. 

PR couples with issue to update CI for v4 tpm2-tools upstream: https://github.com/keylime/keylime/issues/123

### DIFF
--- a/keylime/tpm2.py
+++ b/keylime/tpm2.py
@@ -20,6 +20,7 @@ violate any copyrights that exist in this work.
 import base64
 import binascii
 import ConfigParser
+import distutils.spawn
 import hashlib
 import json
 import os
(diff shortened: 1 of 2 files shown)


jetwhiz:
@lukehinds This should be ready for review now, since all needed tpm2-tool changes are upstreamed to 3.X.  

Can you add a CI test for tpm2-tools v3.X?  It should be the same as the tpm2-tools one except now we need: 
 * tpm2-tss version `2.0.x`
 * tpm2-tools `3.X` from https://github.com/tpm2-software/tpm2-tools/ (not our fork)
 * tpm2-abrmd version `2.0.3`

keylime/tpm2.py
```diff
@@ -404,6 +448,30 @@ def __get_pub_ek(self): # assumes that owner_pw is correct at this point
         
         self._set_tpm_metadata('ek', ek)
 
+    def __get_pub_ak(self):
```
lukehinds: Could we add a docstring here , just a little on what the function achieves, args processed and expected return:

https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

jetwhiz:
> Just a couple of notes (will finalise review after testing locally).
> 
> Could we add a small logger entry just to emit one line to cloudagent.log to show keylime has found either tools 3.x or 4.x - this would be for debug purposes. Let's say a user logs and issue, we can then see in the logs the version selected and look at what versions they have installed (and perhaps see they used a non uniform path for example).

That's a good idea -- I've incorporated your suggested changes into the latest version.

PR #119, opened 2019-04-19
If PR only contains .md files or in docs, abort CI

lukehinds:
This change adds a conditional to check if a pull request
contains only markdown files or is placed into `docs/`.

If the conditional is true, the Travis CI build will be aborted.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,18 @@ language: ruby
 
 sudo: required
 
+# If the change is a markdown file, don't run CI build. Or at a later point
+# we can set a specific job in here (such as a markdown lint)
+before_install:
+- |
+    if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
+      TRAVIS_COMMIT_RANGE="FETCH_HEAD...$TRAVIS_BRANCH"
+    fi
+    git diff --name-only $TRAVIS_COMMIT_RANGE | grep -qvE '(\.md$)|(^doc)/' || {
+      echo "Only documentation files were updated, stopping build process."
+      exit
+    }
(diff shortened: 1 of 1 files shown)


lukehinds:
I keep forgetting to add `[skip-ci]` and travis requires you have it in the small real estate area of the commit subject, lets automate it instead.

.travis.yml
```diff
@@ -3,6 +3,18 @@ language: ruby
...
 sudo: required
 
+# If the change is a markdown file, don't run CI build. Or at a later point
+# we can set a specific job in here (such as a markdown lint)
+before_install:
+- |
+    if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
+      TRAVIS_COMMIT_RANGE="FETCH_HEAD...$TRAVIS_BRANCH"
+    fi
+    git diff --name-only $TRAVIS_COMMIT_RANGE | grep -qvE '(\.md$)|(^docs)/' || {
```
frozencemetery: The subdirectory is "doc", not "docs", isn't it?

.travis.yml
```diff
@@ -3,6 +3,18 @@ language: ruby
...
+      echo "Only markdown files were updated, stopping build process."
```
frozencemetery: Technically, I think this should be "Only documentation files", right?

.travis.yml
lukehinds: well spotted. I will do a later change to rename this to /docs which is more common then just doc.

.travis.yml
lukehinds: Agree!

PR #120, opened 2019-04-19
Implement stale bot in keylime/keylime

lukehinds:
This (free) bot marks issues / PRs as stale when there is no activity
for 60 days. It will initially make a comment giving 7 days notice.

Should someone comment or apply a change to a PR, the issue will
no longer be considered stale and the automated close tag will be
aborted / revoked.

This really helps keep the PR / issue queue clean of issues
considered unimportant or PR's that have become abandoned.

More details are available here [0]

[0] https://probot.github.io/apps/stale/

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,17 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - pinned
+  - key_feature
+# Label to use when marking an issue as stale
+staleLabel: wontfix
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had
+  recent activity. It will be closed after 7 days if no further activity occurs.
+  Thank you for your contributions.
+# Comment to post when closing a stale issue. Set to `false` to disable
+closeComment: false


jetwhiz:
Is this going to wipe out our planned, long-term enhancements (such as "Add support for ECC in TPM 2.0")?

lukehinds:
> Is this going to wipe out our planned, long-term enhancements (such as "Add support for ECC in TPM 2.0")?

Only if we don't respond within the 7 days warning, also I am tracking those myself.

lukehinds:
We can also  use `exemptLabels:` for features we have agreed need to be done.

It's well worth getting going, when momentum builds and the issues start coming in, there will be a lot of of stuff raised , with a lot of it being issues that can't be recreated and the author vanishes. This way all authors get a notification stating it will go stale, so if its some they care about still, all they need to do is comment and it refreshes the 60 days period.

jetwhiz:
Ok, that sounds reasonable.  Can we go with the `exemptLabels` solution and use `key_feature` for such issues going forward then?

lukehinds:
> Ok, that sounds reasonable. Can we go with the `exemptLabels` solution and use `key_feature` for such issues going forward then?

Done! Will add the label and tag the key issues now.

lukehinds:
@jetwhiz , I tagged a few issues, feel free to tag more. I would also say to trust the stale bot, especially if its your own issues. Getting the notifications often prompts you to revisit an issue and add an update, for example (now that we completed this work, we could start looking at this now). If its something that's really no longer irrelevant, you just let stale bot mark it as wontfix.

lukehinds:
Is this ok to merge now @jetwhiz ?

jetwhiz:
Sure it looks good!  I think you've tagged all of the issues that I feel are key features, too

PR #126, opened 2019-06-13
Python 3 Port

lukehinds:
ToDo:

- [x] Unit tests passing rework with http client
- [x] encrypted payloads
- [x] IMA
- [x] encrypted payloads
- [x] whitelist handling performance (very slow and intensive).
- [x] http.client exception handler
- [x] tpm_policy
- [x] keylime_webapp
- [x] clean up code (remove all debug statements)
- [x] `tpm1.py`
- [x] cfssl
- [x] docker images
- [x] ansible roles


lukehinds:
note to self:  Change all realitive imports `from . import common` to using the main namespace `from keylime import common`

lukehinds:
@jetwhiz  Update, fixed the quote file issues.

Having something go wrong with tornado loop

https://paste.fedoraproject.org/paste/V-aO-zqMDVK1kxkhnwMBLQ

```
2019-06-18 09:27:31.886 - keylime.registrar_client - ERROR - Cannot run the event loop while another loop is running
  File "/usr/local/lib/python3.6/site-packages/keylime-1.2-py3.6.egg/keylime/registrar_client.py", line 103, in getKeys
    context=context)
  File "/usr/local/lib/python3.6/site-packages/keylime-1.2-py3.6.egg/keylime/tornado_requests.py", line 25, in request
    http_client = httpclient.HTTPClient()
  File "/usr/lib64/python3.6/site-packages/tornado/httpclient.py", line 94, in __init__
    gen.coroutine(lambda: async_client_class(**kwargs)))
  File "/usr/lib64/python3.6/site-packages/tornado/ioloop.py", line 576, in run_sync
    self.start()
  File "/usr/lib64/python3.6/site-packages/tornado/platform/asyncio.py", line 127, in start
    self.asyncio_loop.run_forever()
  File "/usr/lib64/python3.6/asyncio/base_events.py", line 428, in run_forever
    'Cannot run the event loop while another loop is running')
RuntimeError: Cannot run the event loop while another loop is running
```

lukehinds:
I have not figured out why yet, but python does not like the syntax:

```
def request(method,url,params=None,data=None,context=None):
    http_client = httpclient.AsyncHTTPClient() 
    if params is not None and len(list(params.keys()))>0:
        url+='?'
        for key in list(params.keys()):
            url+="%s=%s&"%(key,params[key])
        url=url[:-1]
    
    if context is not None:
        url = url.replace('http://','https://',1)
    try:
        request = httpclient.HTTPRequest(url=url,
                                         method =method,
                                         ssl_options=context,
                                         body=data)
        response = await http_client.fetch(request)
    except httpclient.HTTPError as e: 
        if e.response is None:
            return tornado_response(500,str(e))
        
        return tornado_response(e.response.code,e.response.body)
    if response is None:
        return None
    return tornado_response(response.code,response.body)
```

```
  File "/usr/lib64/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib64/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/usr/lib64/python3.6/unittest/__main__.py", line 18, in <module>
    main(module=None)
  File "/usr/lib64/python3.6/unittest/main.py", line 94, in __init__
    self.parseArgs(argv)
  File "/usr/lib64/python3.6/unittest/main.py", line 141, in parseArgs
    self.createTests()
  File "/usr/lib64/python3.6/unittest/main.py", line 148, in createTests
    self.module)
  File "/usr/lib64/python3.6/unittest/loader.py", line 219, in loadTestsFromNames
    suites = [self.loadTestsFromName(name, module) for name in names]
  File "/usr/lib64/python3.6/unittest/loader.py", line 219, in <listcomp>
    suites = [self.loadTestsFromName(name, module) for name in names]
  File "/usr/lib64/python3.6/unittest/loader.py", line 153, in loadTestsFromName
    module = __import__(module_name)
  File "/root/keylime/test/test_restful.py", line 82, in <module>
    from keylime import tornado_requests
  File "/usr/local/lib/python3.6/site-packages/keylime-1.2-py3.6.egg/keylime/tornado_requests.py", line 41
    response = await httpclient.fetch(request)
                              ^
SyntaxError: invalid syntax
```

lukehinds:
Issue we have now since using async:

```
  File "/usr/local/lib/python3.6/site-packages/keylime-1.2-py3.6.egg/keylime/registrar_client.py", line 146, in doRegisterAgent

    response_body = response.yaml()

AttributeError: 'coroutine' object has no attribute 'yaml'
```

If we take a look at `response` we can now see its a coroutine

`response: <coroutine object request at 0x7fa3e5f42eb8>`

My guess is we need to close this somehow so we can process it?

jetwhiz:
> Issue we have now since using async:
> 
> ```
>   File "/usr/local/lib/python3.6/site-packages/keylime-1.2-py3.6.egg/keylime/registrar_client.py", line 146, in doRegisterAgent
> 
>     response_body = response.yaml()
> 
> AttributeError: 'coroutine' object has no attribute 'yaml'
> ```
> 
> If we take a look at `response` we can now see its a coroutine
> 
> `response: <coroutine object request at 0x7fa3e5f42eb8>`
> 
> `/root/keylime/keylime//cloud_agent.py:614: RuntimeWarning: coroutine 'request' was never awaited`
> 
> My guess is we need to close this somehow so we can process it?

I think we might be awaiting the wrong thing here, which is causing a coroutine to be returned to "result".  Can you try switching `response = await http_client.fetch(request)` to: 

````python
response = http_client.fetch(request)
await response
````

and see what happens?

**Comment by lukehinds** [comment]
That's what I thought too @jetwhiz

This is the stack trace with the new code:

```
2019-06-20 16:06:20.404 - keylime.cloudagent - INFO - Agent UUID: D432FBB3-D2F1-4A97-9EF7-75BD81C00000

2019-06-20 16:06:20.407 - keylime.cloudagent - ERROR - 'coroutine' object has no attribute 'yaml'

  File "/root/keylime/keylime//cloud_agent.py", line 612, in <module>

    main()

  File "/root/keylime/keylime//cloud_agent.py", line 521, in main

    keyblob = registrar_client.doRegisterAgent(registrar_ip,registrar_port,agent_uuid,tpm_version,ek,ekcert,aik,ek_tpm,aik_name)

  File "/usr/local/lib/python3.6/site-packages/keylime-1.2-py3.6.egg/keylime/registrar_client.py", line 145, in doRegisterAgent

    response_body = response.yaml()

AttributeError: 'coroutine' object has no attribute 'yaml'

/root/keylime/keylime//cloud_agent.py:614: RuntimeWarning: coroutine 'request' was never awaited

  logger.exception(e)

```
I put a debug in and its still be returned as a coroutine `response: <coroutine object request at 0x7fec3d481eb8>` 

I am trying out different variations too, but no luck yet.

**Comment by lukehinds** [comment]
phew! all unit tests passing now @jetwhiz - still more to do, but the tornado port to async works now.

**Comment by jetwhiz** [comment]
> phew! all unit tests passing now @jetwhiz - still more to do, but the tornado port to async works now.

Awesome, great job @lukehinds !  What was the fix with the tornado async port?

**Comment by lukehinds** [comment]
> > phew! all unit tests passing now @jetwhiz - still more to do, but the tornado port to async works now.
> 
> Awesome, great job @lukehinds ! What was the fix with the tornado async port?

You basically have to `await` every call to a function decorated with `async`:

https://github.com/keylime/keylime/blob/ce901a6d37e75a8cc4a21121047f486601a1a28a/keylime/registrar_client.py#L142-L146

Looking at the code it still needs a bit of cleaning up (some debug stuff I have left in), but its getting ever nearer.

**Comment by lukehinds** [comment]
Just an update, now we are using `asyncio` other areas have to be changed to asynchronous code as well (for example sqllite).  This is going to take a bit longer again, 

Here's why , whenever you wrap a function with `async` that function needs to be invoked with `await` which then places it for execution on the event loop.  Any use of `await` also needs to occur from *within* an `async` function or called with `loop.run_until_complete(my_function())`

This invariably means you find yourself walking backwards through the code, to change all function calls  to handle asynchronous handles / coroutines. 

This explains it quite well https://www.aeracode.org/2018/02/19/python-async-simplified/

**Comment by lukehinds** [comment]
@jetwhiz & @nabilschear 

Help me out with this one please, not sure what format we are trying to get:

```
2019-07-08 12:47:25.805 - keylime.tpm - INFO - Checking IMA measurement list...
filedata_hash: <class 'str'>
filedata_hash: 0000000000000000000000000000000000000000
filedata_algo: <class 'str'>
filedata_algo: sha1
2019-07-08 12:47:25.806 - keylime.cloudverifier - ERROR - argument for 's' must be a bytes object
  File "/usr/local/lib/python3.7/site-packages/keylime-1.2-py3.7.egg/keylime/cloud_verifier_tornado.py", line 316, in on_get_quote_response
    process_quote_response_results = await cloud_verifier_common.process_quote_response(agent, yaml_response['results'])
  File "/usr/local/lib/python3.7/site-packages/keylime-1.2-py3.7.egg/keylime/cloud_verifier_common.py", line 239, in process_quote_response
    hash_alg)
  File "/usr/local/lib/python3.7/site-packages/keylime-1.2-py3.7.egg/keylime/tpm2.py", line 1023, in check_quote
    return self.check_pcrs(tpm_policy, pcrs, data, False, ima_measurement_list, ima_whitelist)
  File "/usr/local/lib/python3.7/site-packages/keylime-1.2-py3.7.egg/keylime/tpm_abstract.py", line 364, in check_pcrs
    if self.__check_ima(pcrval, ima_measurement_list, ima_whitelist):
  File "/usr/local/lib/python3.7/site-packages/keylime-1.2-py3.7.egg/keylime/tpm_abstract.py", line 318, in __check_ima
    ex_value = ima.process_measurement_list(ima_measurement_list.split('\n'), ima_whitelist)
  File "/usr/local/lib/python3.7/site-packages/keylime-1.2-py3.7.egg/keylime/ima.py", line 181, in process_measurement_list
    tohash=struct.pack(fmt,len(filedata_hash)+len(filedata_algo)+2,filedata_algo,ord(':'),ord('\0'),filedata_hash,len(path)+1,path,ord('\0'))
struct.error: argument for 's' must be a bytes object
```

**Comment by jetwhiz** [comment]
Hi Luke, 

I am thinking that `filedata_algo` and `filedata_hash` must be encoded to a byte array instead of a string in order to match with the 's' fields in the pack format.  For example: 

````python
tohash=struct.pack(fmt,len(filedata_hash)+len(filedata_algo)+2,filedata_algo.encode('utf-8'),ord(':'),ord('\0'),filedata_hash.encode('utf-8'),len(path)+1,path,ord('\0'))
````

**Comment by lukehinds** [comment]
Hey @jetwhiz 

I had queued up a long post with lots of debug, but I think it might be quicker if you run this yourself.

It does not require you to build and run keylime, you should be able to do the following:

```
git checkout -b lukehinds-python3-port master
git pull https://github.com/lukehinds/keylime.git python3-port
```

And then:

`python3 keylime/keylime/ima.py`

from a directory that has a whitelist.txt

I am beginning to chase my tail here and run further down a rabbit hole, could you take a quick look as you have some history of how things are the way they are.

We should be very close then, I have all the async stuff working with tornado now. 

Big thanks in advance.

**Comment by lukehinds** [comment]
This one in particular has me quizzed:

https://github.com/keylime/keylime/blob/e22b10b4567bcad25bbcfe321520d1b16311fcca/keylime/ima.py#L151

`template_hash` is already decoded. If I add the following line:

print 'template_hash_pre: {0}'.format(tokens[1])
template_hash_pre: 1d8d532d463c9f8c205d0df7787669a85f93e260

So I don't have any insight into what we are hashing too?, so that i can make the same call in Py3.

**Comment by jetwhiz** [comment]
Hi @lukehinds 

For decode, we're essentially taking a string of hex pairs and converting it to binary data.  You should be able to get the same hex decode functionality in Py3 via the [codecs](https://docs.python.org/3/library/codecs.html) or [binascii](https://docs.python.org/3/library/binascii.html#module-binascii) modules.  

For instance: 
````python
template_hash=codecs.decode(tokens[1], "hex").decode("utf-8")
````

The rest of that code is basically performing what IMA (and the TPM) do themselves but in Python -- calculating the hash for each row and extending it into the register storage.  Of course IMA stores this in the TPM, but we're just keeping it locally in the `runninghash` variable.   You can get a rough idea about what's happening in that code on the IMA description where they say how they calculate the template-hash for each entry: https://sourceforge.net/p/linux-ima/wiki/Home/#ima-measurement-list. 

I don't think anything there needs to be modified, though (other than the encoding I mentioned before). 

I'll try to run this later today and see what errors start coming up.

**Comment by lukehinds** [comment]
Any luck @jetwhiz 

Do you have any ideas what we are doing here?

```
#verify template hash. yep this is terrible
fmt = "<I%dsBB%dsI%dsB"%(len(filedata_algo),len(filedata_hash),len(path))
```

**Comment by jetwhiz** [comment]
Hey @lukehinds 

Here are the changes that I made to get this function working: 

[commit-2b83ff8.txt](https://github.com/keylime/keylime/files/3387746/commit-2b83ff8.txt)

Can you try this out and see if this gets it working for you?

**Comment by lukehinds** [comment]
Thanks @jetwhiz - that certainly worked.

Now noticed some more issues with async operations (huge CPU use). I must be approaching this wrong in how I am stacking things on to the event loop, will try to debug what's wrong.

**Comment by lukehinds** [comment]
@jetwhiz need some more of your encoding wizardry:

https://github.com/keylime/keylime/blob/f0cf1f1fefc3c96d791bf28918e42ab419b956fc/keylime/cloud_verifier_common.py#L257-L275

I placed the following debug lines into `crypto.py`

```
def rsa_encrypt(key,message):
    print('key:', type(key))
    print('key:', key)
    print('message:', type(message))
    print('message:', message)
    cipher = PKCS1_OAEP.new(key)
    return cipher.encrypt(message)
```

```
2019-07-16 14:26:34.483 - keylime.cloudverifier - INFO - PUT returning 404 response. agent id: D432FBB3-D2F1-4A97-9EF7-75BD81C00000 not found.
2019-07-16 14:26:34.500 - keylime.tpm - INFO - Checking IMA measurement list...
key: <class 'Cryptodome.PublicKey.RSA.RsaKey'>
key: Public RSA key at 0x7F359D32E908
message: <class 'str'>
message: b"\xe5J\xae\xcd\xbe6\x04'4Fv\x12at\x03\xd9l\xd1\xbc\xe5\xf7\x8a\xfa\x8b\x89\x8e\x9c\xbdo*\\,"
2019-07-16 14:26:34.526 - keylime.cloudverifier - ERROR - Polling thread error: can't concat str to bytes
2019-07-16 14:26:34.526 - keylime.cloudverifier - ERROR - can't concat str to bytes
  File "/usr/local/lib/python3.7/site-packages/keylime-1.2-py3.7.egg/keylime/cloud_verifier_tornado.py", line 405, in process_agent
    await self.invoke_provide_v(agent)
  File "/usr/local/lib/python3.7/site-packages/keylime-1.2-py3.7.egg/keylime/cloud_verifier_tornado.py", line 340, in invoke_provide_v
    v_yaml_message = cloud_verifier_common.prepare_v(agent)
  File "/usr/local/lib/python3.7/site-packages/keylime-1.2-py3.7.egg/keylime/cloud_verifier_common.py", line 267, in prepare_v
    b64_encrypted_V = base64.b64encode(crypto.rsa_encrypt(crypto.rsa_import_pubkey(agent['public_key']),str(base64.b64decode(agent['v']))))
  File "/usr/local/lib/python3.7/site-packages/keylime-1.2-py3.7.egg/keylime/crypto.py", line 70, in rsa_encrypt
    return cipher.encrypt(message)
  File "/usr/local/lib64/python3.7/site-packages/Cryptodome/Cipher/PKCS1_OAEP.py", line 121, in encrypt
    db = lHash + ps + b'\x01' + _copy_bytes(None, None, message)
TypeError: can't concat str to bytes
```

**Comment by jetwhiz** [comment]
Hi @lukehinds 

Can you remove the string casting for the `str(base64.b64decode(agent['v']))` portion on line 267, since the `rsa_encrypt` method is expecting a `bytes` object instead of a string.  It should be just `base64.b64decode(agent['v'])`.

**Comment by lukehinds** [comment]
I could have sworn I tried that, cleaned up all my changes and it worked!

**Comment by lukehinds** [comment]
Next thing we need to tackle @jetwhiz is how to convert the following callback fetch requests to async:

https://github.com/keylime/keylime/blob/fc9307854b2ccab09f6a69e5352f9c2e84742833/keylime/cloud_verifier_tornado.py#L281-L293

If you have any ideas, let me know.

**Comment by nabilschear** [comment]
@lukehinds i took a look through this and started to debug some things myself.  One thing that dawned on me is that none of the keylime apps/services need to be async enabled except for the verifier.  So, agent, registrar and tenant can all use blocking I/O.

**Comment by nabilschear** [comment]
would that simplify things?  it looks like there's alot of code for handling sql and simple web requests asynchronously that aren't necessary

**Comment by lukehinds** [comment]
> @lukehinds i took a look through this and started to debug some things myself. One thing that dawned on me is that none of the keylime apps/services need to be async enabled except for the verifier. So, agent, registrar and tenant can all use blocking I/O.

I was saying just the same in gitter earlier. A lot of services are all using `tornado_requests.py` but they don't need async or even tornado, they could be replaced with something native such as urlib3 or requests.

**Comment by lukehinds** [comment]
adding to the above, the verifier is still not in good health - its timing out a lot and returning malformed packets.

**Comment by lukehinds** [comment]
so I have started to port `tenant.py` with some success to just using the standard 'http.client' library in Python

The question now @jetwhiz / @nabilschear is what needs to be non blocking? Is it just the verifier or the registrar too?

I would have thought just the verifier, as that's what gets a lot of requests (quotes), where as the registrar is fairly light in its request volume? I am thinking I could perhaps port `registrar_client.py` away from being async and using `http.client` as well?

File: keylime/ca_impl_cfssl.py [hunk 1]
```diff
@@ -102,7 +103,7 @@ def mk_cacert():
                    }
                      ]
            }
-    data = json.dumps(csr)
+    data = yaml.dump(csr)
```
jetwhiz: I think these all have to stay JSON (including below), since cfssl is using JSON for inputs and outputs.  

When I ran this I got a lot of errors when generating the certs, which may be related to some of the verifier issues you're seeing.

keylime/tpm_abstract.py
```diff
@@ -401,7 +399,7 @@ def init_system_rand(self):
                     # as fp has a method fileno(), you can pass it to ioctl
```
jetwhiz: The code above is broken for py3: `t = struct.pack("ii%ds"%len(rand_data), 8, len(rand_data), str(rand_data))`

`pack` expects binary instead of a string for rand_data, so we need to remove the `str` conversion. 

````python
t = struct.pack("ii%ds"%len(rand_data), 8, len(rand_data), rand_data)
````

File: keylime/tenant_webapp.py [hunk 3]
```diff
@@ -3,45 +3,45 @@
...
 import ssl
 import json
-import ConfigParser
+import configparser
 import traceback
 import sys
 import tornado.ioloop
 import tornado.web
-import tornado_requests
+from . import tornado_requests
```
jetwhiz: Does this need to be `from keylime import tornado_requests`?

lukehinds: yep, all relative imports need to be using the namespace.

keylime/tpm_abstract.py
lukehinds: thanks

keylime/ca_impl_cfssl.py
lukehinds: I will take a look, main thing is if its going to try and serialize bytes. If so someone was recommending simplejson (on pypi that will allow byte handling in json).

lukehinds:
ok, I am much happier with how this is working.

I gave up fighting async apart from when we use AsyncHTTPClient in tornado.

The only issue I can really see now is a slow response when its a big whitelist. My guess is we need to optimize how we are iterating over that list, or loading it into the DB.

There is also likely to be more work on tpm1.py and cf_ssl  - I have been coding against tmp2 and openssl.

Let me know anything you find!

keylime/tenant_webapp.py
```diff
@@ -61,40 +61,40 @@ def write_error(self, status_code, **kwargs):
...
-        # Get default policies for TPM/vTPM from config as suggestions to user 
+
+        # Get default policies for TPM/vTPM from config as suggestions to user
         tpm_policy = json.dumps(json.loads(config.get('tenant', 'tpm_policy')), indent=2)
         vtpm_policy = json.dumps(json.loads(config.get('tenant', 'vtpm_policy')), indent=2)
-        
+
         self.set_status(200)
-        self.set_header('Content-Type', 'text/html')
+        self.set_header('Content-Type', 'text/x-yaml; charset=utf-8')
```
jetwhiz: This should still be `text/html`, since we're writing out an HTML document to the browser.

lukehinds:
@nabilschear @jetwhiz 

commit 164704787e36b922f7e6e828b4cffd7b64f8dc63 has IMA working, ignore line 93 (that's a bit of debug code that I have removed for the next commit).

I am now on to encrypted payloads, but cannot get this working. I verified my keylime_tenant args and set up were correct against the python 2 code base.

Could you take a look at this when you have a chance?

lukehinds:
I think this might be a possible cause:

```
`        if ex_mac == self.auth_tag:
            logger.info( "Successfully derived K for UUID %s",self.agent_uuid)
            self.final_U = decrypted_U
            self.K = candidate_key
            return True
        else:
            print('ex_mac:', ex_mac)
            print('self.auth_tag:', self.auth_tag)
```

```
2019-07-22 15:01:32.934 - keylime.cloudagent - INFO - GET integrity quote returning 200 response.
ex_mac: b27f6aba4d6b4b78e6933bc955840176212c03a642c249d0a50a30a9db8c5a56319abcd4054f63defe5e6e7c752f0d64
self.auth_tag: d2478f954bb9516d2d6ee7453ac7110fc77a19aaad09883ca40bccc82dc0897fd475e81a2045e1ded7adec1ef90d10ff
```

lukehinds:
fixed the above with `candidate_key = crypto.strbitxor(decrypted_U, decrypted_V)`

lukehinds:
@jetwhiz having some issues with this one:

*tenant.py*

```
    async def do_verify(self):
        """initiaite v, agent_id and ip
        initiate the cloudinit sequence"""
        challenge = TPM_Utilities.random_password(20)

        numtries = 0
        while True:
            try:
                response = tornado_requests.request("GET",
                                            "http://%s:%s/keys/verify?challenge=%s"%(self.cloudagent_ip,self.cloudagent_port,challenge))
                response = await response
            except Exception as e:
                # this is one exception that should return a 'keep going' response
                if tornado_requests.is_refused(e):
                    numtries+=1
                    maxr = config.getint('tenant','max_retries')
                    if numtries >= maxr:
                        raise UserError("Quitting after max number of retries to connect to %s"%(self.cloudagent_ip))
                    retry  = config.getfloat('tenant','retry_interval')
                    logger.info("Connection to %s refused %d/%d times, trying again in %f seconds..."%(self.cloudagent_ip,numtries,maxr,retry))
                    time.sleep(retry)
                    continue
                else:
                    raise e
            response_body = response.yaml()
            if response.status_code == 200:
                if "results" not in response_body or 'hmac' not in response_body['results']:
                    logger.critical("Error: unexpected http response body from Cloud Agent: %s"%str(response.status_code))
                    break
                mac = response_body['results']['hmac']
                ex_mac = crypto.do_hmac(self.K,challenge)
                print('mac:', mac)
                print('ex_mac:', ex_mac)
                if mac == ex_mac:
                    logger.info("Key derivation successful")
                else:
                    logger.error("Key derivation failed")
            else:
                common.log_http_response(logger,logging.ERROR,response_body)
                retry  = config.getfloat('tenant','retry_interval')
                logger.warning("Key derivation not yet complete...trying again in %s seconds...Ctrl-C to stop"%retry)
                time.sleep(retry)
                continue
            break;
```

*crypto.py:*

```
def do_hmac(key,value):
    value = value.encode('utf-8')
    # Let's only encode if its not a byte
    try:
        key = key.encode('utf-8')
    except AttributeError:
        pass
    h = HMAC.new(key,value,digestmod=SHA384.new())
    return h.hexdigest()
```

I require `value = value.encode('utf-8')`

As otherwise we get:

```
  File "/usr/local/bin/keylime_agent", line 11, in <module>
    load_entry_point('keylime==1.2', 'console_scripts', 'keylime_agent')()
  File "/usr/local/lib/python3.7/site-packages/keylime-1.2-py3.7.egg/keylime/cloud_agent.py", line 542, in main
    retval = registrar_client.doActivateAgent(registrar_ip,registrar_port,agent_uuid,key)
  File "/usr/local/lib/python3.7/site-packages/keylime-1.2-py3.7.egg/keylime/registrar_client.py", line 167, in doActivateAgent
    'auth_tag': crypto.do_hmac(base64.b64decode(key),agent_id),
  File "/usr/local/lib/python3.7/site-packages/keylime-1.2-py3.7.egg/keylime/crypto.py", line 93, in do_hmac
    h = HMAC.new(key,value,digestmod=SHA384.new())
  File "/usr/local/lib64/python3.7/site-packages/Cryptodome/Hash/HMAC.py", line 213, in new
    return HMAC(key, msg, digestmod)
  File "/usr/local/lib64/python3.7/site-packages/Cryptodome/Hash/HMAC.py", line 87, in __init__
    self._inner.update(msg)
  File "/usr/local/lib64/python3.7/site-packages/Cryptodome/Hash/SHA384.py", line 90, in update
    c_uint8_ptr(data),
  File "/usr/local/lib64/python3.7/site-packages/Cryptodome/Util/_raw_api.py", line 144, in c_uint8_ptr
    raise TypeError("Object type %s cannot be passed to C code" % type(data))
TypeError: Object type <class 'str'> cannot be passed to C code
```

I digress, so the issue is `do_verify`

Here is the result:

```
keylime_tenant -v 127.0.0.1 -t 127.0.0.1 -f /root/excludes.txt --uuid D432FBB3-D2F1-4A97-9EF7-75BD81C00000 --verify
                                                                           
2019-07-26 15:27:03.080 - keylime.tpm2 - INFO - Using legacy version of tpm2-tools
2019-07-26 15:27:03.094 - keylime.tenant - WARNING - CAUTION: using default password for private key, please set private_key_pw to a strong password
2019-07-26 15:27:03.095 - keylime.tenant - INFO - Setting up client TLS in /var/lib/keylime/cv_ca
2019-07-26 15:27:03.101 - keylime.tenant - INFO - TPM PCR Mask from policy is 0x408000
2019-07-26 15:27:03.101 - keylime.tenant - INFO - vTPM PCR Mask from policy is 0x808000
2019-07-26 15:27:03.271 - keylime.registrar_client - INFO - Setting up client TLS...
2019-07-26 15:27:03.292 - keylime.registrar_client - WARNING - CAUTION: using default password for private key, please set private_key_pw to a strong password
2019-07-26 15:27:03.348 - keylime.tenant - WARNING - DANGER: EK cert checking is disabled and no additional checks on EKs have been specified with ek_check_script option. Keylime is not secure!!
2019-07-26 15:27:03.349 - keylime.tenant - INFO - Quote from 127.0.0.1 validated
2019-07-26 15:27:03.372 - keylime.tenant - ERROR - Response code 400: Bootstrap key not yet available.
2019-07-26 15:27:03.372 - keylime.tenant - WARNING - Key derivation not yet complete...trying again in 1.0 seconds...Ctrl-C to stop
mac: 5b4616341fa141ac24195493a4c1979e7f9155c623e4ad8ad9d1a7b73d50ef6d88adced7e252468d420c89842fb6e30a
ex_mac: e7583f0168264165b9b8aeb4ac39641de81900b03bcfb6aeef5c043ac21af2e15ebca913db7217acf22ec6db039def4c
2019-07-26 15:27:04.386 - keylime.tenant - ERROR - Key derivation failed
```

So key bootpoints, `Response code 400: Bootstrap key not yet available.` and the non match for `mac` and `ex_mac`

Any ideas on possible debug approach we could take?

jetwhiz:
@lukehinds 

Have you managed to track this issue down yet?  I'm still trying to wrap my head around all of the encode-decodes introduced because of py3, so there are a lot of reasons the HMACs might be coming out different (especially since even a tiny change somewhere along the line can result in a completely different HMAC calculation).  

For reference, the `self.server.K` variable is being set here: 

https://github.com/keylime/keylime/blob/436ce3c73f7cc63f8fc2a45ab7fa2da91da40f56/keylime/cloud_agent.py#L460-L466

keylime/cloud_agent.py
```diff
@@ -158,90 +160,90 @@ def do_GET(self):
...
         enc_path = "%s/encrypted_payload"%common.WORK_DIR
-        
+
         dec_payload = None
         enc_payload = None
-        
+
         if self.server.payload is not None:
-            dec_payload = crypto.decrypt(self.server.payload, str(self.server.K))
+            dec_payload = crypto.decrypt(self.server.payload, self.server.K)
```
jetwhiz: Make sure the `elif` block below matches how we're doing it here.  Right now, if we're using a previous payload then we load with `dec_payload = crypto.decrypt(enc_payload,str(self.server.K))` (note the conversion to _str_ still)

lukehinds: This is what we get with `dec_payload = crypto.decrypt(self.server.payload, str(self.server.K))`


```
2019-07-31 12:02:49.703 - keylime.cloudagent - INFO - POST of V key returning 200
----------------------------------------
Exception happened during processing of request from ('127.0.0.1', 58382)
  File "/usr/lib64/python3.7/socketserver.py", line 650, in process_request_thread
    self.finish_request(request, client_address)
  File "/usr/lib64/python3.7/socketserver.py", line 360, in finish_request
    self.RequestHandlerClass(request, client_address, self)
  File "/usr/lib64/python3.7/socketserver.py", line 720, in __init__
    self.handle()
  File "/usr/lib64/python3.7/http/server.py", line 426, in handle
    self.handle_one_request()
  File "/usr/lib64/python3.7/http/server.py", line 414, in handle_one_request
    method()
  File "/usr/local/lib/python3.7/site-packages/keylime-1.2-py3.7.egg/keylime/cloud_agent.py", line 246, in do_POST
    dec_payload = crypto.decrypt(self.server.payload, str(self.server.K))
  File "/usr/local/lib/python3.7/site-packages/keylime-1.2-py3.7.egg/keylime/crypto.py", line 157, in decrypt
    cipher = AES.new(key, AES.MODE_GCM, nonce = nonce)
  File "/usr/local/lib64/python3.7/site-packages/Cryptodome/Cipher/AES.py", line 232, in new
    return _create_cipher(sys.modules[__name__], key, mode, *args, **kwargs)
  File "/usr/local/lib64/python3.7/site-packages/Cryptodome/Cipher/__init__.py", line 79, in _create_cipher
    return modes[mode](factory, **kwargs)
  File "/usr/local/lib64/python3.7/site-packages/Cryptodome/Cipher/_mode_gcm.py", line 622, in _create_gcm_cipher
    return GcmMode(factory, key, nonce, mac_len, kwargs, ghash_c)
  File "/usr/local/lib64/python3.7/site-packages/Cryptodome/Cipher/_mode_gcm.py", line 225, in __init__
    **cipher_params
  File "/usr/local/lib64/python3.7/site-packages/Cryptodome/Cipher/AES.py", line 232, in new
    return _create_cipher(sys.modules[__name__], key, mode, *args, **kwargs)
  File "/usr/local/lib64/python3.7/site-packages/Cryptodome/Cipher/__init__.py", line 79, in _create_cipher
    return modes[mode](factory, **kwargs)
  File "/usr/local/lib64/python3.7/site-packages/Cryptodome/Cipher/_mode_ecb.py", line 215, in _create_ecb_cipher
    cipher_state = factory._create_base_cipher(kwargs)
  File "/usr/local/lib64/python3.7/site-packages/Cryptodome/Cipher/AES.py", line 93, in _create_base_cipher
    raise ValueError("Incorrect AES key length (%d bytes)" % len(key))
ValueError: Incorrect AES key length (102 bytes)
```

keylime/crypto.py
```diff
@@ -79,65 +79,81 @@ def strbitxor(a,b):
...
+    padding = b'\x80'+b'\0'*pad_len
+    result = s + padding
     return s + padding
- 
+
 def _strip_pad(s):
     '''
     Strips the padding from the string
     '''
     return s.rstrip(b'\0')[:-1]
```
jetwhiz: This isn't relevant to the py3 conversion, but shouldn't we be stripping the appended \x80 character as well @nabilschear ?  Under padding mode 2, we add a 1 (always) and then pad with 0's until we hit a multiple of 16 (if necessary).  

In python2 this might not make a difference, since `chr(0x80)` isn't valid 7-bit ASCII and isn't printable, but this might cause some unexpected problems now that we're using UTF-8.

keylime/cloud_agent.py
lukehinds: and debug output for candidate and decrypted_U

```
decrypted_U: b'\x9eM:!\xca\n\xca\xabn"\xc3f\x11f\xb2:5\x9e\x1a<\xf6p\xae\xef\xb9[\xaaE\xd3W\xfb\x13'
candidate_key: bytearray(b'\xdf\x1c(\x0c\xc5\xbd8\x08\xd2\xc5\xc6\x8em\xf3\xde0Y\xbf\n\x15:\xed\x1f.V\x98\x1d\xc9\xb5:\x9f\xf7')
```

jetwhiz: I noticed in the latest commit you added (https://github.com/keylime/keylime/pull/126/commits/436ce3c73f7cc63f8fc2a45ab7fa2da91da40f56#diff-242fc79fc17586955d7b4bcd2c12f4d5R637): 

````python
    async def do_verify(self):
        """initiaite v, agent_id and ip
        """initiaite v, agent_id and ip
        initiate the cloudinit sequence"""
        initiate the cloudinit sequence"""
        challenge = TPM_Utilities.random_password(20)
        challenge = TPM_Utilities.random_password(20)
        numtries = 0
        encoded = base64.b64encode(self.K).decode('utf-8')
````

This might be causing the issue you're seeing, since the crypto libraries are expecting a raw key `self.K` and not base64-encoded. 

Does `encoded = base64.b64encode(self.K).decode('utf-8')` need to be removed?

Comment (jetwhiz) [reply keylime/cloud_agent.py]: Note that I think the way you've fixed the `if` block there is correct, but just don't forget to update the `elif` block as well.

From: 
https://github.com/keylime/keylime/blob/436ce3c73f7cc63f8fc2a45ab7fa2da91da40f56/keylime/cloud_agent.py#L245-L253

To: 
````python
        if self.server.payload is not None:
            dec_payload = crypto.decrypt(self.server.payload, self.server.K)
            enc_payload = self.server.payload
        elif os.path.exists(enc_path):
            # if no payload provided, try to decrypt one from a previous run stored in encrypted_payload
            with open(enc_path,'r') as f:
                enc_payload = f.read()
            try:
                dec_payload = crypto.decrypt(enc_payload,self.server.K)
````

Comment (lukehinds) [reply keylime/cloud_agent.py]: I have removed that for now, still getting a mismatch:

```
mac: 129ce651fabc4942b59a08efb7eb0a1444ddd04909e92e01cac7fce8d31ed9cb91edfeb843e7d9a61c4540622b7a08bf
mac: <class 'str'>
ex_mac: 8b63b4df48a59e9928d2c2f004a0673fd85c67c3692398ab504a3fd8563a58133d734ac4df51417f13f1960ad129047e
ex_mac: <class 'str'>
```

Comment (jetwhiz) [reply keylime/cloud_agent.py]: This seems to be working on my side, but I get hit with a random `Error: HTTP 599: Timeout while connecting` error on the verifier on occasion.  

I do get an issue on dealing with the payload for the cloud_agent though, related to appending byte arrays to each other.

Comment (lukehinds) [reply keylime/cloud_agent.py]: I have made some progress here:

```
2019-07-31 13:26:25.806 - keylime.tenant - ERROR - Response code 400: Bootstrap key not yet available.
2019-07-31 13:26:25.806 - keylime.tenant - WARNING - Key derivation not yet complete...trying again in 1.0 seconds...Ctrl-C to stop
2019-07-31 13:26:26.818 - keylime.tenant - INFO - Key derivation successful
```

However this couples with:

```
  File "/usr/local/lib/python3.7/site-packages/keylime-1.2-py3.7.egg/keylime/cloud_agent.py", line 264, in do_POST
    f.write(self.server.payload)
TypeError: write() argument must be str, not bytes
```

Yet if I encode self.server.payload or add a `b` to 

```
with open(enc_path,'wb') as f:
                f.write(self.server.payload)
```

Key derivation fails. I need to work out end to end how to handle the payloads.

Comment (lukehinds) [reply keylime/cloud_agent.py]: Should I push my new code @jetwhiz ?

File: keylime/cloud_agent.py [hunk 7]
```diff
@@ -254,23 +256,23 @@ def do_POST(self):
...
         # also write out encrypted payload to be decrytped next time
         if enc_payload is not None:
             with open(enc_path,'w') as f:
-                f.write(self.server.payload)
+                f.write(self.server.payload.decode())
 
         # deal with payload
         payload_thread = None
         if dec_payload is not None:
             tomeasure += dec_payload
```
jetwhiz: This might need to be changed to 

````python
tomeasure = tomeasure + dec_payload
````

I was getting an error saying that the `object cannot re-sized` otherwise.

File: keylime/cloud_agent.py (hunk 5 above)
Comment (jetwhiz) [reply keylime/cloud_agent.py]: Your `436ce3c` commit seems to be working for me.  What's changed in your new code?  Just adding that binary flag to the file write for payload? 

Did you also add the `b` to the corresponding read part?  

https://github.com/keylime/keylime/blob/436ce3c73f7cc63f8fc2a45ab7fa2da91da40f56/keylime/cloud_agent.py#L250-L251

Comment (lukehinds) [reply keylime/cloud_agent.py]: I will get you the whole file, its working completely now:

https://gist.github.com/lukehinds/ecf5fe0cceb366bbd771cf6b2e7b2c78

Comment (jetwhiz) [reply keylime/cloud_agent.py]: Nice!  Go ahead and push that commit and I'll take a look

**Comment by lukehinds** [comment]
@jetwhiz c837f41 resolves payload using `--verify`

As far as I can tell, the remaining issue we have now is everything grinding when a large whitelist is used.

A demo here:

https://streamable.com/bglfx

**Comment by jetwhiz** [comment]
Looks like it is working on my side, too ... great job!  

Are you still getting issues with the verifier timing out after a while?  

Also, does this need to be updated to match the binary-write change you made, or does this cause problems for you? 

https://github.com/keylime/keylime/blob/436ce3c73f7cc63f8fc2a45ab7fa2da91da40f56/keylime/cloud_agent.py#L250-L251

(using `'rb'` instead of `'b'`)

**Comment by lukehinds** [comment]
> Looks like it is working on my side, too ... great job!
> 
> Are you still getting issues with the verifier timing out after a while?
> 

No issues here. Which version of tornado do you have?

On my machine:

```
python3 -c 'import tornado; print(tornado.version)'                                                                                                                        
5.0.2
```


> Also, does this need to be updated to match the binary-write change you made, or does this cause problems for you?
> 
> https://github.com/keylime/keylime/blob/436ce3c73f7cc63f8fc2a45ab7fa2da91da40f56/keylime/cloud_agent.py#L250-L251
> 
> (using `'rb'` instead of `'b'`)

Works ok, will make another commit

**Comment by jetwhiz** [comment]
Is it just taking a long time to send the large IMA whitelists, or is the processing taking a while as well?  

I'm wondering if it has to do with sending a giant text blob over the REST interface (from the tenant to the verifier): 

https://github.com/keylime/keylime/blob/4a46fa924d34d54f8f5e64f2e707fb9192d5b609/keylime/tenant.py#L433-L460

If I run a large whitelist directly in IMA (via ima.py) I do not see any slowdowns.

**Comment by lukehinds** [comment]
> Is it just taking a long time to send the large IMA whitelists, or is the processing taking a while as well?
> 
> I'm wondering if it has to do with sending a giant text blob over the REST interface (from the tenant to the verifier):
> 
> https://github.com/keylime/keylime/blob/4a46fa924d34d54f8f5e64f2e707fb9192d5b609/keylime/tenant.py#L433-L460
> 
> If I run a large whitelist directly in IMA (via ima.py) I do not see any slowdowns.

I don't think so. One observation is that performing a delete against a node that was provisioned with a large whitelist also takes a long time and spikes the CPU:

```
keylime_tenant -v 127.0.0.1 -t 127.0.0.1  --uuid D432FBB3-D2F1-4A97-9EF7-75BD81C00000 -c delete 
```

I am beginning to wonder if its maybe the database or processsing yaml instead of json

**Comment by jetwhiz** [comment]
Can you benchmark the line that has `yaml_message = yaml.dump(data)` for a large whitelist and see how long it takes?  You'd expect a slowdown there and in the corresponding `yaml.safe_load` if yaml is the issue.

**Comment by lukehinds** [comment]
> Can you benchmark the line that has `yaml_message = yaml.dump(data)` for a large whitelist and see how long it takes? You'd expect a slowdown there and in the corresponding `yaml.safe_load` if yaml is the issue.

Which one @jetwhiz ?

**Comment by lukehinds** [comment]
daf7673 resolves a lot of performance issues, but we are not there yet.

There is still a spike for sql queries when there is a large whitelist. 

I will try the following:

Handle the whitelist with json, there are no byte objects to handle so this should be fine.

If for some reason the above does not work (I can't see why), then we can try to optimise the SQL queries.

**Comment by lukehinds** [comment]
hi @jetwhiz 

Trying to resolve this in tpm1.py

```
with tempfile.NamedTemporaryFile() as inFile:
            print('get_tpm_metadata:', base64.b64decode(self.get_tpm_metadata('aikpriv')))
            print('get_tpm_metadata:', type(base64.b64decode(self.get_tpm_metadata('aikpriv'))))
            inFile.write(base64.b64decode(self.get_tpm_metadata('aikpriv')))
            inFile.flush()

            retDict = self.__run("loadkey -hp 40000000 -ik %s"%(inFile.name))
            retout = retDict['retout']

            if len(retout)>0 and len(retout[0].split())>=4:
                handle = retout[0].split()[4]
            else:
                raise Exception("unable to process output of loadkey %s"%(retout))

        return handle.upper()
```

```
2019-08-28 13:42:04.065 - keylime.tpm1 - INFO - TPM Owner password confirmed: keylime
2019-08-28 13:42:04.074 - keylime.tpm1 - WARNING - No EK certificate found in TPM NVRAM
get_tpm_metadata: b'\x01\x01\x00\x00\x00\x12\x00\x00\x00\x00\x01\x00\x00\x00\x01\x00\x01\x00\x02\x00\x00\x00\x0c\x00\x00\x08\x00\x00\x00\x00\x02\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01\x00\xa8\xa3*\xae\xdd\x11\xa7\xff\xa8i\x98\x8e\xe7\x8d\x866\rn\xe1\x91\xe2\xe0F\xd1h\xf8}\x9a\xf72\xf7!\xa7r\xb5\x8a<#\xd6\x96\xfa\x877E?W\xa9.\xdd\x18\xe2\x8a\x87?a\x97\xff\x92\x90f4\xd9\x17\xb1q\xd3\xd55\n\x16iZL\xaf:`\\\x11\xf7m\xbdO\\\x8f\xaa\x9d\x15\xb5\xc3\xc4\xbd\xcc\xe7\xbd\x81\xf6M\x17\xfa2\x90a\xfb\x95+\x82\xea\xac\xe1\xf2\xf7\x1d\xb9\x9au\xc3\x199\x1cp\x88\xf1Mn\x03\xc8\xff\xf2Se\xe4\x1cL\x7f\x9f\xb8Z\xf6\xe4ib\xb7\x94\x80\x10\xb2:\xf3\xbcN\x85\xa7"\xf3u@\xb9i\xbeZ/nv\x83\n{P\xd2\x12\xc3\x9aR\xc1\x1f\x93\x93\xc9\xb5`7\xf8\xb3x\xe6\x93z\xdc\xe1\x01;a\'\t\xf4\xbcc\xf1\xf8\xe1\xd0\xf5\xbd\x9b\xb60D\x86h\xd3+C]\xad\x82/!c\xbb\xd2\xa3\xd7\x0cI\x93\'\xef]\xeb\xa5\xecZ\x9b\xd6\xf4\xbbFg\xc8av\x00\x83$\xfci\x1e$v\xde\xf8\xee\x1c(\xd5V\xd5\x00\x00\x01\x00P\xbf,\x1f?\xe7T\xe6\x8eu\xb0.\x8fNLZt\xb1\xb2Z\x965\xee#C\x86\x98\x17\xfd\x046\xaaf\r\x88\x10\t\x84b\x1f;\xec\xd8\xfa\xf5,\xce\xe8n\xb5\x82\xc9\xd8\x81)\xf1\xc5-D\xfb{\x0b\x9c\xfb\x90P\x80\r\x95\xa9e\xaa,\x11E\xe1(\\\xfc8\x9aM\x9c\xce\xbf\r\xc7\x19B\xf67\xf2\x08;\xb8\xfc\x029\x8e\xb5\x81\n\xbdv\xder\xd5\xc3\x19\n\xb9O\x19\x1c\xa8\x02\x0c\x84\xbd\xde\xec\xd0\xc9E$\x9b\x12Bv\x1a\xe0\x95\rs \x8f\x9d\x10\x9d\x16\x85\x12\xf8@LP\xb2)\xbae\x15\xf1\x12\x11\x80\xe8\x87\x1a\xf3\x91\xda\xc3/Yt\xad\xa5\xe9\xdf\x01G]\x83\x0f3\xe8\x97\xec\xe6\xe5\'w\x94\x97\xf1\xde;.R\x89\xd5\xce\xf4\xa1\xc6\x10[q1\xf1B\xfb29*\x06\xca\xb0\xb1\x0e\xa5\xc8\x94A\x95\x16\x0b\xe7\xd2\xa7if\x11\xa57\xb5n\x01\xee\\_\xba\xe2\xc9\xd5\xe7\xe1\x1f\xbe\xac\x9b\x07\x012\xabY<\xc1+\xe9\xb2l\x8e[n\xd3'
get_tpm_metadata: <class 'bytes'>
  File "/usr/local/bin/keylime_agent", line 11, in <module>
    load_entry_point('keylime==1.2', 'console_scripts', 'keylime_agent')()
  File "/usr/local/lib/python3.7/site-packages/keylime-1.2-py3.7.egg/keylime/cloud_agent.py", line 493, in main
    (ek,ekcert,aik,ek_tpm,aik_name) = tpm.tpm_init(self_activate=False,config_pw=config.get('cloud_agent','tpm_ownerpassword')) # this tells initialize not to self activate the AIK
  File "/usr/local/lib/python3.7/site-packages/keylime-1.2-py3.7.egg/keylime/tpm1.py", line 580, in tpm_init
    self.__create_aik(self_activate)
  File "/usr/local/lib/python3.7/site-packages/keylime-1.2-py3.7.egg/keylime/tpm1.py", line 352, in __create_aik
    handle = self.__load_aik()
  File "/usr/local/lib/python3.7/site-packages/keylime-1.2-py3.7.egg/keylime/tpm1.py", line 376, in __load_aik
    retDict = self.__run("loadkey -hp 40000000 -ik %s"%(inFile.name))
  File "/usr/local/lib/python3.7/site-packages/keylime-1.2-py3.7.egg/keylime/tpm1.py", line 170, in __run
    raise Exception("Command: %s returned %d, expected %d, output %s"%(cmd,code,expectedcode,retout))
Exception: Command: loadkey -hp 40000000 -ik /tmp/tmpkj6hw_m0 returned 250, expected 0, output [b'Decryption failure from TPM_LoadKey2\n']
```

**Comment by lukehinds** [comment]
Going go open this for review now as its close to done.

**Comment by lukehinds** [comment]
I think I know the issue with the tenant, an import asyncio might be missing from a false positive. I can fix that shortly.

**Comment by lukehinds** [comment]
@nabilschear , I am still getting the same stack trace:

```
  File "/usr/local/bin/keylime_agent", line 11, in <module>
    load_entry_point('keylime==1.2', 'console_scripts', 'keylime_agent')()
  File "/usr/local/lib/python3.7/site-packages/keylime-1.2-py3.7.egg/keylime/cloud_agent.py", line 491, in main
    (ek,ekcert,aik,ek_tpm,aik_name) = tpm.tpm_init(self_activate=False,config_pw=config.get('cloud_agent','tpm_ownerpassword')) # this tells initialize not to self activate the AIK
  File "/usr/local/lib/python3.7/site-packages/keylime-1.2-py3.7.egg/keylime/tpm1.py", line 577, in tpm_init
    self.__create_aik(self_activate)
  File "/usr/local/lib/python3.7/site-packages/keylime-1.2-py3.7.egg/keylime/tpm1.py", line 349, in __create_aik
    handle = self.__load_aik()
  File "/usr/local/lib/python3.7/site-packages/keylime-1.2-py3.7.egg/keylime/tpm1.py", line 373, in __load_aik
    retDict = self.__run("loadkey -hp 40000000 -ik %s"%(inFile.name))
  File "/usr/local/lib/python3.7/site-packages/keylime-1.2-py3.7.egg/keylime/tpm1.py", line 170, in __run
    raise Exception("Command: %s returned %d, expected %d, output %s"%(cmd,code,expectedcode,retout))
Exception: Command: loadkey -hp 40000000 -ik /tmp/tmpk08go2kd returned 250, expected 0, output [b'Decryption failure from TPM_LoadKey2\n']
```

I don't even need the tenant for this, it happens during keylime_agent start up

**Comment by lukehinds** [comment]
> can you try deleting the tpmdata.yml file and trying again? i think you have a mismatch between your tpm and the stored data that keylime keeps around

Is that in /var/lib/keylime/secure ?

if yes, I unmount and delete the whole lot before running

**Comment by lukehinds** [comment]
Got past the issue @nabilschear 

woking on this one now:

```
2019-08-29 12:16:01.929 - keylime.tpm1 - ERROR - Error encrypting AIK: write to closed file
2019-08-29 12:16:01.930 - keylime.tpm1 - ERROR - write to closed file
  File "/usr/local/lib/python3.7/site-packages/keylime-1.2-py3.7.egg/keylime/tpm1.py", line 401, in encryptAIK
    pubaikFile.write(pubaik.encode('utf-8'))
ValueError: write to closed file
```

**Comment by lukehinds** [comment]
2019-08-29 14:36:10.329 - keylime.cloudagent - INFO - Agent UUID: D432FBB3-D2F1-4A97-9EF7-75BD81C00000
2019-08-29 14:36:10.358 - keylime.registrar_client - INFO - Agent registration requested for D432FBB3-D2F1-4A97-9EF7-75BD81C00000
2019-08-29 14:36:10.367 - keylime.tpm1 - ERROR - Error decrypting AIK: Command: activateidentity -hk b'2EE99B07' -pwdo keylime -pwdk Lya5nKegiow9hbfpsNrQ -if /tmp/tmp77xlhj9p -ok /var/lib/keylime/secure/tmp9n9ouwzr returned 33, expected 0, output [b"ActivateIdentity returned error 'Decryption failure' (0x21).\n"]
2019-08-29 14:36:10.367 - keylime.tpm1 - ERROR - Command: activateidentity -hk b'2EE99B07' -pwdo keylime -pwdk Lya5nKegiow9hbfpsNrQ -if /tmp/tmp77xlhj9p -ok /var/lib/keylime/secure/tmp9n9ouwzr returned 33, expected 0, output [b"ActivateIdentity returned error 'Decryption failure' (0x21).\n"]
  File "/usr/local/lib/python3.7/site-packages/keylime-1.2-py3.7.egg/keylime/tpm1.py", line 462, in activate_identity
    retDict = self.__run(command,outputpaths=secpath)
  File "/usr/local/lib/python3.7/site-packages/keylime-1.2-py3.7.egg/keylime/tpm1.py", line 170, in __run
    raise Exception("Command: %s returned %d, expected %d, output %s"%(cmd,code,expectedcode,retout))
Exception: Command: activateidentity -hk b'2EE99B07' -pwdo keylime -pwdk Lya5nKegiow9hbfpsNrQ -if /tmp/tmp77xlhj9p -ok /var/lib/keylime/secure/tmp9n9ouwzr returned 33, expected 0, output [b"ActivateIdentity returned error 'Decryption failure' (0x21).\n"]
  File "/usr/local/bin/keylime_agent", line 11, in <module>
    load_entry_point('keylime==1.2', 'console_scripts', 'keylime_agent')()
  File "/usr/local/lib/python3.7/site-packages/keylime-1.2-py3.7.egg/keylime/cloud_agent.py", line 543, in main
    retval = registrar_client.doActivateAgent(registrar_ip,registrar_port,agent_uuid,key)
  File "/usr/local/lib/python3.7/site-packages/keylime-1.2-py3.7.egg/keylime/registrar_client.py", line 163, in doActivateAgent
    'auth_tag': crypto.do_hmac(base64.b64decode(key),agent_id),
  File "/usr/lib64/python3.7/base64.py", line 80, in b64decode
    s = _bytes_from_decode_data(s)
  File "/usr/lib64/python3.7/base64.py", line 46, in _bytes_from_decode_data
    "string, not %r" % s.__class__.__name__) from None
TypeError: argument should be a bytes-like object or ASCII string, not 'bool'

**Comment by nabilschear** [comment]
ok @lukehinds this was working for me on my system.  Though, I tested with openssl, had to add simplejson to the setup.py, and turned on DISABLE_EK_CERT_CHECK_EMULATOR in common.  Also mount secure doesn't work inside of a docker container.  SO i set it false in common.py

**Comment by lukehinds** [comment]
tpm1 looks good @nabilschear

**Comment by lukehinds** [comment]
@nabilschear , could you test cfssl, it should be working in commit f4f4d86290e46db137c50ecd026fcecbfa3406e2 , but I am not sure how to test a revocation.

File: keylime/crypto.py (hunk 6 above)
Comment (jetwhiz) [reply keylime/crypto.py]: Do you know if this needs to be updated, @nabilschear?

Comment (nabilschear) [reply keylime/crypto.py]: the `[:-1]` strips out the \x80 char after removing all the zeros.

**Comment by nabilschear** [comment]
i've got tpm1, revocation, and cfssl all working reasonably.  Can someone else give it a try?

**Comment by lukehinds** [comment]
@nabilschear @jetwhiz 

I think we might be there now! 

https://www.youtube.com/watch?v=eMhDQFLwrAA

The only last thing to close down is the following:

https://github.com/keylime/keylime/pull/126#pullrequestreview-284098919

I also need to override or force this branch or PR over the conflicting files in master..Any ideas on the best way of achieving this?

**Comment by lukehinds** [comment]
> KeyError: '7 is not registered'

This is fixed with b67f1666f0a0bb83f611207ab08055b87dbc09e3

**Comment by lukehinds** [comment]
squashed commits.

we just need to find away of overwriting the conflicts. any ideas here welcome.

**Comment by jetwhiz** [comment]
For some reason the nodes never show their status as going past START for me, though they do enter the quote loop successfully.  That is, `CloudAgent_Operational_State` gets stuck at `START` and never progresses.  Not sure if this is an issue with the database not updating correctly?

**Comment by lukehinds** [comment]
ok, that should be all of the conflicts resolved.

**Comment by lukehinds** [comment]
> For some reason the nodes never show their status as going past START for me, though they do enter the quote loop successfully. That is, `CloudAgent_Operational_State` gets stuck at `START` and never progresses. Not sure if this is an issue with the database not updating correctly?

I don't believe I see this. Do you have some recreation steps?

I will capture a video and also try a fresh install. Which Linux dist were you using, perhaps python version as well?

**Comment by nabilschear** [comment]
I also saw the issue with operational state in my testing.  i also suspect a database issue

**Comment by lukehinds** [comment]
> I also saw the issue with operational state in my testing. i also suspect a database issue

This is what I see:

https://drive.google.com/file/d/1edhSlA-nJAAHsuAfpSarJ5pdhKWu-TzV/view?usp=sharing

How are you checking this, just a SQL query or something fails?

**Comment by nabilschear** [comment]
you can do `-c status`

File: keylime/tenant_webapp.py [hunk 8]
```diff
@@ -528,105 +538,106 @@ def post(self):
...
     for uri in dataset_uris:
         fpos = uri.find(",")
-        if fpos == -1: 
+        if fpos == -1:
             return None
-        
+
         try:
-            data.append(base64.b64decode(uri[fpos:]))
+            data.append(base64.b64decode(uri[fpos:])).decode('utf-8'))
```
jetwhiz: This got messed up somehow -- it should be: 

````python
data.append(base64.b64decode(uri[fpos:]).decode('utf-8'))
````

It looks like double parenthesis got added in here: `uri[fpos:])).decode(`

Comment (lukehinds) [reply keylime/tenant_webapp.py]: done!

**Comment by lukehinds** [comment]
@nabilschear / @jetwhiz 

https://github.com/keylime/keylime/pull/126/commits/461d726bc0a48516ac61ece8bfd76b588bb5867a >

[status](https://files.gitter.im/keylime-project/community/JVBj/image.png)

**Comment by jetwhiz** [comment]
@nabilschear I'm getting stuck on trying to get libreswan to work.  it gets installed and configured but doesn't intercept any traffic that's supposed to be ipsec.  racoon appears to be working well, though now.

**Comment by lukehinds** [comment]
@nabilschear / @jetwhiz what's your thoughts on this patch being merge ready now? Could the libreswan connection problems be worked on post merge.?

**Comment by jetwhiz** [comment]
The issues I'm running into aren't related to the python3 conversion, it's mostly just that the `whack` command of ipsec never returns anything.  

Because of that, I haven't been able to complete this portion of the python3 conversion: 

https://github.com/keylime/keylime/blob/2d151bc17bdeac03875ce6ea0642046be7e0c4ac/auto-ipsec/libreswan/src/local_action_crashsa.py#L51-L79

Everything else appears to be working, though.

**Comment by lukehinds** [comment]
> The issues I'm running into aren't related to the python3 conversion, it's mostly just that the `whack` command of ipsec never returns anything.
> 
> Because of that, I haven't been able to complete this portion of the python3 conversion:
> 
> https://github.com/keylime/keylime/blob/2d151bc17bdeac03875ce6ea0642046be7e0c4ac/auto-ipsec/libreswan/src/local_action_crashsa.py#L51-L79
> 
> Everything else appears to be working, though.

Thanks @jetwhiz 

I recommend we merge this patch then, and anything we have missed we can fix post.

I will do it by lazy consensus, so unless anyone objects I will merge in the morning (Tuesday)!

File: keylime/registrar_client.py [hunk 9]
```diff
@@ -92,35 +97,33 @@ def getAIK(registrar_ip,registrar_port,agent_id):
...
+        if response.status != 200:
+            logger.critical("Error: unexpected http response code from Registrar Server: %s"%str(response.status))
+            keylime_logging.log_http_response(logger,logging.CRITICAL,response_body)
+            return None
+
         if "results" not in response_body:
-            logger.critical("Error: unexpected http response body from Registrar Server: %s"%str(response.status_code))
-            return None 
-        
+            logger.critical("Error: unexpected http response body from Registrar Server: %s"%str(response.statu))
```
jetwhiz: It looks like this got cut off

lukehinds: well spotted, will patch that

keylime/ima.py
```diff
@@ -250,91 +255,92 @@ def process_whitelists(wl_data, excl_data):
         tmp = whitelist.get(path,[])
         tmp.append(fhash)
         whitelist[path]=tmp
-    
+
     if whitelist.get('boot_aggregate',None) is None:
         logger.warning("No boot_aggregate value found in whitelist, adding an empty one")
-        whitelist['boot_aggregate'] = [START_HASH.encode('hex')]
-    
+        whitelist['boot_aggregate'] = ['0000000000000000000000000000000000000000']
```
jetwhiz: Should we switch this back to being a variable, instead of hard-coded?

lukehinds: I will need to test this first, I was running into some issues before. If it works ok for you and want to push a patch, go for it. I am just fixing up CI on another branch.

PR #134, opened 2019-07-18
Add systemd service files for keylime verifier, agent, registrar

### LINKED ISSUE #60: Keylime systemd files
Was thinking it might be useful to have systemd service files for the verifier, tenant and registrar.

mrcdb:
In case of single-host deployment, ensures that verifier and
registrar services run before the agent

Fixes #60

### DIFF
--- a/services/keylime_agent.service
+++ b/services/keylime_agent.service
@@ -1,0 +1,9 @@
+[Unit]
+Wants=keylime_verifier.service keylime_registrar.service
+After=network.target keylime_verifier.service keylime_registrar.service
+
+[Service]
+ExecStart=/usr/local/bin/keylime_agent
+
+[Install]
+WantedBy=default.target
--- a/README.md
+++ b/README.md
@@ -410,6 +410,18 @@ The revocation key will be automatically created by the tenant the first time
 you use the CA with keylime.  Currently the CRL is only written back to the CA
 directory, unless IPsec configuration is being used (see [Additional Reading](#additional-reading)).
 
+## Systemd service support
+
+The directory `services/` includes `systemd` service files for the verifier,
+agent and registrar.
+
+You can install the services with the following command:
+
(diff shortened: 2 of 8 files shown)


services/keylime_agent.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
+Wants=keylime_verifier.service keylime_registrar.service
+After=network.target keylime_verifier.service keylime_registrar.service
+
+[Service]
+ExecStart=/usr/local/bin/keylime_agent
```
lukehinds: Looks great and the install script worked well (thanks for your contribution too!)

The only thing I found was the path is different for me.

I am on Fedora 30 and used `python setup.py install` 

This resulted in:

```
Jul 18 14:16:19 localhost.localdomain systemd[4798]: keylime_verifier.service: Failed to execute command: No such file or directory
Jul 18 14:16:19 localhost.localdomain systemd[4798]: keylime_verifier.service: Failed at step EXEC spawning /usr/local/bin/keylime_verifier
```

I edited the service file and changed the path to `/usr/bin/keylime_verifier` and it works well.

I wonder if this is linux distribution specific or the `installer.sh`

Which Linux dist did you test this against, and did you use setuptools or the `installer.sh`? I think once we iron this out, the patch will be good to merge!

lukehinds: > I can update the services installer by checking first where are located the keylime_agent, keylime_verifier and keylime_registrar on the target system and then update the service files accordingly with some sed or similar tools.

I think that's a good idea... have a condition statement to check the file exists and then set the path accordingly. I would be happy with that!

mrcdb:
I just slightly modified the service installation script and the service files to set the path to keylime scripts (if they are available). The service files are not modified inside the repository to avoid creating unstaged changes, but they are directly stored in systemd directory via `sed`.

PR #147, opened 2019-09-10
Implements new docker images for CI

lukehinds:
Now that Python 3 support is merged, this change points towards
the updated v500 docker hub tags which contain images built for
Python 3 support.


lukehinds:
@jetwhiz 

I might have to switch to relative imports so we can get travis working:

https://www.python.org/dev/peps/pep-0328/

Before:

`from keylime import common`

After

`from . import common`

This will be for files that are in `keylime` importing from the same directory.

lukehinds:
@jetwhiz https://travis-ci.org/keylime/keylime/jobs/583224883#L616

Issue #141, opened 2019-07-26
Set all string declarations using f-strings


lukehinds:
Once Python 3 support is in, we should port string declaration to using f-strings.

f-strings are much easier to read. for example:

```
>>> name = "Eric"
>>> age = 74
>>> f"Hello, {name}. You are {age}."
'Hello, Eric. You are 74.'
```

Can be used in function calls:

```
>>> def to_lowercase(input):
...     return input.lower()

>>> name = "Eric Idle"
>>> f"{to_lowercase(name)} is funny."
'eric idle is funny.'
```

There are also much faster then `%s` and `str.format()`

This would make a line such as the following, much easier to read:

https://github.com/keylime/keylime/blob/d46237141be8585d337518ec95f91953c637afff/test/test_restful.py#L559

More details here: https://www.python.org/dev/peps/pep-0498/

amylily1011:
This one sounds like fun! May I take this issue, @lukehinds ?

lukehinds:
Hi @amylily1011 yes please!

I would wait for the Python 3 patch to merge first though, as f-strings only work in Python 3 (we currently working on Python 2!)

amylily1011:
I will keep an eye on #32 , when it gets merged.

amylily1011:
Hey @lukehinds, I got a head start on the f-string on your python-3 branch. It's still WIP, but really fun, so far. I expect to be done with this sometime next week, (I think), how do I push it to your branch?

lukehinds:
Hi @amylily1011

Good question, you should be able to add a commit on top of my patch. If you like we can work together doing this over gitter when I return (about one week). Or if you want I can push the latest and we then merge yours on top (as I expect your work will have less merge conflicts to fix up) 


Thank you for working on this, really appreciate your contributions!

amylily1011:
@lukehinds 
Sure! Let's do this over gitter, I should be done then too.

lukehinds:
hi @amylily1011 - The Python 3 stuff is almost complete now, all the performance issues are resolved. Lets merge that and I can help you layer these on top. That way we can get CI to pass too.

amylily1011:
Hey @lukehinds sounds good! I'm pretty much done with the f-string as well.

There is only one line that I'm not sure how to turn it into the f-string:

`logger.error("IMA ERRORS: template-hash %d fnf %d hash %d good %d"%tuple(errs))`

because it calls the tuple. 

So I left it for now :P

But I'm ready when you're ready.

[...]

PR #153, opened 2019-09-13
Set all string declarations using f-strings

amylily1011:
Resolves: #141


lukehinds:
hi @amylily1011 

thanks for your work, we have a few things to fix up, but surprisingly a small amount for such a voluminous change: https://travis-ci.org/keylime/keylime/builds/584610418?utm_source=github_status&utm_medium=notification

I can go over this with you on Monday!

amylily1011:
Sure @lukehinds !
I'll be online around 4pm BKK time.

amylily1011:
I got one error in the test, but it wasn't relevant to the fstring. So I pushed the new commit for now, so you can have a look at it :)

lukehinds:
Thanks amy, will try to take a look tomorrow.

lukehinds:
Restarting CI Build to see if failures still present.

keylime/ca_util.py
```diff
@@ -56,8 +56,8 @@
 elif common.CA_IMPL=='openssl':
     from keylime import ca_impl_openssl as ca_impl
 else:
-    raise Exception("Unknown CA implementation: %s"%common.CA_IMPL)
-from M2Crypto import X509, EVP, BIO
+    raise Exception(f'Unknown CA implementation: {common.CA_IMPL}')
+    from M2Crypto import X509, EVP, BIO
```
lukehinds: Found the issue making CI Fail.

Line 60 is indented which means the import does not happen (it only happens as part of the exception occurs..."raise Exception"

Just bring the 4 from back 4 spaces and its all good.

amylily1011: Looking at it right now :)

lukehinds:
Let me know if you need some help with the conflicting files.

amylily1011:
I think I found something I might have missed.

Issue #146, opened 2019-09-06
Move installer.sh to using packages


lukehinds:
I recommend we start using packages for the installer.

They are more stable and simplifies the installation a lot!

Packages also work better for Fedora, as the packages include SELinux labels (which means being able to run without needing to disabled SE). I also believe the Ubuntu tpm2-tools package has the needed code release upstream.

We should instead move tools / tss etc compiling to a CI job 

@nabilschear @jetwhiz  agree / disagree?

nabilschear:
i think that would be great, but a lot of work to support for many distros.  I like the idea of keeping the installer for our tier2 distro support and then focus on getting fedora to work just with rpms.

lukehinds:
> i think that would be great, but a lot of work to support for many distros. I like the idea of keeping the installer for our tier2 distro support and then focus on getting fedora to work just with rpms.

I am sure I could do something, I could add a flag to build the tools instead of using packages.

PR #157, opened 2019-09-20
Move installer.sh to using packages

lukehinds:
This change moves the installer to using OS packages, instead
of compiling.

It also adds logic to handle tpm2-abrmd working with an emulator
and provides a message on selinux (if a rhel based OS). This
message is to ask for SELinux to be disabled, as it blocks the
emulator and tpm2-abrmd from communication. Note SELinux disabling
is not required when tpm2-abrmd communicates with a hardware
TPM.

An emulator security warning is also displayed at the end of
installation about production use.

Resolves: #146

# pending items:

- [ ] Ubuntu build for tpm2-abrmd needs a specific branch
- [ ] SELInux labels not working for RHEL family.

### DIFF
--- a/installer.sh
+++ b/installer.sh
@@ -383,7 +383,7 @@ if [[ "$TPM_VERSION" -eq "1" ]] ; then
 elif [[ "$TPM_VERSION" -eq "2" ]] ; then
     echo
     echo "=================================================================================="
-    echo $'\t\t\t\tBuild and install tpm2-tss'
+    echo $'\t\t\t\tInstall tpm2-tss'
     echo "=================================================================================="
    $PACKAGE_MGR -y install $TPM2_SOFTWARE
     if [[ $? > 0 ]] ; then
--- a/installer.sh
+++ b/installer.sh
@@ -434,6 +434,7 @@ elif [[ "$TPM_VERSION" -eq "2" ]] ; then
         fi
         systemctl enable tpm2-abrmd
         systemctl start tpm2-abrmd
+    fi
 
     if [[ "$TPM_SOCKET" -eq "1" ]] ; then
         echo
@@ -539,3 +540,4 @@ if [[ -n "$(command -v dnf)" ]] || [[ -n "$(command -v yum)" ]]; then
         echo "Please also be mindful, that an emulator is not a secure option and should not be used in production!"
     fi
 fi
+
--- a/installer.sh
+++ b/installer.sh
@@ -43,6 +43,13 @@ MIN_PYZMQ_VERSION="14.4"
 MIN_PYCRYPTODOMEX_VERSION="3.4.1"
 MIN_GO_VERSION="1.8.4"
 
+# Determine distibution (using systemd standard `os-release`):
+if [ -f /etc/os-release ]; then
+        . /etc/os-release
+    else
+        echo "Not able to determine your OS or Distribution"
+        exit 1
+fi
 
 # Check to ensure version is at least minversion
 version_checker () {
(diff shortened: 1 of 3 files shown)


installer.sh
```diff
@@ -67,19 +67,22 @@ if [[ -n "$(command -v dnf)" ]]; then
...
     PYTHON_PIPS="pycryptodomex tornado pyzmq m2crypto"
     BUILD_TOOLS="openssl-devel libtool make automake m4 libgcrypt-devel autoconf autoconf-archive libcurl-devel libstdc++-devel uriparser-devel dbus-devel gnulib-devel doxygen"
+    TPM2_SOFTWARE="tpm2-tss tpm2-tools tpm2-abrmd"
 elif [[ -n "$(command -v apt-get)" ]]; then
     PACKAGE_MGR=$(command -v apt-get)
     PYTHON_PREIN="git patch"
     PYTHON_DEPS="python3 python3-pip python3-dev python3-setuptools python3-zmq python3-tornado python3-simplejson python3-requests gcc g++ libssl-dev swig python3-yaml wget"
     PYTHON_PIPS="pycryptodomex m2crypto"
     BUILD_TOOLS="build-essential libtool automake pkg-config m4 libgcrypt20-dev uthash-dev autoconf autoconf-archive libcurl4-gnutls-dev gnulib doxygen libdbus-1-dev"
+    TPM2_SOFTWARE="tpm2-tss tpm2-tools tpm2-abrmd"
```
jetwhiz: I don't think these versions support the features we need (our Keylime-specific modifications). 

For tpm2-tools, our changes were released in [3.2](https://github.com/tpm2-software/tpm2-tools/releases/tag/3.2.0) and are also in the latest/master/4.X, while even the latest available in Ubuntu is 3.1.x. 

This will break Ubuntu below, since we're removing the possibility to build from source.

lukehinds: Hm, good point. I am just really not into compiling code in an installer and who knows how long it will take for Debian to package the right version.

Have you checked against disco / buster?

lukehinds: https://packages.debian.org/buster/tpm2-tools

installer.sh
```diff
@@ -520,3 +487,13 @@ if [[ "$TARBALL" -eq "1" ]] ; then
     fi
     ./make_agent_bundle_tarball.sh $TAR_BUNDLE_FLAGS
 fi
+
+echo "=================================================================================="
+echo $'\t\t\t\tInstallation complete!'
+echo "=================================================================================="
+if [[ -n "$(command -v dnf)" ]] || [[ -n "$(command -v yum)" ]]; then
+    if [[ "$TPM_VERSION" -eq "2" ]] && [[ "$TPM_SOCKET" -eq "1" ]] ; then
+        echo "As you have selected to run an emulator, you will need to disable SELinux"
```
dawud: Set to permissive instead?
If the emulator ever works in an SELinux enabled system, it's easier to set back to enforce than having to relabel from disabled state.

lukehinds: That's a good point @dawud. thanks..will make the change.

jetwhiz: Are we sure this is needed?  I've run the emulator + abrmd in Fedora before without needing to disable SELinux, but you need to install the policy: https://github.com/tpm2-software/tpm2-abrmd/blob/master/INSTALL.md#selinux

installer.sh
jetwhiz: Can we just build from source if they're using apt, otherwise use the package manager?

installer.sh
lukehinds: It always fails for me and I am pretty sure i found a denial.

lukehinds:
```
285. 09/20/2019 19:03:20 tpm2-abrmd system_u:system_r:tabrmd_t:s0 0 udp_socket create system_u:system_r:tabrmd_t:s0 denied 1009
286. 09/20/2019 19:03:20 tpm2-abrmd system_u:system_r:tabrmd_t:s0 0 udp_socket create system_u:system_r:tabrmd_t:s0 denied 1010
287. 09/20/2019 19:03:20 tpm2-abrmd system_u:system_r:tabrmd_t:s0 0 tcp_socket create system_u:system_r:tabrmd_t:s0 denied 1011
288. 09/20/2019 19:03:20 tpm2-abrmd system_u:system_r:tabrmd_t:s0 0 tcp_socket create system_u:system_r:tabrmd_t:s0 denied 1012
```

I will try and get the label applied. 

I need to dig more as well, as javier had a `tpm2-abrdmd-selinux` package as a depency to the main abrmd package.

lukehinds: @jetwhiz I just noticed we don't even build tpm2-abrmd:

https://github.com/keylime/keylime/blob/01785a25077fbb6b2c2220395a3fd99300583d43/installer.sh#L394-L419

lukehinds:
pushed 007576a6b6eb45dc4806d1e119588b411062c662 , but not had a chance to test yet.

installer.sh
jetwhiz: Yeah we don't install it by default with the installer, since I wasn't able to get a cross-platform, automated way to get it working.  The instructions are there to get it working in Ubuntu, and the README also shows how to install it.  It seems to be very finicky, though, especially on Fedora.

installer.sh
```diff
@@ -117,7 +118,7 @@ while getopts ":shotkmp:" opt; do
             echo $'-k \t\t\t\t Download Keylime (stub installer mode)'
             echo $'-o \t\t\t\t Use OpenSSL instead of CFSSL'
             echo $'-t \t\t\t\t Create tarball with keylime_agent'
-            echo $'-m \t\t\t\t Use modern TPM 2.0 libraries (vs. TPM 1.2)'
+            echo $'-m \t\t\t\t Use legacy TPM version 1.2'
```
jetwhiz: The logic is inverted here now, so this will do the opposite of what is listed.  Also, the "-m" flag no longer makes sense (modern vs. legacy).

installer.sh
lukehinds: Ack, makes sense. i remember having to do it differently with Fedora as well. It should be fine now, as its ubuntu specific.

installer.sh
lukehinds: Sorry, that's a long day mistake from me.

I was going to change the default to being tpm2.0, but then thought it better for another patch so we can discuss it first. I will revert that change.

installer.sh
```diff
@@ -380,18 +381,59 @@ if [[ "$TPM_VERSION" -eq "1" ]] ; then
...
+        ./configure --prefix=/usr/local
+        make
+        make install
+        popd # tpm
+        echo "=================================================================================="
+        echo $'\t\t\t\tBuild and install tpm2-abrmd'
+        echo "=================================================================================="
+        useradd --system --user-group tss
+        git clone https://github.com/tpm2-software/tpm2-abrmd.git tpm2-abrmd
+        pushd tpm2-abrmd
```
jetwhiz: We need to `git checkout` the correct version of abrmd that is compatible with the TSS used above.  Note that this block of code was commented out and was intended to be used as a guide.  It might need changes to get it working, and definitely won't work on Fedora (since we are not installing the SELinux policy).

installer.sh
```diff
@@ -377,59 +381,59 @@ if [[ "$TPM_VERSION" -eq "1" ]] ; then
...
```
jetwhiz: We need to git checkout the correct version of abrmd that is compatible with the TSS used above. Note that this block of code was commented out and was intended to be used as a guide. It might need changes to get it working, and definitely won't work on Fedora (since we are not installing the SELinux policy).

lukehinds: agree.

I am throwing these changes up, but yet to test them. I will be running full tests before letting these even get close to being merge worthy (against both Ubuntu and Fedora)

I am just discussing with Javier what the issue might be with SELinux (looks like it might be upstream).

PR #165, opened 2019-09-25
change async calls to support python 3.6

nabilschear:
use this method to call coroutines so that we don't need python3.7 and 3.6 is sufficient.  this means that compatibility with ubuntu 18.04 is easier


lukehinds:
thanks @nabilschear , will test this when I get back from travelling.

This will also be good for rhel 8

lukehinds:
I will sneak this one in for 5.0.0 as its quite key for Python coverage. Just going to test now.

Issue #175, opened 2019-10-01
shebang changes required for rpmlint


lukehinds:
In order for an rpm to be successfully reviewed / accepted, it needs to pass `rpmlint` checks. Currently we have a mix of files with different shebangs, which is causing failures.

Keylime currently has the following errors:

```
keylime.noarch: E: non-executable-script /usr/lib/python3.7/site-packages/keylime/ca_util.py 644 /usr/bin/python3 
keylime.noarch: E: non-executable-script /usr/lib/python3.7/site-packages/keylime/cloud_verifier_common.py 644 /usr/bin/python 
keylime.noarch: E: wrong-script-interpreter /usr/lib/python3.7/site-packages/keylime/httpclient_requests.py /usr/bin/env python3
keylime.noarch: E: non-executable-script /usr/lib/python3.7/site-packages/keylime/httpclient_requests.py 644 /usr/bin/env python3
keylime.noarch: E: non-executable-script /usr/lib/python3.7/site-packages/keylime/provider_platform_init.py 644 /usr/bin/python 
keylime.noarch: E: non-executable-script /usr/lib/python3.7/site-packages/keylime/provider_registrar.py 644 /usr/bin/python 
keylime.noarch: E: non-executable-script /usr/lib/python3.7/site-packages/keylime/provider_vtpm_add.py 644 /usr/bin/python 
keylime.noarch: E: non-executable-script /usr/lib/python3.7/site-packages/keylime/registrar.py 644 /usr/bin/python 
keylime.noarch: E: non-executable-script /usr/lib/python3.7/site-packages/keylime/registrar_common.py 644 /usr/bin/python 
keylime.noarch: E: wrong-script-interpreter /usr/lib/python3.7/site-packages/keylime/revocation_notifier.py /usr/bin/env python
keylime.noarch: E: non-executable-script /usr/lib/python3.7/site-packages/keylime/revocation_notifier.py 644 /usr/bin/env python
keylime.noarch: E: non-executable-script /usr/lib/python3.7/site-packages/keylime/secure_mount.py 644 /usr/bin/python 
keylime.noarch: E: non-executable-script /usr/lib/python3.7/site-packages/keylime/serve_uuid.py 644 /usr/bin/python 
keylime.noarch: E: non-executable-script /usr/lib/python3.7/site-packages/keylime/tenant.py 644 /usr/bin/python 
[...]

PR #176, opened 2019-10-01
Set standard shebang and only for exec files

lukehinds:
This change seeks to set a common shebang for files that are
executable only (e.g. set as entry_points in setup.py)

It is needed to pass `rpmlint` and allow package acceptence
upstream

Resolves: #175


lukehinds:
Thanks @nabilschear

Issue #163, opened 2019-09-25
tpm2-tools in RHEL


lukehinds:
To track issue in RHEL

https://bugzilla.redhat.com/show_bug.cgi?id=1755551

jetwhiz:
Assuming we can pull tpm2-tools up to 3.2 in RHEL8, should we try targeting EPEL8 for Keylime so we can install/release it more easily in RHEL as well as Fedora?

lukehinds:
https://bugzilla.redhat.com/show_bug.cgi?id=1725714

lukehinds:
RHEL has been updated now

PR #177, opened 2019-10-01
Implement multi OS and tpm2 software packages

### LINKED ISSUE #146: Move installer.sh to using packages

nabilschear:
updates to the installer to make it work on fedora 30, 29 and ubuntu 18.04.  there's partial support for centos7, but only for TPM1.2  TPM2.0 build is broken

fixes #146


nabilschear:
something went amiss with all the squashing i think

installer.sh
```diff
@@ -59,27 +59,38 @@ confirm_force_install () {
...
+    	NEED_BUILD_TOOLS=1
+    fi
 elif [[ -n "$(command -v yum)" ]]; then
     PACKAGE_MGR=$(command -v yum)
     $PACKAGE_MGR -y install epel-release
     PYTHON_PREIN="python36 python36-devel python36-setuptools python36-pip git wget patch openssl"
-    PYTHON_DEPS="gcc gcc-c++ openssl-devel swig python36-PyYAML python36-tornado python3-cryptography python36-simplejson python36-requests yaml-cpp-devel"
-    PYTHON_PIPS="tornado pyzmq m2crypto"
-    BUILD_TOOLS="openssl-devel libtool make automake m4 libgcrypt-devel autoconf autoconf-archive libcurl-devel libstdc++-devel uriparser-devel dbus-devel gnulib-devel doxygen"
+    PYTHON_DEPS="gcc gcc-c++ openssl-devel swig python36-PyYAML python36-tornado python36-simplejson python36-requests yaml-cpp-devel"
```
jetwhiz: Can we keep the `cryptography` library coming from the repos instead of pip?  

The current script in `master` has a typo -- this should be targeting `python36-cryptography` and not `python3-cryptography`, which is why this is broken on `master` right now.

installer.sh
```diff
@@ -522,3 +556,31 @@ if [[ "$TARBALL" -eq "1" ]] ; then
...
+    echo $'\t\t\t\tStart TPM emulator'
+    echo "=================================================================================="
+    # starts emulator and IMA stub at boot
+    cd $KEYLIME_DIR/ima_stub_service
+    ./installer.sh
+	if [[ -n "$(command -v service)" ]] ; then
+		service tpm_emulator restart
+	fi
+	if [[ "$TPM_VERSION" -eq "2" ]] ; then
+		echo 'export TPM2TOOLS_TCTI="mssim:port=2321"' >> $HOME/.bashrc && source $HOME/.bashrc
```
jetwhiz: It looks like there are a lot of mixed tabs and spaces for indentation -- can we go with all spaces?

jetwhiz:
I got this working on CentOS 7 by setting the `PKG_CONFIG_PATH` environment variable.  

Can you try this and confirm if it fixes the CentOS 7 issues on your end, too, @nabilschear ? 

````bash
export PKG_CONFIG_PATH=/usr/lib/pkgconfig/
````

Put this right before running `./configure` for the `tpm2-tools` build.

**Comment by lukehinds** [comment]
I don't know how I managed it, but I closed a PR with lots of installer changes here https://github.com/keylime/keylime/pull/157

This one is shaping up OK though. But there are few things need changing.

1. rhel and centos both will use 'yum' going forward, so package managers are no longer a reliable way to establish distribution:

https://github.com/keylime/keylime/blob/a42b3d37535cc581d152cb1e9ebd5c49dda906f9/installer.sh#L46-L52

And then:

https://github.com/keylime/keylime/blob/a42b3d37535cc581d152cb1e9ebd5c49dda906f9/installer.sh#L69-L106

2. We still have some issues with SELinux upstream

Others I will make in review or perhaps push another commit to this branch.

File: installer.sh (hunk 1 above)
Comment (lukehinds) [reply installer.sh]: +1

It would be good to avoid pip whenever we can going forward and stick with package managers (no reliance on an internet connection then as some operators run their own internal package repositories, especially in LABs).  Some of the big distributiors (rhel, opensuse etc), will also only support stuff they have packaged themselves (especially for security SLAs).

File: installer.sh (hunk 2 above)
Comment (nabilschear) [reply installer.sh]: k, i've fixed the settings in my editor to do 4 spaces.  i also replaced all the tabs

File: installer.sh (hunk 1 above)
Comment (nabilschear) [reply installer.sh]: the only one that's not packaged is m2crypto for py36

**Comment by nabilschear** [comment]
@lukehinds i recall seeing some of this but then i couldn't find it!  The OSes that I tested on were rhel/centos 7, ubuntu 18.04, fedora 29 and fedora 30.  I presume that rhel 8 should be on our list too?  

Can you help me merge your os detection stuff and other enhancements into this PR?

**Comment by nabilschear** [comment]
ok, i've updated various things and attempted to merge @lukehinds pr.  I've tested cent8, cent7, ubuntu 18.04, and fedora 29/30.  I've also tested the -so -mo -smo -ms options to cover 1.2, 2.0, emulators, and cfssl.

at this point the only platform not working is centos 8 with tpm2 and an emulator.  i'm getting this error:
```
2019-10-02 17:48:27.745 - keylime.tpm2 - INFO - Using legacy version of tpm2-tools
2019-10-02 17:48:27.772 - keylime.tpm2 - ERROR - Error encrypting AIK: Command: tpm2_makecredential -T none -e /tmp/tmpa0fbqswr -s /tmp/tmpqafjqo0w -n 000b1afb8d67fad9349d18ff3ddd03afd7d687ac70903b534c6298fb417e1af66111 -o /tmp/tmppa1lisxl returned 1, expected 0, output [b'\n', b'** (process:730): CRITICAL **: 17:48:27.771: failed to allocate dbus proxy object: Could not connect: No such file or directory\n', b'ERROR: tcti init allocation routine failed for library: "tabrmd" options: "(null)"\n', b'ERROR: Could not load tcti, got: "tabrmd"\n']
2019-10-02 17:48:27.772 - keylime.tpm2 - ERROR - Command: tpm2_makecredential -T none -e /tmp/tmpa0fbqswr -s /tmp/tmpqafjqo0w -n 000b1afb8d67fad9349d18ff3ddd03afd7d687ac70903b534c6298fb417e1af66111 -o /tmp/tmppa1lisxl returned 1, expected 0, output [b'\n', b'** (process:730): CRITICAL **: 17:48:27.771: failed to allocate dbus proxy object: Could not connect: No such file or directory\n', b'ERROR: tcti init allocation routine failed for library: "tabrmd" options: "(null)"\n', b'ERROR: Could not load tcti, got: "tabrmd"\n']
```

what's strange is that the agent is able to talk to the tpm for other operations (like take ownership etc.)  so it think that the env var config is right.  There's currently no abrmd in the installer (other than rpms)

**Comment by nabilschear** [comment]
ok, i think that my issues with abrmd or lackthereof were because i was doing this testing in containers and no systemd, so abrmd wasn't running properly.  @lukehinds or @jetwhiz can you test this installer in `-smo` mode with Cent/Rhel 8 and Fedora 30?  those will use abrmd.

File: installer.sh [hunk 3]
```diff
@@ -59,33 +59,78 @@ confirm_force_install () {
...
-   exit 1
+# Determine distibution (using systemd standard `os-release`):
+if [ -f /etc/os-release ]; then
+        . /etc/os-release
+    else
+        echo "Not able to determine your OS or Distribution"
+        exit 1
 fi
 
+NEED_BUILD_TOOLS=0
```
jetwhiz: We should define defaults for anything set below that isn't always set to a value.  For instance: 

````bash 
CENTOS7_TSS_FLAGS=
GOPKG=
HAS_GO_PKG=0
NEED_BUILD_TOOLS=0
NEED_PYTHON_DIR=
PYTHON_PIPS=
TPM2_TOOLS_PKGS=
````

File: installer.sh [hunk 4]
```diff
@@ -59,33 +59,78 @@ confirm_force_install () {
...
+        NEED_BUILD_TOOLS=1
+        $PACKAGE_MGR update
+    ;;
+
+    rhel | centos)
+        case "${VERSION_ID}" in 
+            7)
+                echo "${ID} ${VERSION_ID} selected."
+                PACKAGE_MGR=$(command -v yum)
+                $PACKAGE_MGR -y install epel-release
```
jetwhiz: Not sure if we care about changing this or not, but `$PACKAGE_MGR -y install epel-release` here (and below in version 8) will install the EPEL repository even if people only run "./installer.sh -h" to see what the installer options are.  Usually running help shouldn't impact the system.

jetwhiz:
> ok, i think that my issues with abrmd or lackthereof were because i was doing this testing in containers and no systemd, so abrmd wasn't running properly. @lukehinds or @jetwhiz can you test this installer in `-smo` mode with Cent/Rhel 8 and Fedora 30? those will use abrmd.

I think I'm running into similar selinux-based issues as @lukehinds was mentioning while trying to get this running on CentOS 8.  Was this a bug upstream in tpm2-abrmd, Luke?

lukehinds:
> > ok, i think that my issues with abrmd or lackthereof were because i was doing this testing in containers and no systemd, so abrmd wasn't running properly. @lukehinds or @jetwhiz can you test this installer in `-smo` mode with Cent/Rhel 8 and Fedora 30? those will use abrmd.
> 
> I think I'm running into similar selinux-based issues as @lukehinds was mentioning while trying to get this running on Fedora 30. Was this a bug upstream in tpm2-abrmd, Luke?

Yes, BZ here: https://bugzilla.redhat.com/show_bug.cgi?id=1754805

I might even try to fix this myself as it breaks all our installers (or at least necessitates manual intervention)

keylime/crypto.py
```diff
@@ -139,7 +139,7 @@ def strbitxor(a,b):
     retval = bytearray(len(b))
     for i in range(len(a)):
         retval[i] = a[i] ^ b[i]
-    return retval
+    return bytes(retval)
```
lukehinds: We need to check this does not break later versions (python 37 & >).

See @jetwhiz comment in gitter:

bytes-like is only supported in 2.5+ of python cryptography (https://github.com/pyca/cryptography/commit/3c68250ad9bfb275c760fcce4c72c78c99b57c34#diff-72412af902819141ea548cebf0335477), while python36-cryptography is running 2.3.  so we need to use bytes specifically on the older systems

installer.sh
nabilschear: great suggestion!

keylime/crypto.py
nabilschear: i've confirmed that this is fine on fedora 30 with python3.7 and cryptography 2.6.1

installer.sh
```diff
@@ -59,33 +68,77 @@ confirm_force_install () {
...
+            ;;
+            8)
+                echo "${ID} ${VERSION_ID} selected."
+                PACKAGE_MGR=$(command -v dnf)
+                NEED_EPEL=1
+                PYTHON_PREIN="python3 python3-devel python3-setuptools python3-pip"
+                PYTHON_DEPS="gcc gcc-c++ openssl-devel python3-yaml python3-requests swig python3-cryptography wget git"
+                PYTHON_PIPS="tornado==5.0.2 pyzmq m2crypto simplejson"
+                BUILD_TOOLS="git wget patch libyaml openssl-devel libtool make automake m4 libgcrypt-devel autoconf libcurl-devel libstdc++-devel dbus-devel"
+                TPM2_TOOLS_PKGS="tpm2-tss tpm2-tools tpm2-abrmd"
```
jetwhiz: Does CentOS/RHEL 8 have the correct version of tpm2-tools?  It looks like they're running 3.1.1 on mine (and we require 3.2+ to have the Keylime-specific features). 

I think we have to build from source on these until (hopefully) Luke's upstream request is fixed: https://github.com/keylime/keylime/issues/163

installer.sh
```diff
@@ -496,6 +584,9 @@ echo "==========================================================================
 echo $'\t\t\t\tInstall Keylime'
 echo "=================================================================================="
 cd $KEYLIME_DIR
+if [[ "$NEED_PYTHON_DIR" -eq "1" ]] ; then
```
jetwhiz: Is this really needed?  It seems something like this should be handled by something else ...

nabilschear: on my centos 8 box, keylime wouldn't install correctly without this.  I think @lukehinds told me about it?

installer.sh
nabilschear: yeah, i get 3.1.1. too . Maybe that's why it wasn't working.  It's pretty easy to switch it to build by hand.

installer.sh
```diff
@@ -59,33 +68,79 @@ confirm_force_install () {
...
+                PACKAGE_MGR=$(command -v dnf)
+                NEED_EPEL=1
+                PYTHON_PREIN="python3 python3-devel python3-setuptools python3-pip"
+                PYTHON_DEPS="gcc gcc-c++ openssl-devel python3-yaml python3-requests swig python3-cryptography wget git"
+                PYTHON_PIPS="tornado==5.0.2 pyzmq m2crypto simplejson"
+                BUILD_TOOLS="git wget patch libyaml openssl-devel libtool make automake m4 libgcrypt-devel autoconf libcurl-devel libstdc++-devel dbus-devel"
+                #TPM2_TOOLS_PKGS="tpm2-tss tpm2-tools tpm2-abrmd" TODO: still on 3.1.1 tpm2_tools 
+                NEED_BUILD_TOOLS=1
+                NEED_PYTHON_DIR=1
+                POWERTOOLS="--enablerepo=PowerTools install autoconf-archive"
```
jetwhiz: _PowerTools_ might be CentOS-only.  I can't tell how to install the RHEL _CodeReady Builder_ repo, though it sounds like it is different.

nabilschear: ok, i don't have an rhel 8 test system.  please update if you figure out how to get it on there.  I did smoke test this in centos 8 and it works.

installer.sh
nabilschear: committed code to move centos/rhel 8 to build tpm2-tools from source

installer.sh
jetwhiz: This could explain it: https://github.com/pypa/setuptools/pull/1431#issuecomment-406640252

nabilschear: i'm not sure what to do about this since the installer runs as root.  Hopefully it will just go away once we can package for centos 8.

jetwhiz: The majority of things in this script don't even need sudo privileges, so we could probably have it run without it and just sudo the commands that actually need root.  The downside here is that if a user has disabled sudo cache then they'll have to type their password 20 times.

installer.sh
```diff
@@ -43,6 +43,15 @@ MIN_PYZMQ_VERSION="14.4"
 MIN_PYCRYPTOGRAPHY_VERSION="2.1.4"
 MIN_GO_VERSION="1.8.4"
```
jetwhiz: We might want to consider bumping go's minimum to `1.11.13`, since `1.8.4` is pretty old there are various open CVEs for anything before that.

jetwhiz: Also, from above, the current version is `1.13.1`.  Anything before `1.12.8` has an open CVE.  It would be good to at least bump this to `1.12.10`.

lukehinds:
OK, I am going to try and squash 17 commits, please keep your local versions intact in case something goes wrong.

lukehinds:
How do I push to this branch? @jetwhiz what did you do?

lukehinds:
OK, please check it over @nabilschear and @jetwhiz 

It looks like they rebased correctly.

nabilschear:
@lukehinds i think it is still behind master a little ways and missing charlie's go/cryptography min version updates.

lukehinds:
> @lukehinds i think it is still behind master a little ways and missing charlie's go/cryptography min version updates.

Looks like its there:

https://github.com/keylime/keylime/blob/82950804adcc16f4e492fac06f2ce324e117d345/installer.sh#L43

jetwhiz:
It looks like this branch got squashed two commits behind `master` (the new `pkg_installer`'s parent commit is `4de2f32` while master is at `7c6e0ea`).   This should be fixed now.

PR #184, opened 2019-10-07
Systemd changes

lukehinds:
This allows standard naming (required for packaging) and also
renames .examples to .templates (so the installer script can still
be used).

It also removes the Unit dependencies which should not be there.
Any of the three services should not need a reliance on others (as
they could be installed on entirely different hosts)


lukehinds:
Going to land this as its holding up rpm testing upstream.

PR #192, opened 2019-10-14
Add script for creation of QEMU/KVM VM with emulated TPM

ozoder:
This script contains all the necessary steps for creation of a
KVM/QEMU virtual machine with emulated TPM inside. It can also be
used within other VM.

Mostly based on code by Luke Hinds. Only tested on Fedora 30.

### DIFF
--- a/scripts/vtpm_testing_setup.sh
+++ b/scripts/vtpm_testing_setup.sh
@@ -1,0 +1,67 @@
+#!/bin/bash
+
+if [[ $EUID -ne 0 ]]; then
+   echo -e "Please run as the root user"
+   exit 1
+fi
+
+read -p "Name for vtpm vm: " vm_name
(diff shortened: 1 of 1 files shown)


lukehinds:
thanks @ozoder , you removed the sections to install libtpms etc. what's the thinking there, it should be done separately?

ozoder:
I used dnf to install it (lines 37-39), is that not OK for some reason? This was working for me with no issue.

lukehinds:
> I used dnf to install it (lines 37-39), is that not OK for some reason? This was working for me with no issue.

Oh its packaged? I have only just seen that.Well that is even better I agree

lgtm!

nabilschear:
oh, i bet it's selinux... trying to disable now.

nabilschear:
yep, selinux strikes again.

PR #194, opened 2019-10-16
enable provider verifier to run on the same machine as a tenant verifier

nabilschear:
this change allows two verifiers to run on the same host without stepping on each other.  They will share the CA directories.

### DIFF
--- a/keylime/cloud_verifier_common.py
+++ b/keylime/cloud_verifier_common.py
@@ -63,6 +63,8 @@ class CloudAgent_Operational_State:
     TERMINATED = 8
     INVALID_QUOTE = 9
     TENANT_FAILED = 10
+    GET_PROVIDER_QUOTE = 11
+    GET_PROVIDER_QUOTE_RETRY = 12
 
     STR_MAPPINGS = {
         0 : "Registered",
@@ -75,7 +77,9 @@ class CloudAgent_Operational_State:
         7 : "Failed",
         8 : "Terminated",
         9 : "Invalid Quote",
-        10 : "Tenant Quote Failed"
+        10 : "Tenant Quote Failed",
+        11 : "Get Provider Quote",
+        12 : "Get Provider Quote (retry)",
     }
 
 
--- a/keylime/cloud_verifier_tornado.py
+++ b/keylime/cloud_verifier_tornado.py
@@ -321,10 +321,15 @@ class AgentsHandler(BaseHandler):
         else:
             try:
                 json_response = json.loads(response.body)
+                
+                # TODO this should be set by the tenant
+                agent['need_provider_quote'] = True
 
                 # validate the cloud agent response
                 if cloud_verifier_common.process_quote_response(agent, json_response['results']):
-                    if agent['provide_V']:
+                    if agent['need_provider_quote']:
+                        self.process_agent(agent, cloud_verifier_common.CloudAgent_Operational_State.GET_PROVIDER_QUOTE)
+                    elif agent['provide_V']:
                         self.process_agent(agent, cloud_verifier_common.CloudAgent_Operational_State.PROVIDE_V)
                     else:
                         self.process_agent(agent, cloud_verifier_common.CloudAgent_Operational_State.GET_QUOTE)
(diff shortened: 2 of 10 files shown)


keylime/cloud_verifier_tornado.py
```diff
@@ -471,16 +471,16 @@ def start_tornado(tornado_server, port):
     print("Tornado finished")
 
 
-def main(argv=sys.argv):
+def start(port_key,db_key,rev_port_key):
```
lukehinds: This has to stay `main` or you will need a new `main` that calls this function. The reason being its an entry point. You can see CI is failing here over this: 
https://travis-ci.org/keylime/keylime/builds/598673045#L675

nabilschear:
i think this is all set now.  @jetwhiz or @lukehinds can take a look?

keylime/cloud_verifier_tornado.py
```diff
@@ -471,16 +471,16 @@ def start_tornado(tornado_server, port):
...
     """Main method of the Cloud Verifier Server.  This method is encapsulated in a function for packaging to allow it to be
     called as a function by an external program."""
 
     config = configparser.ConfigParser()
     config.read(common.CONFIG_FILE)
 
-    cloudverifier_port = config.get('general', 'cloudverifier_port')
+    cloudverifier_port = config.get('general', port_key)
```
lukehinds: There is no `port_key` on `keylime.conf` , is that a deliberate omission ?

keylime/cloud_verifier_tornado.py
```diff
@@ -471,16 +471,16 @@ def start_tornado(tornado_server, port):
...
 
-    db_filename = "%s/%s"%(common.WORK_DIR,config.get('cloud_verifier','db_filename'))
+    db_filename = "%s/%s"%(common.WORK_DIR,config.get('cloud_verifier',db_key))
```
lukehinds: same again as with `port_key`, no conf entry

nabilschear:
is this one ok to merge?

keylime/cloud_verifier_tornado.py
nabilschear: port_key is a var that's set by the caller to start.  See https://github.com/nabilschear/python-keylime/blob/be234cafb67e928cae20b685b9cd842e76565593/keylime/cloud_verifier_tornado.py#L521

nabilschear: i've added three new config entries to keylime.conf: `provider_verifier_port`, `provider_db_filename`,`provider_revocation_notifier_port` . These get passed into the `start` routine by the new provider verifier main function: 
https://github.com/nabilschear/python-keylime/blob/be234cafb67e928cae20b685b9cd842e76565593/keylime/provider_verifier.py#L32

So, `port_key` and the others are just variables that point either to the old config entries or the new ones depending on who called start . This replicates the functionality of what was there when we only had one verifier.  
https://github.com/nabilschear/python-keylime/blob/be234cafb67e928cae20b685b9cd842e76565593/keylime/cloud_verifier_tornado.py#L521 

Maybe i should change the names of those variables to make it more clear?  (instead of xxx_key)

nabilschear:
@cjustacoder i've uploaded the basic logic for the verifiers to talk to each other.  this all runs on one computer (on localhost).  So you need a verifier, provider verifier, registrar, agent, and tenant.  When you add a new agent, the verifier will get a quote from the agent and then connect to the other verifier.  Right now, it just does a agent listing (which should be none, since that verifier doesn't have anything in it right now).  But, this is the basic structure of how to talk to more than one.  You'll need to figure out what needs to pass from one to the other to make things work.

lukehinds:
need any help with this PR folks?

lukehinds:
@cjustacoder is this pull request needed alongside #214 still? 

I am trying to work out if we need to merge both patches, or does #214 cover us and we can close this PR?

cjustacoder:
@lukehinds, sorry for the late reply. Busy for the final exam week. 
This pull request isn't needed alongside #214. #214 is not related to this PR now. Currently, we run different instances on different VMs, so we don't have port conflict now. Because our prototype needs two full Keylime instances, not only verifiers. 
Btw, I will organize our vtpm REST-API prototype  a little bit and make a PR to vtpm brach.

keylime/cloud_verifier_common.py
```diff
@@ -63,6 +63,8 @@ class CloudAgent_Operational_State:
     TERMINATED = 8
     INVALID_QUOTE = 9
     TENANT_FAILED = 10
+    GET_PROVIDER_QUOTE = 11
+    GET_PROVIDER_QUOTE_RETRY = 12
```
jetwhiz: Note that this will require some WebApp-related updates (though the existing one is being replaced eventually).  

Namely: 
https://github.com/keylime/keylime/blob/5f92021e64e6a08d5cf2f14b1f2e60293ddf60b8/keylime/static/js/webapp.js#L258-L283

and 

https://github.com/keylime/keylime/blob/5f92021e64e6a08d5cf2f14b1f2e60293ddf60b8/keylime/tenant_webapp.py#L402

Issue #195, opened 2019-10-16
Verifier talking testing


cjustacoder:
# Environment
 * OS / version: Fedora 30 on Virtual Box
 * Processor architecture: Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
 * TPM Manufacturer: 
$ tpm2_getcap -c properties-fixed
TPM_PT_FAMILY_INDICATOR:
  as UINT32:                0x08322e3000
  as string:                "2.0"
TPM_PT_LEVEL:               0
TPM_PT_REVISION:            1.46
TPM_PT_DAY_OF_YEAR:         0x000000a7
TPM_PT_YEAR:                0x000007e1
TPM_PT_MANUFACTURER:        0x49424d20
TPM_PT_VENDOR_STRING_1:
  as UINT32:                0x53572020
  as string:                "SW  "
TPM_PT_VENDOR_STRING_2:
  as UINT32:                0x2054504d
  as string:                " TPM"
TPM_PT_VENDOR_STRING_3:
  as UINT32:                0x00000000
  as string:                ""
TPM_PT_VENDOR_STRING_4:
  as UINT32:                0x00000000
  as string:                ""
TPM_PT_VENDOR_TPM_TYPE:     0x00000001
TPM_PT_FIRMWARE_VERSION_1:  0x20160511
TPM_PT_FIRMWARE_VERSION_2:  0x00162800
TPM_PT_INPUT_BUFFER:        0x00000400
TPM_PT_TPM2_HR_TRANSIENT_MIN:    0x00000003
TPM_PT_TPM2_HR_PERSISTENT_MIN:   0x00000002
TPM_PT_HR_LOADED_MIN:       0x00000003
TPM_PT_ACTIVE_SESSIONS_MAX: 0x00000040
TPM_PT_PCR_COUNT:           0x00000018
TPM_PT_PCR_SELECT_MIN:      0x00000003
TPM_PT_CONTEXT_GAP_MAX:     0xffffffff
TPM_PT_NV_COUNTERS_MAX:     0x00000000
TPM_PT_NV_INDEX_MAX:        0x00000800
TPM_PT_MEMORY:              0x00000006
TPM_PT_CLOCK_UPDATE:        0x00001000
TPM_PT_CONTEXT_HASH:        0x0000000c
TPM_PT_CONTEXT_SYM:         0x00000006
TPM_PT_CONTEXT_SYM_SIZE:    0x00000100
[...]

PR #197, opened 2019-10-17
Verifier talk to each other

cjustacoder:
Try to implement the communication between two Verifier. Only focus on tornado part. A simple helloworld try for communication, not involve actual functions or interfaces.

Based on existing 'tornado', adding URL and handler correspond to GET request. Verifier communicate with each other using GET. Tenant verifier will send a GET request(using python requests package) before entering tornadoIOLoop. The rest of changes are comments which help me understand the code.

Resolves: #195

### DIFF
--- a/keylime/cloud_verifier_tornado.py
+++ b/keylime/cloud_verifier_tornado.py
@@ -1,3 +1,4 @@
+#!/usr/bin/python3
 '''
 DISTRIBUTION STATEMENT A. Approved for public release: distribution unlimited.
 
@@ -23,6 +24,7 @@ import traceback
 import sys
 import functools
 import asyncio
+import requests
 
 import tornado.ioloop
 import tornado.web
@@ -48,6 +50,12 @@ if sys.version_info[0] < 3:
 config = configparser.ConfigParser()
 config.read(common.CONFIG_FILE)
 
+# ===============test parameter===============
+provider_ip = "http://10.0.2.4:8881"
+tenant_ip = "http://10.0.2.15:8881"
+identity = input("Enter your identity: provider or tenant: ")
+# ============================================
+
 class BaseHandler(tornado.web.RequestHandler):
 
     def write_error(self, status_code, **kwargs):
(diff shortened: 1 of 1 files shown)


nabilschear:
@cjustacoder it looks like you're mixing the existing non-blocking event-based approach for servicing requests from tenants with the blocking requests library.  Take a look at how this function does things.  https://github.com/keylime/keylime/blob/e2bfff7b1ebcaef9b3dec7d840f18133684e84bf/keylime/cloud_verifier_tornado.py#L295 Since you're operating in the tornado async world, you have to issue get requests to talk to others using the same mechanisms.  I'm also not sure about adding a totally different handler for this feature since all the state management and other stuff is already built in the other handler (see `process_agent`) could this be implemented by adding more operational states and moving through?

nabilschear:
another thought, it would be great if the verifier essentially had a very similar rest interface as the agent.  So when the first verifier talks to the provider verifier, it uses the similar uris and other info.  It all depends on whether the merkle tree data is going to be kept and maintained.

cjustacoder:
Thanks a lot, Nabil@nabilschear! I will dig into this and try to figure it out.

cjustacoder:
Hi Nabil, @nabilschear. I'm still trying to dig the code. There is something I don't understand. 

I try to figure out the workflow so I print out each steps when runs the program, and below is my out put. 
```
[root@localhost keylime]# python3 cloud_verifier_tornado.py 
Enter your identity: provider or tenant: provider
2019-10-24 16:24:09.376 - keylime.cloudverifier - INFO - no agent now(testing)
2019-10-24 16:24:09.376 - keylime.cloudverifier - INFO - Starting Cloud Verifier (tornado) on port 8881, use <Ctrl-C> to stop
2019-10-24 16:24:09.377 - keylime.cloudverifier_common - INFO - Setting up TLS...
2019-10-24 16:24:09.378 - keylime.cloudverifier_common - INFO - Existing CA certificate found in /var/lib/keylime/cv_ca, not generating a new one
2019-10-24 16:24:09.380 - keylime.cloudverifier - INFO - Starting service for revocation notifications on port 8992
entering the loop
entering the loop
get a post
/agents/D432FBB3-D2F1-4A97-9EF7-75BD81C00000
process_agent
1
invoke_get_quote
http://127.0.0.1:9002/quotes/integrity?nonce=SACTsgsJOIyQ9xcWQIAF&mask=0x408000&vmask=0x808000&partial=0
2019-10-24 16:24:25.072 - keylime.cloudverifier - INFO - POST returning 200 response for adding agent id: D432FBB3-D2F1-4A97-9EF7-75BD81C00000
on_get_quote_response
2019-10-24 16:24:25.422 - keylime.cloudverifier - INFO - content of quote
```

It seems like after entering `tornado.ioloop.IOLoop.instance().start()`https://github.com/keylime/keylime/blob/e2bfff7b1ebcaef9b3dec7d840f18133684e84bf/keylime/cloud_verifier_tornado.py#L513, the tornado keeps idling until it receives a POST from agent. Then the verifier invokes a `process agent` function, but I didn't see any invoke from the POST handler. https://github.com/keylime/keylime/blob/e2bfff7b1ebcaef9b3dec7d840f18133684e84bf/keylime/cloud_verifier_tornado.py#L176-L239

So I'm wondering who invoke the 'process_agent' in this process? Thanks a lot!

cjustacoder:
@nabilschear @lukehinds, I have updated some changes, still couldn't get through. Can you please help me to have a look? Thanks a lot!

nabilschear:
process_agent gets called here in post https://github.com/cjustacoder/keylime/blob/f2bb598113bfd263153e6f66b811167cbf8b8250/keylime/cloud_verifier_tornado.py#L269

there's a little state machine that agents go through that is driven by it's operational_state value.  all of the state transitions happen in `process_agent` .  So you should be able to see old state new state transitions all in that function.

nabilschear:
I wasn't able to figure out what was going wrong with this (i only tried a little).  So, i spent a little bit of time to figure out the basic scaffolding for what you'll want to do to do it right.  See #194 for the code @cjustacoder

Issue #160, opened 2019-09-20
yaml.scanner issue when running keylime_agent 


atothRedHat:
# Environment
 * OS / version: 
Fedora 30

 * Processor architecture: 
model name	: Intel(R) Core(TM) i7-8650U CPU @ 1.90GHz

 * TPM Manufacturer: 
$ tpm2_getcap -c properties-fixed
TPM_PT_FAMILY_INDICATOR:
  as UINT32:                0x08322e3000
  as string:                "2.0"
TPM_PT_LEVEL:               0
TPM_PT_REVISION:            1.16
TPM_PT_DAY_OF_YEAR:         0x00000109
TPM_PT_YEAR:                0x000007e0
TPM_PT_MANUFACTURER:        0x49465800
TPM_PT_VENDOR_STRING_1:
  as UINT32:                0x534c4239
  as string:                "SLB9"
TPM_PT_VENDOR_STRING_2:
  as UINT32:                0x36363500
  as string:                "665"
TPM_PT_VENDOR_STRING_3:
  as UINT32:                0x00000000
  as string:                ""
TPM_PT_VENDOR_STRING_4:
  as UINT32:                0x00000000
  as string:                ""
TPM_PT_VENDOR_TPM_TYPE:     0x00000000
TPM_PT_FIRMWARE_VERSION_1:  0x0005003f
TPM_PT_FIRMWARE_VERSION_2:  0x000d1900

 * Keylime version: 
4.0

# Description 
When running keylime_agent, the process dies withthe following error:
 

## Expected behavior vs. actual behavior 
 - keylime.secure_mount - INFO - mounting secure storage location /var/lib/keylime/secure on tmpfs
Traceback (most recent call last):
  File "/usr/local/bin/keylime_agent", line 11, in <module>
    load_entry_point('keylime==1.2', 'console_scripts', 'keylime_agent')()
  File "/usr/local/lib/python3.7/site-packages/keylime-1.2-py3.7.egg/keylime/cloud_agent.py", line 492, in main
    (ek,ekcert,aik,ek_tpm,aik_name) = tpm.tpm_init(self_activate=False,config_pw=config.get('cloud_agent','tpm_ownerpassword')) # this tells initialize not to self activate the AIK
  File "/usr/local/lib/python3.7/site-packages/keylime-1.2-py3.7.egg/keylime/tpm2.py", line 814, in tpm_init
    self.warn_emulator()
[...]

PR #205, opened 2019-11-11
Updates needed for 4.0.x tpm2-tools changes

lukehinds:
In tools commit hash:
https://github.com/tpm2-software/tpm2-tools/commit/45c4883dcdadc754b60832a4    2a5e9e82f847bc7d

`f` has been changed to `o`

`tpm2_pcrlist` is now `tpm2_pcrread`

Resolves: #204, #208

Issue #206, opened 2019-11-11
tpm2-tools 3.2.0 nvread fails with Infineon SLB9665VQ2.0 TPM v2.0


lukehinds:
Keylime agent registration works with 3.2.0 using  an emulator, but fails with the Infineon SLB9665VQ2.0 TPM v2.0

Debug output:

```
Using config file /etc/keylime.conf
2019-11-11 14:36:23.860 - keylime.tpm2 - INFO - Using legacy version of tpm2-tools
('cmd: ', 'tpm2_startup -c')
('cmd: ', 'tpm2_getcap -c algorithms')
('cmd: ', 'mount')
('cmd: ', 'tpm2_getcap -c properties-fixed')
WARN: More data to be queried: capability: 0x6, property: 0x100

2019-11-11 14:36:23.950 - keylime.tpm2 - INFO - Taking ownership with config provided TPM owner password: keylime
('cmd: ', 'tpm2_takeownership -c')
('cmd: ', 'tpm2_takeownership -o keylime -e keylime')
2019-11-11 14:36:24.090 - keylime.tpm2 - INFO - TPM Owner password confirmed: keylime
2019-11-11 14:36:24.091 - keylime.tpm2 - INFO - Flushing old ek handle: 0x81010007
('cmd: ', 'tpm2_getcap -c handles-persistent')
('cmd: ', 'tpm2_getpubek -H 0x81010007 -g rsa -f /tmp/tmpGCMRtx -P U1RKfzVUVpVGfQsxldtb -o keylime -e keylime')
('cmd: ', 'tpm2_readpublic -H 0x81010007 -o /tmp/tmpdU12rp -f pem')
('cmd: ', 'tpm2_nvlist')
('cmd: ', 'tpm2_nvread -x 0x1c00002 -s 1177 -f /tmp/tmppRPCNb')
ERROR: Failed to read NVRAM area at index 0x1c00002 (29360130). Error:0x9a2
ERROR: Unable to run tpm2_nvread
Traceback (most recent call last):
  File "/usr/bin/keylime_agent", line 11, in <module>
    load_entry_point('keylime==1.2', 'console_scripts', 'keylime_agent')()
  File "/usr/lib/python2.7/site-packages/keylime-1.2-py2.7.egg/keylime/cloud_agent.py", line 482, in main
[...]

PR #207, opened 2019-11-11
Introduce Bool for toggling STDOUT on commands

lukehinds:
Commands such as `tpm2_getcap -c properties-fixed` cause a YAML
formatting error when run against an SLB9665VQ2 TPM

This can be seen in issue #160

This fix provides a simple boolean that can be passed on commands
that break with STDOUT removed

Resolves: #160, #206

### DIFF
--- a/keylime/tpm2.py
+++ b/keylime/tpm2.py
@@ -107,10 +107,11 @@ class tpm2(tpm_abstract.AbstractTPM):
         retDict = self.__run("tpm2_startup --version")
 
         code = retDict['code']
-        if code != tpm_abstract.AbstractTPM.EXIT_SUCESS:
-            raise Exception("Error establishing tpm2-tools version using TPM2_Startup: %s"+str(code)+": "+str(output))
-        
         output = ''.join(common.list_convert(retDict['retout']))
+        errout = ''.join(common.list_convert(retDict['reterr']))
+        if code != tpm_abstract.AbstractTPM.EXIT_SUCESS:
+            raise Exception("Error establishing tpm2-tools version using TPM2_Startup: %s"+str(code)+": "+str(errout))
+        
         # Extract the `version="x.x.x"` from tools
         version_str = re.search(r'version="([^"]+)"', output).group(1)
         # Extract the full semver release number.
@@ -135,10 +136,11 @@ class tpm2(tpm_abstract.AbstractTPM):
             retDict = self.__run("tpm2_getcap algorithms")
 
         output = common.list_convert(retDict['retout'])
+        errout = common.list_convert(retDict['reterr'])
         code = retDict['code']
 
         if code != tpm_abstract.AbstractTPM.EXIT_SUCESS:
-            raise Exception("get_tpm_algorithms failed with code "+str(code)+": "+str(output))
+            raise Exception("get_tpm_algorithms failed with code "+str(code)+": "+str(errout))
 
         if tools_version == "3.2":
             # output, human-readable -> json
@@ -245,6 +247,7 @@ class tpm2(tpm_abstract.AbstractTPM):
                 # Package for return
                 returnDict = {
                     'retout': thisRetout,
+                    'reterr': [],
                     'code': thisCode,
                     'fileouts': fileoutEncoded,
                     'timing': thisTiming,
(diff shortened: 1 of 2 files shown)


lukehinds:
Reviewers: please merge #206 first!

keylime/tpm2.py
```diff
@@ -1176,8 +1177,8 @@ def read_ekcert_nvram(self):
             # Read the RSA EK cert from NVRAM (DER format)
             if tools_version == "3.2":
                 retDict = self.__run("tpm2_nvread -x 0x1c00002 -s %s -f %s"%(ekcert_size, nvpath.name), raiseOnError=False, outputpaths=nvpath.name)
-            else:
-                retDict = self.__run("tpm2_nvread 0x1c00002 -s %s -f %s"%(ekcert_size, nvpath.name), raiseOnError=False, outputpaths=nvpath.name)
+            elif tools_version == "4.0":
```
lukehinds: note, this is from fix #205

jetwhiz:
Just to clarify, this is just squashing `STDERR` -> `STDOUT` output and not `STDOUT` -> `STDOUT` for certain TPM calls?

lukehinds:
> Just to clarify, this is just squashing `STDERR` -> `STDOUT` output and not `STDOUT` -> `STDOUT` for certain TPM calls?

Yes, so for example with issue #160 we resolved this by removing `stderr=subprocess.STDOUT` - however doing so breaks calls such as `tpm2_nvread`

With this change we can pipe cmds to run without `STDOUT` (currently only the YAML we grab from `tpm2_getcap -c properties-fixed` gets passed in without `STDOUT`

jetwhiz:
Which `STDERR` outputs are causing the issues with https://github.com/keylime/keylime/issues/160 on the SLB9665VQ2?  In the text you output in the issue I'm not seeing anything that looks like `STDERR`.

jetwhiz:
I'm wondering if we should just separate the stdout and stderr streams and have two different returns: `retout` and `reterr`, for instance: 

https://github.com/keylime/keylime/blob/e2bfff7b1ebcaef9b3dec7d840f18133684e84bf/keylime/cmd_exec.py#L46-L52

We would make a `reterr` that reads from `proc.stderr.readline()` and change `,stderr=subprocess.STDOUT` to `,stderr=subprocess.PIPE`.

lukehinds:
> Which `STDERR` outputs are causing the issues with #160 on the SLB9665VQ2? In the text you output in the issue I'm not seeing anything that looks like `STDERR`.

It's `stderr=subprocess.STDOUT` causing the issues. Have a search for the email `Fedora 30, Nuk, agent issue, python work`

lukehinds:
> I'm wondering if we should just separate the stdout and stderr streams and have two different returns: `retout` and `reterr`, for instance:
> 
> https://github.com/keylime/keylime/blob/e2bfff7b1ebcaef9b3dec7d840f18133684e84bf/keylime/cmd_exec.py#L46-L52
> 
> We would make a `reterr` that reads from `proc.stderr.readline()` and change `,stderr=subprocess.STDOUT` to `,stderr=subprocess.PIPE`.
> 
> This block would need to be updated to check against the `reterr` instead of `retout`:
> 
> https://github.com/keylime/keylime/blob/e2bfff7b1ebcaef9b3dec7d840f18133684e84bf/keylime/tpm2.py#L1199-L1207

Sounds viable to me, do you want to go ahead and push what you mean onto this PR?

jetwhiz:
> Sounds viable to me, do you want to go ahead and push what you mean onto this PR?

Take a look at my commit and see what you think.  Not sure how we should handle all of the logging outputs in tpm2 (do we output both `retout` and `reterr`?)

jetwhiz:
@lukehinds can you test this against #160 and #206 to see if it fixes those issues for you?

lukehinds:
hey @jetwhiz , thanks for the commit. I tested this out against the HW TPM and it works well, its also working well against the emulator.

Give me an ack, and I will merge this.

jetwhiz:
@lukehinds  Most of the tpm2.py logging is only outputting `retout`.  Should we update these to output both `retout` and `reterr`, or do you think it is fine as-is?  

Also, do we need to wait for PR #205 before merging this?

lukehinds:
@jetwhiz I have no issue with how it is, in that it all appears to work as expected to me.

took the liberty of merging #205 already :)

jetwhiz:
@lukehinds - I made some updates to make error logging outputs more useful (since errors will come out in stderr and not stdout).  

How does this look to you?

lukehinds:
Looks good to me @jetwhiz 

Happy to merge this now.

jetwhiz:
Sounds good to me to merge.

PR #215, opened 2019-11-25
Fix bug in tenant's "list" functionality

jetwhiz:
cvstatus only supported 'listing=False' mode


lukehinds:
thanks @jetwhiz , will take a look first thing in the morning.

Issue #218, opened 2019-12-05
Move to using files for EK CA manufactorer certificates


lukehinds:
Move to use .pem files in a certificates/ directory instead of the current `tpm_ek_ca.py`

This way users can add their own intermediary / root certs for EK trust and not have to edit code (which will be installed at `/usr/local/lib/python3.7/site-packages/` and so wiped out everytime an upgrade occurs.

lukehinds:
@amylily1011 , you might like this one. Its a nice secluded bit of code so you won't have to manage lots of merge conflicts. Let me know if it interests you and we can walk over an approach when we next meet.

amylily1011:
@lukehinds Yes yes, please :-)

PR #227, opened 2019-12-12
[WIP]: Move to using files for EK CA certs

amylily1011:
I had a look at issue https://github.com/keylime/keylime/issues/218 and plotted something together just to replace the statically defined certs by loading the files into the same dict. I'd be happy to receive some feedback on how to implement this and make it conform with the rest of the codebase.

@lukehinds


jetwhiz:
This looks great, thanks @amylily1011 !  

Do we also need to include the new `certificates` directory in the setup process now to ensure these certs will be included, @lukehinds ?  

https://github.com/keylime/keylime/blob/a300a3d9818b0e306e61d734cb26911c09d21ccb/setup.py#L102

Also, we should be able to leave out the `.auditing-0`, `NVChip`, and ` testdata.sqlite` files for the PR.

lukehinds:
> This looks great, thanks @amylily1011 !
> 

Agree, thanks @amylily1011 

> Do we also need to include the new `certificates` directory in the setup process now to ensure these certs will be included, @lukehinds ?
> 
> https://github.com/keylime/keylime/blob/a300a3d9818b0e306e61d734cb26911c09d21ccb/setup.py#L102

That's a good point. I expect we do, as we will want these to be included in the packaging. 
> 
> Also, we should be able to leave out the `.auditing-0`, `NVChip`, and ` testdata.sqlite` files for the PR.

keylime/tpm_ek_ca.py
```diff
@@ -18,1005 +18,15 @@
...
-mQ+96q8dwJdNabwBokrZy2beL71CXt/4jYNN0j/N9uYO4vIDBFDKRMWCtUO217+w
-xQTSOv5+mpgFw7UML/QpgpdmZy2i+eZPxDo8dzT+YJXC5vsHVSooA3rWDDzvnoLC
-cmDDiT3pG6AdjAN61MeeHHmoJavV8Tvdoa3g14Sn1lL+TQ1xaznyh520sX0dXPTp
-GqZbDzqEMiVbG7vFECqINE96/rwppJlWK91F1MZikGXr7FeF5C0JutGLb0gaYOmv
-Yau4DQ==
------END CERTIFICATE-----"""
-}
+trusted_certs = {}
 
+for trusted_cert in glob.glob('certificates/**/*.pem'):
```
lukehinds: I wonder if we could use a config parameter here, so users can configure the location themselves (they might want to use a read-only secure mount to somewhere.

Something like `ek_cert_store` in `keylime.conf`?

keylime/tpm_ek_ca.py
```diff
@@ -18,1005 +18,15 @@
...
+    with open(trusted_cert) as f:
+        trusted_cert = trusted_cert[:-4].split('/')
+        content = f.readlines()
+        trusted_certs[trusted_cert[2]] = content
 
 atmel_trusted_keys = {"ATM1": {"key": "E5AFB8C97B6E6D49C624FCFB6558D6B291A17B94EBF4186B590F5039F0E85659268BEB60D86A9CF6A03D477F5686288498AB90DFEE1722F3BE5C516B542BFEF18B849C1208BB8E85117D4B046ECB9D425B920015D29CC73163A084B84320670683FEA43BD7C7FBB29CC1FC0D6A1F992E47DB7E2B37726D3D465386F114AAF0BC8AF19734ED01F633B2FFF0F1674514393270205BBBEE93C87F33AD934F2FB20D5B9697F8E7877BEB2A054CA0EC8BFE233711D147215FABFE524E23F1D1025922782446889D64248E2D42951788BFAD2B6881D720319EA712433596C6536EBBAB00515DAD23EA8C54CDA8E9E4F57FF30CEFE6F89E6CAE06B74E20EE5DBD37A4EF",
```
lukehinds: @jetwhiz do we use these atmel_trusted_keys - they are in a different dictionary structure that I don't think the code that calls `tpm_ek_ca` would use this structure?

jetwhiz: These are still used for the tpm1 implementation: 

https://github.com/keylime/keylime/blob/22a9789e25a36a9b8d0a69a679de63125a2d1fca/keylime/tpm1.py#L515-L520

I think this would be something we can remove when we decide to deprecate TPM 1.2 functionality.

keylime/tpm_ek_ca.py
jetwhiz: Would the installer need to know this location as well so it can install the certs there automatically?

lukehinds:
hi @amylily1011 

Got this working against a HW tpm. Here are the changes I recommend.

Put all the same certificates into a folder in the root of the repository `tpm_cert_store`.

Regarding the `atmel_trusted_keys`, you can put those into individual files, for example

```
$ cat ATM6.pem

C3DB3DCA4A60CC7D418ABCFC25598773F8D7352685B0341833D85C4B7E75DCFEDAAD     2FABFE6BAC994A9F7726D8FC643CF98A8003A219ABEC65199ED65049B1BE8FC79679BF1508F77A96460F7AEE0299BEF1D2D5E     07BECE118D9092F1C7E9911672821FA745DAD9DDA0306518D06E1BE5F26009390D91A7D00D699333623ADEEDA764BAA2050DA     F379CA827E1790C21FB7E32D84EBDC318999B522FAE20BC0359ECF8ABC81D9C3214BED3200C18AFCEAA6D042C8464F30A8047     F977B71455C68404A80F5F064D9BBC63625AF6BDD9C87108BA445FDF827195AF51E1FA529E5AD840F1C3C060C44D0BEACAA32     0D5A7DAE374C7B8E52B2B20F97132A3BA94A4C45
```

Don't worry about `exponent`  etc, we don't use those keys.

`tpm_ek_ca.py`

```
import configparser
import glob
import os
from keylime import common

config = configparser.RawConfigParser()

trusted_certs = {}

tpm_cert_store = config.get('tenant', 'tpm_cert_store')

def cert_loader():
  file_list = glob.glob(os.path.join(os.getcwd(), tpm_cert_store + "/*.pem"))
  trusted_certs = []
  for file_path in file_list:
      with open(file_path) as f_input:
          trusted_certs.append(f_input.read())
  return trusted_certs
```

`keylime.conf`

Add the following to the `[tenant]` section

```
# EK certificate storage location
tpm_cert_store = /var/lib/keylime/tpm_cert_store
```

For both `tpm2.py` and `tpm1.py` the whole of the `verify_ek` function will now look like this

```
    def verify_ek(self, ekcert, ekpem):
        """Verify that the provided EK certificate is signed by a trusted root
        :param ekcert: The Endorsement Key certificate in DER format
        :param ekpem: the endorsement public key in PEM format
        :returns: True if the certificate can be verified, false otherwise
        """
        #openssl x509 -inform der -in certificate.cer -out certificate.pem
        try:
            ek509 = M2Crypto.X509.load_cert_der_string(ekcert)
            ekcertpem = ek509.get_pubkey().get_rsa().as_pem(cipher=None).decode('utf-8')

            # Make sure given ekcert is for their ek
            if str(ekpem) != str(ekcertpem):
                logger.error("Public EK does not match EK certificate")
                return False

            trusted_certs = tpm_ek_ca.cert_loader()
            for cert in trusted_certs:
                signcert = M2Crypto.X509.load_cert_string(cert)
                signkey = signcert.get_pubkey()
                if ek509.verify(signkey) == 1:
                    logger.debug("EK cert matched cert: ", cert)
                    return True
        except Exception as e:
            # Log the exception so we don't lose the raw message
            logger.exception(e)
            raise Exception("Error processing ek/ekcert. Does this TPM have a valid EK?").with_traceback(sys.exc_info()[2])

        logger.error("No Root CA matched EK Certificate")
        return False
```

If you commit these changes I will test again, and we can then update the installers to set up the tpm_cert_store

jetwhiz:
> Don't worry about `exponent` etc, we don't use those keys.

It looks like they're all using the common exponent of `65537` (or `10001` in hex) anyways, so we can just hard-code this into the code (we're dropping support for TPM 1.2 eventually, so this probably doesn't need to be flexible). 

Note that these are not PEM certificates, they are the modulus `n` and exponent `e` values of the RSA key.  It might be good to name them something other than `.pem` to avoid confusion.  

We will also need to keep this code for TPM 1.2 if we want to use them: 

https://github.com/keylime/keylime/blob/22a9789e25a36a9b8d0a69a679de63125a2d1fca/keylime/tpm1.py#L515-L520

lukehinds:
> Note that these are not PEM certificates, they are the modulus n and exponent e values of the RSA key. It might be good to name them something other than .pem to avoid confusion.

@jetwhiz , please go ahead and make a suggestion for a file extension.

jetwhiz:
> > Note that these are not PEM certificates, they are the modulus n and exponent e values of the RSA key. It might be good to name them something other than .pem to avoid confusion.
> 
> @jetwhiz , please go ahead and make a suggestion for a file extension.

Should we just leave them in the dictionary in `tpm_ek_ca.py`?   They will need to be treated specially anyways if we pull them out into a file, since we will have to use the modulus ("key") and exponent to rebuild the public RSA key.  

In the future, we can just strip those lines out of `tpm_ek_ca.py` when we want to drop support for TPM 1.2 (rather than juggling a bunch of files and special code).

lukehinds:
> > > Note that these are not PEM certificates, they are the modulus n and exponent e values of the RSA key. It might be good to name them something other than .pem to avoid confusion.
> > 
> > 
> > @jetwhiz , please go ahead and make a suggestion for a file extension.
> 
> Should we just leave them in the dictionary in `tpm_ek_ca.py`? They will need to be treated specially anyways if we pull them out into a file, since we will have to use the modulus ("key") and exponent to rebuild the public RSA key.
> 
> In the future, we can just strip those lines out of `tpm_ek_ca.py` when we want to drop support for TPM 1.2 (rather than juggling a bunch of files and special code).

Sounds good to me.

keylime/cloud_verifier_common.py
```diff
@@ -142,7 +142,7 @@ def init_mtls(section='cloud_verifier', generatedir='cv_ca'):
         tls_dir = os.path.abspath('%s/%s' % (common.WORK_DIR, 'cv_ca'))
         if not os.path.exists("%s/cacert.crt" % (tls_dir)):
             raise Exception(
-                "It appears that the verifier has not yet created a CA and certificates, please run the verifier first")
+                "It appears that the verifier has not yet created a CA and tpm_cert_store, please run the verifier first")
```
lukehinds: This is actually a different certificate store `ca`  to the `tpm_cert_store`. The ca is where new certs are created for communication between the systems, where as the `tpm_cert_store` is where we keep vendor tpm public certificates.

keylime/tpm1.py
```diff
@@ -488,46 +488,27 @@ def verify_ek(self,ekcert,ekpem):
         """
         #openssl x509 -inform der -in certificate.cer -out certificate.pem
         try:
-            pubekmod = base64.b64decode(self.__get_mod_from_pem(ekpem))
```
lukehinds: I don't think we can remove these lines from 491 to 504, it should be I believe lines 508 to 509 that need replacing. I will try and test this shortly.

keylime/cloud_verifier_common.py
amylily1011: This file made some automatic changes when I renamed the folder from ’certificates’ to ’tpm_cert_store’. I can change it back to it’s initial state manually :).

amylily1011:
@lukehinds done! :-) Fingers crossed.

lukehinds:
> @lukehinds done! :-) Fingers crossed.

yay! 

I hope your laptop is feeling better now. Maybe send it to a health spa for the weekend :)

Will take a look at this sometime over this weekend. Great work, it was a challenging patch to take on and you have nailed it!

PR #229, opened 2019-12-20
Cloudagent ip

rajdroid:
@lukehinds Please review these changes.

### DIFF
--- a/keylime.conf
+++ b/keylime.conf
@@ -7,6 +7,7 @@ cloudverifier_port = 8881
 registrar_port = 8890
 registrar_tls_port = 8891
 registrar_ip = 127.0.0.1
+cloudagent_ip = 127.0.0.1
 cloudagent_port = 9002
 provider_registrar_port = 8990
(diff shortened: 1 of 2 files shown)


lukehinds:
hi @rajdroid , we still need to squash your commits as per comment https://github.com/keylime/keylime/pull/229#pullrequestreview-335150991

Do you need some help with this. I could jump on a google hangouts next week if you like, or perhaps do the squash for you. 

There is a nice article here that goes into it more: https://magonye.github.io/post/git-squash/

The other option is we just make a fresh PR and close this, I understand if you prefer that route, as git can be a bit of pain to unravel when it gets complicated with different branches and origin upstream / master etc. 

The reason we do this (have a single commit) is two fold. 1. It makes the commit history easier to read. 2. It also helps us generate release notes automatically.

Thanks for your work on this, its nice having you come into the community.

rajdroid:
Thanks, @lukehinds for your help. I did squash earlier, but I think I missed something. Could you please check now I squashed again.

PR #234, opened 2019-12-21
Additional STMicroelectronics ek certs

lukehinds:
The commited files were obtained from:

https://secure.globalsign.com/stmtpmekint06.crt
https://secure.globalsign.com/stmtpmekint07.crt


lukehinds:
> where did you dig these up from? i.e., were they linked from stm's site? if so, they it looks legit.

I Googled them :)

I extracted the der from my TPM, did some conversations with openssl and googled a line of the end cert - this took me to the STM site who host their certs with https://secure.globalsign.com

thanks for the approve!

PR #240, opened 2019-12-31
Changes required for Cryptography Deprecation

lukehinds:
Keylime was experiencing the following deprecation warning:

"CryptographyDeprecationWarning: signer and verifier have been
deprecated. Please use sign and verify instead."

This resolves the warning by updating to useing `verify` instead
of `verifier`

Resolves: #222


lukehinds:
Tested this with no issues.

Issue #216, opened 2019-11-27
doRegisterAgent connection handling


lukehinds:
If you start the agent without the registar running, there is no graceful capture of the connection error:

```
2019-11-27 10:07:15.856 - keylime.cloudagent - INFO - Agent UUID: D432FBB3-D2F1-4A97-9EF7-75BD81C00000
Traceback (most recent call last):
  File "/usr/local/bin/keylime_agent", line 11, in <module>
    load_entry_point('keylime==1.2', 'console_scripts', 'keylime_agent')()
  File "/usr/local/lib/python3.7/site-packages/keylime-1.2-py3.7.egg/keylime/cloud_agent.py", line 529, in main
    keyblob = registrar_client.doRegisterAgent(registrar_ip,registrar_port,agent_uuid,tpm_version,ek,ekcert,aik,ek_tpm,aik_name)
  File "/usr/local/lib/python3.7/site-packages/keylime-1.2-py3.7.egg/keylime/registrar_client.py", line 141, in doRegisterAgent
    response_body = json.loads(response.read().decode("utf-8"))
AttributeError: 'int' object has no attribute 'read'
```

rajdroid:
Hmm, I stumbled upon this one while doing my first commit. You can assign it to me @lukehinds. I'll take a look at this one.

lukehinds:
Thanks @rajdroid

PR #242, opened 2020-01-06
Added exception handling in doRegisterAgent

rajdroid:
Added exception handling in doRegisterAgent

Currently, when the registrar is not running and we try to run agent
it exits without any graceful error message. With this change, an error
message is reported to user when the registrar is not running.

Resolves: #216


lukehinds:
thank you @rajdroid

Issue #235, opened 2019-12-22
Refactor arg and configparser use of IP's and Ports.


lukehinds:
This issue first started off as a report on the following issue (created by yours truly not properly testing his patch)

If you run keylime tenant without the `-tp` arg, it fails:

```
keylime_tenant -t 127.0.0.1 -v 127.0.0.1  -f excludes.txt --whitelist ~/whitelist.txt --exclude excludes.txt -c add 
Using config file /etc/keylime.conf
2019-12-22 08:38:30.658 - keylime.tenant - WARNING - CAUTION: using default password for private key, please set private_key_pw to a strong password
2019-12-22 08:38:30.658 - keylime.tenant - INFO - Setting up client TLS in /var/lib/keylime/cv_ca
2019-12-22 08:38:30.665 - keylime.tenant - WARNING - Using default UUID D432FBB3-D2F1-4A97-9EF7-75BD81C00000
2019-12-22 08:38:30.665 - keylime.tenant - INFO - TPM PCR Mask from policy is 0x408000
2019-12-22 08:38:30.665 - keylime.tenant - INFO - TPM PCR Mask from policy is 0x808000
2019-12-22 08:38:30.752 - keylime.ima - WARNING - No boot_aggregate value found in whitelist, adding an empty oneTraceback (most recent call last):
  File "/usr/local/bin/keylime_tenant", line 11, in <module>
    load_entry_point('keylime==1.2', 'console_scripts', 'keylime_tenant')()
  File "/usr/local/lib/python3.7/site-packages/keylime-1.2-py3.7.egg/keylime/tenant.py", line 812, in main
    mytenant.do_cv()
  File "/usr/local/lib/python3.7/site-packages/keylime-1.2-py3.7.egg/keylime/tenant.py", line 492, in do_cv
    keylime_logging.log_http_response(lo
Traceback (most recent call last):
  File "/usr/local/bin/keylime_tenant", line 11, in <module>
    load_entry_point('keylime==1.2', 'console_scripts', 'keylime_tenant')()
  File "/usr/local/lib/python3.7/site-packages/keylime-1.2-py3.7.egg/keylime/tenant.py", line 812, in main
    mytenant.do_cv()
  File "/usr/local/lib/python3.7/site-packages/keylime-1.2-py3.7.egg/keylime/tenant.py", line 492, in do_cv
```

We need a validation check to make sure its captured when the -tp arg is not used and it falls back to the configparser set value.

# Now for the main issue

[...]

PR #245, opened 2020-01-08
Added code to fallback agent port to configparser set value

rajdroid:
Currently, not providing agent port in keylime tenant instantiation
results in a crash. This code change adds fallback to configparser
set value if the port is not provided in the command line.

Resolves: #235


rajdroid:
Thanks @lukehinds. I wanted to discuss with you about how to proceed with the second PR (refactoring) before starting that work. I wanted to work on that in an incremental fashion instead of single change. Let me know your thoughts about it.

And could you please open the second PR also and assign it to me.

Issue #247, opened 2020-01-15
Refactor IP / Ports


lukehinds:
refactor to a better layout and definition of ip's and ports. At the moment most are grouped in together at the top of the file, with no real outline on what is for connecting to a service and what is for a service to bind to. I will create another issue for this.

I have observed a few incorrect behaviours as well, for example for the -c delete command. This calls do_cvdelete which then http posts a DELETE to the verfifier, however it does not pass on the target IP. Instead it takes the IP from the config. Yet if you run the tenant command without a target IP ( -t), it fails. So its basically defining an arg as mandatory, yet not even using it.

```
keylime_tenant -v 127.0.0.1  -f excludes.txt --whitelist ~/whitelist.txt --exclude excludes.txt -c delete
Using config file /etc/keylime.conf
2019-12-22 09:07:36.581 - keylime.tenant - WARNING - CAUTION: using default password for private key, please set private_key_pw to a strong password
2019-12-22 09:07:36.582 - keylime.tenant - INFO - Setting up client TLS in /var/lib/keylime/cv_ca
Traceback (most recent call last):
  File "/usr/local/bin/keylime_tenant", line 11, in <module>
    load_entry_point('keylime==1.2', 'console_scripts', 'keylime_tenant')()
  File "/usr/local/lib/python3.7/site-packages/keylime-1.2-py3.7.egg/keylime/tenant.py", line 786, in main
    raise UserError("-t/--targethost is required for command %s"%args.command)
keylime.tenant.UserError: -t/--targethost is required for command delete
```

This issue can be for working out what requires changing, any new design considerations etc.

lukehinds:
hey @rajdroid , will keep this open for now or if you prefer you can open a new bug for the '-c delete' quirk.

rajdroid:
Yeah, no problem @lukehinds. I will fix that bug under this issue.

PR #251, opened 2020-01-26
Refactored IP address and Port in the configration file

rajdroid:
Currently, the layout of the configuration file is clumsy. This
change try to group logical items together, so that it make sense.

Resolves: #247

### DIFF
--- a/keylime.conf
+++ b/keylime.conf
@@ -1,83 +1,67 @@
 [general]
 #=============================================================================
-# these options are self explanatory ip/port combinations to find the various
-# keylime services on the network.  some services use these, some do not.
-# But they are used frequently enough that they end up here
-cloudverifier_port = 8881
-registrar_port = 8890
-registrar_tls_port = 8891
-registrar_ip = 127.0.0.1
-cloudagent_ip = 127.0.0.1
-cloudagent_port = 9002
-provider_registrar_port = 8990
-provider_registrar_tls_port = 8991
-provider_registrar_ip = 127.0.0.1
-webapp_port = 443
-webapp_ip = 127.0.0.1
-cfssl_ip = 127.0.0.1
-cfssl_port = 8888
 
-# This is the port combination of the notification server.  The cloud verifier
-# will start a notification server automatically if the revocation_notifier
-# option is set to true
-revocation_notifier_ip = 127.0.0.1
-revocation_notifier_port = 8992
-
-# turn on or off TLS keylime wide
+# Turn on or off TLS keylime wide
 enable_tls = True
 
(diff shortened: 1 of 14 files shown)


rajdroid:
Hi @lukehinds could you please review these changes. And I was not able to work on '-c delete' bug that is mentioned because my VM is not able to connect to TPM, so I can't test. I will work on that issue after I resolve my setup.

lukehinds:
Hi Raj,

Just travelling back home, will take a look tomorrow

For tpm, try..

# tpm_serverd
# systemctl restart tpm2-abrmd

It might be underscore if that fails... tpm2_abrmd

Test with # tpm2_pcrread

rajdroid:
Sure @lukehinds whenever you get time. And I'll try your suggested method to resolve the issue.

rajdroid:
Hi @lukehinds, I tried tpm2_pcrread, but it fails as well. Interestingly, I don't find the TPM device in the VM. 
`% ll /dev/ | grep tpm`

jetwhiz:
Hi @rajdroid -- if you're doing this in a VM (and with the tpm2 emulator) then you will need to have tpm2_abrmd using mssim (the TPM will not be available as a hardware device). See: https://github.com/keylime/keylime#tpm-20-resource-manager for how to make this change.

rajdroid:
Hi @jetwhiz, Thanks, I'll try the shared resource, but I am using this resource (https://github.com/keylime/ansible-keylime-tpm-emulator) to provision VM, so I believe this should automatically be handled during provisioning.

lukehinds:
Another option while we fix this for you @rajdroid is to run within a container. Check out https://keylime-docs.readthedocs.io/en/latest/developers.html#docker-development-environment

This will pull down the images we use in CI and mount your local repository.

I would perhaps make a tweak and remove these lines:

```
docker stop "$(cat ${container_id})"
docker rm "$(cat ${container_id})"
```

You could then run `docker ps -a` and get the ID of the container and `docker attach $ID`, you should then be in a fully working environment. Note though that IMA won't work in this context as IMA is not namespaced yet, we need this work to complete:

http://kernsec.org/wiki/index.php/IMA_Namespacing_design_considerations

However http ports and running the keylime services should be fine.

lukehinds:
oh, and change 

```
tpm20tag="v301"
```

to

```
tpm20tag="v501"
```

rajdroid:
Thanks @lukehinds, let me try that

rajdroid:
Hi @lukehinds, docker way does worked, but I'm debugging why tpm2_abrmd is not working.

PR #255, opened 2020-02-26
SQLAlchemy port

lukehinds:
Main changes:

SQLAlchemy performs all CRUD operations and handles sessions.

Two model files are involved `verifier_db.sql` and `registrar_db.sql` 

A `keylime_database.sql` file handles sessions, it also uses `as_future` to provide asynchronous calls to the database when running under the verifiers async decorated functions.

Two config sections are now present in `keylime.conf` to allow database URL / string setting.

For example with postgresql:

```
drivername = postgres
username = keylime
password = allyourbase
host = localhost
port = 5432
database = keylime_db
query = ''
```

and the default using sqlite

```
drivername = sqlite
username = ''
password = ''
host = ''
port = ''
database = reg_data.sqlite
query = ''
```

The only manual sets ups are installing the database software (for everything outside of sqlite) and creating a user and database. SQLAlcemy will create all of the tables based on the model files `verifier_db.sql` and `registrar_db.sql` .

SQL calls are situated in `cloud_verifier_tornado.py` and `registrar_common.py` if someone is strongly convinced they need to be set within functions in `keylime_database.py` I can make that change. 

`metadata` has been renamed to `meta_data` as its a reserved row name in SQLAlchemy

You will also see some stuff using `ast` - this is to reformat lists into dicts as that is how SQLAlchemy is returning them. It works! But if someone smarter than me works out how to not require ast,  a patch is happily accepted.


lukehinds:
This is a bit of a mess, but it works. It is now being refactored to be less of a mess, so please no need to review just yet.

lukehinds:
silly me, unit test amendments incoming

keylime/cloud_verifier_tornado.py
```diff
@@ -50,7 +57,54 @@
 config = configparser.ConfigParser()
 config.read(common.CONFIG_FILE)
 
-class BaseHandler(tornado.web.RequestHandler):
+drivername = config.get('cloud_verifier', 'drivername')
+
+if drivername == 'sqlite':
+    database = "%s/%s%s" % (common.WORK_DIR,
+                            config.get('cloud_verifier', 'database'), '?check_same_thread=False')
+    print(database)
```
rajdroid: Syntactic change:
Maybe this is not required.

lukehinds:
@jetwhiz @nabilschear @amylily1011 @rajdroid 

This is complete now with added testing, please test locally and review.

keylime/ima.py
```diff
@@ -56,106 +56,116 @@
...
-            template['name']='ima'
-        template['desc-fmt']=defined_templates[template['name']]
+            template['name'] = 'ima'
+        template['desc-fmt'] = defined_templates[template['name']]
 
         #tokens = template['desc-fmt'].split('|')
 
-        import pdb; pdb.set_trace()
+        import pdb
+        pdb.set_trace()
```
rajdroid: Why we need debugger here?

lukehinds: Part of the old code reformatted by my IDE's linter... @jetwhiz , do we need pbd still? I imagine not.

jetwhiz: Yeah I expect this was just used for debugging the loading of the binary IMA list.

rajdroid: Ok, @lukehinds. Btw, I haven't started the code review. This caught my attention, that's why I asked the purpose of it. I will do the review in chunks.

lukehinds:
Just found an issue with delete and multiple agents, ready for review still, but not ready for merging.

keylime/ima.py
```diff
@@ -56,106 +56,113 @@
...
+supported_fields = {
+    "n": ima_eventname_parse,
+    "d-ng": ima_eventdigest_ng_parse,
+    "n-ng": ima_eventname_ng_parse,
+    "sig": ima_eventsig_parse,
+}
+
+
+def read_measreument_list_bin(path, whitelist):
     raise Exception("not implementated fully yet")
```
zedr: Typo

keylime/tenant.py
```diff
@@ -543,7 +542,7 @@ def do_cvdelete(self):
             deleted = False
             for _ in range(12):
                 response = httpclient_requests.request("GET", "%s"%(self.cloudverifier_ip), self.cloudverifier_port, params=params, context=self.context)
-                if response.status == 404:
+                if response.status == 200 or 404:
```
zedr: This doesn't look correct. You might want this instead:
`if response.status in (200, 404):`

lukehinds: looks good, thanks @zedr

keylime/ima.py
lukehinds: nice catch, thanks!

keylime/tpm_abstract.py
```diff
@@ -337,12 +346,17 @@ def __check_ima(self, agent_id, pcrval, ima_measurement_list, ima_whitelist):
         return True
 
     def check_pcrs(self, agent_id, tpm_policy, pcrs, data, virtual, ima_measurement_list, ima_whitelist):
+        try:
+            tpm_policy = ast.literal_eval(tpm_policy)
+        except ValueError:
+            pass
```
zedr: If the ValueError is caught here, the next lin will raise a NameError. It's better to raise or provide a default.

zedr: Sorry, my bad. `tpm_policy` is already defined in the function signature.

zedr: However, I would go with another name instead of reusing it. The problem is that `ast.literal_eval()` will rehydrate the string to whatever type it will evaluate to. If the exception is caught then `tpm_policy` will remain a `str` which does not have a `copy()` method. I'd go with:
```
try:
    tpm_policy = ast.literal_eval(tpm_policy)
except ValueError:
    tpm_policy_ = {} # If you expect it to be a dict
pcrWhiteList = tpm_policy_.copy()

File: keylime/cloud_verifier_tornado.py [hunk 6]
```diff
@@ -103,32 +163,33 @@ def get(self):
...
-            logger.warning('GET returning 400 response. uri not supported: ' + self.request.path)
+            logger.warning(
+                'GET returning 400 response. uri not supported: ' + self.request.path)
             return
 
         agent_id = rest_params["agents"]
 
         if agent_id is not None:
-            agent = self.db.get_agent(agent_id)
+            agent = session.query(VerfierMain).filter_by(agent_id=agent_id).one()
```
rajdroid: Can we use exception here?

try:
session query
except sqlalchemy.orm.exc.NoResultFound as e:

rajdroid: Otherwise use `one_or_none()`

keylime/cloud_verifier_tornado.py
```diff
@@ -249,48 +348,61 @@ def put(self):
...
                 common.echo_json_response(self, 400, "uri not supported")
                 logger.warning("PUT returning 400 response. uri not supported")
-
-            agent = self.db.get_agent(agent_id)
+            try:
+                agent = session.query(VerfierMain).filter_by(agent_id=agent_id)
+            except SQLAlchemyError as e:
+                logger.error(f'SQLAlchemy Error: {e}')
 
             if agent is not None:
```
rajdroid: The agent will always be `not None`. Because if it is not, then the above exception will be thrown.

rajdroid: Btw, why we are throwing 404 when we find an agent? Am I missing something?

jetwhiz: I'm thinking this is a typo and should be 
````python 
if agent is None:
````

File: keylime/cloud_verifier_tornado.py [hunk 8]
```diff
@@ -474,36 +633,50 @@ def main(argv=sys.argv):
 
     config = configparser.ConfigParser()
     config.read(common.CONFIG_FILE)
-
     cloudverifier_port = config.get('cloud_verifier', 'cloudverifier_port')
+    os.umask(0o077)
```
jetwhiz: Can we throw some comments in to this section to describe what's happening?

keylime/registrar_common.py
```diff
@@ -217,46 +277,63 @@ def do_POST(self):
...
+
             # Add values to database
-            d={}
-            d['ek']=ek
-            d['aik']=aik
-            d['ekcert']=ekcert
-            d['virtual']=int(ekcert=='virtual')
-            d['active']=int(False)
-            d['key']=key
-            d['tpm_version']=tpm_version
```
jetwhiz: Are we removing the `tpm_version` dictionary entry on purpose here?

keylime/registrar_common.py
```diff
@@ -359,93 +474,81 @@ def do_DELETE(self):
...
 
-
-    db = init_db("%s/%s"%(common.WORK_DIR,dbfile))
-
-    count = db.count_agents()
-    if count>0:
-        logger.info("Loaded %d public keys from database"%count)
-
-    server = ProtectedRegistrarServer(serveraddr, db, ProtectedHandler)
+    config = configparser.ConfigParser()
```
jetwhiz: Comments here as well would be helpful

keylime/tenant.py
```diff
@@ -543,7 +542,7 @@ def do_cvdelete(self):
             deleted = False
             for _ in range(12):
```
jetwhiz: Can we add a comment that this is a retry block (upon failure)?

keylime/tenant.py
```diff
@@ -543,7 +542,7 @@ def do_cvdelete(self):
...
                 response = httpclient_requests.request("GET", "%s"%(self.cloudverifier_ip), self.cloudverifier_port, params=params, context=self.context)
-                if response.status == 404:
+                if response.status in (200, 404):
```
jetwhiz: Do we want to match on `200`?  Realistically, a successfully deleted node should be returning `404` at this point I think?

keylime/tpm1.py
```diff
@@ -813,6 +813,18 @@ def check_quote(self,agent_id,nonce,data,quote,aikFromRegistrar,tpm_policy={},im
 
         return self.check_pcrs(agent_id,tpm_policy,pcrs,data,False,ima_measurement_list,ima_whitelist)
 
+    def sim_extend(self,hashval_1,hashval_0=None):
```
jetwhiz: Is this PR based on an outdated branch?  This patch should already be in `master` ...

keylime/tpm2.py
```diff
@@ -30,6 +30,7 @@
 import time
 import zlib
 import yaml
+import codecs
```
jetwhiz: Same as with `tpm1`, this patch should already be in `master`

keylime/tpm_abstract.py
```diff
@@ -300,6 +298,13 @@ def check_deep_quote(self, agent_id, nonce, data, quote, vAIK, hAIK, vtpm_policy
     def check_quote(self, agent_id, nonce, data, quote, aikFromRegistrar, tpm_policy={}, ima_measurement_list=None, ima_whitelist={}, hash_alg=None):
         pass
 
+    def START_HASH(self, algorithm=None):
```
jetwhiz: Same as `tpm1` and `tpm2`, this patch should already be in `master` ...

keylime/cloud_verifier_tornado.py
jetwhiz: I think this might be related to a `500` error I'm getting when running the WebApp.  When the WebApp asks for an agent that isn't in the Verifier yet (but is in the Registrar), this block is throwing an exception.  Agree we either need to handle this exception or use `one_or_none()` (or both).

keylime/registrar_common.py
```diff
@@ -68,60 +104,70 @@ def do_GET(self):
...
+            if agent.virtual:
+                response['provider_keys'] = agent.provider_keys
 
             common.echo_json_response(self, 200, "Success", response)
             logger.info('GET returning 200 response for agent_id:' + agent_id)
         else:
             # return the available registered uuids from the DB
-            json_response = self.server.db.get_agent_ids()
-            common.echo_json_response(self, 200, "Success", {'uuids':json_response})
+            json_response = session.query(RegistrarMain.agent_id).all()
```
jetwhiz: Note that this breaks our API, since this now returns a dictionary instead of a list, e.g., 
````javascript
[{'agent_id': 'XXXX'}]
````
vs. 

````javascript
['XXXX']
````

We should check to make sure this isn't also the case in other areas of the code, this is the first one I stumbled across.

File: keylime/cloud_verifier_tornado.py (hunk 8 above)
Comment (lukehinds) [reply keylime/cloud_verifier_tornado.py]: agree, sure

File: keylime/registrar_common.py (hunk 9 above)
Comment (lukehinds) [reply keylime/registrar_common.py]: We never retrieve this from the database, so I dropped it

File: keylime/registrar_common.py (hunk 10 above)
Comment (lukehinds) [reply keylime/registrar_common.py]: agree

File: keylime/tenant.py (hunk 11 above)
Comment (lukehinds) [reply keylime/tenant.py]: yep

File: keylime/registrar_common.py (hunk 16 above)
Comment (lukehinds) [reply keylime/registrar_common.py]: ok, interesting..this is why I used literal.eval for a similar query used in the verifier. I will try this out.

File: keylime/cloud_verifier_tornado.py (hunk 8 above)
Comment (lukehinds) [reply keylime/cloud_verifier_tornado.py]: Since `init_dn` has been removed we need to create the directory path for where the sqlite database is stored. I think this is better situated under the sqlite conditional. You will see this in my next commit.

File: keylime/tpm2.py (hunk 14 above)
Comment (lukehinds) [reply keylime/tpm2.py]: I am not following, yes its already in master so I performed a merge of master into this branch?

File: keylime/registrar_common.py (hunk 16 above)
Comment (lukehinds) [reply keylime/registrar_common.py]: How does this get called? I have set some type debug statements but can't replicate the conditions to make this part of the code get called?

File: keylime/tenant.py (hunk 12 above)
Comment (lukehinds) [reply keylime/tenant.py]: reverted.

File: keylime/tpm2.py (hunk 14 above)
Comment (jetwhiz) [reply keylime/tpm2.py]: It's strange that the diff is showing these changes as being new, since they are already merged into master ...

File: keylime/registrar_common.py (hunk 16 above)
Comment (jetwhiz) [reply keylime/registrar_common.py]: I noticed it because the WebApp uses this API to get information about the registered agents in the system.  The WebApp expects to be getting an array of UUIDs back (not an array of dictionaries), so the WebApp is failing right now.

Comment (lukehinds) [reply keylime/registrar_common.py]: understood, thanks @jetwhiz trying to debug this now.

Comment (jetwhiz) [reply keylime/registrar_common.py]: Just for reference, the actual error is occurring here: https://github.com/keylime/keylime/blob/6068e1023de334d46b13661a093d2c5b31592f07/keylime/tenant_webapp.py#L387-L389

The `agent` variable used to be a string, but is now a dictionary (resulting in an exception when trying to perform `agents[agent]`. 

Also note that the `one_or_none()` issue mentioned in a previous review also is breaking the WebApp, so that needs to be fixed as well before you make it to this issue.

Comment (lukehinds) [reply keylime/registrar_common.py]: could you try this if you already have an environment:
```
import ast
for agent in agent_list: 
     ast.literal_eval(agents[agent]) = await self.get_agent_state(agent_id) 
```

Comment (lukehinds) [reply keylime/registrar_common.py]: got the `one_or_none()` issue queued up.

Comment (jetwhiz) [reply keylime/registrar_common.py]: I can't even get it to install when doing that, it gives `SyntaxError: can't assign to function call`

Comment (lukehinds) [reply keylime/registrar_common.py]: Sorry, we might need another approach. literal_eval is more useful for converting to a dict. Just doing some port forwards to allow debug env.,

File: keylime/cloud_verifier_tornado.py (hunk 7 above)
Comment (rajdroid) [reply keylime/cloud_verifier_tornado.py]: ok

File: keylime/cloud_verifier_tornado.py (hunk 6 above)
Comment (rajdroid) [reply keylime/cloud_verifier_tornado.py]: Ok, I think there are two exceptions associated with this query: 1) one no entry is found, 2) multiple entries are found. `one_or_none()` will consume (1), but not the (2).

I think `one_or_none()` will be suffice.

File: keylime/registrar_common.py [hunk 17]
```diff
@@ -152,8 +152,9 @@ def do_GET(self):
         else:
             # return the available registered uuids from the DB
             json_response = session.query(RegistrarMain.agent_id).all()
+            return_response = [item[0] for item in json_response]
```
lukehinds: @jetwhiz this should fix the issue, from testing:

```
2020-03-27 13:05:29.159 - keylime.registrar-common - INFO - GET returning 200 response for agent_id list
result: ['D432FBB3-D2F1-4A97-9EF7-75BD81C00000']
result: <class 'list'>
2020-03-27 13:05:31.160 - keylime.registrar-common - INFO - GET returning 200 response for agent_id list
result: ['D432FBB3-D2F1-4A97-9EF7-75BD81C00000']
result: <class 'list'>
2020-03-27 13:05:33.136 - keylime.registrar-common - INFO - GET returning 200 response for agent_id list
result: ['D432FBB3-D2F1-4A97-9EF7-75BD81C00000']
result: <class 'list'>
```

lukehinds:
@jetwhiz 

[working[(https://i.imgur.com/S1y0jpr.png)

jetwhiz:
> 
> 
> @jetwhiz
> 
> working

Yep, this seems to be provisioning on my Ubuntu VM as well now, nice work!

keylime/cloud_verifier_tornado.py
```diff
@@ -249,48 +365,62 @@ def put(self):
...
+            if agent is None:
                 common.echo_json_response(self, 404, "agent id not found")
-                logger.info('PUT returning 404 response. agent id: ' + agent_id + ' not found.')
+                logger.info(
+                    'PUT returning 404 response. agent id: ' + agent_id + ' not found.')
 
             if "reactivate" in rest_params:
-                agent['operational_state']=cloud_verifier_common.CloudAgent_Operational_State.START
-                asyncio.ensure_future(self.process_agent(agent, cloud_verifier_common.CloudAgent_Operational_State.GET_QUOTE))
+                agent['operational_state'] = cloud_verifier_common.CloudAgent_Operational_State.START
```
jetwhiz: Does this part of the code work?  When an agent fails it executes this line and throws the exception: 
`TypeError: 'Query' object does not support item assignment`

It seems like this should be `agent.operational_state = ....`

lukehinds: I am not that my work station at the moment, but agree this should be `agent.operational_state` - are you able to check that, if not I can in the morning?

jetwhiz: It seems to fix the issue on my box at least (Ubuntu)

lukehinds: Made a ToDo to check this and push a commit tomorrow!

jetwhiz: Ok, make sure it works on your side as well.  I also haven't tested IMA functionality yet.

keylime/cloud_verifier_tornado.py
```diff
@@ -150,31 +222,50 @@ def delete(self):
...
         else:
-            self.db.update_agent(agent_id, 'operational_state',cloud_verifier_common.CloudAgent_Operational_State.TERMINATED)
-            common.echo_json_response(self, 202, "Accepted")
-            logger.info('DELETE returning 202 response for agent id: ' + agent_id)
-
+            try:
+                update_agent = session.query(VerfierMain).get(agent_id)
+                update_agent.operational_state = cloud_verifier_common.CloudAgent_Operational_State.TERMINATED
+                try:
+                    session.add(update_agent)
```
jetwhiz: What does the `add` method here do?  It seems like we're trying to update the operational state of an existing entry, so do we need to be doing some sort of update instead, or can `add` be overloaded this way?

lukehinds: Think of this as an update. Alchemy uses add for also updating row information. So we first of all  ORM Query object on line 254 and then set the operational state to failed for that object as `TERMINATED`  then reload the now updated object `update_agent` into the session.

There is an (albeit old) explanation [here](https://groups.google.com/forum/#!topic/sqlalchemy/ukM3UbHI5KA) and another [here](https://overiq.com/sqlalchemy-101/crud-using-sqlalchemy-orm/#updating-data)

jetwhiz: Ok, just making sure this wasn't a typo ... I'm still getting used to SQLAlchemy and how it works.

lukehinds: > Ok, just making sure this wasn't a typo ... I'm still getting used to SQLAlchemy and how it works.

Me too and the documents are not strongly opinionated. I hope in time others will refactor if there are better ways of doing things. can't see any performance impact at all from when I decorated with a timing function.

lukehinds:
actually there is an issue I just found testing revocation:

!()[blob:https://imgur.com/63379636-e2bb-423b-935d-3c3c43f8cc6d]

can you spot any type issues that are making this fail @jetwhiz ?

jetwhiz:
> 
> 
> actually there is an issue I just found testing revocation:
> 
> 
> 
> can you spot any type issues that are making this fail @jetwhiz ?

Ah, good catch ... I think it's because you renamed `metadata` to `meta_data`, so now this line is failing: 

https://github.com/keylime/keylime/blob/9db670e2d1e6a34cd3679253ae66e46d2ccf6bb1/keylime/ca_util.py#L394

lukehinds:
I think SQLAlchemy is changing types on me again. For some reason is see's meta_data as a str.

This is the result using this branch:

```
2020-04-03 11:05:56.501 - keylime.revocation_notifier - INFO - Lazy loading the revocation certificate from /var/lib/keylime/ca//RevocationNotifier-cert.crt
revocation: <class 'dict'>
revocation: {'type': 'revocation', 'ip': '127.0.0.1', 'agent_id': 'D432FBB3-D2F1-4A97-9EF7-75BD81C00000', 'port': 9002, 'tpm_policy': '{"22": ["0000000000000000000000000000000000000001", "0000000000000000000000000000000000000000000000000000000000000001", "000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001", "ffffffffffffffffffffffffffffffffffffffff", "ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff", "ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff"], "15": ["0000000000000000000000000000000000000000", "0000000000000000000000000000000000000000000000000000000000000000", "000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"], "mask": "0x408400"}', 'vtpm_policy': '{"23": ["ffffffffffffffffffffffffffffffffffffffff", "0000000000000000000000000000000000000000"], "15": ["0000000000000000000000000000000000000000"], "mask": "0x808000"}', 'meta_data': '{"cert_serial": 558554686486569297595124572146091421805223588446, "subject": "/C=US/ST=MA/L=Lexington/O=MITLL/OU=53/CN=D432FBB3-D2F1-4A97-9EF7-75BD81C00000"}', 'event_time': 'Fri Apr  3 11:05:56 2020'}
2020-04-03 11:05:56.504 - keylime.ca-util - ERROR - 'str' object has no attribute 'get'
  File "/usr/local/lib/python3.7/site-packages/keylime-1.2-py3.7.egg/keylime/ca_util.py", line 406, in cmd_listen
    revocation_notifier.await_notifications(revoke_callback,revocation_cert_path=cert_path)
  File "/usr/local/lib/python3.7/site-packages/keylime-1.2-py3.7.egg/keylime/revocation_notifier.py", line 140, in await_notifications
    callback(message)
  File "/usr/local/lib/python3.7/site-packages/keylime-1.2-py3.7.egg/keylime/ca_util.py", line 396, in revoke_callback
    serial = revocation.get("meta_data",{}).get("cert_serial",None)
AttributeError: 'str' object has no attribute 'get'
```

And master (without this branch merged):

```
020-04-03 10:39:46.205 - keylime.revocation_notifier - INFO - Lazy loading the revocation certificate from /var/lib/keylime/ca//RevocationNotifier-cert.crt
revocation: <class 'dict'>
revocation: {'type': 'revocation', 'ip': '127.0.0.1', 'agent_id': 'D432FBB3-D2F1-4A97-9EF7-75BD81C00000', 'port': 9002, 'tpm_policy': {'22': ['0000000000000000000000000000000000000001', '0000000000000000000000000000000000000000000000000000000000000001', '000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001', 'ffffffffffffffffffffffffffffffffffffffff', 'ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff', 'ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff'], '15': ['0000000000000000000000000000000000000000', '0000000000000000000000000000000000000000000000000000000000000000', '000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000'], 'mask': '0x408400'}, 'vtpm_policy': {'23': ['ffffffffffffffffffffffffffffffffffffffff', '0000000000000000000000000000000000000000'], '15': ['0000000000000000000000000000000000000000'], 'mask': '0x808000'}, 'metadata': {'cert_serial': 591295668187536534434898758038109642423527305509, 'subject': '/C=US/ST=MA/L=Lexington/O=MITLL/OU=53/CN=D432FBB3-D2F1-4A97-9EF7-75BD81C00000'}, 'event_time': 'Fri Apr  3 10:39:45 2020'}
2020-04-03 10:39:46.206 - keylime.ca-util - INFO - Revoking certificate: 591295668187536534434898758038109642423527305509
```

lukehinds:
@jetwhiz https://github.com/keylime/keylime/pull/255/commits/8477c37ed3d21464025272e50afb8b790fd3135f shown working here:

https://www.youtube.com/watch?v=B_44uP_J850

Issue #158, opened 2019-09-20
Release script


lukehinds:
Just one for me.

Develop a release script that updates all instances of versions, for example:

`setup.py` 
`version='x.x.x',`

PR #263, opened 2020-04-03
Added config for bump2version and fixed version across project

bu3alwa:
solves #158. Added a config for `bump2version` which can be installed through `pip`

It can be used as follows `bump2version --current-version 5.5.0 minor setup.py docker/*`
Replacing minor with major or patch will autoincrement accordingly. It can also be setup to tag and commit as well. I have it set to false right now but I can switch that if wanted. 

https://github.com/bu3alwa/keylime/blob/c3f6937d59ba3897f11704232b26037176749be8/setup.py#L77
Also, is the project considered in beta still? If not I can move the classifier to production instead.

Also if I missed someplace else where version is set please let me know.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -57,7 +57,7 @@ setup(
     # Versions should comply with PEP440.  For a discussion on single-sourcing
     # the version across setup.py and the project code, see
     # https://packaging.python.org/en/latest/single_source_version.html
-    version='1.2',
+    version='5.5.0',
 
     description='TPM-based key bootstrapping and system integrity measurement system for cloud',
     long_description=long_description,
@@ -74,8 +74,7 @@ setup(
 
     # See https://pypi.python.org/pypi?%3Aaction=list_classifiers
     classifiers=[
-        'Development Status :: 5.3.1 - Beta',
-
+        'Development Status :: 4 - Beta',
         # Indicate who your project is intended for
         'Intended Audience :: Developers',
         'Intended Audience :: System Administrators',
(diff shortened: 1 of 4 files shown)


lukehinds:
Interesting, so I would install bump2version locally, run it and then merge and make the release?

setup.py
```diff
@@ -74,8 +74,7 @@
 
     # See https://pypi.python.org/pypi?%3Aaction=list_classifiers
     classifiers=[
-        'Development Status :: 5.3.1 - Beta',
-
+        'Development Status :: 4 - Beta',
```
lukehinds, line 77: Should this also be 5.5.0?

bu3alwa: That is the list of classifier for pypi https://pypi.python.org/pypi?%3Aaction=list_classifiers. That just tells it its in beta. If we are out of Beta we should change it to `Development Status :: 5 - Production/Stable`

bu3alwa:
@lukehinds yea it's basically just a program that uses regex and replaces the strings as set in the config file. You can also make it tag and commit automatically for release.

setup.py
lukehinds: Got you, let's keep it as beta for now. We will likely hit production level with JWT is implemented, that will be a community decision though.

lukehinds:
> @lukehinds yea it's basically just a program that uses regex and replaces the strings as set in the config file. You can also make it tag and commit automatically for release.

sounds good, many thanks for picking this up!

Issue #272, opened 2020-04-13
Keylime agent fails to start with latest version of `tmp2-tools`


maugustosilva:
# Environment
 * OS / version: Ubuntu / 18.04.4
 * Processor architecture: Intel Xeon E312xx (Sandy Bridge)
 * TPM Manufacturer: SWTPM (https://github.com/stefanberger/swtpm)
 * Keylime version: 1.2
 * TPM2-Tools version: 4.2

# Description 
With the latest (top of git) version of `tpm2-tools`, keylime agent fails to start properly, due to the fact the initialization of the TPM module returns an error. The error seems easily identifiable. A change in the parameter on the `tpm2_createek` command line parameters, as demonstrated on the logs present here.

The main error message is: `Exception: createek failed with code 1: [b"tpm2_createek: invalid option -- 'p'\n"]`

It looks like the problem was introduced in `tpm2-tools` around the commit https://github.ibm.com/k8s4g/tpm2-tools/commit/6192d3cb100df526adc78a8764202547006e8139

Installing `tmp2-tools` with the commit id https://github.ibm.com/k8s4g/tpm2-tools/commit/c28932caef2036039901a91cf55eb7ff093c70f5 seems to be enough to restore functionality.

## Expected behavior vs. actual behavior 
Expected: keylime_agent is expected to start successfully
Actual: keylime_agent fails to run

## Steps to reproduce problem
1. Install (from git) the latest versions of `tpm2-tss` (https://github.com/tpm2-software/tpm2-tss.git), `tpm2-tools`(https://github.com/tpm2-software/tpm2-tools.git) and `tpm2-abrmd`(https://github.com/tpm2-software/tpm2-tools.git)
2. Install (from git) the latest version of Keylime.
3. Attempt to start the agent.

# Relevant logs
```
root@localhost:/keylime# keylime_agent
Using config file /etc/keylime.conf
2020-04-13 15:39:22.293 - keylime.tpm2 - INFO - TPM2-TOOLS Version: 4.2
2020-04-13 15:39:22.418 - keylime.tpm - WARNING - INSECURE: Keylime is using a software TPM emulator rather than a real hardware TPM.
2020-04-13 15:39:22.419 - keylime.tpm - WARNING - INSECURE: The security of Keylime is NOT linked to a hardware root of trust.
2020-04-13 15:39:22.419 - keylime.tpm - WARNING - INSECURE: Only use Keylime in this mode for testing or debugging purposes.
[...]

PR #273, opened 2020-04-14
Fixes incompatibility between keylime and tpm2-tools >= 4.2

maugustosilva:
This commit implements a fix for the keylime_agent when used with a more
recent (03/2020) version of tpm2-tools. At this point, this commit makes
the minimal amount of changes required to have keylime operational, by
simply defining a new "version point" (4.2) and adding the appropriate
conditionals in the multiple if-elif-else statements. Moving forward, a
cleaner refactoring (e.g. using a "command vs tpm2-tools version"
dictionary) looks like a more robust solution in the long term.

Resolves: #272

### DIFF
--- a/keylime/tpm2.py
+++ b/keylime/tpm2.py
@@ -136,7 +136,7 @@ class tpm2(tpm_abstract.AbstractTPM):
 
         if tools_version == "3.2":
             retDict = self.__run("tpm2_getcap -c algorithms")
-        elif tools_version == "4.0" or tools_version == "4.2" :
+        elif tools_version in ["4.0", "4.2"]:
             retDict = self.__run("tpm2_getcap algorithms")
 
         output = common.list_convert(retDict['retout'])
@@ -388,7 +388,7 @@ class tpm2(tpm_abstract.AbstractTPM):
             logger.info("Flushing old ek handle: %s"%hex(current_handle))
             if tools_version == "3.2":
                 retDict = self.__run("tpm2_getcap -c handles-persistent", raiseOnError=False)
-            elif tools_version == "4.0" or tools_version == "4.2":
+            elif tools_version in ["4.0", "4.2"]:
                 retDict = self.__run("tpm2_getcap handles-persistent", raiseOnError=False)
             output = retDict['retout']
             reterr = retDict['reterr']
@@ -469,7 +469,7 @@ class tpm2(tpm_abstract.AbstractTPM):
         if tools_version == "3.2":
             retDict = self.__run("tpm2_takeownership -c", raiseOnError=False)
             retDict = self.__run("tpm2_takeownership -o %s -e %s"%(owner_pw, owner_pw), raiseOnError=False)
-        elif tools_version == "4.0" or tools_version == "4.2" :
+        elif tools_version in ["4.0", "4.2"]:
             retDict = self.__run("tpm2_changeauth -c o %s"%(owner_pw), raiseOnError=False)
             retDict = self.__run("tpm2_changeauth -c e %s"%(owner_pw), raiseOnError=False)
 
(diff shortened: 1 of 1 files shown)


keylime/tpm2.py
```diff
@@ -429,6 +432,8 @@ def __create_ek(self, asym_alg=None):
                 command = "tpm2_getpubek -H 0x81010007 -g {asymalg} -f {ekpubfile} -P {ekpw} -o {opw} -e {epw}".format(**cmdargs)
             elif tools_version == "4.0":
                 command = "tpm2_createek -c - -G {asymalg} -u {ekpubfile} -p {ekpw} -w {opw} -P {epw}".format(**cmdargs)
+            elif tools_version == "4.2":
```
jetwhiz, line 435: I'm wondering if the `-p` EK password option isn't actually doing anything, even in v4.0, which is why it was removed in v4.2 (see: https://github.com/tpm2-software/tpm2-tools/issues/1917).

maugustosilva: Ah, good good... I didn't see the bug report (thanks) in `tpm2-tools`, but I concur, even from the commit reported on the opened issue, that it looks like the option `-p` does not have any effect even before 4.2. My main goal on this very first PR of mine for this project was to simply make the minimal amount of changes to get it back to functionality (and I don't have proper facilities for full, multi-version testing, was gonna ask you guys about it actually). As I mentioned before, I would be willing to attempt a second refactoring of it to have something like a "tpm2-tool vs version command map" (perhaps a python dictionary), but would argue in favor of making it on a separate PR. Would you agree with that?

lukehinds: @maugustosilva I did consider that before, so I concur a version map would be a sound approach. It would save us building a long list of conditionals each time a new set of tools come out.

maugustosilva: Just to confirm: you are ok with that (replacing the long list of conditionals) being done on a separate PR, correct?

jetwhiz: Yeah I agree that sounds reasonable, plus I think we need to look into the consequences of that password option not actually doing anything (and how to fix it).

keylime/tpm2.py
```diff
@@ -133,7 +136,7 @@ def __get_tpm_algorithms(self):
 
         if tools_version == "3.2":
             retDict = self.__run("tpm2_getcap -c algorithms")
-        elif tools_version == "4.0":
+        elif tools_version == "4.0" or tools_version == "4.2" :
```
elfosardo, line 139: tools_version in ["4.0", "4.2"] ?

keylime/tpm2.py
```diff
@@ -385,7 +388,7 @@ def __create_ek(self, asym_alg=None):
             logger.info("Flushing old ek handle: %s"%hex(current_handle))
             if tools_version == "3.2":
                 retDict = self.__run("tpm2_getcap -c handles-persistent", raiseOnError=False)
-            elif tools_version == "4.0":
+            elif tools_version == "4.0" or tools_version == "4.2":
```
elfosardo, line 391: ditto

keylime/tpm2.py
```diff
@@ -464,7 +469,7 @@ def __take_ownership(self, config_pw):
         if tools_version == "3.2":
             retDict = self.__run("tpm2_takeownership -c", raiseOnError=False)
             retDict = self.__run("tpm2_takeownership -o %s -e %s"%(owner_pw, owner_pw), raiseOnError=False)
-        elif tools_version == "4.0":
+        elif tools_version == "4.0" or tools_version == "4.2" :
```
elfosardo, line 472: ditto and same for the other elif-s where you have multiple versions

keylime/tpm2.py
ibmcb: done, pushed a new commit with this suggested modification (for all entries)

PR #277, opened 2020-04-17
Fedora 31 based CI image

### LINKED ISSUE #271: CI Fails on Fedora 31
For some reason unknown as yet, setUpModulee.GreenTestSuite fails on Fedora 31 (issue not present on Fedora 30).

elfosardo:
For some reason the systemd-udev package is not installed by
default in the base container image.
Because of that, we see this error when starting the tpm2-abrmd
service:
Failed to start tpm2-abrmd.service: Unit systemd-udev-settle.service not found.

Also commenting out starting condition that requires /dev/tmp* to
be present and not needed in CI.

Fix #271

### DIFF
--- a/.ci/run_local.sh
+++ b/.ci/run_local.sh
@@ -9,7 +9,7 @@ REPO=${KEYLIME_REPO_PATH:-${1:-/home/${USER}/keylime}}
 tpm12image="lukehinds/keylime-ci-tpm12"
 tpm12tag="v500"
 tpm20image="lukehinds/keylime-ci-tpm20"
-tpm20tag="v501"
+tpm20tag="v552"
 
 echo -e "Grabbing latest images"
 
--- a/.ci/test_wrapper.sh
+++ b/.ci/test_wrapper.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -x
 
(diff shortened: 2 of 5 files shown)


lukehinds:
@elfosardo could I request a couple of amends:

Increment the label in the Dockerfile: `LABEL version="5.5.1"` to 5.5.2 - coincidentally, this used to follow keylimes release numbering, but they have fallen out of sync now. I use this number for the [container tag number](https://hub.docker.com/layers/lukehinds/keylime-ci-tpm20/v552/images/sha256-796399f0bfa948dd9587b9177f3ca7dfedea72491376935552c437f420f08161?context=repo)

Notch the following up to 5.5.2 in `.travis.yml` (note its widely out of sync at present, this was from a revert). I think next time we fix up anything in the Dockerfiles, we could give the container its own release numbers as its never going to keep sync with Keylimes release numbering:

https://github.com/keylime/keylime/blob/fcd372bb5b39162894c9ca88f621cc79248bda9d/.travis.yml#L24

elfosardo:
@lukehinds this works as expected, I can add some words to explain and it will probably require some cleaning, but at least we have a working Fedora 31 image

PR #280, opened 2020-04-20
JWT Auth

lukehinds:


lukehinds:
Will likely close this , keeping it now for reference. 

Main work is now in: https://github.com/keylime/keylime/pull/309

PR #281, opened 2020-04-21
Separate setup config from setup code

elfosardo:
Use a setup.cfg file to ease managing setup configuration and add
a requirements file to simplify tracking dependencies, leaving
setup.py to take care of the simple setup execution.
Also start using pbr to manage versions.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,8 @@
+cryptography>=2.1.4 # BSD/Apache-2.0
+tornado>=5.0.2 # Apache-2.0
+m2crypto>=0.21.1 # MIT
+pyzmq>=14.4 # LGPL+BSD
+pyyaml>=3.11 # MIT
+simplejson>=3.8 # MIT
+requests>=2.6 # Apache-2.0
+sqlalchemy>=1.3 # MIT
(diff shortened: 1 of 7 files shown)


lukehinds:
`ModuleNotFoundError: No module named 'sqlalchemy'`

CI failure might be do to the container tag changes, will see if I can restart

elfosardo:
@lukehinds the failure at the moment is due to systemd-udev package missing in the image, but that's actually intended, please look at #282 , we will need that and a new Fedora 31 image :)

requirements.txt
```diff
@@ -0,0 +1,8 @@
+cryptography>=2.1.4 # BSD/Apache-2.0
```
lukehinds, line 1: It does make sense to start using a requirements.txt, been meaning to do this for a while. We will need to do some updates to add `pip3 install -r requirements`  to the readme and  installer.sh though as well:

https://github.com/keylime/keylime/blob/f4f53fad7b3058da8d04040a454d4d31eae6b473/installer.sh#L598

For the readme you will see those with a <ctr-f> "setup.py" (I can't permalink readme files as github renders them)

PR #284, opened 2020-04-23
Fix arg reference in the print_metadata.py

kaifeng:
The argument reference in print_metadata.execute was changed to
revocation, but the argument itself was not.

### DIFF
--- a/keylime/revocation_actions/print_metadata.py
+++ b/keylime/revocation_actions/print_metadata.py
@@ -35,5 +35,5 @@ config.read(common.CONFIG_FILE)
 
 logger = keylime_logging.init_logging('print_metadata')
 
-def execute(json_revocation):
+def execute(revocation):
     print(json.loads(revocation['meta_data']))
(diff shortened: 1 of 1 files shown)


lukehinds:
thanks @kaifeng !, I just noted this should be an async function (something I missed before). 

could you please add a `import asyncio` and async the function: `async def execute(revocation):`

kaifeng:
Hi @lukehinds, thanks for the review, and it seems I messed up with the git history.. the new commit appears to be still in couple of hours ago.
My past experience was working with gerrit, pull-request seems to be a different way.

lukehinds:
> Hi @lukehinds, thanks for the review, and it seems I messed up with the git history.. the new commit appears to be still in couple of hours ago.
> My past experience was working with gerrit, pull-request seems to be a different way.

no worries @kaifeng 

I remember the differences between gerrit and githubs flow. thanks for the patch and welcome to the community!

Issue #27, opened 2019-01-23
Improve tpm2.verify_ek to speed up cert matching


jetwhiz:
Currently we are brute forcing the entire list of signing certs by TPM manufacturers: 
  see https://github.com/mit-ll/python-keylime/blob/master/keylime/tpm2.py#L625

 ````python
for signer in trusted_certs:
    signcert = M2Crypto.X509.load_cert_string(trusted_certs[signer])
    signkey = signcert.get_pubkey()
    if ek509.verify(signkey) == 1:
        logger.debug("EK cert matched signer %s"%signer)
        return True
````

This can be improved by determining the correct CA to check against based on info from the EK certificate, saving the trouble of a brute force.

lukehinds:
keep this one

lukehinds:
keep alive

lukehinds:
keep

donky106:
Hi, I would like to work on this issue

lukehinds:
hi @donky106 , sure  - thanks.

Are you familiar with what needs doing?

One note, the code has changed a little since the issue was raised, mainly instead of having the certs in one big python file they are now pem files [resident here](https://github.com/keylime/keylime/tree/master/tpm_cert_store)

donky106:
Thanks for assigning. I appreciate that. 
 
currently you extract the CA=certificate authority, i.e. the signer from each certificate in the array [trusted_certs].

With a public key of that CA you verify the X.509 certificate.

To avoid a brute force trying each trusted certificate you would like to determine the correct CA first. And verify with the correct CA.

The X.509 certificates are PEM-encoded.

Am I right?

lukehinds:
Thanks @donky106 , you can ping @jetwhiz if you need to clarify anymore in the [gitter channel](https://gitter.im/keylime-project/community)

lukehinds:
@donky106 you will need a HW tpm to test this (the emulator has no EK). I can do this for you on a machine I have here (unless you have one yourself). Just add me to the PR review. Thanks again!

donky106:
I have no HW tpm to test. Could you please do this for me on your machine? Thanks.

Also I have got this error:
Adding message: Thanks for your PR. Please check the CI travis job to make sure the build passes and fix any failures. Many thanks for your contribution! to pull request 286
##[error]Resource not accessible by integration

But the CI travis job seems to be ok

PR #286, opened 2020-04-23
Improve tpm2.verify_ek to speed up cert matching

donky106:
Extract the issuer (CA) from the EK certificate, then search for the matching subject in the entire list of signing certs by TPM manufacturers. If found verify the EK certificate against this CA.

Resolves: #27


lukehinds:
hi @donky106 , still looking at this. I seem to have an issue with my machine i need to iron out.

lukehinds:
All tested against a HW tpm, thanks!

PR #287, opened 2020-04-24
[WIP] Correct duplicate travis phase

elfosardo:


elfosardo:
as I suspected the script to check changed files doesn't really work as intended and can be dangerous as it is
I'll see if that can be written differently

PR #288, opened 2020-04-28
Improve base http request

kaifeng:
Make timeout configurable from argument, and we can make it
configurable in the configuration file in the future.

Remove retry count as it doesn't work as expected, we can add retry
mechanism after we sort out which kind of status should be retried.

Use the same code path whether there is body in the request because it's
only a difference in the argument.

Also renamed a few arguments to reflect their real meaning to be less
confusing.

### DIFF
--- a/keylime/httpclient_requests.py
+++ b/keylime/httpclient_requests.py
@@ -1,53 +1,39 @@
 import http.client
 
-def request(method, url, port, params=None, data=None,context=None):
-    """
-    http client connection handler
+
+def request(method, host, port, params=None, data=None, context=None,
+            timeout=5):
+    """http client connection handler
 
     :param method: Contains the method type (e.g. GET, POST)
-    :param url: Is the connection IP (could use a rename to `ip`)
+    :param host: The host will be connected to
     :param port: The connection port
-    :param params: Is the url specifics (e.g /v1/agent/)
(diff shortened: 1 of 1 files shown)


keylime/httpclient_requests.py
```diff
@@ -1,53 +1,39 @@
 import http.client
 
-def request(method, url, port, params=None, data=None,context=None):
-    """
-    http client connection handler
+
+def request(method, host, port, url=None, data=None, context=None,
```
lukehinds, line 4: Thanks for the pull request @kaifeng !

I think you might have meant to change `url=None` to `params=None` (this is why CI failed).

kaifeng: Thanks @lukehinds , Ideally I think it's good to rename the params to url, but it appears the method are reference in many places, so will make a change at a later time.

lukehinds: I do agree, its currently not really params and more paths. 

The handler does need to rework. I am introducing an Authorization Bearer so we will need to be able to handle headers. So the whole logic of conditionals checking for `data` , ssl.context and headers needs to be re-thought. I might even move us over to using requests instead of http.client.

Thanks for your contribution

kaifeng: Agreed, the requests is a very good candidate, which is also widely used in the openstack, we can iterate though.

PR #292, opened 2020-04-29
Singletonize configuration reading

kaifeng:
Moves configuration loading into the common module and make
the config loaded a global instance.

Sticks to ConfigParser instead of mixed usage of ConfigParser and
RawConfigParser.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,8 @@ var/
 .installed.cfg
 *.egg
 MANIFEST
+AUTHORS
+ChangeLog
 
 # Installer logs
 pip-log.txt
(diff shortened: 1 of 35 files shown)


lukehinds:
Nice change, thank @kaifeng 

Could you just make the same update to following files, thanks!

```
grep -nr "config = configparser.RawConfigParse"
auto-ipsec/libreswan/src/local_action_update_crl.py:34:config = configparser.RawConfigParser()
auto-ipsec/libreswan/src/local_action_crashsa.py:39:config = configparser.RawConfigParser()
auto-ipsec/racoon/src/local_action_update_crl.py:34:config = configparser.RawConfigParser()
auto-ipsec/racoon/src/local_action_deletesa.py:40:config = configparser.RawConfigParser()
```

PR #294, opened 2020-04-30
Remove faulty script in travis conf

elfosardo:
We should not rely on the value of $? in sections to alter the behavior of travis, that could be dangerous as the commands are manipulated by special functions and they won't work as expected.

In the future, if we want to manipulate the travis workflow, we should use separate shell scripts.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,18 +3,6 @@ dist: bionic
 
 sudo: required
 
-# If the change is a markdown file, don't run CI build. Or at a later point
-# we can set a specific job in here (such as a markdown lint)
-before_install:
-- |
-    if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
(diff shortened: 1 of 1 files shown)


lukehinds:
The script did work well for a while and saved burning up CI builds when someone just made changes to markdown files. 

For the danger aspect, do you mean a shell execution attack? I am not sure how much we are exposed there, travis only has read only access to the repo.

Either way aside to the above, it would still be good to have a means to check its not a txt or md file change, as we don't need to run a full test against those PR types. Also I am interested in seeing if we can port to github actions, so if its a bit to hacky trying to filter out doc files from triggering builds, we could let this go.

elfosardo:
@lukehinds I totally agree.
It's dangerous mainly because it's not reliable.
I'll put this in draft and rework the script as shell script, not embedded in travis, that should work better.

lukehinds:
you could perhaps put it into `keylime\.ci` if that helps?

Issue #297, opened 2020-05-01
Unable to run `keylime_tenant -a add` when the verifier and registrar are running on two different machines


maugustosilva:
# Environment
OS / version: Ubuntu / 18.04.4
Processor architecture: Intel Xeon E312xx (Sandy Bridge)
TPM Manufacturer: SWTPM (https://github.com/stefanberger/swtpm)
Keylime version: 1.2
TPM2-Tools version: 4.2

# Description 
Have a setup with three nodes. Node "V" runs the `verifier`, node "R" runs the `registrar` and node "A" runs the `agent`. When attempted to "deploy" the already running agent on node "R" `keylime_tenant` reports and error.

### Expected: 
command `keylime_tenant -c add -t <ip address of A> -v <ip address of V> -u 64124A95-1E10-543E-9560-DC8BF962E60D --cert /var/lib/keylime/ca` should return exit code 0

### Actual: 
command fails with a (quite ugly) python error about ` 'int' object has no attribute 'read'`

## Steps to reproduce problem
1. Deploy verifier, registrar and agent on three different nodes (my personal opinion: the configurations required for this are, at this point, quite messy and undocumented)
2. Attempt to "add" an agent via `keylime_tenant`

# Relevant logs
```
root@css-flex14vm1:~# /usr/local/bin/keylime_tenant -c update -t <ip address of A> -v <ip address of V> -u 6D98723F-153F-5BD6-9097-152011DD3EE8 --cert /var/lib/keylime/ca
Using config file /etc/keylime.conf
2020-05-01 19:06:03.781 - keylime.tenant - INFO - Setting up client TLS in /root/keylime/verifier_ca/
2020-05-01 19:06:03.784 - keylime.tenant - INFO - TPM PCR Mask from policy is 0x408000
2020-05-01 19:06:03.784 - keylime.tenant - INFO - TPM PCR Mask from policy is 0x808000
Please enter the password to decrypt your keystore:
2020-05-01 19:06:07.023 - keylime.ca-util - INFO - Creating cert package for 6D98723F-153F-5BD6-9097-152011DD3EE8 in 6D98723F-153F-5BD6-9097-152011DD3EE8-pkg.zip
2020-05-01 19:06:07.117 - keylime.ca-util - INFO - Creating cert package for RevocationNotifier in RevocationNotifier-pkg.zip
[...]

PR #299, opened 2020-05-06
Fix for registrar and verifier running on different hosts

maugustosilva:
This commit fixes the inability of running `keylime_tenant` in a setup
where Keylime's Registrar and Verifier daemons run on separate hosts.
There is also an attempt at providing a clearer error message when the
registrar cannot be reached at the specified IP address, TCP port.

Resolves: #297

### DIFF
--- a/keylime/registrar_client.py
+++ b/keylime/registrar_client.py
@@ -121,6 +121,13 @@ def getKeys(registrar_ip,registrar_port,agent_id):
             return None
 
         return response_body["results"]
+
+    except AttributeError as e :
+        if response == 503 :
+            logger.critical("Error: the registrar is not availabled at %s:%s"%(registrar_ip, registrar_port))
+        else :
+            logger.exception(e)
+
     except Exception as e:
         logger.exception(e)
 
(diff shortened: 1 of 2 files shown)


keylime/registrar_client.py
```diff
@@ -121,6 +121,13 @@ def getKeys(registrar_ip,registrar_port,agent_id):
             return None
 
         return response_body["results"]
+
+    except AttributeError as e :
```
kaifeng, line 125: this doesn't look good as I think the httprequest.request doesn't do things consistently, which caused inconsistent handling in the caller, it should be fixed, doesn't need to be in this patch though.

keylime/tenant.py
```diff
@@ -387,7 +387,7 @@ def check_ek(self,ek,ekcert,tpm):
 
     def validate_tpm_quote(self,public_key,quote,tpm_version,hash_alg):
         registrar_client.init_client_tls(config,'tenant')
-        reg_keys = registrar_client.getKeys(self.cloudverifier_ip,self.registrar_port,self.agent_uuid)
+        reg_keys = registrar_client.getKeys(self.registrar_ip,self.registrar_port,self.agent_uuid)
```
lukehinds, line 390: Brilliant catch, we would never have noticed this from running everything over local host.

keylime/registrar_client.py
lukehinds: I fully expect a port in the registrar to requests. I have already moved over to requests in tenant code , will likely do registrar client next.

PR #302, opened 2020-05-07
Move major service entry & cli into keylime/cmd

kaifeng:
This includes verifier, registrar, agent, tenant, etc.

@lukehinds This is a minimum change to paint an idea, provider related code is not changed for it seems they are not maintained atm.


kaifeng:
Calling this[1] will fail, the argument number doesn't match, this make me think the platform_* might not be used or at lease not covered in the tests.

https://github.com/keylime/keylime/blob/master/keylime/provider_registrar.py#L34

lukehinds:
@kaifeng found an issue with `keylime_ca` (not looked into it yet)

```
/usr/local/bin/keylime_ca
  File "/usr/local/bin/keylime_ca", line 6, in <module>
    from keylime.cmd.ca_util import main
ModuleNotFoundError: No module named 'keylime.cmd.ca_util'
```

kaifeng:
@lukehinds Oops, the ca_util was renamed to ca but not changed in the script directive, thanks for verifying, looks like not covered by ci :(

lukehinds:
Yep, we should really add the keylime_ca to CI.

PR #303, opened 2020-05-08
Catch regex exception for invalid exclude list

font:
This change checks that the regular expression provided in an exclude
list can be successfully compiled before attempting to match paths
against it. Prior to this change, if the exclude list contained an
invalid regular expression, an exception was raised when attempting to
match a path with the exclude list regular expression. This exception
was not successfully handled triggering the verifier to crash that could
result in a denial of service attack.

@lukehinds identified this as being a potential DOS attack after we witnessed the verifier crashed during a joint keylime session.

### DIFF
--- a/keylime/ima.py
+++ b/keylime/ima.py
@@ -139,9 +139,15 @@ def process_measurement_list(lines, lists=None, m2w=None):
     else:
         whitelist = None
 
-    combined = None
+    compiled_regex = None
     if exclude_list != []:
-        combined = "(" + ")|(".join(exclude_list) + ")"
+        combined_regex = "(" + ")|(".join(exclude_list) + ")"
+        try:
+            compiled_regex = re.compile(combined_regex)
+        except re.error as regex_err:
+            msg = "Invalid regular expression '" + regex_err.pattern + "': "
+            msg += regex_err.msg + " at position " + str(regex_err.pos) + ". Exclude list will be ignored."
+            logger.error(msg)
 
     for line in lines:
         line = line.strip()
@@ -220,7 +226,7 @@ def process_measurement_list(lines, lists=None, m2w=None):
                 continue
 
             # determine if path matches any exclusion list items
-            if combined is not None and re.match(combined, path):
+            if compiled_regex is not None and compiled_regex.match(path):
                 logger.debug("IMA: ignoring excluded path %s" % path)
                 continue
 


font:
> One small recommendation, I think we should exit with a failure code around line # 151 if the regex is invalid. Otherwise a scenario could play out where machines fail on artefacts that were meant to be ignored. This way they get a chance to fix up their regex first and then try to attest again.

Are you recommending exiting the verifier program entirely i.e. `sys.exit(code)`, or just returning from the function? I'm assuming the latter as I don't think the former would remove the DOS attack surface. For the latter, right now returning a failure (indicated by returning `None`) propagates a `False` all the way back up to the verifier main loop `invoke_get_quote`. This results in the cloud agent operational state machine transitioning to an `INVALID_QUOTE` state and thus failing the agent anyway.

I had thought about how best to handle the invalid exclude list error. My thinking was we should force a failure of the node at some point, because it could also introduce another DOS where an invalid exclude list doesn't necessarily terminate the verifier, but prevents subsequent attestations that could be correctly failing the node. Therefore, I assumed malicious intent so that ignoring the exclude list if it's invalid was just as good of a reason to fail the node. Of course this forces the user to remove and re-add the node via the `keylime_tenant` utility.

An ideal solution might provide a grace period before failing the node possibly with an entirely new `CloudAgent_Operation_State` that suggests input data is invalid and the node will be failed after X amount of time. Currently this would have to get plumbed all the way back up.

lukehinds:
woops! made the wrong assumption thinking we were still operating within the context of the tenant at this juncture - so agree, in no way should we exit the verifier.

I have taken a little time to look into this a bit more now.

I would really prefer to avoid changing any op states or even attempting to start the attestation of the machine in the first place, if the regex is invalid. If something is invalid we should reject it as early as we can and give the user the chance to amend and then run again. 

The verifiers POST handler deals with the json sent by the tenant and that contains the whitelist. This is then loaded into the database on lines 331 to 336. This would be our best position to implement a gate check.

https://github.com/keylime/keylime/blob/b940b4a1a332deefdb4d20f4cf662f5566825a6f/keylime/cloud_verifier_tornado.py#L295-L336

What I am thinking is that a some juncture after we unmarshal (`json_body = json.loads(self.request.body)` we perform a validation of the regex, if this fails, we send a 400 back:

    common.echo_json_response(self, 400, "Invalid whitelist regex")

For how we do the validation, this might need some consideration. `re_compile` has an `is_valid` which will perform a lot of the checks for us, but we also might need a list that contains the `*` bomb we saw during our session together. 

I guess we could then go belt and braces and have new op state as you highlight in your last paragraph, however if we find a solid means pushing any mis-configurations back to the tenant / api client etc, that would be the ideal outcome. 

Hope you're ok working on this still! It's a bit more scope than we originally assumed :)

font:
@lukehinds Thanks for your thorough response! Finally had a chance to read it. :smile: I agree failing early would be best. I had originally thought the tenant would be the earliest, but later backtracked because the tenant utility is really just a reference implementation for that REST API - as you also mentioned. I can look at solving this through the verifier REST API.

I was thinking it might be good to still have these code changes as a fallback to prevent the verifier from failing if for some reason the exclude list was altered during execution, for example either accidentally by a developer bug, deliberately via some other security exploit, or via memory corruption. Therefore, what do you think about merging this PR and another can be created to add the other changes?

Issue #304, opened 2020-05-09
Incorrect text of verifier, should be agent


lukehinds:
# Environment
 * OS / version: N/A
 * Processor architecture: N/A
 * TPM Manufacturer:  N/A
 * Keylime version:  master

# Description 
 Incorrect use of verifer should be agent

## Expected behavior vs. actual behavior 

https://github.com/keylime/keylime/blob/db53181d7b2762d6bdb073dc587709cbce65bb2a/keylime/tenant.py#L690

Text should read as follows:

```
logger.error(f"Cannot connect to Agent at {self.cloudagent_ip} with Port {self.cloudagent_port}. Connection refused.") ```

PR #305, opened 2020-05-09
Incorrect text of verifier, should be agent 

vinaydawani:
Tried to fix this issue https://github.com/keylime/keylime/issues/304

Hope this is what you are looking for.


vinaydawani:
Thank you @lukehinds :)

PR #307, opened 2020-05-10
Perform clean up and archive old items

lukehinds:
The current `scripts` directory contains a lot of stuff we no
longer use, such as old test scripts and legacy vTPM scripts.

Lets archive this and use `scripts` for more utilised objects.

I also move create_whitelist.sh from keylime/keylime which is a
bit an odd home for it to reside and the tpm wrapper scripts.


scripts/create_whitelist.sh
```diff
@@ -0,0 +1,108 @@
+#!/bin/bash
```
kaifeng, line 1: I wonder why this file is not moved from keylime/
I thought move this to scripts/ too, but it seems some scripts depending on it, some are used by CI, some are used by other scripts that I don't have history background on.
If the rename is a two phase step, then it's OK.

lukehinds: This is just a useful script to generate whitelists for IMA. It unarchives an initamfs and iterates over each file with a sha checksum. It only has a reference in the demo scripts.

kaifeng: Ah, I remember I used this script in the evaluation, I guess there is no way to get the the checksum of boot aggregate? It looks like a runtime output, please forgive my limited knowledge on the IMA.

lukehinds: Do you mean get it before first boot? 

It might be the same between identical machines and OSes.

lukehinds: unless this might be of help:

http://linux-ima.sourceforge.net/linux-ima-measurements.html

kaifeng: Thanks for the link, noted. Yep, so the trust solutions as I am aware need profiling a trust node for the target state.

jetwhiz:
@lukehinds @kaifeng  I think there might be an issue with the installer due to removing this block: https://github.com/keylime/keylime/blob/610565d26bfc360ddf9cff39aba6461f70238aa5/installer.sh#L555-L557

Note that the `scripts/` folder here is not the Keylime scripts folder, but is the swtpm's scripts folder.  

This means that when we get down to this block: 

https://github.com/keylime/keylime/blob/610565d26bfc360ddf9cff39aba6461f70238aa5/installer.sh#L573-L586

The `tpm_serverd` and `init_tpm_server` scripts will not exist (since they were not copied over).

PR #309, opened 2020-05-12
Multi Tenancy Support

lukehinds:
This change introduces JWT (JSON WEB TOKENS) and the early foundation for multi tenancy. Multi tenancy will be delivered over several patches to make review easier. Currently it is recommend just to work with a single role `admin`.

The user sets environment variables `KL_API_USER` & `KL_API_PASS`

A `User` table is introduced which salts all user passwords (using `werkzeug`) and contains details on `role_id`, `group_id`, `username` etc.

The user sets a preferred HMAC

`jwt_dsa = HS512`

The user sets a HMAC passphrase used to sign JWT

`jwt_hmac_passphrase = changeme`

* Note, in a later patch RSA & EC etc will be available to use keys to sign a JWT. 

Every request towards the verifier now first requires a valid JWT token, as each of the handlers is protected (except for auth where the user authenticates themselves and is then provided with a token).

Tests are updated and so is the webUI.

* Note , creation of admin password will be handled by installer. `changeme` is just a placeholder and used for CI test

This change also introduces the python `request` library for calls towards the verifier, over time this will completely replace use of `httpclient`


lukehinds:
Will squash commits once approved,

THS-on:
@lukehinds do you want to try rebasing your changes on the latest master or should we close this PR?

lukehinds:
I am happy to close unless the community feels they want multi-tenancy? I will give it a week and if no one states its useful, then lets close.

THS-on:
@lukehinds we discussed this last Wednesday and there is currently no one that intends to use Keylime with multiple different tenants. This is still interesting for the future if someone wants to provide Keylime as a service, but it should be implemented when it is required and somebody wants to maintain it.

PR #311, opened 2020-05-12
Fixes requests to reactivate agent causes verifier exception

### LINKED ISSUE #310: reactivate command broken over type mismatch
When running the `keylime_tenant` with `reactivate` the following stack trace occurs:

```
2020-05-12 09:32:34.709 - keylime.cloudverifier - ERROR - can only concatenate str (not "int") to str 
  File "/usr/local/lib/python3.7/site-packages/keylime/cloud_verifier_tornado.py", line 499, in process_agent
    main_agent_operational_state = agent['operational_state']
  File "/usr/lib64/python3.7/site-packages/sqlalchemy/orm/query.py", line 3048, in __getitem__
    return list(self[item : item + 1])[0]  
TypeError: can only concatenate str (not "int") to str
```

The reason for this is because `agent` object is coming in a SQLAclhemy type and not a dict (or list?. So its not just as case of changing ` main_agent_operational_state = agent['operational_state']` to ` main_agent_operational_state = agent.operational_state`

If we do this, then the standard quote operation will fail

We might instead need to perform a type check and have a conditional to process accordingly.

Steps to recreate

1. Add an agent:
`keylime_tenant -v 127.0.0.1 -t 127.0.0.1 -u D432FBB3-D2F1-4A97-9EF7-75BD81C00000 -f somefile.txt -c add`

2. Reactive
`keylime_tenant -v 127.0.0.1 -t 127.0.0.1 -u D432FBB3-D2F1-4A97-9EF7-75BD81C00000 -f somefile.txt -c reactivate`

The above error will then throw.

kaifeng:
The agent passed to process_agent can be an orm obj (PUT)
or a dict obj (POST), convert the agent to dict if it's not.

Resolves #310

### DIFF
--- a/keylime/cloud_verifier_tornado.py
+++ b/keylime/cloud_verifier_tornado.py
@@ -105,6 +105,22 @@ exclude_db = {
 }
 
 
+def _from_db_obj(agent_db_obj):
+    fields = ['agent_id', 'v', 'ip', 'port',
+              'operational_state', 'public_key',
+              'tpm_policy', 'vtpm_policy', 'meta_data',
+              'ima_whitelist', 'revocation_key',
+              'tpm_version',
+              'accept_tpm_hash_algs',
+              'accept_tpm_encryption_algs',
+              'accept_tpm_signing_algs',
+              'hash_alg', 'enc_alg', 'sign_alg']
+    agent_dict = {}
+    for field in fields:
+        agent_dict[field] = getattr(agent_db_obj, field, None)
+    return agent_dict
+
+
 class BaseHandler(tornado.web.RequestHandler, SessionManager):
     def prepare(self):
         super(BaseHandler, self).prepare()
@@ -494,6 +510,10 @@ class AgentsHandler(BaseHandler):
                 agent, cloud_verifier_common.CloudAgent_Operational_State.GET_QUOTE))
 
     async def process_agent(self, agent, new_operational_state):
+        # Convert to dict if the agent arg is a db object
+        if not isinstance(agent, dict):
+            agent = _from_db_obj(agent)
+
         session = self.make_session(engine)
         try:
             main_agent_operational_state = agent['operational_state']


keylime/cloud_verifier_tornado.py
```diff
@@ -105,6 +105,23 @@
 }
 
 
+# TODO(kaifeng) Move this to db as well as other db objects
```
lukehinds, line 108: Could you make an issue please, rather then mark todo in the code (that way we have it trackable in the project board).

keylime/cloud_verifier_tornado.py
```diff
@@ -105,6 +105,23 @@
...
+              'ima_whitelist', 'revocation_key',
+              'tpm_version',
+              'accept_tpm_hash_algs',
+              'accept_tpm_encryption_algs',
+              'accept_tpm_signing_algs',
+              'hash_alg', 'enc_alg', 'sign_alg']
+    agent_dict = {}
+    for field in fields:
+        agent_dict[field] = getattr(agent_db_obj, field)
+    return agent_dict
```
lukehinds, line 122: Looks good, do you think we could benefit from a try / except here..perhaps check for KeyErrors?

keylime/cloud_verifier_tornado.py
kaifeng: ack, I will remove it on next push.

keylime/cloud_verifier_tornado.py
kaifeng: I don't worry about this, if it's a db object, it should have those fields, except there are other code path I am unaware of. To be defensive, we could add a default value though, it will not raise no matter a field exists or not.

kaifeng:
the ima_whitelist could be a performance issue, ~30M if memory serves. Maybe we should put it into another table.

PR #317, opened 2020-05-23
Move provider scripts into cmd directory

kaifeng:
All entry scripts will be under cmd.


lukehinds:
@kaifeng looks good, just some merge conflicts to resolve.

PR #319, opened 2020-06-03
Update verifier REST API to return error for invalid exclude list

### LINKED ISSUE #314: Handler for invalid exclude lists
As per discussion in https://github.com/keylime/keylime/pull/303 - we need a way for the verifier to validate an excludes list regexp, if invalid it should be rejected and a http response error sent back to the user.

font:
Fixes #314

Exclude list example:

```shell
[root@keylime keylime-dev]# cat ~/excludes.txt
/root/keylime-dev/.*
*
```

Error from the verifier:
```shell
2020-06-05 20:40:21.875 - keylime.cloudverifier - WARNING - Invalid regex: nothing to repeat. Exclude list regex is misformatted. Please correct the issue and try again.
```

Error from the tenant:
```shell
[root@keylime keylime-dev]# python keylime/cmd/tenant.py -v 127.0.0.1 -t 127.0.0.1 -tp 9002 -f /root/excludes.txt -u D432FBB3-D2F1-4A97-9EF7-75BD81C00000 --whitelist /root/whitelist.txt --exclude /root/excludes.txt -c add
Using config file /root/keylime-dev/keylime.conf
2020-06-05 20:40:20.539 - keylime.tenant - WARNING - CAUTION: using default password for private key, please set private_key_pw to a strong password
2020-06-05 20:40:20.539 - keylime.tenant - INFO - Setting up client TLS in /var/lib/keylime/cv_ca
2020-06-05 20:40:20.546 - keylime.tenant - INFO - TPM PCR Mask from policy is 0x408000
2020-06-05 20:40:20.546 - keylime.tenant - INFO - TPM PCR Mask from policy is 0x808000
2020-06-05 20:40:20.671 - keylime.ima - WARNING - No boot_aggregate value found in whitelist, adding an empty one
2020-06-05 20:40:21.876 - keylime.tenant - ERROR - Response code 400: Invalid regex: nothing to repeat. Exclude list regex is misformatted. Please correct the issue and try again.
2020-06-05 20:40:21.876 - keylime.tenant - ERROR - POST command response: 400 Unexpected response from Cloud Verifier: {'code': 400, 'status': 'Invalid regex: nothing to repeat. Exclude list regex is misformatted. Please correct the issue and try again.', 'results': {}}
```

### DIFF
--- a/test/test_restful.py
+++ b/test/test_restful.py
@@ -699,6 +699,40 @@ class TestRestful(unittest.TestCase):
         self.assertIn("port", json_response["results"], "Malformed response body!")
 
 
+    def test_034_cv_agent_post_invalid_exclude_list(self):
+        """Test CV's POST /v2/agents/{UUID} Interface"""
+        self.assertIsNotNone(self.V, "Required value not set.  Previous step may have failed?")
+
+        b64_v = base64.b64encode(self.V)
+        # Set unsupported regex in exclude list
+        ima_whitelist = { 'exclude': ['*'] }
+        data = {
+            'v': b64_v,
+            'cloudagent_ip': tenant_templ.cloudagent_ip,
+            'cloudagent_port': tenant_templ.cloudagent_port,
+            'tpm_policy': json.dumps(self.tpm_policy),
+            'vtpm_policy':json.dumps(self.vtpm_policy),
+            'ima_whitelist':json.dumps(ima_whitelist),
+            'metadata':json.dumps(self.metadata),
+            'revocation_key':self.revocation_key,
+            'accept_tpm_hash_algs': config.get('tenant','accept_tpm_hash_algs').split(','),
+            'accept_tpm_encryption_algs': config.get('tenant','accept_tpm_encryption_algs').split(','),
+            'accept_tpm_signing_algs': config.get('tenant','accept_tpm_signing_algs').split(','),
+        }
+
+        agent_data = json.dumps(data)
+        params = f"/v{self.api_version}/agents/{tenant_templ.agent_uuid}"
+        response = httpclient_requests.request("POST", "%s"%(tenant_templ.cloudverifier_ip),
+                        tenant_templ.cloudverifier_port, params=params,
+                        data=agent_data, context=tenant_templ.context)
+
+        self.assertEqual(response.status, 400, "Successful CV agent Post return code!")
+        json_response = json.loads(response.read().decode())
+
+        # Ensure response is well-formed
+        self.assertIn("results", json_response, "Malformed response body!")
+
+
 
     """Agent Poll Testset"""
 
--- a/keylime/cloud_verifier_common.py
+++ b/keylime/cloud_verifier_common.py
@@ -369,3 +369,15 @@ def notify_error(agent, msgtype='revocation'):
     else:
         tosend['signature'] = "none"
     revocation_notifier.notify(tosend)
+
+def validate_agent_data(agent_data):
+    if agent_data is None:
+        return False, None
+
+    # Validate exlude list contains valid regular expressions
+    lists = ast.literal_eval(agent_data['ima_whitelist'])
+    is_valid, _, err_msg = common.valid_exclude_list(lists.get('exclude'))
+    if not is_valid:
+        err_msg += " Exclude list regex is misformatted. Please correct the issue and try again."
+
+    return is_valid, err_msg
--- a/keylime/cloud_verifier_tornado.py
+++ b/keylime/cloud_verifier_tornado.py
@@ -314,6 +314,12 @@ class AgentsHandler(BaseHandler):
                     agent_data['sign_alg'] = ""
                     agent_data['agent_id'] = agent_id
 
+                    is_valid, err_msg = cloud_verifier_common.validate_agent_data(agent_data)
+                    if not is_valid:
+                        common.echo_json_response(self, 400, err_msg)
+                        logger.warning(err_msg)
+                        return
+
                     try:
                         new_agent_count = session.query(
                             VerfierMain).filter_by(agent_id=agent_id).count()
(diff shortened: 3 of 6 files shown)


lukehinds:
This is really good thanks @font 

It worked well in testing, but noted one thing we may be able to refine.

Is it doable to just report the regex that failed.

I tested with a large whitelist placing the `*` at the start, and the HTTP response (bottom) and verifier log (bottom) contained the whole whitelist:

Screenshot from 2020-06-04 11-59-38

font:
> Is it doable to just report the regex that failed.
> 
> I tested with a large whitelist placing the `*` at the end, and the HTTP response (bottom) and verifier log (bottom) contained the whole whitelist:

@lukehinds ah, good catch! That's not a very helpful error message if the exclude list is that large. In my testing I had a limited number of regular expressions. I'm not sure how large these exclude lists can get, but at the same time I would hope regular expressions would help make the exclude list more concise.

What we can do, at the cost of some performance, is attempt to compile every regular expression in the exclude list separately, and just report the ones that fail. However, if the entire regular expression is put on one line (i.e. each one separated by `|`) instead of separately on each line, we'd have to catch that and perform some splitting first -- all regular expressions on one line is perfectly valid today.

I'm not sure what sort of testing framework we have for keylime, but it'd be great to have some tests for stuff like this. Ideally, there would be e2e and/or at least unit tests that handled all these corner cases so that we could automate the testing for this.

lukehinds:
> but at the same time I would hope regular expressions would help make the exclude list more concise.

I very much agree, this is an edge case so I don't think we should sweat over this one for long. 
I have entries like this:

```
fb01d3787135d1fbc1c333f02edcf5c9ff59f864f3f4d3e51a1cc0315b764bcd  /etc/fonts/conf.d/65-nonlatin.conf
0fe8db4f86a29df2c348b009a26aeb1f7b0686315e7310e3301e1d83466af23b  /etc/fonts/conf.d/25-no-bitmap-fedora.conf
1a17eaeb44bdc367f16939dfb2ca1a00a57a9b788b9057a80c656667fa41721e  /etc/fonts/conf.d/45-latin.conf
54a5935ea79c66e07faf743b81f2647eca849800b2b0422e84cd66527a75066a  /etc/fonts/conf.d/90-synthetic.conf
1fac58aa3c865e9b05088f94f7a4684d8a60d134421690cfd04842eb00fe1332  /etc/fonts/conf.d/10-hinting-slight.conf
```

If a user entered this as a bug, we would tell them to use `\etc/fonts/conf.d/*`. I would rather educate users then take a performance hit. The main thing for me is that we keep badly formatted regex out, and you have achieved that goal with this patch.

Is it possible to just return `Exclude list regex is misformatted. Please correct the issue and try again`, but without the regex, but without the contexts of the exclude list e.g. `'(/root/keylime-dev/.*)|(*)'`

We can then if anything supply a script for people to validate locally, or they can go somewhere like https://regex101.com

I think we should be good to merge then!

font:
> If a user entered this as a bug, we would tell them to use `/etc/fonts/conf.d/*`. I would rather educate users then take a performance hit. The main thing for me is that we keep mangled regex out, and you have achieved that goal with this patch.
> 
> Is it possible to just return `Exclude list regex is misformatted. Please correct the issue and try again`, but without the regex, e.g. remove `'(/root/keylime-dev/.*)|(*)'` to use your example:

@lukehinds I updated the error message per your suggestion. I still kept the simple error message e.g. `nothing to repeat` in this case, that may provide some hints to the user without supplying the entire regex or the exact location. I updated the original sample outputs above to show the new changes. Let me know if you're good with this.

font:
@lukehinds Any other changes you'd like to see here?

lukehinds:
I am good to merge this @font , but need to work out why its failing CI tests. I will try to take a look.

font:
@lukehinds No worries, I will investigate the CI test failure.

font:
@lukehinds I've fixed the CI issue and have added a test for an invalid exclude list. The tests are passing now, but I'm not sure why the travis CI build link is missing...here it is https://travis-ci.org/github/keylime/keylime/builds/708992570. This is ready for review/merge.

test/test_restful.py
```diff
@@ -699,6 +699,40 @@ def test_033_cv_agent_get(self):
         self.assertIn("port", json_response["results"], "Malformed response body!")
 
 
+    def test_034_cv_agent_post_invalid_exclude_list(self):
+        """Test CV's POST /v2/agents/{UUID} Interface"""
+        self.assertIsNotNone(self.V, "Required value not set.  Previous step may have failed?")
+
+        b64_v = base64.b64encode(self.V)
+        # Set unsupported regex in exclude list
+        ima_whitelist = { 'exclude': ['*'] }
```
lukehinds, line 708: Is this right? You're loading the exclude list into the whitelist?

font: This is how the CV REST API currently works. That is, the CV expects an `ima_whitelist` entry that consists of 2 list entries: a `whitelist` and `exclude` entry for each list. This code is just setting it up according to that. The tenant does something similar shown in the below snippets. We should probably consider separating both lists out into separate fields in the future.
- https://github.com/keylime/keylime/blob/4b7be7c56ab1b2e549e3d470710fb0f74780b197/keylime/ima.py#L273
- https://github.com/keylime/keylime/blob/4b7be7c56ab1b2e549e3d470710fb0f74780b197/keylime/tenant.py#L230-L231
- https://github.com/keylime/keylime/blob/4b7be7c56ab1b2e549e3d470710fb0f74780b197/keylime/tenant.py#L478

lukehinds:
Hey Ivan

I tested this out but we still have an issue. The invalid expression is picked up, but it still proceeds to move into measurement of the node. I captured a screenshot for you:

Screenshot from 2020-07-24 12-48-51

Happy to work on this with you, I have not debugged it yet, but will try to find some time to look into it and help out

lukehinds:
nice work @font !

Issue #313, opened 2020-05-15
Keylime agent should be capable of simply using the already created EK in a TPM


maugustosilva:
# Description 
During startup, `keylime_agent` is pretty forceful and uncompromising on "taking over" the TPM device, including a "startup & clear (`tpm2_startup -c`), "change owner, endorsement and lockout authorizations" (`tpm2_changeauth`) and "generate a new Endorsement Key" (`tpm2_createek`). This means that, when the agent submits an EK to registrar, there isn't a specific way for an administrator to tie a particular (EK) to a specific piece of hardware (e.g., whose "chain of provenance" started at a factory, and included an "air-gapped boot measurement station" before going into production). 

I would like explore and discuss the possibility of changing keylime to allow a "primed" TPM to be controlled by the `keylime_agent`. Almost needless to say, this would be an optional behavior controlled by a new configuration parameter. In this scenario, the agent would query the TPM and simply use the pre-existing EK (which was created beforehand by some other method). The plan here wouldn't require us to *prevent* the `keylime_registrar` from recording any EK submitted (as it is today), but rather to make the (`tenant`-side) script `ek_check_script` to be able to check the EK against a list of properly pre-defined (and previously collected) values.

# Goal
Add a new `keylime.conf` option under `[cloud_agent]` specifying `reinitialize_tpm=True|False`

Am I off the mark here, or is there a specific, real-world, use case such as described?

lukehinds:
hi @maugustosilva 

Just a cursory note for now and will reply in full at a later point.

There are quite a few synergies with the following enhancement:

https://github.com/keylime/enhancements/blob/29be79363f96c7c622497bffb22f87003ce835be/1_persist_agent_restart.md

I wonder if we could combine both.

My initial thoughts are I am open to this and see it as a valid case, but we should also try to keep the current approach as an option for a user who is getting a new machine and wants to be sure its refreshed for their use.

maugustosilva:
@lukehinds apologies for not responding sooner. I have read the document, and I believe I understand your requirements. It looks like that the basic logic in `tpm2.py` for both cases will be exactly the same, with the only difference being the criterion to decide when to use the already created EK (should we preserve AIK too?): in my case, it will be if a given flag is found on `keylime.conf`, in yours, it is base on a result from an early REST GET call. Do you agree?

PR #320, opened 2020-06-03
Agent can just "use" an already created EK

maugustosilva:
This commit is a first attempt at giving keylime de ability to use an
    already generated Endorsement Key (EK) on a TPM. The use case and
    rationale for this is well described on #313.

    A new option is introduced under `[cloud_agent]`, name `ek_handle`. When
    the default value of `generate` (wanted to keep the familiar term
    already present on other options, such as `tls_dir`) is found, the
    normal agent startup process, which includes clearing/start, ownership,
    flush/evict and EK creation, will be executed. If `ek_handle` points to
    an actual value (e.g., `0x81000000`), then all the initial steps
    previously described will be skipped and the agent will just start from
    the "create AIK" step. In the latter scenario, the agent will make sure
    that particular handle is not flushed on exit.

    Resolves: #313


lukehinds:
@jetwhiz could you take a look too, see #313 for context.

lukehinds:
how could I try this out @maugustosilva  ?

maugustosilva:
apologies for the long delay answering @lukehinds. Here is how I've been testing..

Step 1: Fully boot a node, and then issue the following commands against a "clean" (v)TPM:

```
tpm2_startup --version
tpm2_startup -c
tpm2_changeauth -c o temp4now
tpm2_changeauth -c e temp4now
tpm2_changeauth -c o -p temp4now temp4now
tpm2_changeauth -c e -p temp4now temp4now
tpm2_getcap handles-persistent
tpm2_createek -c - -G rsa -u /tmp/test -p rsxBpUzLpR28mHPlC5NN -w temp4now -P temp4now
tpm2_getcap handles-persistent
```

Step 2: get the handle from the last command in step 1 (e.g., `0x81000000`), and add it under the section `[cloud_agent`] in `/etc/keylime.conf`

```
...
[cloud_agent]
...
ek_handle = 0x81000000
...
```

Step3: start `keylime_agent`

lukehinds:
Nice! 

If you can, it would be nice to have a section over at keylime/keylime-docs , something I guess like "Agent setup for nodes with existing EK"

maugustosilva:
@lukehinds Documentation PR submitted: https://github.com/keylime/keylime-docs/pull/51

keylime.conf
```diff
@@ -103,6 +103,11 @@ tpm_hash_alg = sha256
 tpm_encryption_alg = rsa
 tpm_signing_alg = rsassa
 
+# If an EK was already created on the TPM (e.g., with "tpm2_createek") and
+# keylime should just use it, just change from "generate" to the actual
```
lukehinds, line 107:
```
# If an EK is already present on the TPM (e.g., with "tpm2_createek") and
# you require keylime to use this EK, change "generate" to the actual EK 
# handle (e.g. "0x81000000"). The keylime agent will then not attempt to 
# create a new EK upon startup, and neither will it flush the EK upon exit
ek_handle = generate
```

> Hope the above changes are ok with you.

PR #323, opened 2020-06-14
Simplify database configuration

kaifeng:
Make the URL configurable so that we don't need to set each part.

I didn't see the the TEST_MODE used in CI, so if it's still used by some one locally, it could be easily resolved by changing the db_connection configuration.

### DIFF
--- a/keylime/registrar_common.py
+++ b/keylime/registrar_common.py
@@ -30,33 +30,13 @@ from keylime.db.registrar_db import RegistrarMain
 from keylime.db.keylime_db import SessionManager
 from sqlalchemy.exc import SQLAlchemyError
 from sqlalchemy import create_engine
-from sqlalchemy.engine.url import URL
+
 
 logger = keylime_logging.init_logging('registrar-common')
 # setup config
 config = common.get_config()
 
-drivername = config.get('registrar', 'drivername')
-
-
-if drivername == 'sqlite':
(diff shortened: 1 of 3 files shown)


keylime/registrar_common.py
```diff
@@ -30,33 +30,13 @@
...
-    )
-else:
-    url = URL(
-        drivername=drivername,
-        username=config.get('registrar', 'username'),
-        password=config.get('registrar', 'password'),
-        host=config.get('registrar', 'host'),
-        database=config.get('registrar', 'database')
-    )
+url = config.get('registrar', 'db_connection')
```
lukehinds, line 39: This won't work, as it only supports SQLite style strings. Anyone wanting to use postgres or mariadb won't be able to.

lukehinds: In its current format it supports all backends, but this PR hard codes a sqlite connection string:

`db_connection = sqlite:////var/lib/keylime/cv_data.sqlite`

It also removes the fields needed by mariadb / postgres etc , such as `username`, `password` and `host`. 

https://keylime-docs.readthedocs.io/en/latest/installation.html#database-support

kaifeng: Oh, this depends on the configuration option, if we want to use mysql, we can set this, for example: mysql://user:pwd@hostname:port/keylime?charset=utf8
So the db_connection here just provides a default value.

PR #328, opened 2020-06-30
WebApp running on a nodes separate from registrar and verifier

maugustosilva:
This commit fixes the inability of running `keylime_tenant` in a setup
where Keylime's Registrar and Verifier daemons run on separate hosts,
as introduced in PR #299.
There is also an attempt at providing a clearer error message when the
registrar cannot be reached at the specified IP address, TCP port.


keylime/tenant.py
```diff
@@ -100,49 +100,50 @@ def __init__(self):
...
-        context.load_cert_chain(
-            certfile=my_cert, keyfile=my_priv_key, password=my_key_pw)
-        context.verify_mode = ssl.CERT_REQUIRED
-        context.check_hostname = config.getboolean(
-            'general', 'tls_check_hostnames')
-        return context
+
+        contexts = []
+        for component in [ '', 'registrar_' ] :
+            ca_cert = config.get('tenant', component + 'ca_cert')
```
lukehinds, line 114: This works ok during my tests so I think we are close to merging, one thing I wanted to check. Does this change require new values in `keylime.conf` ? I am trying to grok what `('tenant', component + 'my_cert')` is doing (mainly the component part?

maugustosilva: No new values required at all. Those are the certs that already exist on `keylime.conf`. Due to an unfortunate (seemingly) historical accident it looks like every verifier-related component (e.g., `my_cert`) does not have a prefix, while registrar-related components (e.g., `registrar_my_cert`) have a proper prefix. Please note that this is true for the sections `[cloud_verifier]` and `[registrar]`, but more importantly to our purposes, this is also true of the `[tenant]` section.

Issue #327, opened 2020-06-23
Keylime project proposal to the CNCF


lukehinds:
Opening this issue for information. I had an introduction call with the CNCF around project adoption of Keylime.

They are very receptive to this and recommend we go forward with a proposal. 

This issue is to inform you as a community. We don't have a formal voting process or governance, so I will go ahead based on lazy consensus, unless someone can communicate a very clear merited objection, which we can then discuss.

https://www.cncf.io/wp-content/uploads/2018/07/cncf-member-benefits-final.pdf

lukehinds:
No opposition or views shared, so this will go ahead.

PR #329, opened 2020-07-13
Port License to Apache 2.0

lukehinds:
As part of the proposal to join the CNCF, this change replaces
license instances of BSD-2-Clause to Apache-2.0

Resolves: #327


lukehinds:
@jetwhiz , over to you. Let me know when we can merge (just approve as a reviewer).

jetwhiz:
@lukehinds  Do we need to make any changes to the LICENSE file to show the original and new licenses?

lukehinds:
How could I have missed the `LICENSE` file, I will make a push. For mentioning the original license we can do so in the readme ?

jetwhiz:
@lukehinds  Do we also want to have a NOTICE file included now (as mentioned in https://www.apache.org/licenses/LICENSE-2.0#apply) 

This will allow us to list out the various copyrights in Keylime, such as the MIT copyrights, your copyrights (e.g., https://github.com/keylime/keylime/blob/master/keylime/db/keylime_db.py#L3), and others as they come.

lukehinds:
> @lukehinds Do we also want to have a NOTICE file included now (as mentioned in https://www.apache.org/licenses/LICENSE-2.0#apply)
> 
> This will allow us to list out the various copyrights in Keylime, such as the MIT copyrights, your copyrights (e.g., https://github.com/keylime/keylime/blob/master/keylime/db/keylime_db.py#L3), and others as they come.

As I understand it a NOTICE is only needed for when distributing the actual Apache software, not using the apache license. We have individual copyrights in each file, so I think that should suffice?

http://www.apache.org/legal/src-headers.html#notice

Here is an example using Openstack:

https://github.com/openstack/nova/blob/c6218428e9b29a2c52808ec7d27b4b21aadc0299/nova/api/openstack/compute/helpers.py#L1

https://github.com/openstack/nova/blob/master/LICENSE

jetwhiz:
> 
> 
> > @lukehinds Do we also want to have a NOTICE file included now (as mentioned in https://www.apache.org/licenses/LICENSE-2.0#apply)
> > This will allow us to list out the various copyrights in Keylime, such as the MIT copyrights, your copyrights (e.g., https://github.com/keylime/keylime/blob/master/keylime/db/keylime_db.py#L3), and others as they come.
> 
> As I understand it a NOTICE is only needed for when distributing the actual Apache software, not using the apache license. We have individual copyrights in each file, so I think that should suffice?
> 
> http://www.apache.org/legal/src-headers.html#notice
> 
> Here is an example using Openstack:
> 
> https://github.com/openstack/nova/blob/c6218428e9b29a2c52808ec7d27b4b21aadc0299/nova/api/openstack/compute/helpers.py#L1
> 
> https://github.com/openstack/nova/blob/master/LICENSE

That makes sense, thanks for clarifying @lukehinds !

PR #338, opened 2020-07-24
setup: Remove PBR once more

ashcrow:
Building seems to work. However, I'd appreciate someone who is more in tune with the project to verify the results are correct :pray: 

PBR is a very helpful tool. However, PBR is not available as a build
dependency in all cases. Here we switch back to the more traditional
in-line setup.py while attempting to honor the the content that
was in setup.cfg.

References:
- PBR added 1ae26fdabb30c54bbf08d8621c52e16020583c10
- Attempted removal cb42ba3b4295f8b0171209f70f1cf2513096150f
- Revert 401d4f9e75be361f91f9a12f004b633d36a0089e


setup.py
```diff
@@ -21,8 +22,53 @@
...
+    author_email='keylime@groups.io',
+    url='https://keylime.dev',
+    python_requires='>=3.6',
+    packages=['keylime'],
+    classifiers=[
+        'Environment :: Console',
+        'Intended Audience :: Developers',
+        'Intended Audience :: Information Technology',
+        'Intended Audience :: System Administrators',
+        'License :: OSI Approved :: BSD License',
```
lukehinds, line 47: This looks good to me @ashcrow , we recently changed to using Apache 2.0, could you just swap this out and we can land this PR..

ashcrow: Switched to `License :: OSI Approved :: Apache Software License` per https://pypi.org/pypi?%3Aaction=list_classifiers

PR #343, opened 2020-09-03
first tentative fix #341

### LINKED ISSUE #341: Inconsistency between IMA logs and PCR10
@galmasi
@maugustosilva 

# Environment
 * OS / version: Ubuntu 18.04.3
 * TPM Manufacturer: swtpm
 * Keylime version: 5.7

# Description 
Keylime is ignoring the fact that IMA logs and PCR 10 are not always in sync, as on the agent side they are retrieved at different times, as requested by the verifier.

The code section:

https://github.com/keylime/keylime/blob/0a921fa48b768d946ea58dbe7674786186518a35/keylime/tpm/tpm_abstract.py#L336-L344

assumes PCR10 and IMA logs to be synchronized when received by the verifier, i.e. it assumes the latest PCR10 received always equals the aggregated hash of IMA measurements up to the last point of the IMA logs just received. However this is not true in general and the test sporadically fails.

## Steps to reproduce problem
Just run a regular Keylime deployment for some time, with a non-trivial number of nodes. We started seeing the issue with 8 nodes or so.

# Proposed Solution
Drop the assumption of PCR10 and IMA logs to be in sync and consider that a given PCR10 value can match any entry in:
- the IMA logs received along with that PCR10 value
- the previous batch of IMA logs (although this will be rare, it cannot be excluded)

droccom:
@mpeters @galmasi @maugustosilva

Please review and comment.

### DIFF
--- a/keylime/ima.py
+++ b/keylime/ima.py
@@ -236,9 +236,7 @@ def process_measurement_list(lines, lists=None, m2w=None, pcrval=None):
 
     # check PCR value has been found
     if not found_pcr:
-        logger.error(
-            "IMA measurement list expected pcr value %s does not match TPM PCR %s"
-            %(codecs.encode(runninghash, 'hex').decode('utf-8'), pcrval))
+        logger.error("IMA measurement list does not match TPM PCR %s" % pcrval)
         return None
 
     # clobber the retval if there were IMA file errors
--- a/keylime/ima.py
+++ b/keylime/ima.py
@@ -114,9 +114,10 @@ def read_measurement_list_bin(path, whitelist):
         #tokens = template['desc-fmt'].split('|')
 
 
-def process_measurement_list(lines, lists=None, m2w=None):
+def process_measurement_list(lines, lists=None, m2w=None, pcrval=None):
     errs = [0, 0, 0, 0]
     runninghash = START_HASH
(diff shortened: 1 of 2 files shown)


lukehinds:
Nice simple fix. I will give this a test run, but its looking like this can be merged. 

@galmasi @maugustosilva , do we want @mpeters to roll a release right after that you can scoop up in your deb package?

keylime/ima.py
```diff
@@ -229,6 +234,13 @@ def process_measurement_list(lines, lists=None, m2w=None):
 
         errs[3] += 1
 
+    # check PCR value has been found
+    if not found_pcr:
+        logger.error(
+            "IMA measurement list expected pcr value %s does not match TPM PCR %s"
+            %(codecs.encode(runninghash, 'hex').decode('utf-8'), pcrval))
```
droccom, line 241: on second thought, this error message looks off, as there is no "expected pcr value".

```suggestion
        logger.error("IMA measurement list does not match TPM PCR %s" % pcrval)
```

lukehinds (approved):
This worked in testing, so I am OK to merge. @mpeters / @jetwhiz , will see if you have any comments.

mpeters (approved):
lgtm

Issue #344, opened 2020-09-04
Postgres does not support connection string check_same_thread


lukehinds:
# Environment
 * OS / version:  fedora 32 / postgres
 * Processor architecture:  N/A
 * TPM Manufacturer: N/A
 * Keylime version: master

# Description 

Database not working with postgres

```
The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/local/bin/keylime_registrar", line 11, in <module>
    load_entry_point('keylime==0.0.0', 'console_scripts', 'keylime_registrar')()
  File "/usr/local/lib/python3.8/site-packages/keylime-0.0.0-py3.8.egg/keylime/cmd/registrar.py", line 20, in main
    registrar_common.start(config.getint(
  File "/usr/local/lib/python3.8/site-packages/keylime-0.0.0-py3.8.egg/keylime/registrar_common.py", line 506, in start
    RegistrarMain.metadata.create_all(engine, checkfirst=True)
  File "/usr/lib64/python3.8/site-packages/sqlalchemy/sql/schema.py", line 4555, in create_all
    bind._run_visitor(
  File "/usr/lib64/python3.8/site-packages/sqlalchemy/engine/base.py", line 2096, in _run_visitor
    with self._optional_conn_ctx_manager(connection) as conn:
  File "/usr/lib64/python3.8/contextlib.py", line 113, in __enter__
    return next(self.gen)
  File "/usr/lib64/python3.8/site-packages/sqlalchemy/engine/base.py", line 2088, in _optional_conn_ctx_manager
    with self._contextual_connect() as conn:
  File "/usr/lib64/python3.8/site-packages/sqlalchemy/engine/base.py", line 2304, in _contextual_connect
    self._wrap_pool_connect(self.pool.connect, None),
  File "/usr/lib64/python3.8/site-packages/sqlalchemy/engine/base.py", line 2341, in _wrap_pool_connect
    Connection._handle_dbapi_exception_noconnection(
  File "/usr/lib64/python3.8/site-packages/sqlalchemy/engine/base.py", line 1583, in _handle_dbapi_exception_noconnection
    util.raise_(
  File "/usr/lib64/python3.8/site-packages/sqlalchemy/util/compat.py", line 182, in raise_
    raise exception
  File "/usr/lib64/python3.8/site-packages/sqlalchemy/engine/base.py", line 2338, in _wrap_pool_connect
    return fn()
[...]

PR #347, opened 2020-09-04
Fixes postgres database configuration

lukehinds:
`check_same_thread` causes an error with postgres and should only
be used with sqlite

Also fixes an issue where the registrar still attempts to create
a sqlite file, even though postgres is configured.

Add missing documentation URL to keylime.conf

Add table creation files for postgres

Resolves: #344


lukehinds:
@maugustosilva  this should also fix the mysql issues

mpeters (approved):
lgtm

Issue #337, opened 2020-07-20
Large whitelist files fail in keylime_tenant


mpeters:
Is your an issue a feature request? If so, please raise it as an [enhancement](https://github.com/keylime/enhancements)

# Environment
 * OS / version: Fedora 32
 * Processor architecture: x86_64
 * TPM Manufacturer: emulated
 * Keylime version: source

# Description 

I have a whitelist file that is very large (196M) and running keylime_tenant with it fails with the message "Cannot connect to Verifier at 127.0.0.1 with Port 8881. Connection refused." The connection isn't actually received, but there is an error in processing the large file and the `keylime/httpclient_requests.py` translates that into a 503 and doesn't give much more information

## Expected behavior vs. actual behavior 

Keylime should be able to handle larger whitelist files. In addition, there should be a better error message when the file is too large. Maybe some pre-checking in keylime_tenant and also better logging for requests in `keylime/httpclient_requests.py`

## Steps to reproduce problem
1. Run keylime_tenant with a very large (196M in my case) whitelist file.

mpeters:
Fixed in https://github.com/keylime/keylime/pull/353

PR #353, opened 2020-09-11
Allow keylime to use larger whitelists

mpeters:
Before the limit was 100MB set by tornado, but now we're explicitly
setting and controlling it ourselves. If this proves to be too
restrictive, we can look at making it configurable in the future.

Fixes issue #337

### DIFF
--- a/keylime/cloud_verifier_common.py
+++ b/keylime/cloud_verifier_common.py
@@ -374,8 +374,10 @@ def validate_agent_data(agent_data):
     if agent_data is None:
         return False, None
 
+    # validate that the ima_whitelist is proper JSON
+    lists = json.loads(agent_data['ima_whitelist'])
+
     # Validate exlude list contains valid regular expressions
-    lists = ast.literal_eval(agent_data['ima_whitelist'])
     is_valid, _, err_msg = common.valid_exclude_list(lists.get('exclude'))
     if not is_valid:
         err_msg += " Exclude list regex is misformatted. Please correct the issue and try again."
--- a/keylime/cloud_verifier_tornado.py
+++ b/keylime/cloud_verifier_tornado.py
@@ -666,6 +666,11 @@ def main(argv=sys.argv):
     config = common.get_config()
     cloudverifier_port = config.get('cloud_verifier', 'cloudverifier_port')
 
+    # allow tornado's max upload size to be configurable
+    max_upload_size = None
+    if config.has_option('cloud_verifier', 'max_upload_size'):
+        max_upload_size = config.get('cloud_verifier', 'max_upload_size')
+
     VerfierMain.metadata.create_all(engine, checkfirst=True)
     session = SessionManager().make_session(engine)
     try:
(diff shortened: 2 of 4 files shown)


keylime/cloud_verifier_common.py
```diff
@@ -374,8 +374,10 @@ def validate_agent_data(agent_data):
     if agent_data is None:
         return False, None
 
+    # validate that the ima_whitelist is proper JSON
+    lists = json.loads(agent_data['ima_whitelist'])
```
lukehinds, line 378: I remember hitting some issues with json.loads, it was the way that sqlaclhemy was formatting the data which is why I used eval. I can't remember what it was now (which is not much use). I will try to remember what it was.

keylime/httpclient_requests.py
```diff
@@ -33,12 +38,20 @@ def request(method, host, port, params=None, data=None, context=None,
             timeout=timeout)
 
     try:
+        if logger is not None:
```
lukehinds, line 41: Is there a case where `logger` could equal `None`?

lukehinds (commented):
Nice change!

keylime/cloud_verifier_common.py
mpeters: Yeah, I didn't hit any issues, but `ast.literal_eval` has this warning: `Warning It is possible to crash the Python interpreter with a sufficiently large/complex string due to stack depth limitations in Python’s AST compiler.` which I was hitting with the large files.

keylime/httpclient_requests.py
mpeters: Good catch. This was from a previous revision where I passed in the logger. Not needed now and removed.

mpeters:
It's now configurable in keylime.conf

keylime/cloud_verifier_common.py
lukehinds: I do feel a bit 'antsy' about using eval, so if we can drop it lets do so.

keylime/cloud_verifier_tornado.py
```diff
@@ -666,6 +666,11 @@ def main(argv=sys.argv):
     config = common.get_config()
     cloudverifier_port = config.get('cloud_verifier', 'cloudverifier_port')
 
+    # tornado defaults to 100MB max upload size, but make it configurable
+    max_upload_size = 104857600
```
lukehinds, line 670: I guess I am nitpicking here, but I would not bother to set the value for `max_upload_size` in the code. We now have it set in the keylime.conf  (I am good with 100mb) so declare it and not bother checking if its present or not. Just have:

cloud_verifier_common.py: `max_upload_size = config.get('cloud_verifier', 'max_upload_size')`

keylime.conf: `max_upload_size = 104857600`

Hope that makes sense, the rest looks great!

lukehinds: In fact it would be good to set the config parse declaration (max_upload_size = config.get....) at the top of the file between these lines somewhere (I guess 39 perhaps?)

https://github.com/keylime/keylime/blob/30b196de45999c89d25490cd00fe88cbd42785f2/keylime/cloud_verifier_tornado.py#L38-L91

mpeters: > In fact it would be good to set the config parse declaration (max_upload_size = config.get....) at the top of the file between these lines somewhere (I guess 39 perhaps?)

I saw this more like the other configuration parameters that control the web server (like cloudverifier_port or multiprocessing_pool_num_workers) and thus put it closer to the code that uses it. If it were used in multiple places, then yes I would agree.

lukehinds (approved):

PR #354, opened 2020-09-14
Readibility and consistency improvements to keylime.conf

axelsimon:
Hi,
Here are quite a few changes to keylime.conf to make it easier to understand (i hope) as well as more consistent.

I've tried to always put the name of an option between single quotes, and the name a value for these options can take between double quotes, ie 'tls_dir', but "default".

Some questions remain:
- options, param(eters), can i use them interchangeanbly?
- what other options for 'tls_dir' exist, apart from "generate"? "default"? 
- Tenant, tls_dir section: "This is for authenticating the CV itself and proving"
  - I find this confusing, as the same wording used for the cloud_verifier, tls_dir section.
  - Is this:
    ```
    The 'tls_dir' option will look for client certificates and
    files for TLS in the /var/lib/keylime/ directory.
    ```
    correct?
- "# Don't send things bigger than the tmpfs where they will be decrypted."
  - Is this a warning? a reminder? also, is it a vulnerability: can i create a DoS by doing so?
- <fully_qualified_name> is this different from <fully_qualified_domain_name>?
- L129: # Set to "generate" to automatically generate a CA/certificates in the 'cv_ca' directory.
  - in the 'cv_ca' subdirectory? ie, a directory in 'tls_dir' (so: 'tls_dir'/'cv_ca' ?)

Happy to make further changes to this PR to reflect any answers to the above questions.


lukehinds (approved):
looks good!

mpeters (approved):
lgtm

lukehinds:
in regards to your questions, tagging @nabilschear who should be able to help (pretty sure he is the author).

nabilschear:
> Hi,
> Here are quite a few changes to keylime.conf to make it easier to understand (i hope) as well as more consistent.
> 
> I've tried to always put the name of an option between single quotes, and the name a value for these options can take between double quotes, ie 'tls_dir', but "default".
> 
> Some questions remain:
> 
> * options, param(eters), can i use them interchangeanbly?

good question!  Only the tenant has as a CLI, so it's the only one that could have them in both places. If memory serves (and my super quick look at tenant.py), there should be no overlap between command line options and what's in keylime.conf.

> * what other options for 'tls_dir' exist, apart from "generate"? "default"?

you can point it to any directory where you want to keep your certificates/keys.  If you use `keylime-ca --init -d /some/dir`, it will set it up for you in /some/dir.  You can also create the certificates with some other tool and place them in an arbitrary directory.  You'd need to follow the naming conventions that keylime-ca uses, but it shouldn't be too hard.  

> * Tenant, tls_dir section: "This is for authenticating the CV itself and proving"
>   
>   * I find this confusing, as the same wording used for the cloud_verifier, tls_dir section.
>   * Is this:
>     ```
>     The 'tls_dir' option will look for client certificates and
>     files for TLS in the /var/lib/keylime/ directory.
>     ```
>     
>     
>     correct?

This is basically what is needed to establish a mutual TLS connection between the tenant and the verifier.  The tenant needs it's client cert/key and the ca-cert/verifier-cert to do so.  

> * "# Don't send things bigger than the tmpfs where they will be decrypted."
>   
>   * Is this a warning? a reminder? also, is it a vulnerability: can i create a DoS by doing so?

yeah, basically the write that the agent does of the payload will fail with an "out of space"-style error.  But this is a self-DoS, not  one that an attacker can control.

> * <fully_qualified_name> is this different from <fully_qualified_domain_name>?

nope, just me being imprecise.

> * L129: # Set to "generate" to automatically generate a CA/certificates in the 'cv_ca' directory.
>   
>   * in the 'cv_ca' subdirectory? ie, a directory in 'tls_dir' (so: 'tls_dir'/'cv_ca' ?)

exactly!

Thanks for reading through this carefully.  I was aspiring to make it really comprehensive all those years ago, but you've up'd the level of clarity substantially!

lukehinds (commented):
I am good with this, but let's merge @mpeters PR to change to allowlist and then re can rebase this patch on top of that.

axelsimon:
Good point, i'll rebase and push the new version.

Issue #359, opened 2020-09-22
PR #353 breaks the code in my testing


maugustosilva:
Prior PR #353, here is how `keylime_tenant` behaves:
```
root@css-flex14vm1:/# /usr/local/bin/keylime_tenant -c status -v 100.64.14.13 -t 100.64.14.16
Using config file /etc/keylime.conf
2020-09-22 16:51:58.888 - keylime.tenant - INFO - Setting up client TLS in /var/lib/keylime/cv_ca
2020-09-22 16:51:58.891 - keylime.tenant - INFO - Setting up client TLS in /var/lib/keylime/cv_ca
2020-09-22 16:51:58.894 - keylime.tenant - WARNING - Using default UUID D432FBB3-D2F1-4A97-9EF7-75BD81C00000
AQUI
2020-09-22 16:51:58.894 - keylime.httpclient_requests - DEBUG - Making HTTP GET request to 100.64.14.13:8881/agents/D432FBB3-D2F1-4A97-9EF7-75BD81C00000
2020-09-22 16:51:58.935 - keylime.tenant - ERROR - Agent D432FBB3-D2F1-4A97-9EF7-75BD81C00000 does not exist on the verifier. Please try to add or update agent
```

Afterwards, here is what I see:

```
root@css-flex14vm1:/KL/keylime# /usr/local/bin/keylime_tenant -c status -v 100.64.14.13 -t 100.64.14.16
Using config file /etc/keylime.conf
2020-09-22 16:52:21.031 - keylime.tenant - INFO - Setting up client TLS in /var/lib/keylime/cv_ca
2020-09-22 16:52:21.034 - keylime.tenant - INFO - Setting up client TLS in /var/lib/keylime/cv_ca
2020-09-22 16:52:21.037 - keylime.tenant - WARNING - Using default UUID D432FBB3-D2F1-4A97-9EF7-75BD81C00000
2020-09-22 16:52:21.037 - keylime.httpclient_requests - DEBUG - Making HTTP GET request to 100.64.14.13:8881/agents/D432FBB3-D2F1-4A97-9EF7-75BD81C00000
2020-09-22 16:52:21.042 - keylime.httpclient_requests - ERROR - ConnectionError: [Errno 104] Connection reset by peer
2020-09-22 16:52:21.042 - keylime.tenant - ERROR - Cannot connect to Verifier at 100.64.14.13 with Port 8881. Connection refused.
```

The problem seems to be a simple one...

[...]

PR #363, opened 2020-09-25
Real fix for issue #359

mpeters:


ashcrow (approved):
Makes sense. The value of `max_upload_size` should be converted to an `int` and stored rather than converting the stored value but not keeping it.

mpeters:
Fixed in another commit to get the tests working again

PR #365, opened 2020-09-25
Fix broken tests

mpeters:
Fixing 2 tests of broken tests at once to get CI working again:

* Fix separate reg cv tests
* Fix botched fix of max_upload_size int conversion

### DIFF
--- a/keylime/cloud_verifier_tornado.py
+++ b/keylime/cloud_verifier_tornado.py
@@ -669,7 +669,7 @@ def main(argv=sys.argv):
     # allow tornado's max upload size to be configurable
     max_upload_size = None
     if config.has_option('cloud_verifier', 'max_upload_size'):
-        max_upload_size = int(config.get('cloud_verifier', 'max_upload_size'))
+        int(max_upload_size = config.get('cloud_verifier', 'max_upload_size'))
 
     VerfierMain.metadata.create_all(engine, checkfirst=True)
     session = SessionManager().make_session(engine)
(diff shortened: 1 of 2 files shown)


lukehinds:
Does this fix the tests that are failing?

mpeters:
Yes, it should get CI back to normal

lukehinds:
here is the build report, unfortunately this max upload size is still failing:

https://travis-ci.org/github/keylime/keylime/jobs/730337434#L396

mpeters:
> here is the build report, unfortunately this max upload size is still failing:
> 
> https://travis-ci.org/github/keylime/keylime/jobs/730337434#L396

Yeah, my other PR fixes that. That one was held up by these failures, so I went to fix those first. Should they all be combined?

lukehinds:
> > here is the build report, unfortunately this max upload size is still failing:
> > https://travis-ci.org/github/keylime/keylime/jobs/730337434#L396
> 
> Yeah, my other PR fixes that. That one was held up by these failures, so I went to fix those first. Should they all be combined?

sure, make another quick commit with your int fix and we should see CI pass as well then.

mpeters:
Here: https://travis-ci.org/github/keylime/keylime/builds/730343055

Should I remove that fix to have them be separate commits or just leave them packaged as one?

lukehinds:
I reckon we can just merge this one and close the other if that works for you?

lukehinds (approved):

PR #367, opened 2020-09-28
Update setup.py

AzharMithani:
Updated with better comments to improve the code readability

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,14 +2,17 @@
 SPDX-License-Identifier: Apache-2.0
 Copyright 2017 Massachusetts Institute of Technology.
 '''
+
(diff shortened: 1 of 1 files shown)


mpeters:
Thanks for the contribution. But I'm not entirely clear on why you think these comments were needed. In general I'm in favor of my comments when they explain the "why" of something, not necessarily the "what". The "what" is usually (though not always) clear in the code, but the "why" is often in our heads and needs to be more explicit.

These comments seem more like the "what" and not "why" to me. For instance, comments like `# importing all the required libraries` would actually be true of any python code that imports things. Why is this different and thus needs more explanation?

Issue #254, opened 2020-02-19
FreeBSD port


lukehinds:
Get Keylime running on FreeBSD

nabilschear:
a first proof of concept https://gist.github.com/nabilschear/4527fa3adf792085879dcbf4f0d30481

PR #371, opened 2020-09-30
park the howto on freebsd in docs

nabilschear:
A while back i wrote up some docs on getting keylime running on freebsd.  I thought it would be good to have these docs in the docs folder in case anyone ever wants to pick this up again.

related to https://github.com/keylime/keylime/issues/254

### DIFF
--- a/doc/freebsd.md
+++ b/doc/freebsd.md
@@ -1,0 +1,101 @@
+# Instructions for Installing Keylime on FreeBSD
+
+These instructions were a proof-of-concept for getting keylime to work.  These instructions may or may not work, depending on the version of FreeBSD or changes to tpm2-tools.
+
+get dependencies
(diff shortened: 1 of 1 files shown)


lukehinds:
this is great, but we should try to use keylime-docs now instead of docs

nabilschear:
Considering this is kinda unofficial documentation, think it's ok going here?  I can certainly mkdocs it.

ashcrow:
Does anyone have feedback/an answer on @nabilschear's question?

lukehinds:
I guess it could go in /doc. I would just like to see our offical docs getting more content. There is an installation section, but its not dist specific (its based on ansible / installer.sh), so I guess just having a freeBSD install guide would look confusing.

Stick it in here for now and we can always migrate in later.

PR #372, opened 2020-10-01
Rename whitelist to allowlist

mpeters:
Some usages of "ima_whitelist" were also shortened to just "allowlist"


mpeters:
This covers issue #372

mpeters:
Currently this change is not backwards compatible. It shouldn't be too hard to change the cli scripts to also take `--whitelist` as an arg that aliases to `--allowlist` and we can also make the JSON APIs copy anything in `ima_whitespace` fields to `allowlist`. The harder part for backcompat is the database field. We'd need to have migrations (using alembic) in place so it can detect the db version and apply the right DDL. It doesn't look like Keylime has ever done this in the past (from what I can tell). It's something we eventually need anyway, so if we think this is as good as any first use case for that I'm happy to tackle that as well.

lukehinds (approved):
LGTM

mpeters:
I'm going to hold off on merging this until after the migrations change lands, then go back and use migrations for the DB change here as the first trial of that.

PR #376, opened 2020-10-08
Clean and minimalistic MySQL compatibility

maugustosilva:
This commit implements the minimal set of changes required to make sure
Keylime can use a MySQL database. This involved mostly small fixes to
Verifier and Registrar DB schema (mostly adding lengths for the "String"
(mapped to "VarChar") columns). The only relevant exception is the
changing of "ima_whitelist" column from "String" to "Text". Tested on
both MySQL 5.7.31 and SQLite.


mpeters (approved):
lgtm

PR #377, opened 2020-10-08
Add DB migrations to keylime

mpeters:
* Use Alembic to manage the migrations
* First abstract the creation of the DB engine into
  keylime_db.DBEngineManager
* Add alembic configuration and env.py that uses
  keylime_db.DBEngineManager to get either registrar or cloud_verifier DB
  depending on the context
* Add initial data migration script
* Add keylime_apply_migrations CLI


mpeters:
We need to figure out the best time to run the migrations. I've created the keylime_apply_migrations script (should it be named keylime_migrations_apply? or just keylime_migrations?) but have left calling it up to the user. 

We probably want to make that automatic. So should it be run when the module is upgraded? When the application starts? Something else?

lukehinds (commented):
nice work @mpeters , there is a fair amount to take in here , so we might be able to do a review on hangouts. This will be useful for when the multi tenancy PR lands, as this introduces a few database changes.

keylime/db/keylime_db.py
```diff
@@ -3,12 +3,55 @@
...
 from sqlalchemy.exc import SQLAlchemyError
+from sqlalchemy.engine.url import URL
 
+from keylime import common
 from keylime import keylime_logging
 
-logger = keylime_logging.init_logging('sql_session_manager')
+
+class DBEngineManager:
+    def make_engine(self, service):
```
ashcrow, line 17: super nit: New line

keylime/db/keylime_db.py
```diff
@@ -3,12 +3,55 @@
...
+            url = URL(
+                drivername=drivername,
+                username=config.get(service, 'username'),
+                password=config.get(service, 'password'),
+                host=config.get(service, 'host'),
+                database=config.get(service, 'database')
+            )
+            engine = create_engine(url)
+
+
```
ashcrow, line 53: super nit: Remove a new line

ashcrow (commented):
Two nits which can be ignored -- though I do see some failing tests which look to be related to the database.

mpeters:
By defaults the migrations occur when the registrar and verifier startup (since technically they are separate databases even if they might be the same in a given implementation). This can be turned off by the `auto_migrate_db` configuration variable in the `[cloud_verifier]` and `[registrar]` sections.

maugustosilva:
@mpeters the PR looks good to me... However, upon trying it, I see that the (otherwise quite verbose in DEBUG level) Verifier log messages just disappears from both the console and `/var/log/keylime/cloudverifier.log`, being replaced by the `alembic` log only. Something is wrong with the log handlers, didn't have time to diagnose yet.

mpeters:
> @mpeters the PR looks good to me... However, upon trying it, I see that the (otherwise quite verbose in DEBUG level) Verifier log messages just disappears from both the console and `/var/log/keylime/cloudverifier.log`, being replaced by the `alembic` log only. Something is wrong with the log handlers, didn't have time to diagnose yet.

I've fixed this problem. Just needed to change the way alembic auto-generated the env.py to not setup it's own logging with `from logging.config import fileConfig` and instead just inherit whatever keylime was doing. This also makes the `alembic.ini` file really simple.

Testing with these changes seems to allow keylime logging to do it's thing

ashcrow (approved):

mpeters:
After chatting with @maugustosilva on their setup, I've now updated the migrations to be able to work even if the registrar and verifier tables are in the same logical DB.

ashcrow (approved):
LGTM

Issue #378, opened 2020-10-12
Seperate out revocation notifier IP:PORT config values


lukehinds:
The `await_notifications` function currently sets up its listening IP & Port from the `cloud_verifier` config section. 

https://github.com/keylime/keylime/blob/ab6397317513b952c090bdd52bbfe6676b8da3c6/keylime/revocation_notifier.py#L81-L95

This might be confusing for users, as the agent and keylime_ca parse out the `'cloud_verifier', 'revocation_notifier_ip'` values, yet they don't run on the verifier, nor are they ** connecting ** to the verifier. 

A better approach is to have values set  in general 

```
#=============================================================================
[general]
#=============================================================================

<snip>

# Revocation IP & Port used by either the cloud_agent or keylime_ca to receive
#revocation events from the verifier
receive_revocation_ip= 12.0.0.1
receive_revocation_port = 8992
```

PR #379, opened 2020-10-12
Seperate out revocation receiver IP:PORT

lukehinds:
This change seeks to clear up potential confusion over what values
are being set and used for the revocation notifier and reciever

We introduce a `receive_revocation_ip` which can be used by either
the agent or keylime_ca

`cloud_verifier, revocation_notifier_ip` remains as it was and is
used by the verifier.

Also remove the hardcoded `*` wildcard for the config value for
`revocation_notifier_ip` which results in listen on all interfaces
Instead use the config value (as a user may wish to listen on
a specfic IP / interface)

Resolves: #378

### DIFF
--- a/keylime.conf
+++ b/keylime.conf
@@ -1,3 +1,4 @@
+#=============================================================================
 [general]
 #=============================================================================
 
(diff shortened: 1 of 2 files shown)


ashcrow (approved):
LGTM -- Let's ensure this is in the release notes/changelog since it does change how the configuration can be set/modified.

mpeters (approved):
lgtm

PR #380, opened 2020-10-15
Initial migration mistakenly left off string lengths

mpeters:


lukehinds (approved):
LGTM, but are we absolultey sure the lengths are sufficient? (likely not fair to bring up on this PR as its not the origin). I guess @maugustosilva researched this?

mpeters:
> LGTM, but are we absolultey sure the lengths are sufficient? (likely not fair to bring up on this PR as its not the origin). I guess @maugustosilva researched this?

I can't say for all of the fields (I did spot check them as I was doing this work and they looked good), but for the allowlist is definitely big enough.

PR #381, opened 2020-10-17
Extract algorithm consts and helper into utilities

kaifeng:
This should be something in the common but there is already
a common file thus renamed to utils. Other common stuff can
be put to the utils as needed.

### DIFF
--- a/keylime/utils/algorithms.py
+++ b/keylime/utils/algorithms.py
@@ -1,0 +1,60 @@
+"""
+SPDX-License-Identifier: Apache-2.0
+Copyright 2020 Kaifeng Wang
+"""
+
+
+def is_accepted(algorithm, accepted):
+    """Check whether algorithm is accepted
+
+    @param algorithm: algorithm to be checked
+    @param accepted: a list of acceptable algorithms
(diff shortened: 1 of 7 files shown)


mpeters (approved):
lgtm

keylime/utils/algorithms.py
```diff
@@ -0,0 +1,60 @@
+"""
+SPDX-License-Identifier: Apache-2.0
+Copyright 2017 Massachusetts Institute of Technology.
```
lukehinds, line 3: @kaifeng you can put you're own name there is you like.  May as well make it 2020 as well (as its a new file).

kaifeng: okay, just haven't seen many personal copyright in opensource projects.

mpeters:
@kaifeng do you want to merge this yourself? I think in general we are going to let the committer do the merge (if they have permissions) as this gives them the most flexibility for timing, etc depending on what other changes they have in the pipeline.

kaifeng:
@mpeters well, it seems I can't merge myself.

Issue #131, opened 2019-07-08
Investigate removal of shell=True calls


lukehinds:
`shell=True` is susceptible to shell execution attacks from unsanitized input [0]

Investigate what hardening we can do.

[0] https://security.openstack.org/guidelines/dg_avoid-shell-true.html

bu3alwa:
Is there a reason why we use Shell=True?

lukehinds:
```
def ping(myserver):
    return subprocess.check_output('ping -c 1 %s' % myserver, shell=True)

>>> ping('8.8.8.8')
64 bytes from 8.8.8.8: icmp_seq=1 ttl=58 time=5.82 ms
```

```
>>> ping('8.8.8.8; rm -rf /')
64 bytes from 8.8.8.8: icmp_seq=1 ttl=58 time=6.32 ms
rm: cannot remove `/bin/dbus-daemon': Permission denied
rm: cannot remove `/bin/dbus-uuidgen': Permission denied
rm: cannot remove `/bin/dbus-cleanup-sockets': Permission denied
rm: cannot remove `/bin/cgroups-mount': Permission denied
rm: cannot remove `/bin/cgroups-umount': Permission denied
```

:)

https://medium.com/python-pandemonium/a-trap-of-shell-true-in-the-subprocess-module-6db7fc66cdfd

lukehinds:
never answered your question, we should not be using it, its better to use `False`

```
def ping(myserver):
    args = ['ping', '-c', '1', myserver]
    return subprocess.check_output(args, shell=False)
```
```
>>> ping('8.8.8.8; rm -rf /')
ping: unknown host 8.8.8.8; rm -rf /
```

bu3alwa:
@lukehinds yea we should use False. I am just wondering why we are using True in the first place? Is there anything in keylime that needs the tag True?

lukehinds:
No, no reason. It's just code from when the project was more of a prototype whereas now we are trying to refactor things to be a bit more secure and optimised.

kaifeng:
I am taking this assignment o/

PR #385, opened 2020-10-29
Remove shell=True from exec_cmd

kaifeng:
Removed shell=True from exec_cmd, currently in a tradeoff
of converting string argument of tpm.__run to list by split.

Issue #131

### DIFF
--- a/keylime/tpm/tpm1.py
+++ b/keylime/tpm/tpm1.py
@@ -127,9 +127,15 @@ class tpm1(tpm_abstract.AbstractTPM):
         while True:
             if lock:
                 with self.tpmutilLock:
-                    retDict = cmd_exec.run(cmd=cmd, expectedcode=expectedcode, raiseOnError=False, lock=lock, outputpaths=outputpaths, env=env)
+                    retDict = cmd_exec.run(
+                        cmd=cmd, expectedcode=expectedcode,
+                        raiseOnError=False, lock=lock,
+                        outputpaths=outputpaths, env=env, shell=True)
             else:
-                retDict = cmd_exec.run(cmd=cmd, expectedcode=expectedcode, raiseOnError=False, lock=lock, outputpaths=outputpaths, env=env)
+                retDict = cmd_exec.run(
+                    cmd=cmd, expectedcode=expectedcode, raiseOnError=False,
+                    lock=lock, outputpaths=outputpaths, env=env, shell=True)
+
             t0 = retDict['timing']['t0']
             t1 = retDict['timing']['t1']
             code = retDict['code']
--- a/keylime/tpm/tpm2.py
+++ b/keylime/tpm/tpm2.py
@@ -254,10 +254,9 @@ class tpm2(tpm_abstract.AbstractTPM):
         while True:
             if lock:
                 with self.tpmutilLock:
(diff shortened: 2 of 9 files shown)


kaifeng:
Judging by code, STUB_IMA is not used, thus it can be inferred that STUB_TPM is also not in use. I am wondering if it's safe to remove them, this will make shell removal a lesser work.

lukehinds:
@nabilschear / @jetwhiz , quicker reminder, what was `STUB_IMA` for and is it needed still?

kaifeng:
> @nabilschear / @jetwhiz , quicker reminder, what was `STUB_IMA` for and is it needed still?

As far as I can see, it's used for (probably an earlier) emulation.

mpeters:
@kaifeng If you can rebase we can go ahead and merge this. We can deal with the STUB stuff in a different PR

kaifeng:
@mpeters ack, I am good to do it in follow ups.

keylime/tpm/tpm1.py
```diff
@@ -127,9 +127,15 @@ def __run(self, cmd, expectedcode=tpm_abstract.AbstractTPM.EXIT_SUCESS, raiseOnE
         while True:
             if lock:
                 with self.tpmutilLock:
-                    retDict = cmd_exec.run(cmd=cmd, expectedcode=expectedcode, raiseOnError=False, lock=lock, outputpaths=outputpaths, env=env)
+                    retDict = cmd_exec.run(
+                        cmd=cmd, expectedcode=expectedcode,
+                        raiseOnError=False, lock=lock,
+                        outputpaths=outputpaths, env=env, shell=True)
```
mpeters, line 133: In addition to changing formatting, this is adding `shell=True` which is the opposite of what the change was trying to do right?

keylime/tpm/tpm2.py
```diff
@@ -254,10 +254,9 @@ def __run(self, cmd, expectedcode=tpm_abstract.AbstractTPM.EXIT_SUCESS, raiseOnE
         while True:
             if lock:
                 with self.tpmutilLock:
-                    retDict = cmd_exec.run(cmd=cmd, expectedcode=expectedcode, raiseOnError=False, lock=lock, outputpaths=outputpaths, env=env)
+                    retDict = cmd_exec.run(cmd=cmd, expectedcode=expectedcode, raiseOnError=False, lock=lock, outputpaths=outputpaths, env=env, shell=True)
```
mpeters, line 257: Again, why are we adding `shell=True` here?

keylime/tpm/tpm1.py
kaifeng: Yep, there are extra works in the tpm code, using shell here is a workaround actually, because spliting them into iterables is not working from CI response at the first time, that's why I am asking if we could remove stub first, I simply want to avoid updating code that actually not in use and can be removed.

kaifeng:
Checking tpm code and found many duplicated command execution there, I am thinking about wraping the tpm2-tools as a module to do the execution, we can drop duplication and also handle various tpm tools versions there.

mpeters (approved):
lgtm

mpeters:
> Checking tpm code and found many duplicated command execution there, I am thinking about wraping the tpm2-tools as a module to do the execution, we can drop duplication and also handle various tpm tools versions there.

This sounds like a good idea to me.

PR #386, opened 2020-10-29
Minimalistic changes to allow successful execution of `keylime_ca -c listen` with `openssl`

maugustosilva:
Basically, the problem stems from the fact that `openssl` produces an
**empty** `cacrl.der` (and `cacrl.pem`).

Resolves: ##261

### DIFF
--- a/keylime/ca_util.py
+++ b/keylime/ca_util.py
@@ -319,9 +319,10 @@ def cmd_revoke(workingdir, name=None, serial=None):
         write_private(priv)
 
         # write out the CRL to the disk
-        with open('cacrl.der', 'wb') as f:
-            f.write(crl)
-        convert_crl_to_pem("cacrl.der", "cacrl.pem")
+        if os.stat('cacrl.der').st_size :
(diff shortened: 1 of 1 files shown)


mpeters (commented):
LGTM, but probably needs eyeballs from @lukehinds  since I'm not completely familiar with this code

lukehinds:
code looks ok, but we did test everything here, mainly revocation operates OK still...I can test this early next week (for me to approve I would just need to give it a test drive first).

mpeters:
@lukehinds did you get a chance to test this? If not, I could give it a go. Could you give me a brief outline of what steps you were thinking about for testing this?

lukehinds:
> @lukehinds did you get a chance to test this? If not, I could give it a go. Could you give me a brief outline of what steps you were thinking about for testing this?

I was incorrect before, there is no revocation with openssl at the moment, so the patch seems to good to me. @nabilschear / @jetwhiz , any thoughts on this one?

mpeters (approved):
There doesn't seem to be anything holding this up and I've verified it works on my end

PR #388, opened 2020-10-30
Remove the DEVELOP_IN_ECLIPSE support

### LINKED ISSUE #383: Drop logics related with DEVELOP_IN_ECLIPSE
In our scenario we use keylime as root, I am wondering if the DEVELOP_IN_ECLIPSE is actually in use because it seems a bit old, but no sure if it's in use in the CI (as far as I see the CI is using container as root). If we no longer need that, we probably could drop to make the maintenance easier.

kaifeng:
The support to develop in eclipse is no longer supported
and removed.

Resolves #383


kaifeng:
@lukehinds Do we still use STUB_TPM and TPM_CANNED_VALUES in the CI?

mpeters (approved):
lgtm

PR #390, opened 2020-10-30
Restrict verifier and registrar port listening

### LINKED ISSUE #389: Verifier and registrar are hardcoded to bind 0.0.0.0 address
This is a security issue that we should fix.

kaifeng:
verifier and registrar is listening 0.0.0.0 address which could expose
services to unwanted networks. Since there are configurations for the
specific addresses to bind to, let's use it instead.

Resolves #389


mpeters (approved):
lgtm

PR #391, opened 2020-11-03
Improve logging time

kaifeng:
Change logger format to use asctime, created is epoch time
which is not user friendly.


kaifeng:
Just found agent is not logged into file, not sure it's intended or not.

mpeters (approved):
lgtm

PR #392, opened 2020-11-05
Missing return in CV API

kaifeng:
DELETE /agents is missing a return control and causes
unnecessary db query and improper log.


mpeters (approved):
lgtm

Issue #384, opened 2020-10-29
Incorporate pep8 coverage


kaifeng:
Adding pep8 checking helps to find potential issues that we are not running in the code path, and apply consistent code style to the project.

lukehinds:
We did have this with codacy, but there seems to be some CI issues at the moment. I will try to investigate.

stefanberger:
The following command line would let us cleanly transition from `pep8` to `pylint` and then crank up the checks by removing what is currently disabled:

```
pylint \
  --disable C0103,C0115,C0116,C0123,C0200,C0325,C0301,C0302,C0330,C0411,C0412,C0414,C0415,C0114,C0122 \
  --disable W0127,W0201,W0223,W0235,W1201,W1202,W0611,W0603,W0613,W0703,W0613,W0102,W0702,W0622,W0212,W1505,W0511,W0631,W0632,W0621,W0612,W0105,W0715,W0101,W0221,W0107,W1509,W0614,W0401,W0601 \
  --disable E0401,E0611,E1300,E1101,E0602,E1205,E1120,E1121,E1123,E1111,E1136 \
  --disable R0801,R0902,R0205,R0903,R0912,R0914,R0915,R1702,R1705,R1711,R1722,R1724,R1720,R0201,R0911,R1723,R1716,R0913,R1714,R0124,R1710,R0904,R1703 \
*.py $(find ./keylime ./test -name '*.py')
```

Everybody can grab a cookie! (#define cookie 'disabled warning')

mpeters:
Yeah, I like this idea as it would definitely help us find problems in code that is either undertested, really hard to test (bad ROI) and utility stuff that isn't applicable to all platforms.

stefanberger:
Problem is it takes a lot longer with pylint (59s) versus pep8 (3.4s).

mpeters:
That's not a huge problem. It's mostly going to be run in the pipeline and the tests already take a lot longer to run than that. And adding it to your workflow as a developer is still not as taxing as running the tests either.

So that extra 55s isn't a problem IMO

stefanberger:
i can send a PR for that then https://travis-ci.com/github/stefanberger/keylime/builds/203187797

stefanberger:
Maybe we can close this issue since this has landed and pylint provides much more than the original thought of using pep8.

mpeters:
Agreed

PR #394, opened 2020-11-09
Run code style checks on python files

stefanberger:
This PR adds a script for launching either pep8 or pycodestyle-3 to check the python code style. Also a Makefile is added with a check target for running the script either on a local machine or by Travis. We now also run it as part of the Travis build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,10 @@
 dist: bionic
 
 sudo: required
+addons:
+  apt:
+    packages:
+      - pycodestyle
 
 # If the change is a markdown file, don't run CI build. Or at a later point
 # we can set a specific job in here (such as a markdown lint)
@@ -27,6 +31,7 @@ services:
   - docker
 
 before_install:
+  - make check
   - 'docker pull ${tpm12image}:${tpm12tag}'
   - 'docker pull ${tpm20image}:${tpm20tag}'
 
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,4 @@
+
+.PHONY: check
+check:
+	scripts/check_codestyle.sh
--- a/scripts/check_codestyle.sh
+++ b/scripts/check_codestyle.sh
@@ -1,0 +1,13 @@
+#!/usr/bin/env bash
+
+TOOL=$(type -P pycodestyle)
+if [ -z "${TOOL}" ]; then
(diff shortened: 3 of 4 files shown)


stefanberger:
So far the syntax check is only done on the root directory after a fix to setup.py. A lot of fixes are needed for the other directories and they can come either later or we add them to this PR. Just let me know how you want to run the code style checker...

stefanberger:
You should move to travis-ci.com (from .org) to get some build threads.
I have a successful build here: https://travis-ci.com/github/stefanberger/keylime/builds/199470651

kaifeng:
I'd love to have a checking, #384 was proposed to track this ;)
It would be a heavy change if we do it in one PR, I personally would prefer to have a basic checking and interate.

lukehinds:
We are just on the verge of porting to github actions (as travis has caused a lot of issues before with failed webhooks and service just degrading over time). However, I think this means we will need this code even more, as we might no longer be using codacy then. 

@mbestavros when you implement https://github.com/keylime/enhancements/issues/18 could you see where pep8 checks could be leveraged using this patch.

lukehinds:
marked as key_feature to stop stalebot from trying to close.

stefanberger:
> We are just on the verge of porting to github actions (as travis has caused a lot of issues before with failed webhooks and service just degrading over time). However, I think this means we will need this code even more, as we might no longer be using codacy then.

I also saw that Travis was taking a long time to build. The issue is caused by them preparing to close travis-ci.org and wanting everyone to move over to travis-ci.com. On '.com' things are 'as expected'.

mpeters:
Even if we are in the process of moving off of Travis, it might be a few weeks before we get there. And only 1 part of this change is travis related. Using github actions would probably still use the same make target and script. So I'm ok with merging this (after a rebase)

stefanberger:
Rebased.
.... because this will come up sooner or later: what is the number of characters per line limit?

kaifeng:
<=79 in case of pep8

mpeters:
> <=79 in case of pep8

Yeah, this is super annoying IMO and I'd be up for having our standards be a little longer, but it's a minor nit and something we should bring up somewhere else :)

mpeters (approved):

PR #396, opened 2020-11-10
Remove redundant registrar server class

kaifeng:
Protected/Unprotected share the same server, just using different socket.


mpeters (approved):
lgtm

PR #397, opened 2020-11-10
Support using system-uuid from dmidecode as agent uuid

kaifeng:
system-uuid from dmidecode can be considered unique and doesn't
change from machine to machine.

### DIFF
--- a/keylime/keylime_agent.py
+++ b/keylime/keylime_agent.py
@@ -25,6 +25,7 @@ import shutil
 
 from keylime import common
 from keylime import keylime_logging
+from keylime import cmd_exec
 from keylime import crypto
 from keylime import openstack
 from keylime import revocation_notifier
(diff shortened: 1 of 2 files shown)


mpeters:
dmidecode isn't installed by default on most systems (that I'm aware of) so do we need to have this listed as an optional dependency for the installer(s)? Also, AFAIK on most systems it requires root to run. There's another conversation about trying to remove the requirement for the agent to run as root and run as a less privileged user (as long as we have permissions to read the IMA measurements log we don't need root). This would complicate it as an optional feature that would require root.

kaifeng:
Actually we are running keylime agent on bare metals directly and as a service with root privilege. As long as it's optional, I think it wouldn't block others not using it,  but could be utilized to satisfy different needs, the requirement of dmidecode is satisfied in our cases. Probably we can add sort of configuration check during agent startup?

mpeters:
It's true that even when we have a separate non-keylime user as the default it would still be possible to run keylime as root. We aren't taking that option away. I think I would be ok with this change if we add a check during startup to verify that if dmidecode is the chosen configuration that it's installed and the current user can execute it.

keylime/keylime_agent.py
```diff
@@ -488,6 +489,14 @@ def main(argv=sys.argv):
         logger.critical("This process must be run as root.")
         return
 
+    if config.get('cloud_agent', 'agent_uuid') == 'dmidecode':
+        ret = cmd_exec.run('which dmidecode', raiseOnError=False)
+        if ret['code'] != 0 or os.getuid() != 0:
```
mpeters, line 494: Is it possible to split this condition out? Check first that we are root and give an error message for that, then check that dmidecode is available and error on that. Seems better to have 2 different error codes to the 2 different conditions rather than combine them together.

mpeters (changes requested):

keylime/keylime_agent.py
kaifeng: sure, doable

mpeters (approved):
lgtm

PR #398, opened 2020-11-11
Enable IMA PCR by adding it to the mask using 'logical or'

stefanberger:
Enable the IMA PCR by adding it to the mask using a 'logical or'
rather than 'adding' it.

Signed-off-by: Stefan Berger <stefanb@linux.ibm.xom>


stefanberger:
... certainly just a minor thing.

nabilschear (approved):

PR #414, opened 2020-11-25
Stop checking simplejson in code

kaifeng:
requirements should be guaranteed by packaging, not code,
unless we also handle alternatives.


mpeters (approved):
lgtm

PR #418, opened 2020-11-27
Move stub code out of tpm.__run

kaifeng:
Move stub and metric handling out of tpm.__run, also
handles cmd as a string or iterable.


mpeters (approved):
lgtm

PR #419, opened 2020-11-27
Rename common to config

kaifeng:
So that we will have a common directory to hold common codes,
the previous common is actually a mix with configurations, debug
flags, helpers. We can further move those helpers into common module
and let conf only handles configurations.


mpeters:
I like the idea of separating them from the actual common code, but I don't like the name 'conf' as it makes it seem like it's related to keylime.conf handling. Is there a better more descriptive name we could use? flags? globals?

kaifeng:
The intent is to extract configuration unrelated codes out to a common module, so the original file is renamed to conf, it will only do configuration related things, probably having keylime.conf autogeneration in this file in the future.
So, the name conf seems fit for this purpose, config is also fine. Using global variables as configuration options is a good choice at early development, I guess it would be good to gradually migrate those into configuration file.

lukehinds:
I like config and have seen that a lot before.

@kaifeng , nothing to do with this PR, but are you able to jump on our cncf slack channel and ping me when you're there?

mpeters (approved):
lgtm

Issue #395, opened 2020-11-10
EE Key too small on CentOS 8


kaifeng:
Keylime verifier failed to run on CentOS 8:
Nov 09 17:07:11 TPM-client-1 keylime_verifier[1329931]: Traceback (most recent call last):
Nov 09 17:07:11 TPM-client-1 keylime_verifier[1329931]:   File "/usr/bin/keylime_verifier", line 11, in <module>
Nov 09 17:07:11 TPM-client-1 keylime_verifier[1329931]:     load_entry_point('keylime==0.0.0', 'console_scripts', 'keylime_verifier')()
Nov 09 17:07:11 TPM-client-1 keylime_verifier[1329931]:   File "/usr/lib/python3.6/site-packages/keylime-0.0.0-py3.6.egg/keylime/cmd/verifier.py", line 23, in main
Nov 09 17:07:11 TPM-client-1 keylime_verifier[1329931]:     cloud_verifier_tornado.main()
Nov 09 17:07:11 TPM-client-1 keylime_verifier[1329931]:   File "/usr/lib/python3.6/site-packages/keylime-0.0.0-py3.6.egg/keylime/cloud_verifier_tornado.py", line 668, in main
Nov 09 17:07:11 TPM-client-1 keylime_verifier[1329931]:     context = cloud_verifier_common.init_mtls()
Nov 09 17:07:11 TPM-client-1 keylime_verifier[1329931]:   File "/usr/lib/python3.6/site-packages/keylime-0.0.0-py3.6.egg/keylime/cloud_verifier_common.py", line 153, in init_mtls
Nov 09 17:07:11 TPM-client-1 keylime_verifier[1329931]:     certfile=my_cert, keyfile=my_priv_key, password=my_key_pw)
Nov 09 17:07:11 TPM-client-1 keylime_verifier[1329931]: ssl.SSLError: [SSL: EE_KEY_TOO_SMALL] ee key too small (_ssl.c:3542)
Nov 09 17:07:11 TPM-client-1 systemd[1]: keylime_verifier.service: Main process exited, code=exited, status=1/FAILURE
Nov 09 17:07:11 TPM-client-1 systemd[1]: keylime_verifier.service: Failed with result 'exit-code'.

This typically indicates the private key is weak, CentOS 8 is using CipherString @SECLEVEL=3, can work around by lowering security level to 1 or 2. Current code already generates 2048 bits certifcate so I am probably missing something here.

kaifeng:
[...]

PR #422, opened 2020-12-01
Improve ee_key_too_small handling

kaifeng:
Capture ssl exception for EE_KEY_TOO_SMALL and provide hint
about how to resolve the issue.

Resolves: #395

### DIFF
--- a/keylime/cloud_verifier_common.py
+++ b/keylime/cloud_verifier_common.py
@@ -6,12 +6,12 @@ Copyright 2017 Massachusetts Institute of Technology.
 '''
 
 import ast
-from urllib.parse import urlparse
 import base64
-import time
 import os
 import ssl
 import socket
+import time
+from urllib.parse import urlparse
 
 import simplejson as json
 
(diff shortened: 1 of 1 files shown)


keylime/cloud_verifier_common.py
```diff
@@ -76,6 +92,7 @@ def __exit__(self, *args):
             print('elapsed time: %f ms' % self.msecs)
 
 
+@convert_exceptions
```
mpeters, line 95: What is the reason to wrap the whole function with convert_exceptions vs using a try/except block around the ssl functions which are throwing the exception? IMO it's easier to see what's going on when the try/except is closer to the code that will throw the exception.

kaifeng: Since this method is ssl operation intensive, the intent to have an exception wrapper is to have a common exception handling logic that can be extended, currently we only handle this special exception though. It works for me if we just try-catch load_cert_chain at the moment.

mpeters (approved):
lgtm

PR #423, opened 2020-12-01
Fixes an error with `keylime_tenant -c reactivate`

maugustosilva:
Fixes an error where `keylime_tenant -c reactivate` is called, 
and the`verifier` returns an error complaining about the lack of the key
`provide_V` on the "agent" object. With this small change, `reactivate`
can be properly issued, even after a `verifier` restart, and a new U/V
(and pubkey) exchange will be done.


mpeters (approved):
lgtm

PR #424, opened 2020-12-02
Optionally validate client certificate

kaifeng:
Currently it's not configurable, this PR adds two configuration
options named check_client_cert and defaults to true for verifier
and registrar, when set to false will not validate client certificate.

### DIFF
--- a/keylime.conf
+++ b/keylime.conf
@@ -160,6 +160,9 @@ private_key = default
 # generated CA private key.
 private_key_pw = default
 
+# Whether verifier validates client certificate
+check_client_cert = True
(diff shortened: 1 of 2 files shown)


mpeters (approved):
lgtm

mpeters:
This looks good but needs to be rebased and conflicts fixed. Also there were some failing tests. I don't think this change caused it because it should just keep the old behaviour as the default correct? But we'll need to get those passing just to make sure.

kaifeng:
Previous commit did have issues, should be ok now.

PR #425, opened 2020-12-02
Fix `keylime_tenant -c list`.

maugustosilva:
When this command is called, the agent uuid passed to `do_cvstatus` must
be made empty. If not, instead of correctly issuing `GET /agents/`,
returning a list of all agents on a verifier, the aforementioned command
will issue `GET /agents/<default UUID in case one is not specified>`,
resulting in error.


mpeters (approved):
lgtm

Issue #412, opened 2020-11-23
Password for client-private.pem not being used


stefanberger:
# Environment
 * OS / version: Linux
 * Keylime version: master (a242f2ce12547d473b82bb39ae5fee061d2a95c3)

# Description 

> keylime_tenant -c delete -u 'D432FBB3-D2F1-4A97-9EF7-75BD81C00000'
Using config file /etc/keylime.conf
2020-11-23 16:59:38.313 - keylime.tenant - INFO - Setting up client TLS in /var/lib/keylime/cv_ca
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/dist-packages/urllib3/connectionpool.py", line 677, in urlopen
    chunked=chunked,
  File "/usr/local/lib/python3.6/dist-packages/urllib3/connectionpool.py", line 381, in _make_request
    self._validate_conn(conn)
  File "/usr/local/lib/python3.6/dist-packages/urllib3/connectionpool.py", line 978, in _validate_conn
    conn.connect()
  File "/usr/local/lib/python3.6/dist-packages/urllib3/connection.py", line 371, in connect
    ssl_context=context,
  File "/usr/local/lib/python3.6/dist-packages/urllib3/util/ssl_.py", line 368, in ssl_wrap_socket
    raise SSLError("Client private key is encrypted, password is required")
urllib3.exceptions.SSLError: Client private key is encrypted, password is required

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.6/dist-packages/requests/adapters.py", line 449, in send
    timeout=timeout
  File "/usr/local/lib/python3.6/dist-packages/urllib3/connectionpool.py", line 727, in urlopen
    method, url, error=e, _pool=self, _stacktrace=sys.exc_info()[2]
  File "/usr/local/lib/python3.6/dist-packages/urllib3/util/retry.py", line 439, in increment
    raise MaxRetryError(_pool, url, error or ResponseError(cause))
urllib3.exceptions.MaxRetryError: HTTPSConnectionPool(host='127.0.0.1', port=8881): Max retries exceeded with url: /agents/D432FBB3-D2F1-4A97-9EF7-75BD81C00000 (Caused by SSLError('Client private key is encrypted, password is required',))

## Expected behavior vs. actual behavior 

Should work.

## Steps to reproduce problem
1. keylime_tenant -c delete -u 'D432FBB3-D2F1-4A97-9EF7-75BD81C00000'

Issue #420, opened 2020-11-27
Registrar key password is never used


lukehinds:
`my_key_pw` is never used:

https://github.com/keylime/keylime/blob/be4b874bf9bc232e6e1dcf747f4a8cd329a33858/keylime.conf#L185

https://github.com/keylime/keylime/blob/be4b874bf9bc232e6e1dcf747f4a8cd329a33858/keylime/registrar_client.py#L41

mpeters:
@maugustosilva and @lukehinds I don't see this config variable in the code anymore so I assume it was removed. It does however still exist in the `keylime.conf` in the repo, so I assume we can remove it, correct?

PR #426, opened 2020-12-02
Fully remove all passwords for private keys

maugustosilva:
Fix issues #412 and #420

### DIFF
--- a/keylime.conf
+++ b/keylime.conf
@@ -135,7 +135,7 @@ cloudverifier_port = 8881
 #
 # Set to "generate" to automatically generate a CA/certificates in the 'cv_ca'
 # directory.
-# Note: if set to "generate", the 'ca_cert', 'my_cert', and 'private_key'
+# Note: if set to "generate", the 'ca_cert', and 'my_cert'
 # options must all be set to "default".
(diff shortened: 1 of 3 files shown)


keylime.conf
```diff
@@ -253,12 +241,6 @@ private_key = default
 # EK certificate storage location (directory).
 tpm_cert_store = /var/lib/keylime/tpm_cert_store/
 
-# Set the password needed to decrypt the private key file.
-# This should be set to a strong password.
-# if you are using the auto generated keys from the CV, set the same password
-# here as you did for 'private_key_pw' in the [cloud_verifier] section.
-private_key_pw = default
```
kaifeng, line 260: We can't simply remove the password, there is validation at ca_utils, when it's None will not generate certifcates.
Tried '' as password, but haven't tested if it works.

maugustosilva: Understood. Can you give me more details on the failure?

kaifeng: @maugustosilva the verification code is here:
https://github.com/keylime/keylime/blob/master/keylime/ca_util.py#L73
this will cause the cert generation fail.

PR #427, opened 2020-12-03
Simplify database configuration

kaifeng:
Currently database configuration is breaked into several options,
we could just pass a connection string to SQLAlchemy instead of
constructing the url.

E.g.:

sqlite:////var/lib/keylime/cv_data.sqlite
mysql+pymysql://keylime:keylime@db_host/verifier

### DIFF
--- a/keylime.conf
+++ b/keylime.conf
@@ -192,7 +192,16 @@ registrar_private_key_pw = default
 # Database URL Configuration
 # See this document https://keylime-docs.readthedocs.io/en/latest/installation.html#database-support
 # for instructions on using different database configurations.
-# The default is "sqlite" and will be located at "/var/lib/keylime/cv_data.sqlite".
+# Database configuration via drivername, username, password, etc is deprecated
+# and superseded by database_url, when database_url is set, other database
+# fields will be ignored.
+# An example of database_url value for using sqlite:
+#   sqlite:////var/lib/keylime/cv_data.sqlite
+# An example of database_url value for using mysql:
+#   mysql+pymysql://keylime:keylime@keylime_db:[port]/verifier?charset=utf8
+# If database_url is not set, the deprecated default is "sqlite" and will be
+# located at "/var/lib/keylime/cv_data.sqlite".
+database_url =
 drivername = sqlite
 username = ''
 password = ''
@@ -200,6 +209,7 @@ host = ''
 port = ''
 database = cv_data.sqlite
 query = ''
+
 auto_migrate_db = True
 
 
@@ -452,7 +462,16 @@ check_client_cert = True
 # Database URL Configuration
 # See this document https://keylime-docs.readthedocs.io/en/latest/installation.html#database-support
 # for instructions on using different database configurations.
-# The default is "sqlite" and will be located at "/var/lib/keylime/reg_data.sqlite".
+# Database configuration via drivername, username, password, etc is deprecated
+# and superseded by database_url, when database_url is set, other database
+# fields will be ignored.
+# An example of database_url value for using sqlite:
+#   sqlite:////var/lib/keylime/reg_data.sqlite
+# An example of database_url value for using mysql:
+#   mysql+pymysql://keylime:keylime@keylime_db:[port]/registrar?charset=utf8
+# If database_url is not set, the deprecated default is "sqlite" and will be
+# located at "/var/lib/keylime/reg_data.sqlite".
+database_url =
 drivername = sqlite
 username = ''
 password = ''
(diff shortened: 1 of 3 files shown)


mpeters:
I kind of like this idea but I do have a concern about the drivername. I believe we use it in a couple of places to do DB specific things (like UTF8 encoding for mysql). So even if someone is using the url version we need to be able to derive the drivername from that. Also we need to make sure it remains backwards compatible and for a better user experience make sure that users can't have both the new and old options present at the same time.

kaifeng:
> I kind of like this idea but I do have a concern about the drivername. I believe we use it in a couple of places to do DB specific things (like UTF8 encoding for mysql). So even if someone is using the url version we need to be able to derive the drivername from that. Also we need to make sure it remains backwards compatible and for a better user experience make sure that users can't have both the new and old options present at the same time.

Charset can be specified in the url, e.g.: mysql+pymysql://keylime:keylime@db_host/verifier?charset=utf8
Actually this is the standard strategy in openstack. For backwards compatibility I think we could use the new option if it's set, or construct from old options if not, I'd advocate to deprecate the old one.

mpeters:
> Charset can be specified in the url, e.g.: mysql+pymysql://keylime:keylime@db_host/verifier?charset=utf8
> Actually this is the standard strategy in openstack. For backwards compatibility I think we could use the new option if it's set, or construct from old options if not, I'd advocate to deprecate the old one.

I have mixed feelings about this one. If keylime needs UTF8 to work properly then is it right to rely on the user to configure it correctly? It's also true that if the user borks their db configuration (wrong pw or hostname) then it won't work right either. The difference is that the user won't necessarily know it's wrong because things will look like they work but the data could be slightly off.

Maybe the compromise is to just do better verification so we can make sure that UTF8 is specified? How would that work for other db connection strings? Do we also need to make sure the drivers are all supported?

kaifeng:
> 
> 
> > Charset can be specified in the url, e.g.: mysql+pymysql://keylime:keylime@db_host/verifier?charset=utf8
> > Actually this is the standard strategy in openstack. For backwards compatibility I think we could use the new option if it's set, or construct from old options if not, I'd advocate to deprecate the old one.
> 
> I have mixed feelings about this one. If keylime needs UTF8 to work properly then is it right to rely on the user to configure it correctly? It's also true that if the user borks their db configuration (wrong pw or hostname) then it won't work right either. The difference is that the user won't necessarily know it's wrong because things will look like they work but the data could be slightly off.
> 
> Maybe the compromise is to just do better verification so we can make sure that UTF8 is specified? How would that work for other db connection strings? Do we also need to make sure the drivers are all supported?

I guess correct configurations should be guaranteered by deployers, we can't cover everything and do anything like wrong password or hostname. Different databases are just details and exactly the abstraction provided by sqlalchemy for us, typically we don't need to handle their differences except in extremely cases, but I have seen none at the moment.
Basically utf8 is considered the default behavior nowadays, by default sqlalchemy/dialet uses utf8 encoding [1][2] so I think no special treatment is required.

[1] https://docs.sqlalchemy.org/en/13/core/engines.html#sqlalchemy.create_engine
[2] https://docs.sqlalchemy.org/en/13/dialects/postgresql.html#psycopg2-unicode

maugustosilva:
@mpeters allow me to join @kaifeng on supporting this change. I have been working with external databases for quite a while now, and I welcome this change due to significant simplification that it brings to the configuration file.

mpeters:
If we are going forward with this, I'd rather we explicitly deprecate the old options. Can we add a log message to that effect if the DB connection info doesn't come from `database_url` ?

keylime.conf
```diff
@@ -193,13 +193,16 @@ registrar_private_key_pw = default
 # See this document https://keylime-docs.readthedocs.io/en/latest/installation.html#database-support
 # for instructions on using different database configurations.
 # The default is "sqlite" and will be located at "/var/lib/keylime/cv_data.sqlite".
+# database_url takes precedence and when set, other database fields are ignored.
```
mpeters, line 196: Can we have an example sqlite value for database_url in the comments? Or as the default?

mpeters (changes requested):

keylime.conf
kaifeng: NP to document in the comments, but make it default needs to remove the special logic on root/non-root because they use different paths. Sinc we agree to make verifier/registrar running as non-root, I guess we can just go ahead and set the fixed path to /var/lib/keylime/cv_data.sqlite?

mpeters: Yeah, we want to remove running as root for verifier/registrar but I don't want to hold this change up for that. So maybe just documenting it in the comments will be good enough for now.

mpeters (approved):
lgtm

PR #429, opened 2020-12-03
End of PTL term for @lhinds and propose @mpeters

lukehinds:
This PR is for me to step down as PTL of the project and let
someone else have a chance to run as PTL for a term.

I will still 100% be active on the project.

I would also like to make a proposal of @mpeters taking over from
me. Michael has by all accounts been driving the project for a
while now, making lots of reviews and working to design and deliver
some key features.

The process is outlined here:

https://github.com/keylime/keylime/blob/master/GOVERNANCE.md#changes-in-project-lead

Typically the PR should be opened no earlier then 6 weeks before
the end of term, however I think we can use a caveat that I have
been in the role for well over a year now.

At the same time, anyone else is welcome to raise a PR with a
nomination of their own.

I will leave this PR open for 2 weeks (unless objections are made
and a community member feels this should be longer). 4 weeks is
max it can stay open.

I will also mail the list with details of my stepping down.

### DIFF
--- a/MAINTAINERS
+++ b/MAINTAINERS
@@ -1,4 +1,4 @@
-Project Lead: @lukehinds
+Project Lead: @mpeters
 
 Maintainers:
 
@@ -6,7 +6,6 @@ Maintainers:
 @jetwhiz
 @nabilschear
 @amylily1011
-@mpeters
 @maugustosilva
 @font
 @axelsimon


ashcrow (approved):

mpeters (approved):
lgtm

MAINTAINERS
```diff
@@ -1,12 +1,11 @@
-Project Lead: @lukehinds
+Project Lead: @mpeters
 
 Maintainers:
 
 @lukehinds
 @jetwhiz
 @nabilschear
 @amylily1011
-@mpeters
```
mpeters, line 9: Just out of curiosity, is it correct to be removed from the list of maintainers? I noticed you were present in both before this change.

lukehinds: Its neither correct / incorrect and has no bearing. I just thought it might be more tidy, but happy to go either way if you have an idea on how it could work.

amylily1011 (approved):

MAINTAINERS
mpeters: Nope, just curious

axelsimon (approved):
+1
Seems like a sensible move to me.

font (approved):
Great choice! :+1:

nabilschear (approved):

jetwhiz (approved):

maugustosilva:
a very late +1 :-)

PR #430, opened 2020-12-04
Allow specify absolute path for certifcates

kaifeng:
Currently cert related configurations are assumed to be under the
same path, but it's not usually satisfied when certificate is supplied
externally.

This patch provides support to use a cert from configuration option when
the value is an absolute path.


mpeters (approved):
lgtm

PR #434, opened 2020-12-08
Support hostname as agent id

kaifeng:
hostname is considered unique and can be used as agent id.
Other forms of identifiers need to be manually retrieved
thus hard to automate.


mpeters (approved):
lgtm

PR #437, opened 2020-12-09
Remove registrar_require_cert

kaifeng:
This option was introduced but not actually used, should be a
left over of renames in the previous patch.


mpeters (approved):
lgtm

PR #439, opened 2020-12-10
Overhaul container infra to support migrating testing to GH Actions

mbestavros:
These changes should support running Keylime's TPM 2.0 tests within a container without needing any of the hacks that were present previously.

Tagging @lukehinds and @puiterwijk for review.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,11 +1,11 @@
 ##############################################################################
 # keylime TPM 2.0 Dockerfile
 #
-# This file is for automatic test running of Keylime.
+# This file is for automatic test running of Keylime and rust-keylime.
 # It is not recommended for use beyond testing scenarios.
 ##############################################################################
 
-FROM fedora:31
+FROM fedora:latest
 MAINTAINER Luke Hinds <lhinds@redhat.com>
 LABEL version="1.0.1" description="Keylime - Bootstrapping and Maintaining Trust in the Cloud"
 
@@ -18,6 +18,8 @@ ENV TPM_HOME ${HOME}/swtpm2
 
 # Packaged dependencies
 ENV PKGS_DEPS="automake \
+cargo \
+clang-devel \
 dbus-devel \
 dnf-plugins-core \
 gcc \
@@ -25,8 +27,10 @@ git \
 glib2-devel \
 glib2-static \
 gnulib \
+kmod \
 libselinux-python3 \
 libtool \
+libtpms \
 make \
 openssl-devel \
 procps \
@@ -35,6 +39,7 @@ python3-dbus \
 python3-devel \
 python3-m2crypto \
 python3-pip \
+python3-requests \
 python3-setuptools \
 python3-sqlalchemy \
 python3-tornado \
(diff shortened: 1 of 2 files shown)


lukehinds (commented):
For the `Dockerfile`

May as well bump this to `fedora33`

Rename it from `docker/Dockerfile-tpm20` to `docker/Dockerfile`

And we need the following packages added: python3-requests

@mpeters anything else we need as of late?

mbestavros:
@lukehinds Done!

docker/Dockerfile
```diff
@@ -5,7 +5,7 @@
 # It is not recommended for use beyond testing scenarios.
 ##############################################################################
 
-FROM fedora:31
+FROM fedora:33
 MAINTAINER Luke Hinds <lhinds@redhat.com>
```
puiterwijk, line 9: Would it be an idea to use an email list for Keylime here perhaps?

docker/Dockerfile
```diff
@@ -5,7 +5,7 @@
 # It is not recommended for use beyond testing scenarios.
 ##############################################################################
 
-FROM fedora:31
+FROM fedora:33
```
puiterwijk, line 8: Maybe instead of `:33`, which we will forget to update probably, just use `:latest`?

puiterwijk (commented):
I think you want to make sure to add `swtpm` and `swtpm-tools` to the `PKGS_DEPS` package list.
Also, if we want to use the same container for rust-keylime, we'd want the rust toolchain and other deps (See [here](https://github.com/keylime/rust-keylime/blob/master/docker/fedora/fedora.Dockerfile)) as well probably.

docker/Dockerfile
```diff
@@ -54,28 +61,6 @@ RUN dnf makecache && \
...
-  tar -zxvf ibmtpm1119.tar.gz
-WORKDIR ${TPM_HOME}/src
-RUN make && \
-  install -c tpm_server /usr/local/bin/tpm_server
-
-VOLUME [ "/sys/fs/cgroup" ]
-
-ENTRYPOINT ["/lib/systemd/systemd"]
+# Install Rust
+RUN curl https://sh.rustup.rs -sSf | bash -s -- -y
```
puiterwijk, line 65: If we're using Fedora, I'd strongly suggest to just add `rust` and `cargo` to the package list instead of installing it via Rustup.

puiterwijk (approved):
LGTM.

lukehinds (approved):

docker/Dockerfile
mpeters: Yeah, that's probably a good idea to do in a lot of places. We have a domain name, so we should be able to get some emails for it and forward those to different people at different times.

mpeters (commented):
> @mpeters anything else we need as of late?

Nope that's the only new one after sqlalchemy that was added.

But taking a step back, maybe we don't need these to be built into the image itself. If we update the requirements.txt file in keylime the keylime build should add the right packages right? Do we really need to have that in 2 places?

PR #440, opened 2020-12-10

stefanberger:
This PR fixes another set of issues detected by pylint


keylime/tpm/tpm2.py
```diff
@@ -328,6 +328,9 @@ def __fingerprint(cmd):
             pass
         return fprt
 
+    def run(self, cmd):
```
kaifeng, line 331: Ideally we should expose lock=True here, since __run defaults to lock=True, the caller (vtpm_manager) is responsible for passing lock=False.
But I think it's ok here since there is only one caller here, and the __run methods is planned to be refactored in the shell issue.

kaifeng (approved):
Thank you!

Issue #441, opened 2020-12-11
Separate configuration options for each service


kaifeng:
Currently each service has their own section in the configuration file, but used without limit, like registrar accesses verifier's options, agent accesses registrar's options, tenant accesses sections of other services. This cause difficulties when we deploy the service under proxy like load balancer, because the public endpoint of a service is different from the backend address.
It could be resolved by having each service only reads its own section, [general] is a global section so it can be used for common options applied to every service.

lukehinds:
As in the address to contact the registrar maybe different to the address the registrar binds to (so we may want two values) so we split it out?

registrar.conf
`listen_ip = 127.0.0.1`  # registrar bind address

verifier.conf
`registar_ip = 192.168.0.1` # registrar endpoint address

kaifeng:
> 
> 
> As in the address to contact the registrar maybe different to the address the registrar binds to (so we may want two values) so we split it out?
> 
> registrar.conf
> `listen_ip = 127.0.0.1` # registrar bind address
> 
> verifier.conf
> `registar_ip = 192.168.0.1` # registrar endpoint address

Yes, we could use different configuration files but this has more impact, so a better way is to add missing options in each section.

lukehinds:
understood, this is good proposal. 

I am open to having different configs, but I do see your point, it would make it more complicated change as we would need to fix occurances of configparser calling and redo the setup.py

maugustosilva:
+1 on this proposal. The current configuration system quickly becomes cumbersome when dealing with more "exotic" configurations (e.g., multiple verifiers, load balancers, external databases). I ended up writing a rather complex python configurator to deal with these redundant information on different sections (to automatically generate a configuration file), but never bothered submitting it, since it was clearly a temporary solution for a much bigger problem. @kaifeng's proposal would be an important step in fixing it.

maugustosilva:
@lukehinds @kaifeng It might look as a pet peeve of mine, but in my experience it made the initial process of understanding the inner workings of keylime (unnecessarily) harder: some options which specific to the verifier (e.g., `tls_dir`) are not explicitly marked as so (i.e., it should be `cloudverifier_tls_dir` or `verifier_tls_dir`, while options specific to the registrar are (i.e., we have `registrar_tls_dir`). In any re-organization of the sections (and options) on the configuration file, it would be good to take some time to sanitize the names of these parameters.

kaifeng:
[...]

PR #448, opened 2020-12-12
Fill in service addresses for each section

kaifeng:
Adds service addresses that was reading from section of other service,
so that each service reads his own section to get the full picture.
Default value is populated as dev env.

Relates-to: #441


mpeters (approved):
This is something that's bitten me a couple of times. Thanks for fixing!
LGTM

PR #450, opened 2020-12-14
Allow using absolute path of cert for registrar client

kaifeng:
Allow using absolute path of cert files for the registrar client.

Relates-to: #430


mpeters (approved):
lgtm

PR #453, opened 2020-12-15
Revert changes to test_wrapper.sh to make Travis work

mbestavros:


kaifeng (approved):
lgtm, it seems swtpm is not shipped with bionic, we probably have to shift to using fedora when apply this change.

PR #458, opened 2020-12-16
Remove COPY command causing container build to fail

mbestavros:


mbestavros:
Not actually necessary; closing.

PR #461, opened 2020-12-17

stefanberger:
This PR fixes another set of issues detected by pylint.


keylime/migrations/env.py
```diff
@@ -1,3 +1,5 @@
+#pylint: disable=no-member
```
kaifeng, line 1: It appears no-memer is a useful check, probably because we are running pylint without requirements satisfied?
I think we should should consider using virtualenv with all requiremennts installed and reopen this check.

kaifeng (approved):
lgtm

kaifeng:
@stefanberger not related with this PR, do you happen to know how the PCR is populated as a qemu guest? I don't know if SeanBIOS is doing some measurement at booting.

keylime/migrations/env.py
mpeters: There also appear to be a lot of scenarios where pylint can't tell the objects type (or it could be multiple types) and there are lots of false positives with no-member check.

I'm actually fine skipping that entirely if we want to. It's trying to treat python like a static language and there are tons of places in real, good, python code where that isn't possible. And I'd prefer to drop the check rather than have tons of `disable=no-member` comments sprinkled throughout the code TBH.

mpeters: Or maybe a `TYPECHECK` section of our pylintrc file such as https://www.lesinskis.com/pylint-false-positives.html ?

stefanberger:
> @stefanberger not related with this PR, do you happen to know how the PCR is populated as a qemu guest? I don't know if SeanBIOS is doing some measurement at booting.

SeaBIOS and EDk2 (UEFI) are both taking measurements and are populating PCRs 0-7.

keylime/migrations/env.py
stefanberger: I can drop this patch (for now) if you prefer... We'll have to somehow start marking the remaining suppressed pylint checks so we don't revisit them again. I have a list that should be possible to fix but some are really a lot of work, like documentation...

stefanberger: This is how alembic's op.py looks like. I can imagine that this creates a lot of the issues shown below: 
```
from .operations.base import Operations

# create proxy functions for
# each method on the Operations class.
Operations.create_module_class_proxy(globals(), locals())
```

```
************* Module keylime.migrations.versions.eeb702f77d7d_allowlist_rename
keylime/migrations/versions/eeb702f77d7d_allowlist_rename.py:41:4: E1101: Module 'alembic.op' has no 'alter_column' member (no-member)
keylime/migrations/versions/eeb702f77d7d_allowlist_rename.py:47:4: E1101: Module 'alembic.op' has no 'alter_column' member (no-member)
************* Module keylime.migrations.versions.8a44a4364f5a_initial_database_migration
keylime/migrations/versions/8a44a4364f5a_initial_database_migration.py:34:4: E1101: Module 'alembic.op' has no 'create_table' member (no-member)
keylime/migrations/versions/8a44a4364f5a_initial_database_migration.py:50:4: E1101: Module 'alembic.op' has no 'drop_table' member (no-member)
keylime/migrations/versions/8a44a4364f5a_initial_database_migration.py:56:4: E1101: Module 'alembic.op' has no 'create_table' member (no-member)
keylime/migrations/versions/8a44a4364f5a_initial_database_migration.py:81:4: E1101: Module 'alembic.op' has no 'drop_table' member (no-member)
```

I think we should enable the check anyway since it is useful for much other code.

mpeters: I don't want to hold up some of your other good fixes because we're in flux about how to proceed with this one. So if you want to drop the E1101 change for now that would be fine.

stefanberger: > I don't want to hold up some of your other good fixes because we're in flux about how to proceed with this one. So if you want to drop the E1101 change for now that would be fine.

Dropped it now.

mpeters (approved):
lgtm

kaifeng:
> 
> 
> > @stefanberger not related with this PR, do you happen to know how the PCR is populated as a qemu guest? I don't know if SeanBIOS is doing some measurement at booting.
> 
> SeaBIOS and EDk2 (UEFI) are both taking measurements and are populating PCRs 0-7.

Get it, thank you!

kaifeng:
We could disable certain types of lint that don't suitable at the moment, I haven't hit issues like no-member at flake8 (feel it's more widely).

PR #465, opened 2020-12-17
Implement a basic semantic release configuration

mbestavros:
This should be a starting point for an automated release process. Uses https://github.com/marketplace/actions/action-for-semantic-release.


mpeters (approved):
I'll probably need you to walk me through all of the automation possibilities this opens up

PR #468, opened 2020-12-18

stefanberger:
This PR fixes another set of issues detected by pylint.


auto-ipsec/racoon/src/local_action_update_crl.py
```diff
@@ -43,7 +43,7 @@ async def execute(json_revocation):
         oldcrl = f.read()
 
     updated = False
-    for i in range(10):
+    for _ in range(10):
```
kaifeng, line 46: I kind of feel this type of lint is overchecking, rarely seen tbh

test/test_restful.py
```diff
@@ -147,7 +147,7 @@ def setUpModule():
                 print("Found dbus service:", str(service))
                 try:
                     print("Restarting tpm2-abrmd.service.")
-                    job = manager.RestartUnit('tpm2-abrmd.service', 'fail')
+                    _ = manager.RestartUnit('tpm2-abrmd.service', 'fail')
```
kaifeng, line 150: Since the return value is not used, we can just skip the value assignment, does pylint complains too?

auto-ipsec/racoon/src/local_action_update_crl.py
stefanberger: `i` is unused ...

kaifeng (commented):
generally looks good, but I feel uncomfortable on a few type of changes, feeling it's overchecking.

test/test_restful.py
stefanberger: It would not complain, so removed it entirely.

stefanberger:
> generally looks good, but I feel uncomfortable on a few type of changes, feeling it's overchecking.

A c or go compiler would also complain about unused variables

auto-ipsec/racoon/src/local_action_update_crl.py
kaifeng: yeah, it's not used, just never hit this kind of comlains :)

kaifeng (approved):

Issue #447, opened 2020-12-11
The `keylime_tenant` tool no longer allows the overriding of the address of the `verifier` via command line option


maugustosilva:
This becomes a real problem (i.e., bug) in a multi-`verifier` environment. The scenario for me is: a single "deployer" node from where multiple `agents` are added to multiple `verifiers`. I will give a simple example with `-c status` (same goes for `-c add` or `-c update`, of course.

```
[root@cssf4-uc0 /]# /usr/local/bin/keylime_tenant -c status -t 100.64.4.7 --cv 100.64.4.2 -u 44D88F7F-C85E-55B8-8B16-6E19109D3458
Using config file /etc/keylime.conf
2020-12-11 20:13:58.755 - keylime.tenant - INFO - Setting up client TLS in /var/lib/keylime/cv_ca
100.64.4.2:8881
2020-12-11 20:13:58.761 - urllib3.connectionpool - DEBUG - Starting new HTTPS connection (1): 100.64.4.2:8881
/usr/lib/python3.8/site-packages/urllib3/connectionpool.py:997: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings
  warnings.warn(
2020-12-11 20:13:58.818 - urllib3.connectionpool - DEBUG - https://100.64.4.2:8881 "GET /agents/44D88F7F-C85E-55B8-8B16-6E19109D3458 HTTP/1.1" 404 60
2020-12-11 20:13:58.819 - keylime.tenant - ERROR - Agent 44D88F7F-C85E-55B8-8B16-6E19109D3458 does not exist on the verifier. Please try to add or update agent
```

Now, this is a fully expected behavior, because even though I had `100.64.4.1` listed on my `/etc/keylime.conf`, I dynamically overrode it via the (well documented) CLI option `--cv`.

Now this is the behavior that I see currently.

```
[root@cssf4-uc0 /]# /usr/local/bin/keylime_tenant -c status -t 100.64.4.7 --cv 100.64.4.2 -u 44D88F7F-C85E-55B8-8B16-6E19109D3458
Using config file /etc/keylime.conf
2020-12-11 20:21:20.677 - keylime.tenant - INFO - Setting up client TLS in /var/lib/keylime/cv_ca
100.64.4.1:8881
[...]

PR #469, opened 2020-12-19
Fix the issue that tenant can't override verifier address from comman line

kaifeng:
Create verifier_base_url on demand.

Resolves: #447

### DIFF
--- a/keylime/tenant.py
+++ b/keylime/tenant.py
@@ -79,17 +79,16 @@ class Tenant():
     def __init__(self):
         """ Set up required values and TLS
         """
-        self.agent_ip = None
         self.nonce = None
+        self.agent_ip = None
+        self.agent_port = config.get('cloud_agent', 'cloudagent_port')
         self.verifier_ip = config.get('cloud_verifier', 'cloudverifier_ip')
         self.verifier_port = config.get('cloud_verifier', 'cloudverifier_port')
-        self.agent_port = config.get('cloud_agent', 'cloudagent_port')
+        self.registrar_ip = config.get('registrar', 'registrar_ip')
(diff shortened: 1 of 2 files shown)


kaifeng:
I wonder if there is a mechanism to recheck the travis job, the failure doesn't seem related with the change and the log implies a build error.

kaifeng:
> 
> 
> I wonder if there is a mechanism to recheck the travis job, the failure doesn't seem related with the change and the log implies a build error.

things turned out that the test is not consistent with code.

maugustosilva:
I've tested it, and the very same bug persists. The `keylime_tenant` CLI is still ignoring the `-v` option (I actually don't see this CLI parameter being processed on the code).

kaifeng:
> 
> 
> I've tested it, and the very same bug persists. The `keylime_tenant` CLI is still ignoring the `-v` option (I actually don't see this CLI parameter being processed on the code).

This is stange, the -v is processed here https://github.com/keylime/keylime/blob/master/keylime/tenant.py#L938-L939
Any idea why this change didn't work for you?

kaifeng:
> 
> 
> > I've tested it, and the very same bug persists. The `keylime_tenant` CLI is still ignoring the `-v` option (I actually don't see this CLI parameter being processed on the code).
> 
> This is stange, the -v is processed here https://github.com/keylime/keylime/blob/master/keylime/tenant.py#L938-L939
> Any idea why this change didn't work for you?

Well, it appears we have naming issues here, cloudverifier_ip and verifier_ip are both used and easily get wrong..
Will push an update soon.

maugustosilva:
Just tested it, and can confirm it is working now. My vote is to merge it as is.

mpeters (approved):
lgtm

PR #473, opened 2020-12-21
Remove lock in the command execution

kaifeng:
A lock is aleady held when executing tpm commands, there is no
need to lock in the underlying command, and this is typically not
a good idea.


mpeters:
@kaifeng just so I understand this a bit more, you say that a lock is already being held in the underlying code, do you mean in the tpm tools code we are calling? Or in another part of the keylime code?

kaifeng:
> 
> 
> @kaifeng just so I understand this a bit more, you say that a lock is already being held in the underlying code, do you mean in the tpm tools code we are calling? Or in another part of the keylime code?

In the tpm.__run we have a tpmutilLock which protects places we need, but when lock is True, the exec_cmd.run will lock again, this is kind of double lock and I think we should avoid, I don't mean there are issues at the moment, just a general consideration.

mpeters:
Makes sense, thanks.

mpeters (approved):

PR #474, opened 2020-12-22
Address issues related to C0411

stefanberger:
This PR addresses issues related to C0411 and extends the .gitignore with temporary editor files that and with '~'.


kaifeng (approved):
lgtm

Issue #436, opened 2020-12-09
Make keylime running as non-root


kaifeng:
keylime can run as root or non-root, but we have to treat them differently for path, validation, etc.
The proposal is to persistently make keylime running as non-root and remove REQUIRE_ROOT, for actions that requires root privilege, sudo should be used.
This will simplify our implementation and have better security control.

lukehinds:
There is an enhancement in place for this work: https://github.com/keylime/enhancements/blob/master/23_dedicated_keylime_user_account.md

I agree it definitely needs to happen and there is nothing stopping work getting underway on the registrar and verifier, the agent is a little more involved as TPM functions require root. Do you know how health rootwrap is @kaifeng , i remember it from openstack. 

tagging @axelsimon who was looking into this.

kaifeng:
@lukehinds openstack is migrating from rootwrap to privsep, rootwrap is considered less secure. Some projects have finished this transition while some are on the way.

axelsimon:
Thanks for bringing this up @kaifeng. Completely agree with your proposal.
i'm looking into it currently. Thanks for mentioning privsep, i'll look into it.

Currently, `keylime_registrar` and `keylime_verifier` run as a keylime user, although they fail to log.
`keylime_webapp` doesn't run, and i'm looking into it.
`keylime_agent` likely needs root for the TPM access, so i've not investigated that one yet.

maugustosilva:
@axelsimon @kaifeng @mpeters So, I tested running the `agent`, both from the command line and as a Systemd service, as non-root (in particular, I tried with user `tss` as this one allows for access to `/dev/tpmrm0`), and was successful in both cases. I believe the time has come to remove the check in https://github.com/keylime/keylime/blob/master/keylime/keylime_agent.py#L453

Please note that we do have the environment variable `KEYLIME_REQUIRE_ROOT` in https://github.com/keylime/keylime/blob/master/keylime/config.py#L71, but it is used mainly for testing, and changes the logging destination quite dramatically.

Thoughts?

mpeters:
I agree now is a good time to remove this. Is this change big enough to need an enhancement proposal to make sure we have all our ducks in a row?

maugustosilva:
I don't think it is particularly big, since we are talking about four loc only. My plan was to leave, at least for now, the `REQUIRE_ROOT` in place (since it is used for testing in different ways), and only remove that particular check.

maugustosilva:
[...]

PR #479, opened 2020-12-23
Allow non-root logging

kaifeng:
When running verifier with keylime user, it complains running as root,
but the /var/log/keylime/cloudverifier.log has correct privilge set
during package installation. Since REQUIRE_ROOT is hardcoded in code,
there is no means to treat verifier/registrar and agent differently,
this patch removes root checking and rely on the proper configuration
at deployment.

Also fixes the issue that registrar never logs to file.

Relates-to: #436


mpeters (approved):
lgtm

kaifeng:
The lint issue will be addressed by #504

PR #484, opened 2020-12-30
codestyle: Address issue related to C0411

stefanberger:
I doscovered this problem on an upgraded machine. For some reason this
problem was not found before with pylint 2.6.0.

This patch addresses the following issue detected by pylint

************ Module keylime.revocation_notifier
keylime/revocation_notifier.py:15:0: C0411: third party import "import simplejson as json" should be placed before "import zmq" (wrong-import-order)

### DIFF
--- a/keylime/revocation_notifier.py
+++ b/keylime/revocation_notifier.py
@@ -10,9 +10,9 @@ import time
 import os
 import sys
 import signal
-import zmq
 
 import simplejson as json
+import zmq
 
 from keylime import config
 from keylime import crypto


kaifeng (approved):
lgtm, the import-order in many files is not alphabically ordered though, not sure if this checking is under disabled state or pylint doesn't check this at all.

stefanberger:
> gtm, the import-order in many files is not alphabically ordered though, not sure if this checking is under disabled state or pylint doesn't check this at all.

pylint cares about the order of system headers versus 3rd part modules and so on. I was surprised to find it complaining about this patche's issue as if it cared about alphabetical order now but only in this file.

PR #489, opened 2021-01-07
codestyle: Address issues relate to C0200

stefanberger:
This patch addresses the following issued detected by pylint:

************* Module keylime.cryptodome
keylime/cryptodome.py:75:4: C0200: Consider using enumerate instead of iterating with range and len (consider-using-enumerate)
************* Module keylime.crypto
keylime/crypto.py:139:4: C0200: Consider using enumerate instead of iterating with range and len (consider-using-enumerate)

The conversion make use of the fact that the following are producing equivalent ouput:

a=bytearray([11,22,55])
for i in range(len(a)):
for i, _ in enumerate(a):
     print(i)

and

a=bytearray([11,22,55])
for i, _ in enumerate(a):
     print(i)

### DIFF
--- a/keylime/crypto.py
+++ b/keylime/crypto.py
@@ -136,7 +136,7 @@ def strbitxor(a, b):
     a = bytearray(a)
     b = bytearray(b)
     retval = bytearray(len(b))
-    for i in range(len(a)):
+    for i, _ in enumerate(a):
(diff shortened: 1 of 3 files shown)


mpeters (approved):
lgtm

PR #490, opened 2021-01-08
Improve tenant cli for status query

kaifeng:
status command doesn't need agent address, exclude it
from the argument validation so we don't need to remember
agent addresses when doing status query.


mpeters (approved):
lgtm

Issue #491, opened 2021-01-08
Add support to use TPM 2.0 Preboot Event log on Keylime


maugustosilva:
Adding all the relevant parties here: @puiterwijk  @mpeters @lukehinds @kaifeng 

# Motivation
We have a scenario with literally hundreds of different types of nodes, and simply "getting the contents of the boot aggregate (PCR 10) from a list of well-known values" has shown itself to be unfeasible from an operational standpoint (just to give some perspective: our nodes are netbooted, and the value of every PCR1 is different as it encodes the MAC address of the NIC). Our experiments with the "TPM2 Preboot Event log" demonstrated that, with a "recent enough" Kernel (e.g. 5.4.X), and a "recent enough" version of `tpm2-tools`, we can access the contents of the aforementioned log to re-create (or debug) the values of PCRs [0-9],14 without requiring a pre-measurement (i.e., a "golden value"). Therefore, I would like to propose we add this capability to Keylime.

# Summary
In a nutshell all we need is for the `agent` to be able to read the (binary) contents of `/sys/kernel/security/tpm0/binary_bios_measurements` and send it back to the `verifier`, where it would be "processed" in a manner similar to what we already do with the IMA runtime measurements.

# Estimated impact
This is not dissimilar to what is already done with IMA in the sense that: a) it is an optional capability, which should not interfere with the "regular" operation of KL and b) adds, to each interaction `verifier`<->`agent` interaction, no more than tens of KB (which is the total size of the "TPM2 Preboot Event log"). While we do expect very little impact on KeyLime's scalability by adding this capability, it is important to remember that we are constantly testing KL in a configuration with 5K nodes (with IMA), and can provide experimental evidence to back this hypothesis.  

# Plan of action
If you all agree with this change, I would like to propose a 3-phase implementation (i.e., roughly 3 individual PRs):

1) Retrieve boot logs in the `agent`, but do nothing with them in the `verifier` (`verifier` just receives log and ignores it), with new `keylime_tenant` CLI options (that was a request from @mpeters) to signal that this "tpm2 boot event log" should be read. This should be very similar to what we do currently with IMA, a new CLI option, hard-coded PCR (10 in case of IMA, [0-9], 14 in case of boot log) and hard-coded paths on the `/sys` (`/sys/kernel/security/ima/ascii_runtime_measurements` in case of IMA, `/sys/kernel/security/tpm0/binary_bios_measurements` in case of boot log) filesystem.

[...]

PR #494, opened 2021-01-12
PR 1 (of 3) to address #491

maugustosilva:
* The `agent` will now retrieve TPM 2.0 preboot event log and will ship
it to the `verifier`. The latter will do nothing with it (just receives
and ignores it)
* Adds a new `keylime_tenant` CLI option, `--mb_refstate` (that was a
request from @mpeters) to signal the need for the agent to collect the
the event log. At this point, `keylime_tenant` do not actually process
this option (it shall be processed in a manner similar to IMA in a
future PR), it is simply used to signal the agent that "tpm2 preboot
event log" should be read.
* Hard-coded PCRs [0-9], 14 and path
/sys/kernel/security/tpm0/binary_bios_measurements for the event log
* The `verifier` will output a debug logging message stating that it
received the "tpm2 event log", and will send it down to the `check_pcrs`
method, but there is no actual processing there (it will be done in
subsequent PR)


mpeters:
The CI failures are related to pylint failures. I haven't looked at why, whether this change needs linting or if it's something else

maugustosilva:
Most of the problems listed by pylint are due to the fact that we declare new parameters (i.e., `mb_measurement_list`, `mb_intended_state`) on the function definition of `check_quote` which are not **yet** used. Please remember that this is the first of three PRs which will implement the required functionality, and thus this particular pylint error is merely temporary.

mpeters:
> Please remember that this is the first of three PRs which will implement the required functionality, and thus this particular pylint error is merely temporary.

Sure, but we can't change CI for this one commit. You'll need to add pylint comments so pylint will ignore the violations temporarily which will be removed in subsequent commits.

keylime/config.py
```diff
@@ -295,6 +295,10 @@ def valid_regex(regex):
 
 IMA_PCR = 10
 
+# measured boot addons
+MEASUREDBOOT_PCRS = [ 0,1,2,3,4,5,6,7,8,9,14 ]
```
stefanberger, line 299: What uses PCR 14 ?

keylime/tenant.py
```diff
@@ -180,11 +180,21 @@ def init_add(self, args):
...
-        al_data = None
         if "allowlist" in args and args["allowlist"] is not None:
 
+            _policy_pcrs = list(self.tpm_policy.keys())
+            _policy_pcrs.remove('mask')
+
+            _protected_pcrs = [ config.IMA_PCR ]
+
+            for _pcr in _policy_pcrs :
+                if int(_pcr) in _protected_pcrs :
```
stefanberger, line 194: You are looking though all the _policy_pcrs to compare against the single IMA_PCR. Is this for future purposes? Because it would seem easier to look the other way around, trying `if is config.IMA_PCR in _protected_pcrs`...

keylime/tenant.py
```diff
@@ -219,6 +229,34 @@ def init_add(self, args):
...
+        # Read command-line path string TPM2 event log template
+        if "mb_refstate" in args and args["mb_refstate"] is not None:
+            _policy_pcrs = list(self.tpm_policy.keys())
+            _policy_pcrs.remove('mask')
+            _protected_pcrs = config.MEASUREDBOOT_PCRS
+
+            for _pcr in _policy_pcrs :
+                if int(_pcr) in _protected_pcrs :
+                    logger.error(f"WARNING: PCR {_pcr} is specified in \"tpm_policy\", but will in fact be used by TPM2 measured boot. Please remove it from policy")
+                    exit()
```
stefanberger, line 241: This would be an intersection of two lists of `int`s, no?
https://www.kite.com/python/answers/how-to-find-the-intersection-of-two-lists-in-python

keylime/cloud_verifier_common.py
```diff
@@ -206,7 +209,11 @@ def process_quote_response(agent, json_response):
                                      agent['tpm_policy'],
                                      ima_measurement_list,
                                      agent['allowlist'],
+                                     mb_measurement_list,
+                                     #agent['mb_intended_state'],
```
stefanberger, line 213: You probably want to remove this (and other commented code below). Otherwise rest LGTM.

keylime/tenant.py
stefanberger: Or maybe you have to do 
```
_policy_pcrs = [int(x) for x in _policy_pcrs]
if config.IMA_PCR in _policy_pcrs`:
```
or
```
_policy_pcrs = set([int(x) for x in _policy_pcrs])
is = set.intersection(set(_protected_pcrs), _policy_pcrs)
if len(is) > 0: ...
```

maugustosilva:
I just realized that I made too many mistakes (including not signing) on this PR. Will create a new one, cleaner

PR #496, opened 2021-01-12
Implementation of 'IMA signature verification with the 'ima-sig' template'

stefanberger:
This PR implements support for 'IMA signature verification with the 'ima-sig' template.'

With this PR, Keylime's verifier can use file signature verification on a Linux system that has IMA appraisal enabled in either enforcement or 'fix mode' (non-enforcing). The Keylime tenant tool allows users to add file signature verification keys in various formats (DER or PEM encode public or private keys or x509 certs) and of various key types (RSA, EC, DSA). The verifier then uses these keys to verify signatures it finds in the IMA measurement log entries 'ima-sig' template. It basically finds a new column there where IMA copied a file's extended `security.ima` file attribute into.

Each IMA measurement entry is either skipped when it is part of the exclude list or must have a valid signature or a valid hash in the allowlist. If a signature is available, a key must be available to verify it, otherwise the system is marked as being out-of-policy (marked red in gui). If no signature is available, it's hash must be part of the allowlist. If a valid signature was found and the allowlist has an entry for the file, the hash from the allowlist must match the hash from the measurement entry.


lukehinds:
nice PR! so the  benefit here is we can use ima sigs for noisey / mutable files that are hard to measure from their ever changing state?

stefanberger:
> nice PR! so the benefit here is we can use ima sigs for noisey / mutable files that are hard to measure from their ever changing state?

IMA signatures only work for unmutable files but you would only need one signature verification key to verify the signature on all of the unmutable files assuming you installed RPMs that contain file signatures created with one key. The signatures will be written out when the RPM is installed -- they are/can be part of an RPM. If on top of signature verfication you would want version verification, you would add an allowlist as well that covers the hashes and thus the version of the files/rpms.

stefanberger:
Mimi just gave me this link. Fedora 34 with signatures in the RPM files: https://lwn.net/SubscriberLink/842002/69b82eb2bd3ead6f/

stefanberger:
One thing to remember: If one signs the files with sha256, or the signature from a sha256 is stored in an RPM for example, then IMA has to be configure to use sha256 for hashing, otherwise nothing will verify.

mpeters (approved):
This looks very useful and well executed. Thanks!

Issue #485, opened 2021-01-05
Store allowlist in a separate table


kaifeng:
Current allowlist is saved with the agent record, there are two issues here:
1. An allowlist can't be shared between agents, so we can't apply the same data to identical OSes.
2. Since the ima measurements could be very large, it has impact on query efficiency when the allowlist does not participate in the processing.

The proposal is to create a new table named ``allowlists`` to save allowlist, and alter the ``agent.allowlist`` to a foreign key to the allowlist table. A full allowlist management and API is beyond this proposal.

maugustosilva:
I strongly support this move, for several additional reasons. One of the most relevant for me is the fact that I am working on support for "TPM2 event log" processing on KeyLime (which should allow a more intelligent verification of PCRs [0-9,14]) and this would require the introduction of a "golden event log template" or some such, which would be also shared across multiple nodes.

lukehinds:
For this change we should factor in multi-tenancy which is underway, so I expect we would need to consider somehow having a relational map to a `group_id` at some point in the future. This way in multi tenant scenario users could implement versioning to groups , e.g. a beta release list could go into a `testing` group and in turn not available to use a production system. 

https://github.com/keylime/enhancements/issues/7

lukehinds:
> I strongly support this move, for several additional reasons. One of the most relevant for me is the fact that I am working on support for "TPM2 event log" processing on KeyLime (which should allow a more intelligent verification of PCRs [0-9,14]) and this would require the introduction of a "golden event log template" or some such, which would be also shared across multiple nodes.

Sorry I should not hijack this thread, but if this requires changes to the agent code we should consider this will need to go into the rust agent, its likely worth syncing with @puiterwijk who has some plans to optimising tpm2 quotes in the rust project.

mpeters:
I agree, this sounds like a good idea for all of the right reasons :)

kaifeng:
> 
> 
> > I strongly support this move, for several additional reasons. One of the most relevant for me is the fact that I am working on support for "TPM2 event log" processing on KeyLime (which should allow a more intelligent verification of PCRs [0-9,14]) and this would require the introduction of a "golden event log template" or some such, which would be also shared across multiple nodes.
> 
> @maugustosilva
> 
> Sorry I should not hijack this thread, but if this requires changes to the agent code we should consider this will need to go into the rust agent, its likely worth syncing with @puiterwijk who has some plans to optimising tpm2 quotes in the rust project.

[...]

PR #499, opened 2021-01-13
Allowlist management

kaifeng:
This PR is part of #485 that will create an allowlist table and implement basic CRD API.
The next step will be drop tpm_policy/allowlist from the ``verifiermain`` table and reference the new allowlist table.

### DIFF
--- a/keylime/migrations/versions/eb869a77abd1_create_allowlist_table.py
+++ b/keylime/migrations/versions/eb869a77abd1_create_allowlist_table.py
@@ -1,0 +1,50 @@
+"""create_allowlist_table
+
+Revision ID: eb869a77abd1
+Revises: ae898986c6e9
+Create Date: 2021-01-12 10:54:45.263268
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = 'eb869a77abd1'
+down_revision = 'ae898986c6e9'
+branch_labels = None
+depends_on = None
+
+
+def upgrade(engine_name):
+    globals()["upgrade_%s" % engine_name]()
+
+
+def downgrade(engine_name):
+    globals()["downgrade_%s" % engine_name]()
+
+
+def upgrade_registrar():
+    pass
+
+
+def downgrade_registrar():
+    pass
+
+
+def upgrade_cloud_verifier():
+    op.create_table('allowlists',
+                    sa.Column('id', sa.Integer(), nullable=False),
+                    sa.Column('name', sa.String(255), nullable=False),
+                    sa.Column('tpm_policy', sa.Text(), nullable=True),
+                    sa.Column('vtpm_policy', sa.Text(), nullable=True),
+                    sa.Column('ima_policy', sa.Text(length=429400000), nullable=True),
+                    sa.PrimaryKeyConstraint('id'),
+                    sa.UniqueConstraint('name', name='uniq_allowlists0name'),
+                    mysql_engine='InnoDB',
+                    mysql_charset='UTF8'
+                    )
+
+
+def downgrade_cloud_verifier():
+    op.drop_table('allowlists')
--- a/keylime/tenant.py
+++ b/keylime/tenant.py
@@ -20,7 +20,6 @@ import tempfile
 import requests
 
 from cryptography.hazmat.primitives import serialization as crypto_serialization
-import simplejson as json
 
 from keylime.requests_client import RequestsClient
 from keylime.common import states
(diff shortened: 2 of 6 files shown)


keylime/migrations/versions/eb869a77abd1_create_allowlist_table.py
```diff
@@ -0,0 +1,50 @@
...
+def upgrade_cloud_verifier():
+    op.create_table('allowlists',
+                    sa.Column('id', sa.Integer(), nullable=False),
+                    sa.Column('name', sa.String(255), nullable=False),
+                    sa.Column('tpm_policy', sa.Text(), nullable=True),
+                    sa.Column('vtpm_policy', sa.Text(), nullable=True),
+                    sa.Column('allowlist', sa.Text(length=429400000), nullable=True),
+                    sa.PrimaryKeyConstraint('id'),
+                    sa.UniqueConstraint('name', name='uniq_allowlists0name'),
+                    mysql_ENGINE='InnoDB',
```
mpeters, line 44: Should we have these MySQL specific options in there? The other migrations we are using don't have them. I know InnoDB is needed for things like referential integrity, etc. What does Alembic do if they are left off?

keylime/migrations/versions/eb869a77abd1_create_allowlist_table.py
```diff
@@ -0,0 +1,50 @@
...
+                    mysql_DEFAULT_CHARSET='UTF8'
+                    )
+
+
+def downgrade_cloud_verifier():
+    pass
```
mpeters, line 50: Shouldn't the downgrade remove the `allowlists` table?

keylime/tenant.py
```diff
@@ -864,6 +864,65 @@ def do_verify(self):
                 continue
             break
 
+    def do_add_allowlist(self, name, tpm_policy=None, vtpm_policy=None,
+                         allowlist=None, exclude_list=None):
+
+        _tpm_policy = None
```
mpeters, line 870: I don't think we need to prefix these variables with an underscore. They aren't class variables and can't be accessed outside of the method anyway. And it's inconsistent that some variables have this convention while other internal vars (like `al_data`, `excl_data`, etc) don't.

mpeters (changes requested):
Other than some of the things I mentioned in my review, this overall looks like a useful change.

keylime/migrations/versions/eb869a77abd1_create_allowlist_table.py
kaifeng: Technically, yes, but in practice it's not used that much, we normally ignore downgrade implementation :)
Can do if desired.

keylime/migrations/versions/eb869a77abd1_create_allowlist_table.py
kaifeng: This argument will be consumed by dialect, should have no effect when working with backends other than mysql.
The behavior depends on the default of database backend if we don't provide it, so it's a safeguard to ensure things are expected.
But the arg name seems changed, will update.

keylime/migrations/versions/eb869a77abd1_create_allowlist_table.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def upgrade_cloud_verifier():
+    op.create_table('allowlists',
+                    sa.Column('id', sa.Integer(), nullable=False),
+                    sa.Column('name', sa.String(255), nullable=False),
+                    sa.Column('tpm_policy', sa.Text(), nullable=True),
+                    sa.Column('vtpm_policy', sa.Text(), nullable=True),
+                    sa.Column('allowlist', sa.Text(length=429400000), nullable=True),
```
kaifeng, line 41: I am wondering if we rename this field to ima_policy? It might be confusing to have an allowlist field in an allowlists table :)

keylime/tenant.py
```diff
@@ -936,11 +991,14 @@ def main(argv=sys.argv):
...
                         help='Block on cryptographically checked key derivation confirmation from the agent once it has been provisioned')
+    parser.add_argument('--allowlist-name', help='The name of allowlist to operate with')
 
     args = parser.parse_args(argv[1:])
     mytenant = Tenant()
 
-    if args.command not in ['list', 'regdelete', 'reglist', 'delete', 'status'] and args.agent_ip is None:
+    if args.command not in ['list', 'regdelete', 'delete', 'status',
+                            'addallowlist', 'deleteallowlist',
+                            'showallowlist'] and args.agent_ip is None:
```
kaifeng, line 1001: Subparsers can be used to provide better experience, we can do some refactoring work  when the whole feature is implemented.

lukehinds:
where is the enhancement for this?

keylime/migrations/versions/eb869a77abd1_create_allowlist_table.py
mpeters: You're right that it's not used much in practice (fail forward is a much better model than downgrading) but we do provide it for the other migrations if I'm not mistaken. So we should at least be consistent. I would be fine if we decided to skip downgrading all together and remove that from the other migrations as well. But that would be a separate issue.

So I'll let you pick which you prefer: add downgrade code to this migration or add a ticket to remove the downgrade from the other migrations.

keylime/migrations/versions/eb869a77abd1_create_allowlist_table.py
mpeters: I'm fine with it as allowlist. I see this field as the main content of the allowlist and the others as meta-data. In that kind of model, double naming is ok.

keylime/tenant.py
mpeters: yeah, that would be a good thing to refactor later.

kaifeng:
ok, will do.

kaifeng:
@lukehinds @mpeters the proposal is up https://github.com/keylime/enhancements/issues/38
Basically matching current implementation with a minor difference: the name of the ``verifiermain.allowlist`` will be renamed to ``allowlists.ima_policy``.

lukehinds:
> @lukehinds @mpeters the proposal is up [keylime/enhancements#38](https://github.com/keylime/enhancements/issues/38)
> Basically matching current implementation with a minor difference: the name of the `verifiermain.allowlist` will be renamed to `allowlists.ima_policy`.

awesome, thank you. I think its a good change. There is a pull request from @mpeters to implement JWT, with this change in we can decorate your new handles to provide some access control. An example here: https://github.com/keylime/keylime/blob/4117509c72acc25d174b49d8597de08f3cf91d45/keylime/cloud_verifier_tornado.py#L211-L212

@mpeters , I could not dig out your patch..did it get merged or you're working on locally?

mpeters:
> @mpeters , I could not dig out your patch..did it get merged or you're working on locally?

Wow, sorry I totally forgot about this. I broke your previous work up into 2 different branches and got the first one up to date and merged but the 2nd is sitting on private fork. I'll get that rebased and up to snuff this week so we can merge it.

keylime/migrations/versions/eb869a77abd1_create_allowlist_table.py
kaifeng: I am asking because ``allowlist`` actually represents IMA in our current implementation, ``ima_policy`` woud match that usage.
There could be other types of allowlist (Pre-Boot ML for example?) that we could not expect at this moment, picking suitable word may help avoid confusion in the future. Ideas?

maugustosilva: I vote to change the name, precisely to accommodate the case where we have an "allowlist" that would refer to the TPM2 Pre-Boot Event log. Let's make each column clearly refer to the mechanism (IMA or PBEL)

mpeters: After thinking about it, I agree with both you now. Let's rename to ima_policy

mpeters (changes requested):

keylime/migrations/versions/eb869a77abd1_create_allowlist_table.py
kaifeng: Will do, I am unable to reach GitHub for a while, hope things will get stabilized soon.

mpeters (approved):

PR #504, opened 2021-01-14
Move lint check to a job

kaifeng:
Move pylint check to a job so it doesn't block
functional tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,11 +36,12 @@ before_install:
   - sudo pip3 install pylint --upgrade
   - sudo pip3 install tornado-requests
   - sudo pip3 install -r requirements.txt
-  - make check
   - 'docker pull ${tpm12image}:${tpm12tag}'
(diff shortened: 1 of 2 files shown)


kaifeng:
pylint is raising issues that were exposed before, not sure what is happening because the version is still 2.6.0.

kaifeng:
> 
> 
> pylint is raising issues that were not exposed before, not sure what is happening because the version is still 2.6.0.

It was introduced from another patch that was passed checking before W0707 enabled.

mpeters:
> It was introduced from another patch that was passed checking before W0707 enabled.

Ah, so because there wasn't a conflict a previously passing PR merged that caused this new failure after it was rebased in GH? I guess that's an argument for not rebasing in GH but on the actual commit to retrigger CI... but it is very convenient.

mpeters (approved):
lgtm

Issue #498, opened 2021-01-13
Verifier db can't hold ipv6 address


kaifeng:
Current ip can only hold up to 15 characters:

```
ip = Column(String(15))
```

PR #505, opened 2021-01-14
Extend capacity of verifiermain.ip field

kaifeng:
Add migration script to extend the ip field capacity to hold up
to 255 characters.

Resolves: #498


mpeters (approved):
lgtm

PR #508, opened 2021-01-15
Fix response handling in tenant

kaifeng:
Response object from requests doesn't have status attribute,
some code path will hit this:

AttributeError: 'Response' object has no attribute 'status'

Response doesn't have read() method either, changed to use text
attribute.


kaifeng:
I'd say GHA is pretty fast, no queuing :)

mpeters (approved):
lgtm

PR #511, opened 2021-01-18
List agents from registrar with tenant cli

kaifeng:
Helpful command to list activated agents from registrar.
Would be more useful if ip address is returned as well, but not
available from the db yet.


mpeters (approved):
lgtm

PR #512, opened 2021-01-18
Remove revocation notifier's ZMQ frontend socket file before killing the process

axelsimon:
I had initially planned to also close the socket, but it appears that's not necessary, if my understanding of the PyZMQ documentation is correct.

So this patch simply removes the lingering frontend socket file before the revocation notifier process is killed.

Signed-off-by: axelsimon <github@axelsimon.net>


mpeters (approved):
lgtm

PR #513, opened 2021-01-19
Fix brached migration path

kaifeng:
There were two database schema changes merged but not dependent on each
other, so current migration path is branched.

Update dependency according to merge sequence.


kaifeng:
Combined into #499

PR #514, opened 2021-01-19
Revert "Simplify database configuration"

lukehinds:
Reverts keylime/keylime#427

This adds no value at all, its half written with TODOs in it and the previous code was fine. We were using sqlachemy generics to construct the connection url, now all we are doing is have the user hardcode it instead. It does not even update the user docs.

Let's please not deprecate working code unless there is a compelling reason to. I recommend we revert this and think it through and discuss it again properly , what merits such a change as this?  Saying 'simplify' is not saying much.

### DIFF
--- a/keylime.conf
+++ b/keylime.conf
@@ -198,16 +198,7 @@ registrar_private_key_pw = default
 # Database URL Configuration
 # See this document https://keylime-docs.readthedocs.io/en/latest/installation.html#database-support
 # for instructions on using different database configurations.
-# Database configuration via drivername, username, password, etc is deprecated
-# and superseded by database_url, when database_url is set, other database
-# fields will be ignored.
-# An example of database_url value for using sqlite:
-#   sqlite:////var/lib/keylime/cv_data.sqlite
-# An example of database_url value for using mysql:
-#   mysql+pymysql://keylime:keylime@keylime_db:[port]/verifier?charset=utf8
-# If database_url is not set, the deprecated default is "sqlite" and will be
-# located at "/var/lib/keylime/cv_data.sqlite".
-database_url =
+# The default is "sqlite" and will be located at "/var/lib/keylime/cv_data.sqlite".
 drivername = sqlite
 username = ''
 password = ''
@@ -215,7 +206,6 @@ host = ''
 port = ''
 database = cv_data.sqlite
 query = ''
-
 auto_migrate_db = True
 
 
@@ -476,16 +466,7 @@ check_client_cert = True
 # Database URL Configuration
 # See this document https://keylime-docs.readthedocs.io/en/latest/installation.html#database-support
 # for instructions on using different database configurations.
-# Database configuration via drivername, username, password, etc is deprecated
-# and superseded by database_url, when database_url is set, other database
-# fields will be ignored.
-# An example of database_url value for using sqlite:
-#   sqlite:////var/lib/keylime/reg_data.sqlite
-# An example of database_url value for using mysql:
-#   mysql+pymysql://keylime:keylime@keylime_db:[port]/registrar?charset=utf8
-# If database_url is not set, the deprecated default is "sqlite" and will be
-# located at "/var/lib/keylime/reg_data.sqlite".
-database_url =
+# The default is "sqlite" and will be located at "/var/lib/keylime/reg_data.sqlite".
 drivername = sqlite
 username = ''
 password = ''
@@ -493,9 +474,7 @@ host = ''
 port = ''
 database = reg_data.sqlite
 query = ''
-
 auto_migrate_db = True
-
 
 # The file to use for SQLite persistence of provider hypervisor data.
 prov_db_filename = provider_reg_data.sqlite
--- a/keylime/db/keylime_db.py
+++ b/keylime/db/keylime_db.py
@@ -12,8 +12,6 @@ from sqlalchemy.engine.url import URL
 from keylime import config
 from keylime import keylime_logging
 
-logger = keylime_logging.init_logging('keylime_db')
-
 
 class DBEngineManager:
 
@@ -26,16 +24,8 @@ class DBEngineManager:
         """
         self.service = service
 
-        database_url = config.get(service, 'database_url')
-        if database_url:
-            engine = create_engine(database_url)
-            return engine
-
-        # TODO(kaifeng) Remove following code as well as related configuration
-        # options when the deprecation period is reached.
-        logger.warning('database_url is not set, using deprecated database '
-                       'configuration options')
         drivername = config.get(service, 'drivername')
+
         if drivername == 'sqlite':
             database = "%s/%s" % (config.WORK_DIR,
                                   config.get(service, 'database'))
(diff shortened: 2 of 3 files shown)


lukehinds:
It was also discussed here before: https://github.com/keylime/keylime/pull/323

kaifeng:
Well, I think leaving TODO is just a way to keeping things unforgotten, it doesn't mean the code is half written. One TODO is for the deprecation, the other is I believe not required but didn't get back to it. If the deprecation is not desired, we can stop the deprecation, support both of them is not a big deal.

From the point of a deployer, I would like to have less configuration variables, determined results, not the various combinations of options. As to the database url, I don't think there is any difference between the unified URL and the construction of several options, both of them are hardcoded and determined during deployment.

Probably it's easy to say no value because of different use cases, my use case needs URL form because it's encrypted as a whole.

lukehinds:
@kaifeng the previous code was  inline with sqlachemy's API: https://docs.sqlalchemy.org/en/14/core/engines.html#sqlalchemy.engine.URL

I hear your encrypted case, that is valid, I am willing to look at that - but removing the old values when we might be able to address your specific case is not an optimal approach.

maugustosilva:
@lukehinds Allow me to voice my support for @kaifeng main point: from a devops standpoint, less configuration variables is, I would argue, a better approach. In addition to being less error prone, some options where named in a confusing way, such as `host`, to indicate the database's IP address. Plus, the encrypted case is highly relevant to any production, real-world deployment. Given that there are people/companies gearing up to start using Keylime at scale, I offer the argument that it is better to go through the pain of switching this particular configuration option sooner rather than later.

lukehinds:
do we know for sure encrypted connections are not possible? I am no convinced about the devops argument, swapping out key / values in an ini is not a challenge.

kaifeng:
> 
> 
> do we know for sure encrypted connections are not possible? I am no convinced about the devops argument, swapping out key / values in an ini is not a challenge.

It is certain that sqlalchemy does not support encrypted URLs, we have consistent handling of encrypted configration so the plain text only exists in memory while keeping encrypted text in the configuration file.

lukehinds:
> > do we know for sure encrypted connections are not possible? I am no convinced about the devops argument, swapping out key / values in an ini is not a challenge.
> 
> It is certain that sqlalchemy does not support encrypted URLs, we have consistent handling of encrypted configration so the plain text only exists in memory while keeping encrypted text in the configuration file.

From a quick look at the docs you can use `query` which we have exposed in keylime.conf

`query={"ssl_ca": "your_cert.crt.pem"},`

kaifeng:
> 
> 
> > > do we know for sure encrypted connections are not possible? I am no convinced about the devops argument, swapping out key / values in an ini is not a challenge.
> > 
> > 
> > It is certain that sqlalchemy does not support encrypted URLs, we have consistent handling of encrypted configration so the plain text only exists in memory while keeping encrypted text in the configuration file.
> 
> From a quick look at the docs you can use `query` which we have exposed in keylime.conf
> 
> `query={"ssl_ca": "your_cert.crt.pem"},`

I am not saying URL can achieve something that current options can't provide, they are identical and sqlalchemy doc is using both interchangeably, there are many examples https://docs.sqlalchemy.org/en/14/core/engines.html#database-urls
What I can't understand is why you prefer constructing URL and exclude the possibility to use user provided URL. By default there is path handling with regard to the root/non-root use case, but we all agree verifier/registrar should run as non-root, this removes the variable path for sqlite3 database, so in the end either filling parts of an URL or provide one URL has no difference.
This is not my specific use case, I know there are many people would prefer to use one URL.

lukehinds:
> What I can't understand is why you prefer constructing URL and exclude the possibility to use user provided URL

You're missing the point. It is not about personal preference, its about not deprecating code without a substantial reason and so far no one has provided me with one yet. This is not an issue so much with the method itself, as you just outlined both are acceptable (which makes my point even more weighted), its about not disrupting users with depreciation of code that works perfectly ok, unless you have a clear reason why the change would be substantially better to merit a depreciation. So far all you have provided me with is its simpler (which I disagree with) and encryption and we have already seen encryption is supported as I highlight above. This is about stablity and being non disruptive to users. If we are going to deprecate code , it should be for a measurement reason.

lukehinds:
adding maintainers to this topic:

@lukehinds
@jetwhiz
@nabilschear
@amylily1011
@maugustosilva
@font
@axelsimon
@ashcrow
@kaifeng

lukehinds:
one way forwards is to implement a `database_raw_url`, but it should not be the default config. It would be an empty and a conditional is put into the code so that is someone populates an hard url the sql engine users that. This way you can use a hardcoded url and we don't force users to change from their current set up.

kaifeng:
> 
> 
> one way forwards is to implement a `database_raw_url`, but it should not be the default config. It would be an empty and a conditional is put into the code so that is someone populates a hard url the sql engine users that. This way you can use a full url and we don't force users to change from their current set up.

I get your point, in fact the merged change is upgrades/backwards compatible. If the consensus is to not deprecated old options,  removing the deprecation comment is enough.
I know the importance of stability, while from my experience as a stacker, it can't always be guaranteed during the project evolvement. it's better to be practical than ideal, if some breaking change should happen, I believe it's better to do it early. It would be hard to maintain for a widely used project while keeping everything compatible, as I can see there are still many gaps for real production use.

maugustosilva:
I vote to keep both configuration options available then, with a simple `if` which will detect if the `database_raw_url` is present on the `keylime.conf`, and switch to the processing of the original database configuration method if not. 

The only extra concession that I ask for is to please change the names of the attributes on the original configuration method to make these less confusing to both configuration parsers and human users reading it: basically lets add the prefix `database_` to `username`, `password`, `host`, `port` and `database` (this last one, we can change from `database` to `database_name`)

lukehinds:
> I vote to keep both configuration options available then, with a simple `if` which will detect if the `database_raw_url` is present on the `keylime.conf`, and switch to the processing of the original database configuration method if not.
> 
> The only extra concession that I ask for is to please change the names of the attributes on the original configuration method to make these less confusing to both configuration parsers and human users reading it: basically lets add the prefix `database_` to `username`, `password`, `host`, `port` and `database` (this last one, we can change from `database` to `database_name`)

I 2nd this approach.

axelsimon:
> I vote to keep both configuration options available then, with a simple `if` which will detect if the `database_raw_url` is present on the `keylime.conf`, and switch to the processing of the original database configuration method if not.
> 
> The only extra concession that I ask for is to please change the names of the attributes on the original configuration method to make these less confusing to both configuration parsers and human users reading it: basically lets add the prefix `database_` to `username`, `password`, `host`, `port` and `database` (this last one, we can change from `database` to `database_name`)

I agree that this all seems like a good way forward. I'm also very much in favour of naming configuration options in a way that "as descriptive as possible" (within reason).

mpeters:
I think the consensus is the following:

- Drop the deprecation part of the origin PR. We'll allow both configuration methods to live side-by-side
- Rename the database configuration directives to start with something like `db_` or `database_` prefixes

If we're in agreement I think that means we can close this PR and if so, I'll open a ticket for those 2 issues.

kaifeng:
The proposal looks good to me, to be specific, let's use the ``database_`` prefix for old options and ``database_raw_url`` for the URL form as @lukehinds proposed.

maugustosilva:
@mpeters @lukehinds @kaifeng with PR #654 merged, everything discussed here (and more) has been implemented. I believe we can close this PR.

PR #517, opened 2021-01-21
First PR (of a total of 3) to address #491

maugustosilva:
* The `agent` will now retrieve TPM 2.0 preboot event log and will ship
it to the `verifier`. The latter will do nothing with it (just receives
and ignores it)
* Adds a new `keylime_tenant` CLI option, `--mb_refstate` (that was a
request from @mpeters) to signal the need for the agent to collect the
the event log. At this point, `keylime_tenant` do not actually process
this option (it shall be processed in a manner similar to IMA in a
future PR), it is simply used to signal the agent that "tpm2 preboot
event log" should be read.
* Hard-coded PCRs [0-9], 14 and path
/sys/kernel/security/tpm0/binary_bios_measurements for the event log
* The `verifier` will output a debug logging message stating that it
received the "tpm2 event log", and will send it down to the `check_pcrs`
method, but there is no actual processing there (it will be done in
subsequent PR)

### DIFF
--- a/keylime/cloud_verifier_common.py
+++ b/keylime/cloud_verifier_common.py
@@ -214,7 +214,6 @@ def process_quote_response(agent, json_response):
                                      hash_alg,
                                      ima_keyring,
                                      mb_measurement_list,
-                                     #agent['mb_intended_state'],
                                      {})
     if not validQuote:
         return False
--- a/keylime/cloud_verifier_common.py
+++ b/keylime/cloud_verifier_common.py
@@ -133,12 +133,15 @@ def process_quote_response(agent, json_response):
         quote = json_response["quote"]
 
         ima_measurement_list = json_response.get("ima_measurement_list", None)
+        mb_measurement_list = json_response.get("mb_measurement_list", None)
 
         logger.debug("received quote:      %s" % quote)
         logger.debug("for nonce:           %s" % agent['nonce'])
         logger.debug("received public key: %s" % received_public_key)
         logger.debug("received ima_measurement_list    %s" %
                      (ima_measurement_list is not None))
+        logger.debug("received boot log    %s" %
+                     (mb_measurement_list is not None))
     except Exception:
         return None
 
@@ -209,7 +212,10 @@ def process_quote_response(agent, json_response):
                                      ima_measurement_list,
                                      agent['allowlist'],
                                      hash_alg,
-                                     ima_keyring)
+                                     ima_keyring,
+                                     mb_measurement_list,
+                                     #agent['mb_intended_state'],
+                                     {})
     if not validQuote:
         return False
 
--- a/keylime/config.py
+++ b/keylime/config.py
@@ -295,6 +295,10 @@ else:
 
 IMA_PCR = 10
 
+# measured boot addons
+MEASUREDBOOT_PCRS = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 14]
+MEASUREDBOOT_ML = '/sys/kernel/security/tpm0/binary_bios_measurements'
+
 # this is where data will be bound to a quote, MUST BE RESETABLE!
 TPM_DATA_PCR = 16
 
--- a/keylime/tenant.py
+++ b/keylime/tenant.py
@@ -201,6 +201,16 @@ class Tenant():
         al_data = None
         if "allowlist" in args and args["allowlist"] is not None:
 
+            _policy_pcrs = list(self.tpm_policy.keys())
+            _policy_pcrs.remove('mask')
+
+            _protected_pcrs = [ config.IMA_PCR ]
+
+            for _pcr in _policy_pcrs :
+                if int(_pcr) in _protected_pcrs :
+                    logger.error(f"WARNING: PCR {_pcr} is specified in \"tpm_policy\", but will in fact be used by IMA. Please remove it from policy")
+                    sys.exit()
+
             # Auto-enable IMA (or-bit mask)
             self.tpm_policy['mask'] = "0x%X" % (
                 int(self.tpm_policy['mask'], 0) | (1 << config.IMA_PCR))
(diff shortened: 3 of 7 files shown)


maugustosilva:
Before @lukehinds even asks :-), here is the link to the enhancement: https://github.com/keylime/enhancements/issues/40

maugustosilva:
Complete doc for enhancement submitted as a [PR](https://github.com/keylime/enhancements/pull/41) on the appropriate repo.

keylime/cloud_verifier_common.py
```diff
@@ -209,7 +212,10 @@ def process_quote_response(agent, json_response):
                                      ima_measurement_list,
                                      agent['allowlist'],
                                      hash_alg,
-                                     ima_keyring)
+                                     ima_keyring,
+                                     mb_measurement_list,
+                                     #agent['mb_intended_state'],
```
mpeters, line 217: Just to not confuse this PR, can we remove this commented out code?

keylime/tenant.py
```diff
@@ -201,6 +201,16 @@ def init_add(self, args):
         al_data = None
         if "allowlist" in args and args["allowlist"] is not None:
 
+            _policy_pcrs = list(self.tpm_policy.keys())
```
mpeters, line 204: We don't need to prefix method/function variables with underscore. That convention in python is just for package level things so they aren't exported.

keylime/tenant.py
```diff
@@ -235,6 +245,34 @@ def init_add(self, args):
             # Process allowlists
             self.allowlist = ima.process_allowlists(al_data, excl_data)
 
+        # Read command-line path string TPM2 event log template
+        if "mb_refstate" in args and args["mb_refstate"] is not None:
+            _policy_pcrs = list(self.tpm_policy.keys())
+            _policy_pcrs.remove('mask')
+            _protected_pcrs = config.MEASUREDBOOT_PCRS
+
+            for _pcr in _policy_pcrs :
```
mpeters, line 254: We're doing this check twice (line 209 above). I know it's in different methods, but just wanted to check that we need to have it in both places. And if we do, maybe it should be in it's own helper method?

keylime/tenant.py
```diff
@@ -235,6 +245,34 @@ def init_add(self, args):
...
+                self.tpm_policy['mask'] = "0x%X" % (
+                    int(self.tpm_policy['mask'], 0) | (1 << _pcr))
+
+            logger.info(f"TPM PCR Mask automatically modified is {self.tpm_policy['mask']} to include IMA/Event log PCRs")
+
+            if isinstance(args["mb_refstate"], str):
+                if args["mb_refstate"] == "default":
+                    args["mb_refstate"] = config.get('tenant', 'mb_refstate')
+                al_data = 'test'
+#                al_data = mb.read_allowlist(args["allowlist"]) <------ read the actual intended state
```
mpeters, line 270: Let's remove this commented code in this first PR

mpeters (changes requested):
Some relatively small changes requested

mpeters (approved):
This LGTM. I'll hold off on merging until @lukehinds has a chance to review early next week

keylime/cloud_verifier_common.py
```diff
@@ -209,7 +212,9 @@ def process_quote_response(agent, json_response):
                                      ima_measurement_list,
                                      agent['allowlist'],
                                      hash_alg,
-                                     ima_keyring)
+                                     ima_keyring,
+                                     mb_measurement_list,
+                                     {})
```
lukehinds, line 217: Just curious what the empty dict is for?

keylime/config.py
```diff
@@ -295,6 +295,10 @@ def valid_regex(regex):
 
 IMA_PCR = 10
 
+# measured boot addons
+MEASUREDBOOT_PCRS = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 14]
```
lukehinds, line 299: not a blocker by any means, but I wonder if this might be useful exposed in keylime.conf. I guess not and you would know better here.

keylime/tenant.py
```diff
@@ -368,6 +392,15 @@ def init_add(self, args):
             raise UserError("Payload size %s exceeds max size %d" % (
                 len(self.payload), config.getint('tenant', 'max_payload_size')))
 
+    def enforce_pcrs(self, policy_pcrs, protected_pcrs, pcr_use) :
+        policy_pcrs = list(self.tpm_policy.keys())
+        policy_pcrs.remove('mask')
+
+        for _pcr in policy_pcrs :
+            if int(_pcr) in protected_pcrs :
+                logger.error(f"WARNING: PCR {_pcr} is specified in \"tpm_policy\", but will in fact be used by {pcr_use}. Please remove it from policy")
```
lukehinds, line 401: Just to clarify my understanding, does mean users can cherry pick PCRs to either tpm_policy or measure boot (mb)?

keylime/tenant.py
```diff
@@ -368,6 +392,15 @@ def init_add(self, args):
...
+                sys.exit()
```
lukehinds, line 402: I would change this to `sys.exit(1)` as it defaults to `0` which is success.  This would be useful then for CI systems etc that might wrap around the tenant CLI.

https://docs.python.org/3/library/sys.html#sys.exit

keylime/tpm/tpm1.py
```diff
@@ -739,8 +739,8 @@ def check_deep_quote(self, agent_id, nonce, data, quote, vAIK, hAIK, vtpm_policy
                 pcrs.append(line)
 
         # don't pass in data to check pcrs for physical quote
-        return self.check_pcrs(agent_id, tpm_policy, pcrs, None, False, None, None, None) and \
-            self.check_pcrs(agent_id, vtpm_policy, vpcrs, data, True, ima_measurement_list, allowlist, None)
+        return self.check_pcrs(agent_id, tpm_policy, pcrs, None, False, None, None, None, None, {}) and \
```
lukehinds, line 742: be good to understand the empty dicts again

keylime/tpm/tpm_abstract.py
```diff
@@ -230,13 +230,16 @@ def __check_ima(self, agent_id, pcrval, ima_measurement_list, allowlist, ima_key
...
-    def check_pcrs(self, agent_id, tpm_policy, pcrs, data, virtual, ima_measurement_list, allowlist, ima_keyring):
+    def check_pcrs(self, agent_id, tpm_policy, pcrs, data, virtual, ima_measurement_list, allowlist, ima_keyring, mb_measurement_list, mb_intended_state):
         try:
             tpm_policy_ = ast.literal_eval(tpm_policy)
         except ValueError:
             tpm_policy_ = {}
         pcr_allowlist = tpm_policy_.copy()
 
+        if mb_measurement_list or mb_intended_state :
+            logger.info("Measured boot information received, but for now it will not be processed")
```
lukehinds, line 241: This could perhaps steer the user more towards the reason why. If I saw this message I would think 'ok, when will it be processed and why?'.

lukehinds (changes requested):
I think the patch is in good order, just some comments and one change request.

keylime/cloud_verifier_common.py
maugustosilva: in the subsequent PRs, it will be changed to the "measured boot reference state" supplied by the `tenant` CLI

keylime/config.py
maugustosilva: I thought about it, and decided to keep all values "hard-coded" mostly because - very much like IMA - "measured boot" refers to a very specific set of PCRs. Granted, there is some latitude on PCR 14 (which is updated only when secure booting), but I plan to revisit it after the basic set (of 3) PRs gets merged.

keylime/tenant.py
maugustosilva: Correct

keylime/tenant.py
maugustosilva: good catch! I have done this now, and later will I make a PR to have non-zero exit code for other errors (unrelated to this PR)

keylime/tpm/tpm1.py
maugustosilva: in the subsequent PRs, it will be changed to the "measured boot reference state" supplied by the tenant CLI

keylime/tpm/tpm_abstract.py
galmasi: the message will disappear when PR2 shows up. Probably should have marked this as "temporary"

lukehinds (approved):

lukehinds:
you should have merge rights @maugustosilva , if not ping me or @mpeters

PR #520, opened 2021-01-23
Fix evictcontrol for tpm2 tools 3.2

kaifeng:
It appears that for tpm2-tools 3.2 the tpm2_evictcontrol accepts
handle from the -H argument instead of -c.

  # tpm2_evictcontrol -A o -c 0x81010007 -P keylime
  ERROR: Invalid arguments
  ERROR: Unable to run tpm2_evictcontrol

Suddenly hits this issue in a working host.

### DIFF
--- a/keylime/tpm/tpm2.py
+++ b/keylime/tpm/tpm2.py
@@ -414,7 +414,7 @@ class tpm2(tpm_abstract.AbstractTPM):
             outjson = config.yaml_to_dict(output)
             if outjson is not None and current_handle in outjson:
                 if self.tools_version == "3.2":
-                    cmd = ["tpm2_evictcontrol", "-A", "o", "-c",
+                    cmd = ["tpm2_evictcontrol", "-A", "o", "-H",
                            hex(current_handle), "-P", owner_pw]
                     retDict = self.__run(cmd, raiseOnError=False)
                 else:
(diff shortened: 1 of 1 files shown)


keylime/tpm/tpm2.py
```diff
@@ -414,7 +414,7 @@ def __create_ek(self, asym_alg=None):
             outjson = config.yaml_to_dict(output)
             if outjson is not None and current_handle in outjson:
```
kaifeng, line 415: I am also hitting an issue here, I suspect this is never executed otherwise we would find out.

```
  File "/usr/bin/keylime_agent", line 11, in <module>
  File "/usr/lib/python3.6/site-packages/keylime/cmd/agent.py", line 15, in main
    keylime_agent.main()
  File "/usr/lib/python3.6/site-packages/keylime/keylime_agent.py", line 492, in main
    'cloud_agent', 'tpm_ownerpassword'))  # this tells initialize not to self activate the AIK
  File "/usr/lib/python3.6/site-packages/keylime/tpm/tpm2.py", line 935, in tpm_init
    self.__create_ek()
  File "/usr/lib/python3.6/site-packages/keylime/tpm/tpm2.py", line 416, in __create_ek
    if outjson is not None and current_handle in outjson:
TypeError: 'in <string>' requires string as left operand, not int
```

The handle is save as integer in the tpmdata.yml, but the persistent handles get from tpm2 tools are string, we probably should hex() it for the ``in`` test.

kaifeng:
I believe the argument should be correct for 3.2 version, because I am experiencing a few similar issues with the one in the comment, each time I manually purge the persistent handles the old command line complains the usage is wrong.
There should be two issue still exist, one is the output parsing and the other is mismatching states between tpmdata.yml and TPM (which is caused by some exception path), but I'd like to address it after I gather more information.

mpeters (approved):
lgtm

Issue #521, opened 2021-01-25
ans1 package needed in CI and installation scripts


lukehinds:
The [IMA file signing patch](https://github.com/keylime/keylime/pull/496) introduced a new dependency asn1, but this was not added to the docker images for CI and the installer.sh (and ansible role), it is currently only in the requirements.txt

Luckily this is packaged in Fedora as python3-pyasn1

This issue means the changes should be made. I also think we should think of a way handling additional dependencies in a way that we can make sure they are supported and packaged correctly. I expect we should do something in enhancements.

lukehinds:
another example of why we need this: https://github.com/keylime/keylime/issues/519#issuecomment-766951400

stefanberger:
I just sent PR #524 that gets rid of this package and rather unused DSA keys.

stefanberger:
Some householding: Is this still an issue or can we close it?

PR #522, opened 2021-01-25
Add comments that changes first require an enhancement

lukehinds:
This change relates to issue #521

Signed-off-by: Luke Hinds lhinds@redhat.com


mpeters (approved):
This is probably a good baseline for whether an enhancement is necessary. Dependencies increase complexity and add some burden to keep up with in the future, so seems like a good threshold to me.

mpeters:
@luke, can you amend your commit with being signed off?

lukehinds:
> @luke, can you amend your commit with being signed off?

I am not sure I can, I did it on the UI. I might be able to pull it down and amend (if git does not tie me up in knots).

PR #523, opened 2021-01-25
Fix for ans1 unpackaged

lukehinds:
Resolves: #521

I am sure will need work still, but want to try this in CI and get some discussion going as this is a high priority to fix.

Note I remove, deep quote - no one is using this at all (I am certain), so I am happy to follow up and remove the rest.

### DIFF
--- a/keylime.conf
+++ b/keylime.conf
@@ -19,6 +19,9 @@ ca_implementation = openssl
 receive_revocation_ip = 127.0.0.1
 receive_revocation_port = 8992
 
+# Enable IMA file signing
+ima_file_signing = False
+
 #=============================================================================
 [cloud_agent]
 #=============================================================================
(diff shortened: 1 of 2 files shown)


lukehinds:
@stefanberger

dnoliver:
I am not 100% aware of what `ima_file_signing` does. Is that intended to check the IMA Signature of the measured files?

If that is the case, I have an objection for removing it :)

Starting with Fedora 34, the RPM contents will be signed for IMA. 

https://fedoraproject.org/wiki/Changes/Signed_RPM_Contents

Are you removing the capability to remotely verify those signatures?

stefanberger:
I think the easiest would be to move the import of asn1 to the part that actually needs. No more DSA keys.

lukehinds:
hold off on this one for https://github.com/keylime/keylime/pull/524 instead.

mpeters:
Are we still going to go forward with this? It still has the feature of turning IMA signature support completely off, but is that desired?

lukehinds:
I am happy to close this now.

PR #525, opened 2021-01-25
Address some more IMA file signature verification issues

stefanberger:
This PR addresses some more IMA file signature verification 'teething' issues:

- It reduces the logging output on file signatures evaluation. Particularly a log entry indicating a good signature is only shown on logger.debug level
- We now account for bad file signatures in a separate error field and report it
- If an allow list entry is not available then an error is only reported if its signature has not already been evaluated; this allows us to only sign a file and not have an allow list entry on top of it


mpeters (approved):
lgtm

PR #480, opened 2020-12-23
Move agents handling out of tornado request handler

kaifeng:
It's not necessarily tight with request handler, by moving
it to the outter scope, it is possible to re-trigger agent
processing during verifier startup.

Resolves: #393

PR #527, opened 2021-01-26
Additional approved states for auto-reactivate agent upon verifier restart 

apoorvemohan:
Currently, only the agents left in  "GET_QUOTE" state (in the verifier database) can be reactivated upon a verifier restart. 
This patch extends this functionality (added in PR #480) with more (presumably safer) states.


lukehinds:
@apoorvemohan thanks, you need to pass --signoff to your commit. You might be able to do a git commit --amend --signoff

I think this needs a maintainer to peform some offline tests first before we merge this as its not captured in CI at present. I am going to mark as needs changes, but that is no reflection of the code quality, its just as a means to insure its checked out with a local test first.

lukehinds (changes requested):
adding request changes until validated by a maintainer.

mpeters:
> I think this needs a maintainer to peform some offline tests first before we merge this as its not captured in CI at present.

Would it be possible to write tests for this? In general I would like to move things to requiring tests before being merged for both new features and bug fixes. We've had some instability lately because CI didn't have a test for something that should probably be testable.

apoorvemohan:
> @apoorvemohan thanks, you need to pass --signoff to your commit. You might be able to do a git commit --amend --signoff
@lukehinds done

kaifeng:
The change looks good, we do support have thoses state transitions, so it should work. Some states can not be easily constructed, so I think unit test coverage should help, but we have none at the moment :(

mpeters (approved):
lgtm

PR #534, opened 2021-02-03
Second PR (of a total of 3) to address #491

maugustosilva:
* The most important change here is the fact that the TPM 2.0 preboot
  event log , shipped to the `verifier` from the `agent`, will now be
authenticated, by replaying the aforementioned log and comparing it
against the values of PCRs [0-9],14.
* If these do not match, then the node will be considered as failed at
  attestation.
* A small utility (enrich.py) is added to perform the log replay. In the
  future, it could be replaced by newer versions of tpm2-tools capable
of performing the same activity (several open PRs against the tool with
fixes and improvements)


keylime/enrich.py
```diff
@@ -0,0 +1,289 @@
...
+#
+#  typedef struct _EFI_SIGNATURE_DATA {
+#      uint8_t SignatureOwner[16];
+#      uint8_t SignatureData[];
+#  } EFI_SIGNATURE_DATA;
+#
+##################################################################################
+
+##################################################################################
+# Print out EFI_SIGNATURE_DATA
```
mpeters, line 103: This method doesn't print anything, it returns a structure of the owner and data fields

keylime/enrich.py
```diff
@@ -0,0 +1,289 @@
...
+#       UINT16 FilePathListLength;
+#       // CHAR16 Description[];
+#       // EFI_DEVICE_PATH_PROTOCOL FilePathList[];
+#       // UINT8 OptionalData[];} EFI_LOAD_OPTION;
+#   } EFI_LOAD_OPTION;
+#
+##################################################################################
+
+##################################################################################
+# Print additional information about variables BootOrder and Boot####
```
mpeters, line 135: Again, misleading comment since this it's not printing anything

keylime/tpm/tpm_abstract.py
```diff
@@ -230,21 +230,37 @@ def __check_ima(self, agent_id, pcrval, ima_measurement_list, allowlist, ima_key
...
-        if mb_measurement_list or mb_intended_state :
+        if mb_measurement_list or mb_refstate :
             logger.info("Measured boot information received, but for now it will not be processed. A future update will enable the full processing of it.")
 
         if 'mask' in pcr_allowlist:
             del pcr_allowlist['mask']
         # convert all pcr num keys to integers
         pcr_allowlist = {int(k): v for k, v in list(pcr_allowlist.items())}
 
+        if mb_refstate not in (False, None) and mb_measurement_list not in (False, None) :
```
mpeters, line 248: You shouldn't need the False or None check, just checking it's "truthiness" should work:

`if mb_refstate and mb_measurement_list`

keylime/tpm/tpm_abstract.py
```diff
@@ -230,21 +230,37 @@ def __check_ima(self, agent_id, pcrval, ima_measurement_list, allowlist, ima_key
...
+            mb_measurement_data = self.parse_bootlog(mb_measurement_list)
+            if not isinstance(mb_measurement_data, dict):
```
mpeters, line 250: What scenarios exist where this variable wouldn't be a dict?

keylime/tpm/tpm_abstract.py
```diff
@@ -281,15 +297,34 @@ def check_pcrs(self, agent_id, tpm_policy, pcrs, data, virtual, ima_measurement_
 
                 return False
 
-            # check whether this is a MB PCR -- do *not* compare measured boot PCRs against a reference state, if one exists
-            if pcrnum in config.MEASUREDBOOT_PCRS and mb_intended_state :
+            # PCRs set by measured boot get their own special handling
+
+            if pcrnum in config.MEASUREDBOOT_PCRS :
+
+                if mb_refstate not in (False, None):
```
mpeters, line 304: Similar to a prev comment, you shouldn't need the `not in (False, None)` check, just `if mb_refstate`

keylime/tpm/tpm_abstract.py
```diff
@@ -281,15 +297,34 @@ def check_pcrs(self, agent_id, tpm_policy, pcrs, data, virtual, ima_measurement_
...
+
+                    if mb_measurement_list in (False, None) :
```
mpeters, line 306: Same as above

keylime/tpm/tpm_abstract.py
```diff
@@ -307,6 +342,13 @@ def check_pcrs(self, agent_id, tpm_policy, pcrs, data, virtual, ima_measurement_
         if len(missing) > 0:
             logger.error("%sPCRs specified in policy not in quote: %s" % (("", "v")[virtual], missing))
             return False
+
+        if mb_refstate not in (False, None) and mb_measurement_list not in (False, None):
```
mpeters, line 346: Same comment about boolean tests

keylime/tpm/tpm_abstract.py
```diff
@@ -307,6 +342,13 @@ def check_pcrs(self, agent_id, tpm_policy, pcrs, data, virtual, ima_measurement_
...
+            missing = list(set(config.MEASUREDBOOT_PCRS).difference(pcrsInQuote))
+            if len(missing) > 0:
```
mpeters, line 348: I assume this is what we were talking about by making this more tolerant of missing PCRs, correct?

mpeters (changes requested):

keylime/tpm/tpm_abstract.py
maugustosilva: The "tolerance" was for the opposite case: a PCR (with value zero) was part of the quote sent by the agent, but the parsing of the boot produce a `None` value for that particular PCR (since it was not extended at any moment). On this part of the code we actually fail if the PCR # was not included in the quote.

keylime/enrich.py
```diff
@@ -0,0 +1,289 @@
+#!/usr/bin/python3
```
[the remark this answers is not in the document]
maugustosilva: fixed

keylime/enrich.py
```diff
@@ -0,0 +1,289 @@
...
+
+
+def representer(_, data):
+    return yaml.ScalarNode('tag:yaml.org,2002:int', hex(data))
+
+
+yaml.add_representer(hexint, representer)
+
+##################################################################################
+#
```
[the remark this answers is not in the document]
maugustosilva: fixed

keylime/enrich.py
maugustosilva: fixed

keylime/enrich.py
maugustosilva: fixed

keylime/tpm/tpm_abstract.py
maugustosilva: there was reason for this (we discussed the possibility of admitting an empty dictionary as a reference state), but decided against if for this PR

keylime/tpm/tpm_abstract.py
maugustosilva: Ah, this is a bit of defensive programming: intel's tpm2-tools (still used, even without all the required features) might output badly formed data (with an exit code 0 mind you) which will be misinterpreted. We are checking if it is indeed a dict in order to detect failure.

keylime/tpm/tpm_abstract.py
maugustosilva: there was reason for this (we discussed the possibility of admitting an empty dictionary as a reference state), but decided against if for this PR

keylime/tpm/tpm_abstract.py
maugustosilva: there was reason for this (we discussed the possibility of admitting an empty dictionary as a reference state), but decided against if for this PR

keylime/tpm/tpm_abstract.py
maugustosilva: there was reason for this (we discussed the possibility of admitting an empty dictionary as a reference state), but decided against if for this PR

keylime/tpm/tpm_abstract.py
mpeters: That makes sense for the things inside of `mb_measurement_data` (like the 'pcrs' key used later). But `mb_measurement_data` is created by `parse_bootlog` which eventually calls `config.yaml_to_dict` and there can't be a case where `yaml_to_dict` where it returns a non-None value that isn't a dict, right? Also, if we need to do defensive programming on what `yaml_to_dict` could return that isn't a dict it should probably be done where `yaml_to_dict` is itself called.

keylime/tpm/tpm_abstract.py
mpeters: gotcha, thanks

mpeters (changes requested):
One more minor change from my side

keylime/tpm/tpm_abstract.py
galmasi: Yes, this particular test is an overreach. By way of excuses, I have just this: we have discussed ad nauseam how to *not trust* anything coming from the agent. Eventually the whole parsing process will be done in tpm2_eventlog and the code will slim down a lot. 

Hai Huang is spending a lot of time with the Intel guys doing the first line of defense in tpm2_eventlog: if a malicious agent submits a malformed quote tpm2_eventlog could crash or output garbage.

keylime/tpm/tpm_abstract.py
mpeters: Even if you pass an empty dictionary it would still work as just `if mb_refstate`. False, None, an empty list or empty dict all evaluate to `false` and True, any value, any non-empty list/dict would be `True`.

keylime/tpm/tpm2.py
```diff
@@ -1261,3 +1262,50 @@ def read_key_nvram(self):
...
+        '''Parse and enrich a BIOS boot log
+
+        The input is the binary log.
+        The output is the result of parsing and applying other conveniences.'''
+        with tempfile.NamedTemporaryFile() as log_bin_file:
+            log_bin_file.write(log_bin)
+            log_bin_filename = log_bin_file.name
+            retDict = self.__run(['tpm2_eventlog', log_bin_filename])
+        log_parsed_strs = retDict['retout']
+        log_parsed_data = config.yaml_to_dict(log_parsed_strs, add_newlines=False)
```
mpeters, line 1300: Your other defensive coding got me thinking... if the log parsed strs coming from the agent are corrupted in some way what happens here? Does it crash the verifier process with a stack trace about yaml_to_dict problems? Even if it doesn't crash it wouldn't necessarily be a good error message. Not necessarily holding up this PR, but something to think about improving

mpeters (approved):
lgtm. @lukehinds did you want to take a look over this as well (like you did the 1st PR)?

keylime/tpm/tpm2.py
maugustosilva: Point taken. Will re-work it for PR3 with a better (logged) error handling

keylime/tpm/tpm_abstract.py
MikeSpreitzer: Again, if we are not trusting the external tool, it is not given that `yaml_to_dict` returns a dict.  All we know is that it returns something that `yaml.load` can return.

keylime/config.py
```diff
@@ -242,9 +242,10 @@ def list_to_dict(alist):
     return params
 
 
-def yaml_to_dict(arry):
+def yaml_to_dict(arry, add_newlines=True):
     arry = convert(arry)
-    return yaml.load("\n".join(arry), Loader=SafeLoader)
+    sep = "\n" if add_newlines else ""
+    return yaml.load(sep.join(arry), Loader=SafeLoader)
```
MikeSpreitzer, line 248: FYI, we can see here that `yaml_to_dict` does not necessarily return a `dict`; it can return anything that `yaml.load` can.

keylime/tpm/tpm2.py
```diff
@@ -1261,3 +1259,50 @@ def read_key_nvram(self):
...
```
MikeSpreitzer, line 1297: Since `config.yaml_to_dict` is not actually guaranteed to return a `dict`, there is a gap in the defensive programming here --- no checking that we actually got a `dict`.

keylime/tpm/tpm_abstract.py
```diff
@@ -281,15 +291,31 @@ def check_pcrs(self, agent_id, tpm_policy, pcrs, data, virtual, ima_measurement_
...
 
-            # check whether this is a MB PCR -- do *not* compare measured boot PCRs against a reference state, if one exists
-            if pcrnum in config.MEASUREDBOOT_PCRS and mb_intended_state :
+            # PCRs set by measured boot get their own special handling
+
+            if pcrnum in config.MEASUREDBOOT_PCRS :
+
+                if mb_refstate :
+
+                    if mb_measurement_list :
```
MikeSpreitzer, line 300: Missing negation here.

maugustosilva: oops, this was a mistake

lukehinds (changes requested):
no changes as yet, just want to pause this while I test (doing so now).

lukehinds:
I hit the following on a fresh fedora 33 vm

```
  File "/usr/local/bin/keylime_verifier", line 33, in <module>
    sys.exit(load_entry_point('keylime==0.0.0', 'console_scripts', 'keylime_verifier')())
  File "/usr/local/bin/keylime_verifier", line 25, in importlib_load_entry_point
    return next(matches).load()
  File "/usr/lib64/python3.9/importlib/metadata.py", line 77, in load
    module = import_module(match.group('module'))
  File "/usr/lib64/python3.9/importlib/__init__.py", line 127, in import_module
  File "<frozen importlib._bootstrap>", line 1030, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 680, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 790, in exec_module
  File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
  File "/usr/local/lib/python3.9/site-packages/keylime-0.0.0-py3.9.egg/keylime/cmd/verifier.py", line 10, in <module>
    from keylime import cloud_verifier_tornado
  File "/usr/local/lib/python3.9/site-packages/keylime-0.0.0-py3.9.egg/keylime/cloud_verifier_tornado.py", line 22, in <module>
    from keylime import cloud_verifier_common
  File "/usr/local/lib/python3.9/site-packages/keylime-0.0.0-py3.9.egg/keylime/cloud_verifier_common.py", line 21, in <module>
    from keylime.tpm import tpm_obj
  File "/usr/local/lib/python3.9/site-packages/keylime-0.0.0-py3.9.egg/keylime/tpm/tpm_obj.py", line 10, in <module>
    from keylime.tpm import tpm1, tpm2
  File "/usr/local/lib/python3.9/site-packages/keylime-0.0.0-py3.9.egg/keylime/tpm/tpm2.py", line 23, in <module>
    from keylime import tpm_bootlog_enrich
  File "/usr/local/lib/python3.9/site-packages/keylime-0.0.0-py3.9.egg/keylime/tpm_bootlog_enrich.py", line 50, in <module>
    efivarlib_functions = CDLL(efivarlib)
  File "/usr/lib64/python3.9/ctypes/__init__.py", line 374, in __init__
    self._handle = _dlopen(self._name, mode)
```

lukehinds:
ok, missing depedency on fedora is: `efivar-devel`

@maugustosilva can you send another commit with this added to `installer.sh` we also need it added to the ansible role, if you ask @axelsimon nicely he can do that, or if you want to do it yourself it needs to be added to packages.yml

Please triage what services need this dep, e.g. verifier | registrar | agent , in case we need to think about impact in the rust agent. 

@lukehinds this will need to be added to the fedora rpm packaging. 

@mbestavros / @lukehinds update the CI docker image for this. 

When we have the above complete (well the rpm packaging can be a post merge step), we can consider shipping this (just runing a few more checks in the meantime).

maugustosilva:
@lukehinds done, and I have also opened a PR on Keylime-ansible.

keylime/tpm/tpm_abstract.py
mpeters: If there's a possibility that we're giving data to `yaml.load` inside of `yaml_to_dict` that could not be a dict, then that needs to be handled in `yaml_to_dict` not everywhere it could be used. So I agree that if we're not trusting it we should be defensive, just not about sprinkling it all over the place, but centralized.

keylime/config.py
mpeters: Then maybe `yaml_to_dict` needs to verify what `yaml.load` is returning and fail if it's not a dict.

MikeSpreitzer: I looked at all the uses of `yaml_to_dict`, and it looks to me like they do indeed assume that the result is a dict (or maybe, in some cases, a value whose truthiness is false).  So, yes, I think that adding a check inside `yaml_to_dict` makes sense; it could raise an Exception when the result is not a dict.

galmasi:
Not quite sure what is holding up merging ... anything any of us can do to facilitate?

mpeters:
@maugustosilva there are now some conflicts that need to be resolved because of the changes to fix CVE-2021-3406. I don't think they will be too substantial, but we'll need this code to be updated against that now.

PR #536, opened 2021-02-04
Added developer conveniences to .gitignore

MikeSpreitzer:
This PR adds to `.gitignore`:
- `.venv`, a common name for a Python virtual environment;
- `.vscode`, the name for the directory that a popular IDE creates for itself; and
- `workspace.code-workspace`, the name suggested by that IDE when it says I should save my current workspace.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,14 @@
 /cloudinit_script_encrypted.txt
+
+# For Python virtual environments
+.venv
+
+# Created automatically by a popular IDE
+.vscode
+
+# Suggested by a popular IDE when it tells me I should save my workspace
+workspace.code-workspace
+
 # Backup files
 *.~
 *~


.gitignore
```diff
@@ -1,4 +1,10 @@
 /cloudinit_script_encrypted.txt
+
+# Conveniences maintained locally by developers
```
mpeters, line 3: Can we put some of the comments from the commit in the file itself instead of this comment? I think most people know .venv but it wouldn't hurt to mention what it was. And then one comment for both of the vscode things to show they are together with vscode?

mpeters (changes requested):

.gitignore
MikeSpreitzer: done

mpeters (approved):

PR #538, opened 2021-02-08
Fix wrong option used in the restful test

kaifeng:
retry_interval should be used instead of max_retries.


lukehinds (approved):
good catch!

PR #539, opened 2021-02-08
Deprecate tpm1 and deep quote

lukehinds:
This change removes all deep quote code and tpm1 functionality

Various tpm values such as `tpm_version` are also removed.


keylime/tenant.py
```diff
@@ -434,19 +436,18 @@ def check_ek(self, ek, ekcert, tpm):
                 logger.warning(
                     "No EK cert provided, require_ek_cert option in config set to True")
                 return False
-            elif not tpm.verify_ek(base64.b64decode(ekcert), ek):
+            elif not self.tpm_instance.verify_ek(base64.b64decode(ekcert), ek):
```
kaifeng, line 439: I suggest we move verify_ek (as well as other hardware unrelated implementations) out of tpm, tenant doesn't need a hardware tpm to run.

kaifeng (commented):
We also need to remove tpm_version from the database and agent processing I guess.

keylime/tenant.py
lukehinds: thanks for noticing that, would explain why my quotes were failing.

on the topic of moving verify_ek we should discuss that in an issue.

lukehinds:
> We also need to remove tpm_version from the database and agent processing I guess.

Yep, I have a migration to do that.

keylime/tpm/tpm_main.py
```diff
@@ -870,6 +869,8 @@ def verify_ek(self, ekcert, ekpem):
             ekcertpem = ek509.get_pubkey().get_rsa().as_pem(cipher=None).decode('utf-8')
 
             # Make sure given ekcert is for their ek
+            print('ekpem: ', str(ekpem))
```
mpeters, line 872: Was this left in for debugging? Should it be sent to the logger instead?

test/test_restful.py
```diff
@@ -450,35 +442,36 @@ def test_011_reg_agent_activate_put(self):
...
-            cert="",
-            verify=False
-        )
-
-        self.assertEqual(response.status_code, 200, "Non-successful Registrar agent vActivate return code!")
-        json_response = response.json()
-
-        # Ensure response is well-formed
-        self.assertIn("results", json_response, "Malformed response body!")
+    # @unittest.skipIf(not vtpm, "Registrar's PUT /v2/agents/{UUID}/vactivate only for vTPMs!")
```
mpeters, line 445: Why is this being commented out? Should it be deleted instead?

mpeters (changes requested):
This is missing the DB migration change for deleting the tpm_version from the DB

test/test_restful.py
lukehinds: It should , was just testing first...will remove this for now

keylime/tpm/tpm_main.py
lukehinds: ack, this is debug residue.

mpeters (approved):
lgtm

nabilschear:
sadness for all that code.  But, life must move on!

PR #544, opened 2021-02-12
Allow specifying REQUIRE_ROOT, LOG_DIR and WORK_DIR via environment variables

puiterwijk:
This allows running Keylime registrar and verifier without root and without manually changing code files.

An example executing of the verifier would be: `KEYLIME_CONFIG=`pwd`/keylime.conf KEYLIME_DIR=`pwd`/workdir/work KEYLIME_REQUIRE_ROOT=off KEYLIME_LOGDIR=`pwd`/workdir/logs/ keylime_verifier`


mpeters (approved):
lgtm

PR #547, opened 2021-02-16
Write boot_aggregate in allowlist

### LINKED ISSUE #546: scripts/create_allowlist.sh doesn't produce a boot_aggregate item and IMA attestation requires it

I believe it is an issue with the current implementation.

# Environment
 * OS / version: Fedora Server 33
 * TPM Manufacturer: Microsoft HyperV vTPM
 * Keylime version: keylime-5.8.0-1.fc33.noarch

# Description 

`scripts/create_allowlist.sh` does not add a boot_aggregate item, nor have an option to add the current one.
After generating the allowlist and using `keylime_tenant` to set up IMA attestation, I get the following message:

```shell
2021-02-12 16:03:32.281 - keylime.ima - WARNING - No boot_aggregate value found in allowlist, adding an empty one
```

Then, attestation fails because it compares against 0.

To workaround this, add a "boot_aggregate" item with the corresponding digest. To add the current boot_aggregate for the current system:

```shell
grep boot_aggregate /sys/kernel/security/ima/ascii_runtime_measurements | awk '{ sub("sha256:","",$4); print $4 "  " $5 }' >> allowlist.txt
```

Expected behaviour is one of the following:

1. You should not need to add a boot_aggregate in allowlist. The expected boot_aggregate should be taken from the TPM Quote, by concatenating and digesting the reported PCR 0 to 7 values. The IMA Measurement Log boot_aggregate should be compared with the recontructed boot_aggregate from the TPM Quote. If that matches, then we can start trusting whatever is in the IMA Measurement Log
2. The tool should warn about the missing boot_aggregate item, or ask to include the current one (it is taking the current digest of the binaries, so getting the current boot_aggregate make sense here as well)
3. Reconstruct the boot_aggregate from the current PCR values an add it to the allowlist?

## Steps to reproduce problem
1.  Run `scripts/create_allowlist.sh allowlist.txt sha256sum`
2.  Use `allowlist.txt` to add a host with the `keylime_tenant` command
3.  IMA Attestation will fail 

# Relevant logs

```shell
2021-02-12 16:03:32.281 - keylime.ima - WARNING - No boot_aggregate value found in allowlist, adding an empty one
```

dnoliver:
Fix #546 

Signed-off-by: Nicolas Oliver <nicolasoliver03@gmail.com>

### DIFF
--- a/scripts/create_allowlist.sh
+++ b/scripts/create_allowlist.sh
@@ -60,6 +60,13 @@ rm -f $OUTPUT
 
 echo "Writing allowlist to $OUTPUT with $ALGO..."
 
+# Add boot_aggregate from /sys/kernel/security/ima/ascii_runtime_measurements (IMA Log) file.
+# The boot_aggregate measurement is always the first line in the IMA Log file.
+# The format of the log lines is the following:
+#     <PCR_ID> <PCR_Value> <IMA_Template> <File_Digest> <File_Name> <File_Signature>
+# File_Digest may start with the digest algorithm specified (e.g "sha1:", "sha256:") depending on the template used.
+head -n 1 /sys/kernel/security/ima/ascii_runtime_measurements | awk '{ print $4 "  boot_aggregate" }' | sed 's/.*://' >> $OUTPUT
+
 # Add all appropriate files under root FS to allowlist
 cd /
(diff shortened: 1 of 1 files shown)


mpeters (approved):
lgtm

Issue #556, opened 2021-02-26
The current code does not allow the deployment of agents onto verifiers (regression)


maugustosilva:
Using latest Keylime (Feb 24th 2021 commits) and attempting to deploy an `agent` on a `verifier`.

```
/usr/local/bin/keylime_tenant -c update -t 100.64.14.16 -v 100.64.14.15 -u 0FC2F2DC-E946-5F94-9D03-5A542D237927 --cert /var/lib/keylime/ca/0FC2F2DC-E946-5F94-9D03-5A542D237927 --tpm_policy '{"23":"0000000000000000000000000000000000000000000000000000000000000000"}'
Using config file /etc/keylime.conf
2021-02-25 20:45:21.849 - keylime.tpm - INFO - TPM2-TOOLS Version: 5.0
2021-02-25 20:45:21.853 - keylime.tenant - INFO - Setting up client TLS in /var/lib/keylime/cv_ca
2021-02-25 20:45:21.853 - keylime.tenant - INFO - TPM PCR Mask from policy is 0x800000
2021-02-25 20:45:21.853 - keylime.tenant - INFO - TPM PCR Mask from policy is 0x808000
Please enter the password to decrypt your keystore:
2021-02-25 20:45:25.781 - keylime.ca-util - INFO - Creating cert package for 0FC2F2DC-E946-5F94-9D03-5A542D237927 in 0FC2F2DC-E946-5F94-9D03-5A542D237927-pkg.zip
2021-02-25 20:45:25.877 - keylime.ca-util - INFO - Creating cert package for RevocationNotifier in RevocationNotifier-pkg.zip
2021-02-25 20:45:25.884 - urllib3.connectionpool - DEBUG - Starting new HTTPS connection (1): 100.64.14.15:8881
/usr/lib/python3.8/site-packages/urllib3/connectionpool.py:997: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings
  warnings.warn(
2021-02-25 20:45:25.925 - urllib3.connectionpool - DEBUG - https://100.64.14.15:8881 "DELETE /agents/0FC2F2DC-E946-5F94-9D03-5A542D237927 HTTP/1.1" 200 49
2021-02-25 20:45:25.927 - keylime.tenant - INFO - Agent 0FC2F2DC-E946-5F94-9D03-5A542D237927 deleted from the CV
[...]

PR #557, opened 2021-03-01
Fix missing tpm2 commands

lukehinds:
During the deprecation of tpm1 in 45cb879, the handling of
`need_hw_tpm=True` was removed, which in turn resulted in
no run of `tpm2_startup` and `tpm2_getcap`

Resolves: #556

### DIFF
--- a/keylime/cmd/ima_emulator_adapter.py
+++ b/keylime/cmd/ima_emulator_adapter.py
@@ -14,7 +14,6 @@ from keylime.tpm.tpm_abstract import config, hashlib
 from keylime.common import algorithms
 
 # Instaniate tpm
-global tpm_instance
 tpm_instance = tpm(need_hw_tpm=True)
 
 start_hash = ('0000000000000000000000000000000000000000')
@@ -23,6 +22,7 @@ ff_hash = ('ffffffffffffffffffffffffffffffffffffffff')
 
 def ml_extend(ml, position, searchHash=None):
     global start_hash
+    global tpm_instance
     f = open(ml, 'r')
     lines = itertools.islice(f, position, None)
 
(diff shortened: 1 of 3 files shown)


lukehinds:
@maugustosilva please test the above.

I have this in draft mode as I need to check a few other things

lukehinds:
tests are failing as there appears to be a bug in swtpm, @puiterwijk is going to raise this with @stefanberger

stefanberger:
@lukehinds What has changed? I just fixed an issue related to saving of contexts over the weekend but that was in libtpms. Are you running swtpm tests?

lukehinds:
> @lukehinds What has changed? I just fixed an issue related to saving of contexts over the weekend but that was in libtpms. Are you running swtpm tests?

@puiterwijk is opening an issue right now in swtpm, should be there any second..he knows the details.

stefanberger:
The solution is to NOT use `--allow-signing` since this will create a non-standard EK that has been specified and per Infrastructure working group is 'possible' but it requires an EK template to be used for re-creating the EK afterwards. Hardware TPMs won't likely have it, either, so you cannot rely on it being generally available.

keylime/cmd/ima_emulator_adapter.py
```diff
@@ -13,13 +13,16 @@
 from keylime.tpm.tpm_abstract import config, hashlib
 from keylime.common import algorithms
 
+# Instaniate tpm
+global tpm_instance
```
mpeters, line 17: Do we need to use a global for this so that the state is shared between them or the initialization happens at the beginning? Or is it to make sure that all these calls to `tpm()` send the right options? If the latter, then I think it's just as clear to pass those options to the calls, it's just a couple of places. Or maybe to create a new function that's called as a wrapper around `tpm()`. I know we use globals in various places already, but just looking at whether they are always required.

maugustosilva:
@lukehinds with this PR, my internal end-to-end testing is now successful. LGTM

PR #559, opened 2021-03-01
tpm_main: make activate_identity return the correct thing on error

puiterwijk:
Right now, the activate_identity function is returning `False`, while
keylime_agent is checking for `None` to return failure.
This means that the `False` is tried to be encoded as a result if the
credential activation procedure failed, which causes very vague errors.

The resulting error:
TypeError: key: expected bytes or bytearray, but got 'bool'

Signed-off-by: Patrick Uiterwijk <patrick@puiterwijk.org>


mpeters:
@puiterwijk can you rebase and re-push? Should kick off CI again

mpeters (approved):
lgtm

PR #563, opened 2021-03-01
Second PR (of a total of 4) to address #491

maugustosilva:
* The most important change here is the fact that the TPM 2.0 preboot
event log , shipped to the verifier from the agent, will now be
authenticated, by replaying the aforementioned log and comparing it
against the values of PCRs [0-9],14.
* If these do not match, then the node will be considered as failed at
attestation.
* A small utility (tpm_bootlog_enrich.py) is added to perform the log replay.
In the future, it could be replaced by newer versions of tpm2-tools capable
of performing the same activity (several open PRs against the tool with
fixes and improvements)

### DIFF
--- a/keylime/cloud_verifier_common.py
+++ b/keylime/cloud_verifier_common.py
@@ -207,7 +207,6 @@ def process_quote_response(agent, json_response):
         ima_keyring,
         mb_measurement_list,
         {})
-
     if not validQuote:
         return False
 
--- a/keylime/cloud_verifier_common.py
+++ b/keylime/cloud_verifier_common.py
@@ -207,6 +207,7 @@ def process_quote_response(agent, json_response):
         ima_keyring,
         mb_measurement_list,
         {})
+
     if not validQuote:
         return False
 
--- a/keylime/tpm/tpm_main.py
+++ b/keylime/tpm/tpm_main.py
@@ -21,7 +21,6 @@ import simplejson as json
 
 from keylime import cmd_exec
 from keylime import config
-from keylime import tpm_bootlog_enrich
 from keylime import keylime_logging
 from keylime import secure_mount
 from keylime import tpm_bootlog_enrich
--- a/keylime/tpm/tpm_main.py
+++ b/keylime/tpm/tpm_main.py
@@ -21,6 +21,7 @@ import simplejson as json
 
 from keylime import cmd_exec
 from keylime import config
+from keylime import tpm_bootlog_enrich
 from keylime import keylime_logging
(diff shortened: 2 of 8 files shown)


keylime/cloud_verifier_common.py
```diff
@@ -207,6 +207,7 @@ def process_quote_response(agent, json_response):
         ima_keyring,
         mb_measurement_list,
         {})
+
```
mpeters, line 210: Is this blank line to make pylint happy? If so, it should be a separate commit as it's not related to this change.

keylime/tpm/tpm_main.py
```diff
@@ -1219,3 +1217,49 @@ def read_key_nvram(self):
             logger.debug("Invalid key length from NVRAM: %d" % (len(output)))
             return None
         return output
+
+    def stringify_pcrs(self, log: dict) -> None:
```
mpeters, line 1221: This is probably just a personal preference, but this method is modifying it's arguments directly instead of returning a new struct. I'm not a fan of methods that only exist for their side effect. I know it's only called once inside of `parse_binary_bootlog`, but I think I'd be more comfortable with it if we made it a private method. 

Also, again probably nit picking, but the name makes me think it would return some string representation of all of the PCRs instead of adjusting the keys in a dict. Maybe if it's renamed to something like `_stringify_pcr_keys' that would make me less squeamish on both counts.

mpeters (changes requested):
Some small changes requested

keylime/cloud_verifier_common.py
maugustosilva: No, I think it was an artifact of an application of a patch. I can remove if needed.

mpeters: Yes please. There's no reason to include files in the git history that weren't functionally changed.

mpeters:
Seems there's a test failure with loading a dependency:

```
```

maugustosilva:
> Seems there's a test failure with loading a dependency:
> 
> ```
> OSError: libefivar.so: cannot open shared object file: No such file or directory
> ```

True! Forgot to add the proper dependency to `installer.sh`, as requested by @lukehinds ... just did it now

mpeters (approved):
lgtm

mpeters:
@maugustosilva seems there are still some conflicts after your most recent push

mpeters:
> @maugustosilva seems there are still some conflicts after your most recent push

Maybe it's the Dockerfile doing the same that was previously merged?

PR #564, opened 2021-03-02
Adds sm3_256 to the known hash algorithms

kaifeng:
This algorithm is seen in Nationz chips, and is in the TCG standard.
Adds sm3_256 to known algorithms, otherwise keylime refuses to run
when the tpm_hash_alg is set to sm3_256.


mpeters (approved):
lgtm

PR #565, opened 2021-03-02
Replace simplejson imports

puiterwijk:
During Python2.6 development, the "simplejson" module got pulled into the standard
library as "json".
https://docs.python.org/3/whatsnew/2.6.html#the-json-module-javascript-object-notation

This means that as of Python 2.6, "json" is always available, and is identical
to what previously was called simplejson.

Signed-off-by: Patrick Uiterwijk <patrick@puiterwijk.org>

### DIFF
--- a/auto-ipsec/libreswan/src/local_action_crashsa.py
+++ b/auto-ipsec/libreswan/src/local_action_crashsa.py
@@ -13,7 +13,7 @@ import keylime.config as common
 import keylime.keylime_logging as keylime_logging
 import keylime.cmd_exec as cmd_exec
 
-import simplejson as json
+import json
 
 # read the config file
 config = common.get_config()
--- a/auto-ipsec/racoon/src/local_action_deletesa.py
+++ b/auto-ipsec/racoon/src/local_action_deletesa.py
@@ -15,7 +15,7 @@ import keylime.config as common
 import keylime.keylime_logging as keylime_logging
 import keylime.cmd_exec as cmd_exec
 
-import simplejson as json
+import json
 
 # read the config file
 config = common.get_config()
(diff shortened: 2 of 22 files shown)


mpeters:
There is a test failure related to your change:

```
Error in test.test_verifier_db.TestVerfierDB.test_set_operation_state
  File "/usr/lib64/python3.9/site-packages/sqlalchemy/engine/base.py", line 1204, in _execute_context
    context = constructor(dialect, self, conn, *args)
  File "/usr/lib64/python3.9/site-packages/sqlalchemy/engine/default.py", line 855, in _init_compiled
    param.append(processors[key](compiled_params[key]))
  File "/usr/lib64/python3.9/site-packages/sqlalchemy/sql/sqltypes.py", line 1709, in process
    value = dumps(value, protocol)
TypeError: dumps() takes 1 positional argument but 2 were given
```

stefanberger:
You should be able to remove simplejson from requirements.txt and test/test-requirements.txt after doing the same change as in your patch to test/test_restful.py.

stefanberger:
Maybe that the build error can be overcome with this type of fix: https://github.com/python-social-auth/social-storage-sqlalchemy/commit/39d129c1718dfab6be33bec69b1c44853e66da86  ... hopefully it doesn't have any other side effects.

stefanberger:
> If not, would you mind if I gave it a try?

I for sure would suggest to give it a try...

sergio-correia:
> > If not, would you mind if I gave it a try?
> 
> I for sure would suggest to give it a try...

Alright, I submitted #789.

THS-on:
With #789 merged I think this now can be closed.

PR #567, opened 2021-03-03
Use the quay.io version of fedora for CI tests

mpeters:
We hit an error when quay was building a new version of our image:

    Could not pull base image: API error (500): toomanyrequests:
    You have reached your pull rate limit. You may increase the
    limit by authenticating and upgrading:
    https://www.docker.com/increase-rate-limit

So docker hub was throttling quay and killing our build. So switching to
the quay.io version of the same image will keep everything inside of quay
and avoid throttling.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -5,7 +5,7 @@
 # It is not recommended for use beyond testing scenarios.
 ##############################################################################
 
-FROM fedora:latest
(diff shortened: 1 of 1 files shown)


mbestavros (approved):
LGTM! This can be pretty annoying to run into, so glad it's being fixed.

maugustosilva:
+1 LGTM

PR #568, opened 2021-03-03
Third PR (of a total of 4) to address #491

maugustosilva:
* The "measured boot reference state" (mb_refstate), provided via
  `tenant` CLI with the option `--mb_refstate` is now parsed (JSON) and
stored on the `verifier` database in a new column (as text)
* The supplied mb_refstate is not yet fully used (but the log is
  replayed against and the values of PCRs [0-9][11-15] are checked)

### DIFF
--- a/keylime/cloud_verifier_common.py
+++ b/keylime/cloud_verifier_common.py
@@ -206,7 +206,7 @@ def process_quote_response(agent, json_response):
         hash_alg,
         ima_keyring,
         mb_measurement_list,
-        {})
+        agent['mb_refstate'])
     if not validQuote:
         return False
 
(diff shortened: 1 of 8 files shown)


stefanberger (commented):
Squash this into the previous file or at least give it a better title...

keylime/cloud_verifier_common.py
```diff
@@ -270,6 +270,17 @@ def process_get_status(agent):
         al_len = len(allowlist['allowlist'])
     else:
         al_len = 0
+
+    try :
+        mb_refstate = ast.literal_eval(agent.mb_refstate)
+    except Exception as e:
+        logger.warning(f'Non-fatal problem ocurred while attempting to evaluate agent attribute "mb_refstate" ({e.args}). Will just consider the value of this attribute to be "None"')
```
mpeters, line 277: I know we're dealing with an AST, and language errors are hard, but is it possible to include more information here? I can imagine trying to debug this being a pain. Maybe we could put some details from the exception in a debug log statement?

mpeters (changes requested):

keylime/cloud_verifier_common.py
maugustosilva: Would something like `str(agent.mb_refstate)` be enough for a `debug` statement? I am asking because the `warning` already brings the args from the exception

mpeters: Yeah, I think if we have the exception args in the warning and the refstate in a debug (could even put the debug in the `try`) that should be good enough to start.

mpeters (approved):
lgtm

PR #569, opened 2021-03-03
More specific base docker image tag

mpeters:
The fedora images in quay dont use generic tags like "latest" or even
"33", so neither should we


mbestavros (approved):

axelsimon (approved):
Looks good, let's go!

PR #580, opened 2021-03-10
Fourth and final PR to address #491

maugustosilva:
On this PR, we finally introduce the concept of "policy engine" in its complete form. Please refer to the item 3 of #491 for the details.

Signed-off-by: Mike Spreitzer <mspreitz@us.ibm.com>

### DIFF
--- a/keylime/config.py
+++ b/keylime/config.py
@@ -315,6 +315,8 @@ IMA_PCR = 10
 # measured boot addons
 MEASUREDBOOT_PCRS = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 11, 12, 13, 14, 15]
 MEASUREDBOOT_ML = '/sys/kernel/security/tpm0/binary_bios_measurements'
+MEASUREDBOOT_IMPORTS = get_config().get('cloud_verifier', 'measured_boot_imports', fallback='').split(',')
+MEASUREDBOOT_POLICYNAME = get_config().get('cloud_verifier', 'measured_boot_policy_name', fallback='accept-all')
 
 # this is where data will be bound to a quote, MUST BE RESETABLE!
 TPM_DATA_PCR = 16
--- a/keylime/elchecking/tests.py
+++ b/keylime/elchecking/tests.py
@@ -1,0 +1,588 @@
+import abc
+import re
+import typing
+
+# This module defines the abstraction of a Test (of JSON data)
+# and several specific test classes.
+# A Test can be used multiple times, even concurrently.
+
+# Data is the type of Python data that corresponds to JSON values.
+Data = typing.Union[int, float, str, bool, typing.Tuple['Data', ...],
+                    typing.Mapping[str, 'Data'], None]
+
+# Globals is a dict of variables for communication among tests.
+# There is a distinct dict for each top-level use of a test.
+Globals = typing.Mapping[str, Data]
+
+# PCR_Contents maps digest name to map from PCR index to PCR value.
+# Here digest name is something like 'sha256'.
+# Each PCR index is a decimal string, so that this can be JSON data
+PCR_Contents = typing.Mapping[str, typing.Mapping[str, int]]
+
+
+class Test(metaclass=abc.ABCMeta):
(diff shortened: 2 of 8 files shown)


keylime/elchecking/tests.py
```diff
@@ -0,0 +1,594 @@
...
+            elt_test(elt)
+        return True
+    return test
+
+
+def dict_test(dom_test: typing.Callable[[typing.Any], bool], rng_test: typing.Callable[[typing.Any], bool]) -> typing.Callable[[typing.Any], bool]:
+    """Return a lambda that tests for dict with certain type key and value"""
+    def test(dat: typing.Any) -> bool:
+        type_test(dict)(dat)
+        for dom, rng in dat.items():
```
mpeters, line 63: I'm curious about the naming here. What are `dom` and `rng` references to and why use them instead of something more familiar like `key`/`value` ?

keylime/config.py
```diff
@@ -315,6 +315,8 @@ def valid_regex(regex):
 # measured boot addons
 MEASUREDBOOT_PCRS = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 11, 12, 13, 14, 15]
 MEASUREDBOOT_ML = '/sys/kernel/security/tpm0/binary_bios_measurements'
+MEASUREDBOOT_IMPORTS = get_config().get('cloud_verifier', 'measured_boot_imports', fallback='').split(',')
+MEASUREDBOOT_POLICYNAME = get_config().get('cloud_verifier', 'measured_boot_policy_name', fallback='accept-all')
```
mpeters, line 319: Right now this patch makes the measured boot policy global. Will that be too inflexible going forward? Should it be a per-machine setting set with the tenant command? I don't think it should necessarily hold up this PR, but something to look into for the future.

mpeters (changes requested):
Not really requesting changes (yet), just looking for some feedback to questions.

keylime/config.py
maugustosilva: @mpeters you're absolutely right... next step on it is to make the "measured_boot" policy a twin of the - about to be merged - "ima_policy" (i.e., these policy might be global or per-node).

mpeters:
@maugustosilva one more thing that would be helpful: I know we've talked about having full documentation for this feature, which will come in later, but can we have a short documented way to use this (besides just the `accept-all` policy)? The tests options are quite extensive, so seeing this in action, even in a small way, would make it easier to visualize.

keylime/elchecking/tests.py
maugustosilva: @MikeSpreitzer can you comment on that?

MikeSpreitzer: `dom` and `rng` are short for `domain` and `range`, terms I thought were common for the two sides of a function (be it finite or otherwise).  If you think it matters, these can easily be changed to `key` and `value`.

mpeters: You're right those are common names for the 2 sides of a function, it just wasn't obv to me that's what we were testing. The method is called `dict_test` and the comment mentions testing the dict for a certain key and value. That's pretty generic and wouldn't necessarily be fore testing the the domain and range of another function. Also because it wasn't immediately obv to me this was what was happening, `rng` looked like `RNG` or random-number generator and I was confused even more :)

Unless there's something specific about testing a domain and range here (which I don't see, but might be more obv in how it's used) I would suggest something more generic like `key`/`value` or `k`/`v`.

maugustosilva:
> @maugustosilva one more thing that would be helpful: I know we've talked about having full documentation for this feature, which will come in later, but can we have a short documented way to use this (besides just the `accept-all` policy)? The tests options are quite extensive, so seeing this in action, even in a small way, would make it easier to visualize.

I have opened a PR with an initial draft of the full documentation on `keylime-docs` repository.

keylime/elchecking/tests.py
maugustosilva: @mpeters I proposed we merge as is, and we will fix it on the "PR5", the one which will also include measured boot-specific tests, as we previously discussed.

mpeters (approved):

PR #583, opened 2021-03-13
Fixes #582

### LINKED ISSUE #582: keylime_agent broken for non-x86_64 architectures due to libefivar.so not being available
# Environment
 * OS / version: Fedora 32
 * Processor architecture: PowerPC64
 * TPM Manufacturer: Nuvoton
 * Keylime version: today's master (1766957a748140d6040df70f)

# Description 

The keylime_agent fails to start on ppc64. The needed package providing the library is not available on this platform.

```
# keylime_agent                       
  File "/usr/local/bin/keylime_agent", line 11, in <module>
  File "/usr/lib/python3.8/site-packages/pkg_resources/__init__.py", line 489, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/usr/lib/python3.8/site-packages/pkg_resources/__init__.py", line 2852, in load_entry_point
    return ep.load()
  File "/usr/lib/python3.8/site-packages/pkg_resources/__init__.py", line 2443, in load
    return self.resolve()
  File "/usr/lib/python3.8/site-packages/pkg_resources/__init__.py", line 2449, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/usr/local/lib/python3.8/site-packages/keylime/cmd/agent.py", line 9, in <module>
    from keylime import keylime_agent
  File "/usr/local/lib/python3.8/site-packages/keylime/keylime_agent.py", line 37, in <module>
    from keylime.tpm.tpm_main import tpm
  File "/usr/local/lib/python3.8/site-packages/keylime/tpm/tpm_main.py", line 26, in <module>
    from keylime import tpm_bootlog_enrich
  File "/usr/local/lib/python3.8/site-packages/keylime/tpm_bootlog_enrich.py", line 50, in <module>
    efivarlib_functions = CDLL(efivarlib)
  File "/usr/lib64/python3.8/ctypes/__init__.py", line 373, in __init__
    self._handle = _dlopen(self._name, mode)
```

keylime_agent should work

## Steps to reproduce problem
1. See above

maugustosilva:
Do not import tpm_bootlog_enrich (which depends on libefivar.so),
on non-x86_64 architectures

### DIFF
--- a/keylime/config.py
+++ b/keylime/config.py
@@ -316,6 +316,8 @@ IMA_PCR = 10
 MEASUREDBOOT_PCRS = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 11, 12, 13, 14, 15]
 MEASUREDBOOT_ML = '/sys/kernel/security/tpm0/binary_bios_measurements'
 
+LIBEFIVAR="libefivar.so" # formerly "/usr/lib/x86_64-linux-gnu/libefivar.so"
+
 # this is where data will be bound to a quote, MUST BE RESETABLE!
 TPM_DATA_PCR = 16
 
--- a/keylime/tpm/tpm_abstract.py
+++ b/keylime/tpm/tpm_abstract.py
@@ -232,6 +232,9 @@ class AbstractTPM(metaclass=ABCMeta):
 
         if mb_refstate and mb_measurement_list :
             mb_measurement_data = self.parse_bootlog(mb_measurement_list)
+            if not mb_measurement_data :
+                logger.error("Unable to parse measured boot event log. Check previous messages for a reason for error.")
+                return False
             log_pcrs = mb_measurement_data.get('pcrs')
             if not isinstance(log_pcrs, dict):
                 logger.error("Parse of measured boot event log has unexpected value for .pcrs: %r", log_pcrs)
--- a/keylime/tpm/tpm_main.py
+++ b/keylime/tpm/tpm_main.py
@@ -23,11 +23,13 @@ from keylime import cmd_exec
 from keylime import config
 from keylime import keylime_logging
 from keylime import secure_mount
-from keylime import tpm_bootlog_enrich
 from keylime.tpm import tpm_abstract
 from keylime import tpm_ek_ca
 from keylime.common import algorithms
 from keylime.tpm import tpm2_objects
+
+if os.uname()[4] == 'x86_64' :
+    from keylime import tpm_bootlog_enrich
 
 logger = keylime_logging.init_logging('tpm')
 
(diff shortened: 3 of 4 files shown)


mpeters:
This means that libefivar is still required for the agent to run even though it's never used by the agent. It also means that running on non-x86-64 will fail on the server side for weird reasons (missing python imports) even if they have managed to compile libefivar for their platform.

Let's instead just load `tpm_bootlog_enrich` if it's going to be used even if that means delaying until run time.

maugustosilva:
> This means that libefivar is still required for the agent to run even though it's never used by the agent. It also means that running on non-x86-64 will fail on the server side for weird reasons (missing python imports) even if they have managed to compile libefivar for their platform.
> 
> Let's instead just load `tpm_bootlog_enrich` if it's going to be used even if that means delaying until run time.

Due to the way the code is implemented, running `verifier` or `registrar` on non-x86_64, as long as it is done **without** "measured boot", will not fail (@stefanberger kindly confirmed that on a Z node). I can certainly create a second patch to ensure that "measured boot" is not allowed in case `keylime_tenant` is not run from an x86_64 node, but I suggest we do this as a solution to an entirely different issue (i.e., activating "measured boot" on non-x86_64 is not technically feasible)

stefanberger:
It solved an issue on 'Z' and ppc64 for me. Another question would x86_64 would be a QEMU node running SeaBIOS and how would it react there? I have not tried there, either. `/sys/firmware/efi` identifies UEFI.

mpeters:
@maugustosilva I think the problem I'm having here is 2 fold:

1) The agent shouldn't be dependent on libefivar regardless of it's host architecture. It doesn't use it and there's no reason for any of the measured boot code loaded by the agent
2) Checking if the server side is x86_64 in order to be able to do measured boot of an agent that might be x86_64 seems wrong. We don't actually care about the architecture, we care about libefivar being available on the server side. I realize the code that enhances the boot log needs libefivar and currently it only really runs on x86_64, but that's more of an indirect fact rather than a dependency.

So I'd rather see this change do the following:
1) Only load `tpm_bootlog_enrich` if `tpm.tpm_main.parse_binary_bootlog()' is called so it's never loaded on the agent
2) Optionally add a check in `tpm_bootlog_enrich` for loading libefivar and if it's not installed, throw a helpful exception/error

I'm more concerned about first than the latter though.

stefanberger:
>     2\. Optionally add a check in `tpm_bootlog_enrich` for loading libefivar and if it's not installed, throw a helpful exception/error

I think x86_64 when using UEFI should try to load `libefivar.so.1` (with the .1 appended since for example Fedora only has `libefivar.so` in case the -devel package is installed) on the agent side and reported an error if not found. All other architectures and x86_64+BIOS should silently return.

FYI: aarch64 does seem to have this library as well and *maybe* could profit from it as well, but we'd need someone to test this: https://fedora.pkgs.org/33/fedora-aarch64/efivar-libs-37-9.fc33.aarch64.rpm.html

maugustosilva:
> @maugustosilva I think the problem I'm having here is 2 fold:
> 
> 1. The agent shouldn't be dependent on libefivar regardless of it's host architecture. It doesn't use it and there's no reason for any of the measured boot code loaded by the agent
> 2. Checking if the server side is x86_64 in order to be able to do measured boot of an agent that might be x86_64 seems wrong. We don't actually care about the architecture, we care about libefivar being available on the server side. I realize the code that enhances the boot log needs libefivar and currently it only really runs on x86_64, but that's more of an indirect fact rather than a dependency.
> 
> So I'd rather see this change do the following:
> 
> 1. Only load `tpm_bootlog_enrich` if `tpm.tpm_main.parse_binary_bootlog()' is called so it's never loaded on the agent
> 2. Optionally add a check in `tpm_bootlog_enrich` for loading libefivar and if it's not installed, throw a helpful exception/error
> 
> I'm more concerned about first than the latter though.

Fair enough, this is certainly a way of fixing the problem, in a non-minimalistic (but much more rounded) manner... will implement something along those lines

mpeters (approved):
lgtm

PR #589, opened 2021-03-16
Add pylint version to check_codestyle output

mpeters:
This can help when tracking down differences between different runs,
say one in CI and the other in local dev


ashcrow (approved):
Good idea!

Issue #585, opened 2021-03-15
Occasional tpm2_quote failures


stefanberger:
# Environment
 * OS / version:  Fedora 32 
 * Processor architecture: PowerPC64
 * TPM Manufacturer: Nuvoton
 * Keylime version: git master ()
 * tpm2-tools: 4.1.3-1.fc32

# Description 

I see occassional tpm2_quote failures. I don't know what is causing these failures. Maybe it's activity on the system, but that's just a guess.

```
2021-03-14 20:11:18.695 - keylime.cloudagent - INFO - GET identity quote returning 200 response.
2021-03-14 20:11:18.727 - keylime.cloudagent - INFO - POST of U key returning 200
----------------------------------------
Exception happened during processing of request from ('9.47.152.88', 48098)
Traceback (most recent call last):
  File "/usr/lib64/python3.8/socketserver.py", line 650, in process_request_thread
    self.finish_request(request, client_address)
  File "/usr/lib64/python3.8/socketserver.py", line 360, in finish_request
    self.RequestHandlerClass(request, client_address, self)
  File "/usr/lib64/python3.8/socketserver.py", line 720, in __init__
    self.handle()
  File "/usr/lib64/python3.8/http/server.py", line 427, in handle
    self.handle_one_request()
  File "/usr/lib64/python3.8/http/server.py", line 415, in handle_one_request
    method()
  File "/usr/local/lib/python3.8/site-packages/keylime/keylime_agent.py", line 117, in do_GET
    quote = tpm_instance.create_quote(
  File "/usr/local/lib/python3.8/site-packages/keylime/tpm/tpm_main.py", line 953, in create_quote
    retDict = self.__run(command, lock=False, outputpaths=[quotepath.name, sigpath.name, pcrpath.name])
  File "/usr/local/lib/python3.8/site-packages/keylime/tpm/tpm_main.py", line 377, in __run
    raise Exception("Command: %s returned %d, expected %d, output %s, stderr %s" % (cmd, code, expectedcode, retout, reterr))
[...]

PR #590, opened 2021-03-19
agent: Fix the checking for a specific error returned by tpm2_quote

stefanberger:
cmd_exec.run returns a list of byte-like objects containing the error strings
that a command may have returned. The current check must have assumed that a
simple string was returned. Fix this by implementing array_contains_substring
that checks whether any byte-like object or string in an array contains a given
substring.

This fixes issue #585.

### DIFF
--- a/keylime/cmd_exec.py
+++ b/keylime/cmd_exec.py
@@ -57,3 +57,12 @@ def run(cmd, expectedcode=EXIT_SUCESS, raiseOnError=True, outputpaths=None,
         'timing': timing,
     }
     return returnDict
+
+
+# list_contains_substring checks whether a substring is contained in the given
+# list. The list may be the reterr from 'run' and contains bytes-like objects.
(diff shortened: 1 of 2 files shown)


maugustosilva:
@mpeters @lukehinds since we don't have a proper fix on `tpm2-tools` for the "non-atomic" operation of getting the quote (and only after that getting the values), I believe this solution proposed by @stefanberger is an acceptable compromise. I fear that without it, we might end up with a steady stream of "false positives" in these large production clusters that we are about to deploy.

stefanberger:
> @mpeters @lukehinds since we don't have a proper fix on `tpm2-tools` for the "non-atomic" operation of getting the quote (and only after that getting the values), I believe this solution proposed by @stefanberger is an acceptable compromise. I fear that

especially since this is a fix to existing code that for some reason did not work anymore

mpeters (approved):
lgtm

PR #592, opened 2021-03-22
scripts: Also use pylint-3 if pylint is not installed

stefanberger:
Fedora installs pylint as pylint-3 and only gets pylint via pip3. We can also
use pylint-3.


mpeters (approved):
lgtm

mpeters:
Seems this breaks things when there is a `pylint` but no `pylint-3` - https://github.com/keylime/keylime/pull/592/checks?check_run_id=2167474143

PR #593, opened 2021-03-22
 run_local.sh: Run unit tests in keylime/tpm/tpm2_objects.py 

stefanberger:
When running local tests using run_local.sh the unit tests in
tpm2_object.py did not run since we were in the wrong directory then.
This patch fixes this problem.

python3: can't open file 'keylime/tpm/tpm2_objects.py': [Errno 2] No such file or directory

Also set the executable flag on run_local.sh so that we can run it like
this for example:

./run_local.sh ${PWD}/../


mpeters (approved):
lgtm

Issue #578, opened 2021-03-09
What version of RHEL does keylime support


kgold2:
Is your an issue a feature request? If so, please raise it as an [enhancement](https://github.com/keylime/enhancements)

# Environment
 * OS / version: RHEL 7.9
 * Processor architecture: x86
 * TPM Manufacturer: n/a
 * Keylime version: master

# Description 

https://github.com/keylime/keylime/blob/master/README.md says that RHEL 7 and 8 are supported.  However, the install script says:  
Version 7.9 of rhel not supported

## Expected behavior vs. actual behavior 

I expected RHEL 7.9 to be supported.

## Steps to reproduce problem

1. clone master
2. /installer.sh -o -m

# Relevant logs

Output says
`Version 7.9 of rhel not supported`

stefanberger:
This would probably fix this *step*:

```
diff --git a/installer.sh b/installer.sh
index 7fb2d58..3212e53 100755
--- a/installer.sh
+++ b/installer.sh
@@ -72,7 +72,7 @@ case "$ID" in

     rhel | centos)
         case "${VERSION_ID}" in
-            7)
+            7*)
                 echo "${ID} ${VERSION_ID} selected."
                 PACKAGE_MGR=$(command -v yum)
                 NEED_EPEL=1
```

I cannot tell whether you would run into other issues with this older version of RHEL since I haven't tried installing it there myself.

The trouble with the VERSION_ID comes from this here:
RHEL identifies itself like this:
```
$ cat /etc/os-release
NAME="Red Hat Enterprise Linux Server"
[...]
VERSION_ID="7.9"
```

CentOS like this:
```
$ cat /etc/os-release
NAME="CentOS Liux"
[...]
VERSION_ID="7"
[...]
```

kgold2:
Hi, stalebot.  Please don't close issues that aren't fixed.

PR #596, opened 2021-03-23
Add python3-gnupg and efivar-dev packages to installer

mpeters:


installer.sh
```diff
@@ -77,7 +77,7 @@ case "$ID" in
                 PACKAGE_MGR=$(command -v yum)
```
stefanberger, line 77: If RHEL 7 / CentOS 7 is still supported can you/we also change the above to `7*)` following my  comments from to issue https://github.com/keylime/keylime/issues/578

mpeters: Good idea, done

stefanberger: The same is probably true for RHEL/ CentOS 8.

stefanberger:
LGTM.

PR #597, opened 2021-03-23
Add signature verification support for local and remote IMA signature verification keys

stefanberger:
This PR adds signature verification support for local and remote IMA signature verification keys.

### DIFF
--- a/keylime/gpg.py
+++ b/keylime/gpg.py
@@ -1,0 +1,32 @@
+'''
+SPDX-License-Identifier: Apache-2.0
+Copyright 2017 Massachusetts Institute of Technology.
+'''
+
+import gnupg
+
+from keylime import keylime_logging
+
+logger = keylime_logging.init_logging('gpg')
+
+def gpg_verify_filesignature(gpg_key_file, filename, gpg_sig_file, file_description):
+    """
+       Verify the file signature (gpg_sig_file) using a public GPG (gpg_key_file)
+       with the file (filename).
+    """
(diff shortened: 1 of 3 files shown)


stefanberger:
@puiterwijk @mpeters So this PR is now the equivalent to the recent extensions for signed allowlists. I make the same assumptions that a GPG key will be used for verification but I don't support checksums. Do you have any comments? 

There's more explanation on the 3rd patch on how to use the options.

stefanberger:
@lukehinds Do you have any comments about this? Are AllowList URLs already available?

lukehinds:
> @lukehinds Do you have any comments about this? Are AllowList URLs already available?

@mpeters might know, he did that patch?

mpeters (approved):
lgtm, appreciate the gpg refactoring/reuse

mpeters:
@stefanberger this looks good to me as it extends a pattern we're already using into other things. Let me know what else you have left to move out of WIP state.

stefanberger:
@mpeters For the remote allow lists you require a checksum or signature. I don't currently require a signature on a remote key, but should I make this a requirement?

mpeters:
That's a good question. I wanted to default to requiring one or the other just to get people in the habit of doing something more securely, but I don't have a firm stance either way.

stefanberger:
I will leave it as it is for now and may come back later to enforce signatures, especially once they exist.

maugustosilva:
@mpeters LGTM, I just tested it on my end-to-end CI successfully

Issue #599, opened 2021-03-24
Add interval-configuration to [webapp] section in keylime.conf


FengqiQiao:
Fix:
Currently keylime_webapp only supports hard-coded "agent populating intervals" inside static/js/webapp.js(line 469-471). This makes GUI hard to configure and test out performance. Add support for keylime_webapp service to allow reading "intervals-configuration" from keylime.conf.

FengqiQiao:
please assign

PR #601, opened 2021-03-24
Add support for keylime_webapp service to read intervals from keylime.conf

FengqiQiao:
This PR is an implementation for #599

### DIFF
--- a/keylime.conf
+++ b/keylime.conf
@@ -544,6 +544,11 @@ cfssl_port = 8888
 webapp_ip = 127.0.0.1
 webapp_port = 443
 
+# The intervals for populating agents, updating agents and updating terminal.
+populate_agents_interval = 4000
+update_agents_interval = 1500
+update_terminal_interval = 2000
+
 #=============================================================================
 # GLOBAL LOGGING CONFIGURATION
 #=============================================================================
--- a/keylime/static/js/webapp.js
+++ b/keylime/static/js/webapp.js
@@ -454,19 +454,3 @@ function updateTerminal() {
         appendToTerminal(response["log"]);
     });
 }
-
-// Attach dragging capabilities for payload upload functionality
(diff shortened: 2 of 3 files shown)


maugustosilva:
Looks good, LGTM

mpeters (approved):
lgtm

dcmiddle:
@mpeters @FengqiQiao 
Not sure if you require sign-offs, but I just happened to notice this one seems to accidentally have been from an unconfigured git client..
`Signed-off-by: root <root@localhost.localdomain>`
As it's the head commit you could maybe still rebase it w/o too much difficulty. If you care. :)

mpeters:
Sorry, I didn't notice this bad signature. We do require sign-offs going forward so I'm not totally sure of the right way to correct this. We don't want to amend the head because it's now published and any changes we make will break others checkouts. I'm fine making changes to git branches/forks that aren't the primary main branch of a project, but once it's out there like this, it's kind of permanent.

@lukehinds and/or @axelsimon do you have any thoughts on the best way to fix this? If nothing else, maybe the DCO bot should be updated to be able to catch these since `root@localhost.localdomain` is never a person that can sign off on where the code came from.

FengqiQiao:
So sorry about that, my bad. `root@localhost.localdomain` is actually me. 
In order to get a clean dev environment, I created a new VM and worked on this PR under root user. However, I forgot to set up git using `git config --global user.name` and `git config --global user.email`. So when I pushed the commit, it resulted in the bad signature.

dcmiddle:
@mpeters I think you can revert which will lay down a new commit. @FengqiQiao can then amend their commit, fetch upstream, and then locally rebase on top of the new head and then push. The offending commit will still be in history but will be immediately adjacent to the reversion and the new commit. 
Probably other options too.

PR #607, opened 2021-04-06
ima_emulator_adapater: Remove unnecessary global statement

stefanberger:
Remove an unnecessary global statement for start_hash since this
variable is accessed read-only. After replaying the log we end up
with the same result in PCR 10 as before this patch.


mpeters (approved):
lgtm

PR #609, opened 2021-04-07
Remove _cLime.c

stefanberger:
Remove _cLime.c and its import. It doesn't seem to be used anywhere in the
active code but only referenced by code in the archive.


kaifeng (approved):
lgtm

stefanberger:
I should be good to merge now.

PR #610, opened 2021-04-07
Very simple fix for the agent (re: measured boot)

maugustosilva:
The agent code does not need to import "measured boot policies"


mpeters (approved):
lgtm

PR #611, opened 2021-04-07
Addresses #436

maugustosilva:
There is still a bit of extra setup work to be done in order to run the
`agent` as non-root, but imo, those should be done outside of the main
keylime code (it is a "packaging" issue, not code issue).

1) It is assumed that the non-root user which is running keylime has the
   ability to read the TPM device.

2) We need to mount `/var/lib/keylime/secure/` as root, since the mount
   command uses `--options`. The solution is to run it as root, and
   chown it to correct non-root user.

3) We need to make sure that the non-root user which is running keylime
   can read `/sys/kernel/security/ima/ascii_runtime_measurements` and
   `/sys/kernel/security/tpm0/binary_bios_measurements`. This could be
   done as simple `chown` or as an udev rule.

### DIFF
--- a/agent_non_root_condition.sh
+++ b/agent_non_root_condition.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+
+KEYLIME_GROUP=tss
+KEYLIME_USER=keylime
+
+# creating group if it isn't already there
+if ! getent group $KEYLIME_GROUP >/dev/null; then
+    addgroup --system $KEYLIME_GROUP
+fi
+
+# creating keylime user if he isn't already there
+if ! getent passwd keylime >/dev/null; then
+    adduser --system --ingroup tss --shell /bin/false \
+    --home /var/lib/keylime --no-create-home \
+    --gecos "Keylime remote attestation" \
+    keylime
+fi
+
+# Create keylime operational directory
+if [ ! -d /var/lib/keylime ]; then mkdir -p /var/lib/keylime/secure
+fi
+
+# Only root can mount tmpfs with `-o`
+if ! grep -qs '/var/lib/keylime/secure ' /proc/mounts ; then mount -t tmpfs -o size=1m,mode=0700 tmpfs /var/lib/keylime/secure
+fi
+
+# Setting ownership for keylime operational directory
+if [ -d /var/lib/keylime ] && getent passwd $KEYLIME_USER >/dev/null; then
+chown -R $KEYLIME_USER:$KEYLIME_GROUP /var/lib/keylime
+fi
+
+# Setting ownership for /sys/kernel/security/<x>, giving keylime agent access to it
+chown -R $KEYLIME_USER:$KEYLIME_GROUP /sys/kernel/security/tpm0
+chown -R $KEYLIME_USER:$KEYLIME_GROUP /sys/kernel/security/ima
--- a/agent_non_root_condition.sh
+++ b/agent_non_root_condition.sh
@@ -1,7 +1,7 @@
-#!/bin/bash
+#!/bin/sh
 
-KEYLIME_GROUP=tss
-KEYLIME_USER=keylime
+KEYLIME_GROUP=${KEYLIME_GROUP:-tss}
+KEYLIME_USER=${KEYLIME_USER:-keylime}
 
 # creating group if it isn't already there
 if ! getent group $KEYLIME_GROUP >/dev/null; then
@@ -10,10 +10,10 @@ fi
 
 # creating keylime user if he isn't already there
 if ! getent passwd keylime >/dev/null; then
-    adduser --system --ingroup tss --shell /bin/false \
+    adduser --system --ingroup $KEYLIME_GROUP --shell /bin/false \
     --home /var/lib/keylime --no-create-home \
     --gecos "Keylime remote attestation" \
-    keylime
+    $KEYLIME_USER
 fi
 
 # Create keylime operational directory
--- a/keylime/keylime_agent.py
+++ b/keylime/keylime_agent.py
@@ -450,7 +450,7 @@ class CloudAgentHTTPServer(ThreadingMixIn, HTTPServer):
 
 
 def main():
-    for ML in [ config.MEASUREDBOOT_ML, config.IMA_ML ] : 
+    for ML in [ config.MEASUREDBOOT_ML, config.IMA_ML ] :
         if not os.access(ML, os.F_OK) :
             logger.warning("Measurement list path %s not accessible by agent. Any attempt to instruct it to access this path - via \"keylime_tenant\" CLI - will result in agent process dying", ML)
 
--- a/keylime/keylime_agent.py
+++ b/keylime/keylime_agent.py
@@ -450,10 +450,6 @@ class CloudAgentHTTPServer(ThreadingMixIn, HTTPServer):
 
 
 def main():
-    if os.getuid() != 0 and config.REQUIRE_ROOT:
-        logger.critical("This process must be run as root.")
-        return
-
     if config.get('cloud_agent', 'agent_uuid') == 'dmidecode':
         if os.getuid() != 0:
             raise RuntimeError('agent_uuid is configured to use dmidecode, '
--- a/keylime/keylime_agent.py
+++ b/keylime/keylime_agent.py
@@ -450,6 +450,10 @@ class CloudAgentHTTPServer(ThreadingMixIn, HTTPServer):
 
 
 def main():
+    for ML in [ config.MEASUREDBOOT_ML, config.IMA_ML ] : 
+        if not os.access(ML, os.F_OK) :
+            logger.warning("Measurement list path %s not accessible by agent. Any attempt to instruct it to access this path - via \"keylime_tenant\" CLI - will result in agent process dying", ML)
+
     if config.get('cloud_agent', 'agent_uuid') == 'dmidecode':
         if os.getuid() != 0:
             raise RuntimeError('agent_uuid is configured to use dmidecode, '


keylime/keylime_agent.py
```diff
@@ -450,10 +450,6 @@ def decrypt_check(self, decrypted_U, decrypted_V):
 
 
 def main():
-    if os.getuid() != 0 and config.REQUIRE_ROOT:
```
stefanberger, line 453: This process needs to run as root afaik due to these here:

```
$ ls -l /sys/kernel/security/ima/
total 0
-r--r-----. 1 root root 0 Mar 26 17:59 ascii_runtime_measurements
-r--r-----. 1 root root 0 Mar 26 17:59 binary_runtime_measurements
-rw-------. 1 root root 0 Mar 26 18:09 policy
-r--r-----. 1 root root 0 Mar 26 17:59 runtime_measurements_count
-r--r-----. 1 root root 0 Mar 26 17:59 violations
```

stefanberger:
>     3\. We need to make sure that the non-root user which is running keylime
>         can read `/sys/kernel/security/ima/ascii_runtime_measurements` and
>         `/sys/kernel/security/tpm0/binary_bios_measurements`. This could be
>         done as simple `chown` or as an udev rule.

The tenant tool should check whether the current user can access (https://man7.org/linux/man-pages/man2/access.2.html) the files it needs rather than faulting later on when it tries to access the file but cannot.

maugustosilva:
> > ```
> > 3\. We need to make sure that the non-root user which is running keylime
> >     can read `/sys/kernel/security/ima/ascii_runtime_measurements` and
> >     `/sys/kernel/security/tpm0/binary_bios_measurements`. This could be
> >     done as simple `chown` or as an udev rule.
> > ```
> 
> 
> The tenant tool should check whether the current user can access (https://man7.org/linux/man-pages/man2/access.2.html) the files it needs rather than faulting later on when it tries to access the file but cannot.

As I discussed with @mpeters, the steps outlined on my PR are a "packaging issue", and do not belong in the main keylime code. I've opened a PR on `keylime-deb` repository with a "postinst" script that takes care of these tasks (still refining it). In addition to that, we cannot know beforehand if the paths in `/sys/kernel/security` will even be accessed (it depends on the parameters passed by `keylime_tenant` CLI), and should not prevent the `agent` from starting if these are not readable.

maugustosilva:
@mpeters @lukehinds I would like your opinion: should I include a small script that sets up the whole environment (mount `/var/lib/keylime/secure` as root and then changes ownership of everything, including `/sys/kernel/securiy/{ima,tpm0}` to a `keylime` user) as part of this PR?

stefanberger:
> As I discussed with @mpeters, the steps outlined on my PR are a "packaging issue", and do not belong in the main keylime code. I've opened a PR on `keylime-deb` repository with a "postinst" script that takes care of these tasks (still refining it). In addition to that, we cannot know beforehand if the paths in `/sys/kernel/security` will even be accessed (it depends on the parameters passed by `keylime_tenant` CLI), and should not prevent the `agent` from starting if these are not readable.

Are you going to adapt `installer.sh` as well?

lukehinds:
I might be a little late to raise this point now, but if there a real need to run the agent as a non-root user? I understand it makes sense with the verifier / registrar.

stefanberger:
> I might be a little late to raise this point now, but if there a real need to run the agent as a non-root user? I understand it makes sense with the verifier / registrar.

It would definitely be better than running it with high privileges.

stefanberger:
> As I discussed with @mpeters, the steps outlined on my PR are a "packaging issue", and do not belong in the main keylime code. I've opened a PR on `keylime-deb` repository with a "postinst" script that takes care of these tasks (still refining it). In addition to that, we cannot know beforehand if the paths in `/sys/kernel/security` will even be accessed (it depends on the parameters passed by `keylime_tenant` CLI), and should not prevent the `agent` from starting if these are not readable.

Also... if your postinst script is correct then the access() checks on all files will succeed and everything will work as expected. If someone doesn't have the system setup this way an error message would be appropriate. Just my $0.02.

kaifeng:
> I might be a little late to raise this point now, but if there a real need to run the agent as a non-root user? I understand it makes sense with the verifier / registrar.

I concur, Agent running as non-root is not as necessary as verifier/verifier, the initial proposal is mainly to remove different logic regarding to root/non-root users which complicates use cases.
Since agent is responsible for collecting necessary system states, this unavoidable require more privileges.
In the long term I think we could still consider making agent running under non-root since agent could be compromised, privilege control is considered an improvement but not mandatory requirements. The end decision probably falls to the hand of deployers but not keylime itself.

maugustosilva:
> > I might be a little late to raise this point now, but if there a real need to run the agent as a non-root user? I understand it makes sense with the verifier / registrar.
> 
> I concur, Agent running as non-root is not as necessary as verifier/verifier, the initial proposal is mainly to remove different logic regarding to root/non-root users which complicates use cases.
> Since agent is responsible for collecting necessary system states, this unavoidable require more privileges.
> In the long term I think we could still consider making agent running under non-root since agent could be compromised, privilege control is considered an improvement but not mandatory requirements. The end decision probably falls to the hand of deployers but not keylime itself.

@kaifeng my rationale is this question is straightforward to the point bluntness: you and I might have our opinions on whether the agent should run as non-root, but nevertheless we should give the keylime deployer the option to run as non-root/root if he choses to do so. This PR is all about this one very simple aspect: choice.

maugustosilva:
> > As I discussed with @mpeters, the steps outlined on my PR are a "packaging issue", and do not belong in the main keylime code. I've opened a PR on `keylime-deb` repository with a "postinst" script that takes care of these tasks (still refining it). In addition to that, we cannot know beforehand if the paths in `/sys/kernel/security` will even be accessed (it depends on the parameters passed by `keylime_tenant` CLI), and should not prevent the `agent` from starting if these are not readable.
> 
> Also... if your postinst script is correct then the access() checks on all files will succeed and everything will work as expected. If someone doesn't have the system setup this way an error message would be appropriate. Just my $0.02.

@stefanberger Let’s split the difference :-) I will not prevent the agent from starting, but will output the appropriate `WARNING` messages. Would that address your concerns?

stefanberger:
> @stefanberger Let’s split the difference :-) I will not prevent the agent from starting, but will output the appropriate `WARNING` messages. Would that address your concerns?

How will it react when it needs to access one of those sysfs files and it cannot?

stefanberger:
An alternative for solving the high privilege problem would be to open the files first (which may work without high privileges if the sysfs is adjusted accordingly), then dropping privileges if run as root, and then access those files using those file descriptors.

stefanberger:
> An alternative for solving the high privilege problem would be to open the files first (which may work without high privileges if the sysfs is adjusted accordingly), then dropping privileges if run as root, and then access those files using those file descriptors.

Along those lines here's a draft PR: https://github.com/keylime/keylime/pull/626

maugustosilva:
> > @stefanberger Let’s split the difference :-) I will not prevent the agent from starting, but will output the appropriate `WARNING` messages. Would that address your concerns?
> 
> How will it react when it needs to access one of those sysfs files and it cannot?

It will fail, and the agent will quit. The user chose to ignore the `WARNING`,  and this should result in a problem

stefanberger:
> Along those lines here's a draft PR: #626

I will not persue this PR for now.

galmasi:
This PR has become a busy place, but I would like to answer @lukehinds question of "is it really necessary to run keylime as a non-root user". In the view of our "production" people this is absolutely necessary, maybe even more so than running the keylime verifier and registrar in unprivileged mode. Their rationale is that keylime running as root (and with an open http[s] port) is a security violation in waiting. Same argument why no sane person today runs apache as root.

mpeters:
I agree that being able to run the agent as non-root is necessary, if for no other reason like @galmasi says, it's running a web server which opens it up to outside attacks. In a system being monitored with keylime, every host would have the agent and if there were a vulnerability then every host is at risk. That's a pretty big risk profile and we should do everything we can to diminish it with reduced privileges.

keylime/keylime_agent.py
mpeters, line 453: Shouldn't we also remove `config.REQUIRE_ROOT` ?

mpeters:
One more note @maugustosilva, can we change the first line in the PR? In git the first line is considered the "title" of the commit and is used in lots of places like `git blame`, etc. And if someone is looking at things in the future they won't be able to tell much by just "Fixes bug #X" especially because in the future we might use a different bug tracking system. It's one thing if the link in the comment doesn't work, but another if the title is useless.

So something like "Remove requirement to run as root" or similar.

agent_non_root_condition.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+
+KEYLIME_GROUP=tss
```
mpeters, line 3: Can we change this to set only if not previously set? That would allow different packagers to use different users and groups if they'd like. I imagine most will use "keylime" as the user, but might have different groups.

Something like:

```
KEYLIME_GROUP=${KEYLIME_GROUP:-'tss'}
KEYLIME_USER=${KEYLIME_USER:-'keylime'}

File: agent_non_root_condition.sh [hunk 3]
```diff
@@ -0,0 +1,34 @@
...
+KEYLIME_USER=keylime
+
+# creating group if it isn't already there
+if ! getent group $KEYLIME_GROUP >/dev/null; then
+    addgroup --system $KEYLIME_GROUP
+fi
+
+# creating keylime user if he isn't already there
+if ! getent passwd keylime >/dev/null; then
+    adduser --system --ingroup tss --shell /bin/false \
```
mpeters, line 13: I think `--ingroup tss` should be `--ingroup ${KEYLIME_GROUP}` right?

mpeters (changes requested):

agent_non_root_condition.sh
maugustosilva: my mistake... just pushed a fix for it

mpeters:
> One more note @maugustosilva, can we change the first line in the PR? 

You can ignore this for this PR. Since it's multiple commits I'll be squashing it anyway and then I can fix the subject.

agent_non_root_condition.sh
maugustosilva: absolutely, just did it exactly it.

mpeters (approved):
lgtm

PR #612, opened 2021-04-07
WIP: Some cleanups

stefanberger:
This patch fixes some nits:
- use '|' rather than '+' to build a mask of bits (assuming sanity that if bit n is set twice it is not supposed to lead to bit n+1 to be set as was the case with the '+') 
- explicitly test for tpm2_tools version 4.0 and 4.2 in else branches to be consistent with other cases where the same testing is done in many places in tpm_main.py already. This may be the pessimistic approach that future versions of the tpm2_tools may again be incompatible and need another else branch. The optimistic approach would be to adjust the other cases where we explicitly test for 4.0 and 4.2 and use a simple else instead.
- removes the unused read_measurement_list_bin function attempting to parse the binary measurement list. Is someone persuing parsing the binary measurement list or can we stick with the existing ascii list? If nobody is persuing this then let's remove it and in the worst case we can find it in the history later on.

### DIFF
--- a/keylime/ca_util.py
+++ b/keylime/ca_util.py
@@ -63,11 +63,6 @@ else:
 global_password = None
 
 
-# def globalcb(*args):
-#    global global_password
-#    return global_password.encode()
-
-
 def setpassword(pw):
     global global_password
     if len(pw) == 0:
--- a/keylime/ima.py
+++ b/keylime/ima.py
@@ -80,45 +80,6 @@ supported_fields = {
 }
 
 
-def read_measurement_list_bin(path):
-    raise Exception("not implementated fully yet")
-    # pylint: disable=W0101
-    f = open(path, 'rb')
-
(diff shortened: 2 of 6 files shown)


mpeters (approved):
lgtm

stefanberger:
Sorry, was still WIP. Need to revert part of it. Sorry.

stefanberger:
> Sorry, was still WIP. Need to revert part of it. Sorry.

Actually no, the bug I am seeing seems to be elsewhere .

mpeters:
Oh, sorry, I was doing a lot of reviews and I completely missed the WIP part... my bad. They seemed like useful, harmless cleanups.

stefanberger:
@mpeters I think the issue is not coming from this series but from what was merged on top of it. See keylime channel on slack.

PR #614, opened 2021-04-07
installer: Only install efi library on x86_64 systems

stefanberger:


mpeters:
Does this mean we can only do measured boot on backends that are 64bit? Probably not a bit deal, but seems like we should document it somewhere...

stefanberger:
The library only exists on x86_64 and possibly ARM, not on s390 or ppc64.

mpeters (approved):
lgtm

PR #615, opened 2021-04-07
installer: Add libuuid-devel/uuid-dev to BUILD_TOOLS for tpm2_tools b…

stefanberger:
…uild

If we need to build tpm2_tools we need libuuid-devel/uuid-dev installed:

https://github.com/tpm2-software/tpm2-tools/blob/master/configure.ac#L63


mpeters (approved):
lgtm

PR #617, opened 2021-04-08
installer: Add json-c-devel/json-c-dev to BUILD_TOOLS for tpm2-tss build

stefanberger:
Add json-c-devel/json-c-dev in case we need to compile tpm2-tss:

https://github.com/tpm2-software/tpm2-tss/blob/master/configure.ac#L157


mpeters (approved):
lgtm

PR #621, opened 2021-04-09
Flush tpm keys upon registration failure - Fixes #620

### LINKED ISSUE #620: Whenever Keylime `agent` starts and cannot contact the `registrar`, it fails and quits *without* flushing create EK handles
# Environment
 * OS / version: Fedora Core 34
 * TPM Manufacturer: Infineon
 * Keylime version: 6.1.0

# Description 
- `agent` starts and creates a new EK and a new AIK
- `agent` attempts to contact the `registrar` and is unable to
- `agent` quits without flushing the just created EK/AIK
- After going through this process several subsequent occasions, TPM device runs out of storage space for new AIKs

- The `agent` should proceed to a TPM cleanup (`instance_tpm.flush_keys()`) in case of failure to contact the `registrar`

maugustosilva:
The simplest fix is to just call instance_tpm.flush_keys() whenever the
agent fails during the initial setup

### DIFF
--- a/keylime/keylime_agent.py
+++ b/keylime/keylime_agent.py
@@ -526,12 +526,14 @@ def main():
         registrar_ip, registrar_port, agent_uuid, ek_tpm, ekcert, aik_tpm)
 
     if keyblob is None:
+        instance_tpm.flush_keys()
         raise Exception("Registration failed")
 
     # get the ephemeral registrar key
     key = instance_tpm.activate_identity(keyblob)
 
     if key is None:
+        instance_tpm.flush_keys()
         raise Exception("Activation failed")
 
     # tell the registrar server we know the key
@@ -540,6 +542,7 @@ def main():
         registrar_ip, registrar_port, agent_uuid, key)
 
     if not retval:
+        instance_tpm.flush_keys()
         raise Exception("Registration failed on activate")
 
     serveraddr = (config.get('cloud_agent', 'cloudagent_ip'),


kaifeng (commented):
I think it would be good to handle this at a single place so we don't need to process at each branch point, how about wrapping main with the exception handling and doing it there?

kaifeng (commented):
I think it would be good to handle this at a single place so we don't need to process at each branch point, how about wrapping main with the exception handling and doing it there?

galmasi:
I do not disagree with the sentiment, but it might require pervasive changes to the code. We are in main() and registration failure is raising exceptions that will not have any try/catch around them. 

One solution I can imagine is 
* put the whole registration process (between creating the registration client and successful activation) into a try/catch block
* the exception handler can log/print the error, flush TPM handles and then call exit(1) -- since registration/activation failed, we are kind of done running the agent. Might as well exist semi-gracefully.

But that will certainly be > 3 LOC change.

galmasi:
Also confirming that fix_620 works. Specifically, on a cluster where the keylime services are deliberately left unstarted and the patch for `fix_620` is applied, the TPM device's handle space has stopped filling up. The keylime agent service can now attempt indefinite restarts up until the registrar starts accepting requests.

mpeters:
This looks good. My only complaint is the commit title (first line). We should mention that it fixes bug #620 in the text of the commit, but the first line should be something like "Flush tpm keys upon registration failure" or something similar.

mpeters (approved):
lgtm

Issue #584, opened 2021-03-14
agent activation issue: Unable to run tpm2_policysecret


stefanberger:
# Environment
 * OS / version: Fedora 32
 * Processor architecture: PowerPC64
 * TPM Manufacturer: Nuvoton
 * Keylime version: git master (1766957a748)

# Description 

Start of keylime_agent at some point results in the following issue. Typically a reboot resolves the issue but I don't know whether this requires a reboot. It could be an issue with the TPM 2 itself or an issue related to some cleanup of the TPM 2 not being done.

# keylime_agent              
Using config file /etc/keylime.conf
2021-03-14 18:37:08.880 - keylime.tpm - INFO - TPM2-TOOLS Version: 4.1.3
2021-03-14 18:37:08.995 - keylime.tpm - INFO - TPM2-TOOLS Version: 4.1.3
2021-03-14 18:37:09.514 - keylime.tpm - INFO - Taking ownership with config provided TPM owner password
2021-03-14 18:37:11.906 - keylime.tpm - INFO - TPM Owner password confirmed: keylime
2021-03-14 18:37:11.906 - keylime.tpm - INFO - Flushing old ek handle: 0x81000001
2021-03-14 18:37:18.230 - keylime.tpm - INFO - Flushing old ak handle: /var/lib/keylime/secure/tmp_1g2t4nm
2021-03-14 18:37:23.506 - keylime.tpm - WARNING - TPM randomness not added to system entropy pool: [Errno 22] Invalid argument
2021-03-14 18:37:23.506 - keylime.cloudagent - INFO - Agent UUID: D432FBB3-D2F1-4A97-9EF7-75BD81C00000
2021-03-14 18:37:23.710 - keylime.registrar_client - INFO - Agent registration requested for D432FBB3-D2F1-4A97-9EF7-75BD81C00000
2021-03-14 18:37:23.899 - keylime.tpm - ERROR - Error decrypting AIK: Command: ['tpm2_policysecret', '-S', '/var/lib/keylime/secure/tmp8_tw5j1u', '-c', '0x4000000B', 'keylime'] returned 1, expected 0, output [], stderr [b'ERROR: Could not read session type\n', b'ERROR: Unable to run tpm2_policysecret\n']
[...]

PR #624, opened 2021-04-11
tpm_main: Invoke tpm2_evictcontrol for 4.0 and 4.2 tools if aik_handl…

stefanberger:
…e exists

tpm2_evictcontrol was not invoked for tpm2_tools 4.0 or 4.2 since aik_handle
is not a hex number but a filename and that filename could not be found in
the 'outjson' since there we only find hex numbers. So, adjust the test for
4.0 and 4.2 tools to check whether the aik_handle file exists and invoke
tpm2_evictcontrol if it is there.

Also remove the old aik_handle file after running tpm2_evictcontrol. This
ensures that only one file in /var/lib/keylime/secure is an aik_handle. This
now resolves issue #584 for me but the mystery remains why multiple files in
that directory would have caused an issue for running tpm2_policysecret
while removing them, as pointed out in issue #584, resolved the issue.
Fact is that the following command now runs for hours:

while :; do timeout -s SIGINT 60 keylime_agent ; done

### DIFF
--- a/keylime/tpm/tpm_main.py
+++ b/keylime/tpm/tpm_main.py
@@ -594,13 +594,18 @@ class tpm(tpm_abstract.AbstractTPM):
                 output = [output]
 
             outjson = config.yaml_to_dict(output)
-            if outjson is not None and aik_handle in outjson:
+            if self.tools_version == "3.2":
+                evict_it = outjson is not None and aik_handle in outjson
+            elif self.tools_version in ["4.0", "4.2"]:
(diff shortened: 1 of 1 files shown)


stefanberger:
I tested this now inside a VM. It should be good to go.

maugustosilva:
This looks good, I appreciate that @stefanberger took the time to track down this annoying problem. LGTM

mpeters (approved):
lgtm

PR #625, opened 2021-04-12
Include PCR16 in the quote only when needed

kaifeng:
Currently the PCR16 is included in the quote unconditionally, but
it only makes sense when the data is not None.


nabilschear (approved):

mpeters (approved):
lgtm

PR #626, opened 2021-04-12
RFC: Drop privileges of keylime_agent when run as root

stefanberger:
This PR enables the keylime agent to drop privileges by switching to another user.

### DIFF
--- a/keylime.conf
+++ b/keylime.conf
@@ -124,6 +124,10 @@ tpm_signing_alg = rsassa
 # create a new EK upon startup, and neither will it flush the EK upon exit.
 ek_handle = generate
 
+# The user account to switch to to drop privileges when started as root
+# If left empty, the agent will keep running with high privileges.
+run_as = nobody:nobody
+
 #=============================================================================
 [cloud_verifier]
 #=============================================================================
--- a/keylime/config.py
+++ b/keylime/config.py
@@ -155,7 +155,9 @@ def get_config():
     return _CURRENT_CONFIG
 
 
-def get(section, option):
+def get(section, option, fallback=None):
+    if fallback is not None:
+        return get_config().get(section, option, fallback=fallback)
     return get_config().get(section, option)
 
 
--- a/keylime/keylime_agent.py
+++ b/keylime/keylime_agent.py
@@ -144,13 +144,9 @@ class Handler(BaseHTTPRequestHandler):
             # other possibilities would include adding additional data to rest_params to trigger boot log retrievals
             # generally speaking, retrieving the 15Kbytes of a boot log does not seem significant compared to the
             # potential Mbytes of an IMA measurement list.
-            if TPM_Utilities.check_mask(imaMask, config.MEASUREDBOOT_PCRS[0]):
-                if not os.path.exists(config.MEASUREDBOOT_ML):
-                    logger.warning("TPM2 event log not available: %s", config.MEASUREDBOOT_ML)
-                else:
-                    with open(config.MEASUREDBOOT_ML, 'rb') as f:
-                        el = base64.b64encode(f.read())
-                    response['mb_measurement_list'] = el
+            if TPM_Utilities.check_mask(imaMask, config.MEASUREDBOOT_PCRS[0]) and self.server.tpm_log_fd >= 0:
+                os.lseek(self.server.tpm_log_fd, 0, os.SEEK_SET)
+                response['mb_measurement_list'] = base64.b64encode(self.server.tpm_log_file.read())
 
             config.echo_json_response(self, 200, "Success", response)
             logger.info('GET %s quote returning 200 response.', rest_params["quotes"])
@@ -333,7 +329,7 @@ class CloudAgentHTTPServer(ThreadingMixIn, HTTPServer):
     final_U = None
     agent_uuid = None
 
-    def __init__(self, server_address, RequestHandlerClass, agent_uuid, ima_log_fd):
+    def __init__(self, server_address, RequestHandlerClass, agent_uuid, ima_log_fd, tpm_log_fd):
         """Constructor overridden to provide ability to pass configuration arguments to the server"""
         secdir = secure_mount.mount()
         keyname = "%s/%s" % (secdir, config.get('cloud_agent', 'rsa_keyname'))
@@ -367,6 +363,11 @@ class CloudAgentHTTPServer(ThreadingMixIn, HTTPServer):
         self.ima_log_file = None
         if ima_log_fd >= 0:
             self.ima_log_file = os.fdopen(self.ima_log_fd)
+        # we'll use tpm_log_fd to seek() and tpm_log_file to read()
+        self.tpm_log_fd = tpm_log_fd
+        self.tpm_log_file = None
+        if self.tpm_log_fd >= 0:
+            self.tpm_log_file = os.fdopen(self.tpm_log_fd)
 
     def add_U(self, u):
         """Threadsafe method for adding a U value received from the Tenant
@@ -528,6 +529,12 @@ def main():
     else:
         ima_log_fd = os.open(config.IMA_ML, os.O_RDONLY)
 
+    tpm_log_fd = -1
+    if not os.path.exists(config.MEASUREDBOOT_ML):
+        logger.warning("TPM2 event log not available: %s", config.MEASUREDBOOT_ML)
+    else:
+        tpm_log_fd = os.open(config.MEASUREDBOOT_ML, os.O_RDONLY)
+
     # register it and get back a blob
     keyblob = registrar_client.doRegisterAgent(
         registrar_ip, registrar_port, agent_uuid, ek_tpm, ekcert, aik_tpm)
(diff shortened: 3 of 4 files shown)


stefanberger:
After running the following command:

```
sudo chown -R nobody:nobody /var/lib/keylime/
```

We now can run keylime_agent and have it drop privileges to nobody:nobody on the way:

```
# keylime_agent
2021-04-12 17:15:24.675 - keylime.tpm - INFO - TPM2-TOOLS Version: 4.3.0
2021-04-12 17:15:24.754 - keylime.tpm - INFO - TPM2-TOOLS Version: 4.3.0
2021-04-12 17:15:24.763 - keylime.cloudagent - INFO - Agent UUID: D432FBB3-D2F1-4A97-9EF7-75BD81C00001
2021-04-12 17:15:24.768 - keylime.cloudagent - INFO - Dropped privileges to nobody:nobody
2021-04-12 17:15:24.799 - keylime.tpm - WARNING - INSECURE: Keylime is using a software TPM emulator rather than a real hardware TPM.
2021-04-12 17:15:24.800 - keylime.tpm - WARNING - INSECURE: The security of Keylime is NOT linked to a hardware root of trust.
2021-04-12 17:15:24.801 - keylime.tpm - WARNING - INSECURE: Only use Keylime in this mode for testing or debugging purposes.
2021-04-12 17:15:24.801 - keylime.tpm - INFO - Taking ownership with config provided TPM owner password
2021-04-12 17:15:24.969 - keylime.tpm - WARNING - Creating tpm metadata file without root. Sensitive trust roots may be at risk!
2021-04-12 17:15:24.972 - keylime.tpm - INFO - TPM Owner password confirmed: keylime
2021-04-12 17:15:24.973 - keylime.tpm - INFO - Flushing old ek handle: 0x8100000e
2021-04-12 17:15:25.363 - keylime.tpm - WARNING - Creating tpm metadata file without root. Sensitive trust roots may be at risk!
2021-04-12 17:15:25.365 - keylime.tpm - WARNING - Creating tpm metadata file without root. Sensitive trust roots may be at risk!
2021-04-12 17:15:25.367 - keylime.tpm - WARNING - Creating tpm metadata file without root. Sensitive trust roots may be at risk!
2021-04-12 17:15:25.369 - keylime.tpm - WARNING - Creating tpm metadata file without root. Sensitive trust roots may be at risk!
2021-04-12 17:15:25.400 - keylime.tpm - WARNING - Creating tpm metadata file without root. Sensitive trust roots may be at risk!
2021-04-12 17:15:25.510 - keylime.tpm - WARNING - Creating tpm metadata file without root. Sensitive trust roots may be at risk!
2021-04-12 17:15:25.513 - keylime.tpm - INFO - Flushing old ak handle: /var/lib/keylime/secure/tmpiq6auxl6
2021-04-12 17:15:25.887 - keylime.tpm - WARNING - Creating tpm metadata file without root. Sensitive trust roots may be at risk!
2021-04-12 17:15:25.889 - keylime.tpm - WARNING - Creating tpm metadata file without root. Sensitive trust roots may be at risk!
2021-04-12 17:15:25.891 - keylime.tpm - WARNING - Creating tpm metadata file without root. Sensitive trust roots may be at risk!
2021-04-12 17:15:25.922 - keylime.tpm - WARNING - TPM randomness not added to system entropy pool: [Errno 1] Operation not permitted
2021-04-12 17:15:27.214 - keylime.registrar_client - INFO - Agent registration requested for D432FBB3-D2F1-4A97-9EF7-75BD81C00001
2021-04-12 17:15:27.417 - keylime.tpm - INFO - AIK activated.
2021-04-12 17:15:27.846 - keylime.registrar_client - INFO - Registration activated for agent D432FBB3-D2F1-4A97-9EF7-75BD81C00001.
2021-04-12 17:15:27.961 - keylime.cloudagent - INFO - Starting Cloud Agent on 127.0.0.1:9002 use <Ctrl-C> to stop
2021-04-12 17:15:27.964 - keylime.revocation_notifier - INFO - Waiting for revocation messages on 0mq 127.0.0.1:8992
2021-04-12 17:15:40.064 - keylime.cloudagent - INFO - TERM Signal received, shutting down...
```

stefanberger:
We can now successfully drop privileges to `nobody:nobody` for example. keylime_agent now runs like this:

```
# keylime_agent
2021-04-13 09:49:09.552 - keylime.tpm - INFO - TPM2-TOOLS Version: 4.3.0
2021-04-13 09:49:09.631 - keylime.tpm - INFO - TPM2-TOOLS Version: 4.3.0
2021-04-13 09:49:09.671 - keylime.tpm - WARNING - INSECURE: Keylime is using a software TPM emulator rather than a real hardware TPM.
2021-04-13 09:49:09.672 - keylime.tpm - WARNING - INSECURE: The security of Keylime is NOT linked to a hardware root of trust.
2021-04-13 09:49:09.672 - keylime.tpm - WARNING - INSECURE: Only use Keylime in this mode for testing or debugging purposes.
2021-04-13 09:49:09.673 - keylime.tpm - INFO - Taking ownership with config provided TPM owner password
2021-04-13 09:49:09.857 - keylime.tpm - INFO - TPM Owner password confirmed: keylime
2021-04-13 09:49:09.857 - keylime.tpm - INFO - Flushing old ek handle: 0x81000000
2021-04-13 09:49:10.389 - keylime.tpm - INFO - Flushing old ak handle: /var/lib/keylime/secure/tmp6jkwltd0
2021-04-13 09:49:10.780 - keylime.cloudagent - INFO - Agent UUID: D432FBB3-D2F1-4A97-9EF7-75BD81C00001
2021-04-13 09:49:10.788 - keylime.cloudagent - INFO - Dropped privileges to nobody:nobody
2021-04-13 09:49:11.798 - keylime.registrar_client - INFO - Agent registration requested for D432FBB3-D2F1-4A97-9EF7-75BD81C00001
2021-04-13 09:49:12.007 - keylime.tpm - INFO - AIK activated.
2021-04-13 09:49:12.308 - keylime.registrar_client - INFO - Registration activated for agent D432FBB3-D2F1-4A97-9EF7-75BD81C00001.
2021-04-13 09:49:12.423 - keylime.cloudagent - INFO - Starting Cloud Agent on 127.0.0.1:9002 use <Ctrl-C> to stop
2021-04-13 09:49:12.426 - keylime.revocation_notifier - INFO - Waiting for revocation messages on 0mq 127.0.0.1:8992
```

```
$ ps aux | grep agent | grep -v grep
nobody      6864  0.3  1.1 225260 46152 pts/0    Sl+  09:49   0:00 /usr/bin/python3 /usr/local/bin/keylime_agent
```

This still needs run-time testing to make sure it doesn't blow up later on.

keylime/utils.py
```diff
@@ -0,0 +1,88 @@
+#!/usr/bin/python3
```
mpeters, line 1: Pretty nitpicky, but maybe rename this file to user_utils. In general, I find things just named "utils" leads to poor code organization in the future as people just dump things in there without really thinking about where they should really belong.

keylime.conf
```diff
@@ -124,6 +124,9 @@ tpm_signing_alg = rsassa
 # create a new EK upon startup, and neither will it flush the EK upon exit.
 ek_handle = generate
 
+# The user account to switch to to drop privileges when started as root
+runas = nobody:nobody
```
mpeters, line 128: Maybe change this to `run_as` so it's clearer it's 2 words.

stefanberger: Fixed.
I noticed that if this field is missing the config.get() will throw an exception. Are we supposed to surround this with try/except considering how an RPM may handle the update of a config file (I think it won't overwrite it)?

keylime/utils.py
stefanberger: Renamed.

keylime.conf
mpeters: That's a good question. I don't think Keylime has been consistent in the past when it's added new configuration directives, but I think going forward we need to be better. If we are adding a new directive, unless absolutely necessary it should have a reasonable default and not be required. So I would say if it's now set or empty, just keep running as root.

stefanberger: > That's a good question. I don't think Keylime has been consistent in the past when it's added new configuration directives, but I think going forward we need to be better. If we are adding a new directive, unless absolutely necessary it should have a reasonable default and not be required. So I would say if it's now set or empty, just keep running as root.

I now modified config's get() function to take a fallback parameter that I think we should always give with any new option we are introducing. The docs call it fallback parameter. I would prefer default parameter... 

Also just recently again with some webapp changes I think the webapp errorred out if it didn't have an up-to-date keylime.conf file .

stefanberger:
Ok, so I quote-tested this now with quotes on the IMA measurement log.  This works and it should work with the TPM measurement list as well, even though I did not test it.

The seek() that is now being used to reset the position for reading the IMA measurement log could come in handy for incremental attestation where the verifier would send the seek position to the agent to read from the last position and only send the measurement log from that position on rather than the whole 10k entries. For this to work, though, we would need the verifier to become stateful in relation to walking the list of PCRs and remembering what the last state of the PCRs was. Stateful here doesn't mean having to write this into the DB but remembering while running. The startup of the verifier may be expensive though since it would start with position 0 for every client[, even more expensive if we were to write the measurement list into a DB or so].

Issue #636, opened 2021-04-22
keylime_webapp front-end no pagination, default update interval too small, outdated Ajax syntax


MartinMa28:
Is your an issue a feature request? If so, please raise it as an [enhancement](https://github.com/keylime/enhancements)

# Environment
 * OS / version: 
 * Processor architecture: 
 * TPM Manufacturer: 
 * Keylime version: 

# Link to PR

#637 

# Description 
This issue list a few existing problems in keylime_webapp front-end:

1. No pagination. It's hard to go through more than 20 agents. And, a high-level overview of the system can be helpful. For example, having a pie-chart to show the percentage of agents in each state.
2. The default update interval is way too small. Based on our test in OpenStack MOC cluster, the maximum throughput of keylime_verifier is around 12~13 requests/sec (verifier has 8 vCPUs, 16 GB RAM). So, when having 10 agents, calling keylime_verifier every 750ms is really close to the max throughput.
3. XMLHttpRequest for Ajax: Javascript fetch API has been out since 2015. After 2017, it's well supported by browsers. It's the modern way to do Ajax stuff.

4. Not using tornado template: This one is a little rougher than the first 3 above. Currently, keylime_webapp (the only GUI in keylime) works in a 'brutal' way. When a browser wants to get an HTML page, keylime_webapp back-end writes a gigantic multi-line string [code link](https://github.com/keylime/keylime/blob/8da88eb479a8edd147d03c6c41cc16aaa5050cf6/keylime/tenant_webapp.py#L125) containing all of HTML tags and a few JS scripts. Although it works surprisingly well in our tests, I've never seen any website or visualization project send HTML to the browser by returning a huge string in production. Sure, the original author understood how it works under the hood pretty well. To be honest, it's not fun to write/read front-end logic in a string. No offense to anyone, seems like keylime_webapp hasn't been touched for years.

There are two common ways to do this kind of thing. 1) Use the framework template. Specifically, we could use tornado template. It's kind of like plain HTML files but with some special syntax to let developers insert dynamic content to the web page. 2) Use a front-end framework like React, Vue, Angular. Using a front-end framework goes overkill a little bit for keylime in my opinion since we only have one page. It shines when there are a lot of reusable components that need to be rendered on multiple pages. But, using a front-end framework can make keylime fully RESTful. No more HTML stuff in keylime back-end. The GUI can be separated into another repo or sub-repo to keylime.

[...]

PR #637, opened 2021-04-22
New webapp GUI

MartinMa28:
Thanks for [Chencheng](https://github.com/Gcccccccc) and [Fengqi](https://github.com/FengqiQiao), in the new front-end GUI, we have the following features:
1. Have a pie chart to show the overview of the system.
2. The selected agents are paginated.
3. The requests are staggered into smaller batches to avoid using larger intervals when more and more agents are registered. Within each interval, only a batch of requests is sent to webapp back-end so that the throughput is manageable.
4. Use JS Fetch API in pie chart and pagination. The remaining legacy code still uses XMLHttpRequest API.

The way of HTML rendering is not changed because it requires significant changes in keylime's building process.

Issue:
 #636 

webapp_failed
webapp_get_quote
webapp_registered

Signed-off-by: MartinMa28 <1135485226@qq.com>

### DIFF
--- a/keylime.conf
+++ b/keylime.conf
@@ -544,10 +544,20 @@ cfssl_port = 8888
 webapp_ip = 127.0.0.1
 webapp_port = 443
 
-# The intervals for populating agents, updating agents and updating terminal.
-populate_agents_interval = 4000
-update_agents_interval = 1500
-update_terminal_interval = 2000
+# The intervals for rendering chart and updating terminal.
+render_chart_interval = 5000
+update_terminal_interval = 10000
+
+# The amount of agents in each page
+page_size = 10
+
+# Instead of sending all of requests at once, webapp use a staggered way.
+# Requests are grouped into multiple batches. keylime_webapp front-end sends
+# one batch per render_chart_interval. Note that batch_size and render_chart_interval
+# together determines the expected throughput. If the overall latency keeps increasing
+# or becomes unacceptable, it's recommended to use a smaller batch_size or
+# larger render_chart_interval.
+batch_size = 10
 
 #=============================================================================
 # GLOBAL LOGGING CONFIGURATION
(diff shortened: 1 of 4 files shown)


lukehinds:
That's really cool, thank you authors.

This shows there is an interest in a GUI. I wonder if its time we restocked work on https://github.com/keylime/enhancements/blob/master/7_multi-tenancy.md

https://github.com/keylime/keylime/pull/309

This way we could add an auth system to the front of the webUI? Would you folks be interested in working on the front end aspects if a token system is on the backend?

MartinMa28:
> That's really cool, thank you authors.
> 
> This shows there is an interest in a GUI. I wonder if its time we restocked work on https://github.com/keylime/enhancements/blob/master/7_multi-tenancy.md
> 
> #309
> 
> This way we could add an auth system to the front of the webUI? Would you folks be interested in working on the front end aspects if a token system is on the backend?

Thanks for your invitation. For me, I am interested in the multi-tenancy feature, but I am afraid that we cannot spend so much time on keylime in the near future. We will start working full-time/internship this May. So, if multi-tenancy is something really masters at the moment, probably we cannot get it done in one or two months. If it's more of an enhancement to keylime open source community, I would love to look into it.

MartinMa28:
> This looks awesome, great job everyone!
> 
> Regarding the more long-term vision, as we eventually move over to using something like tornado templates as you mention, is it possible to also move over to a more open source friendly charts implementation (such as https://github.com/chartjs/Chart.js)?
> 
> The Google-based charts require a persistent connection to the internet (which might not always be available for Keylime running on a local intranet with no connectivity to the outside). It is also a violation of Google's TOS to download their charts source code and host it locally.

Sure, we have no problem moving to Chart.js. We might need a few more days to refactor since we are in the middle of something. By next Tuesday or Wednesday, the refactoring should be done.

mpeters:
@MartinMa28 any news on the refactor to Chart.js?

mpeters:
@MartinMa28, @Gcccccccc and @FengqiQiao: any news on the refactor to Chart.js?

lukehinds:
Its almost a year this has been left hanging, so I will close (it can always be re-opened if OP wishes to still work on it).

PR #648, opened 2021-04-26
Cleanup and fix error case in encryptAIK

stefanberger:
This function cleans up encryptAIK and fixes the error case where we were not returning enough parameters to the caller. The latter would cause an unpack error on the caller's side in case encryptAIK failed.


mpeters (approved):
lgtm

PR #650, opened 2021-04-27
ima_file_signatures: Extract keyidv2 from x509 certs

stefanberger:
When an x509 certificate is given, attempt to extract the keyidv2 from the
certificate's subject key identifier digest's last 4 bytes. This is safer
than recalculating it with the sha1 hash since other methods may exist.
For public and public keys extracted from private keys we have to use the
sha1 method.

Also adjust how we store the state of the keys into the database. We now
also store an array of the keyidv2's of all keys.

### DIFF
--- a/keylime/ima_file_signatures.py
+++ b/keylime/ima_file_signatures.py
@@ -134,9 +134,13 @@ class ImaKeyring:
         keydigest = digest.finalize()
         return int.from_bytes(keydigest[16:], 'big')
 
-    def add_pubkey(self, pubkey):
-        """ Add a public key object to the keyring. """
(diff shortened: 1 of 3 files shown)


maugustosilva:
@mpeters  @stefanberger This PR is "broken" according with my new "private" end-to-end CI/CD, but I firmly believe this is due to a bug  already fixed in PR #669. If you rebase it, I will be happy to try it again.

stefanberger:
Thanks for letting me know. Rebased.

maugustosilva:
+1 LGTM

mpeters (approved):

PR #649, opened 2021-04-26
installer.sh: Do not delete the local copy of the certificates.

kgold2:
Copy, rather than move, the certificates to
/var/lib/keylime/tpm_cert_store/.  This way, git status
does not show that all the certificates have been deleted.

Signed-off-by: Ken Goldman <kgold@linux.ibm.com>

PR #651, opened 2021-04-27
 installer: Fix back level Intel TSS configure issues for Ubuntu groovy

kgold2:
The first configure issue was reported here.  The recommendation
was to try 2.4.x, as 2.0.x is no longer supported.  This should
be tested against keylime 'next' builds for compatibility.

https://github.com/tpm2-software/tpm2-tss#2066

The second configure issue was reported here.  The recommendation was
to disable building of the man pages, because 2.4.x expects an older
version of doxygen.

https://github.com/tpm2-software/tpm2-tss#2067

### DIFF
--- a/installer.sh
+++ b/installer.sh
@@ -605,7 +605,7 @@ echo "==========================================================================
 if [ ! -d "/var/lib/keylime/tpm_cert_store" ]; then
   echo "Creating new tpm_cert_store"
   mkdir -p /var/lib/keylime
-  mv $KEYLIME_DIR/tpm_cert_store /var/lib/keylime/tpm_cert_store
+  cp $KEYLIME_DIR/tpm_cert_store /var/lib/keylime/tpm_cert_store
 else
   echo "Updating existing cert store"
   cp -n $KEYLIME_DIR/tpm_cert_store/* /var/lib/keylime/tpm_cert_store/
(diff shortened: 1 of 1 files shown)


installer.sh
```diff
@@ -605,7 +605,7 @@ echo "==========================================================================
 if [ ! -d "/var/lib/keylime/tpm_cert_store" ]; then
   echo "Creating new tpm_cert_store"
   mkdir -p /var/lib/keylime
-  mv $KEYLIME_DIR/tpm_cert_store /var/lib/keylime/tpm_cert_store
+  cp $KEYLIME_DIR/tpm_cert_store /var/lib/keylime/tpm_cert_store
```
stefanberger, line 608: This should be `cp -r`.

stefanberger:
Aslo see the `cp -r` issue report in https://github.com/keylime/keylime/pull/649

installer.sh
mpeters: @kgold2 can you update this PR with the `-r` option to `cp`?

mpeters (changes requested):

installer.sh
mpeters: Actually this has been fixed in another PR. @kgold2 can you rebase this commit off of master so your specific fixes can be applied?

mpeters (changes requested):

installer.sh
stefanberger:
```
git fetch --all
git rebase origin/master
```

THS-on:
@mpeters I think is can be closed because most issues that this PR addresses are fixed in the current installer.sh script.

PR #652, opened 2021-04-28
tpm_abstract: Fix ioctl number for adding entropy on x86_64

stefanberger:
The previous change fixed the invalid argument error for the ioctl to add
entropy to /dev/random on ppc64le but broke it on x86_64 since the ioctl
numbers are architecture-dependent. This patch fixes it now for x86_64 and
ppc64le and skips all other systems that have an unknown ioctl number.

### DIFF
--- a/keylime/tpm/tpm_abstract.py
+++ b/keylime/tpm/tpm_abstract.py
@@ -320,12 +320,20 @@ class AbstractTPM(metaclass=ABCMeta):
 
     # tpm_random
     def init_system_rand(self):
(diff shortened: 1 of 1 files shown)


stefanberger:
I apologize for the noise but didn't know the ioctl numbers were architecture-dependent...

A good alternative for this particular ioctl may be to remove the code entirely and rely on the Linux kernel mixing various entropy sources -- unless this ioctl is a selling point to run the agent on every system :-)

stefanberger:
Please consider #653 as the alternative.

PR #653, opened 2021-04-28
Remove adding entropy to system rng

stefanberger:
Trust the Linux kernel that it gets entropy from various sources.

Recent Fedora kernels for example are compiled with CONFIG_HW_RANDOM_TPM
that add the TPM as an entropy provider:

https://elixir.bootlin.com/linux/v5.12/source/drivers/char/tpm/Kconfig#L30


kaifeng (approved):
lgtm

PR #654, opened 2021-04-28
Fixes and improvements for database configuration

maugustosilva:
- Implements all the changes and improvements discussed at length during
  PR 514, making it obsolete.
- Add the option to specify, via `keylime.conf`, limits for database
  connection pool size. Absolutely required for large-scale deployments.

### DIFF
--- a/keylime.conf
+++ b/keylime.conf
@@ -198,23 +198,25 @@ registrar_private_key_pw = default
 # Database URL Configuration
 # See this document https://keylime-docs.readthedocs.io/en/latest/installation.html#database-support
 # for instructions on using different database configurations.
-# Database configuration via drivername, username, password, etc is deprecated
-# and superseded by database_url, when database_url is set, other database
-# fields will be ignored.
+# There are two options for the specification of the database connection configuration.
+# 1) A tuple of parameters (e.g., database_drivername, database_username, database_password)
+# 2) A single string in database_url. This, when set, takes precedence over the previous option.
 # An example of database_url value for using sqlite:
 #   sqlite:////var/lib/keylime/cv_data.sqlite
 # An example of database_url value for using mysql:
 #   mysql+pymysql://keylime:keylime@keylime_db:[port]/verifier?charset=utf8
-# If database_url is not set, the deprecated default is "sqlite" and will be
+# If database_url is not set, the default is "sqlite", using method 1, and will be
 # located at "/var/lib/keylime/cv_data.sqlite".
 database_url =
-drivername = sqlite
-username = ''
-password = ''
-host = ''
-port = ''
-database = cv_data.sqlite
-query = ''
+database_drivername = sqlite
+database_username = ''
+database_password = ''
+database_host = ''
+database_port = ''
+database_name = cv_data.sqlite
+database_query = ''
+# Limits for DB connection pool size in sqlalchemy (https://docs.sqlalchemy.org/en/14/core/pooling.html#api-documentation-available-pool-implementations)
+database_pool_sz_ovfl=5,10
 
 auto_migrate_db = True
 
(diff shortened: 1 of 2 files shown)


stefanberger:
The good thing is the tests are testing this -- some are currently failing.

maugustosilva:
> The good thing is the tests are testing this -- some are currently failing.

Very insightful

keylime.conf
```diff
@@ -198,23 +198,25 @@ registrar_private_key_pw = default
...
-query = ''
+database_drivername = sqlite
+database_username = ''
+database_password = ''
+database_host = ''
+database_port = ''
+database_name = cv_data.sqlite
+database_query = ''
+# Limits for DB connection pool size in sqlalchemy (https://docs.sqlalchemy.org/en/14/core/pooling.html#api-documentation-available-pool-implementations)
+database_pool_sz_ovfl=5,10
```
stefanberger, line 219: While these may be better names for these config items, I am wondering what will happen on a system that is upgraded from the old config to the new config where these items are renamed. An rpm for example treats config files differently not overwriting the user's config, thus leading to a malfunctioning system.

mpeters: I agree. I would like to see the old config values still be valid but deprecated and printing an info message. Then we can completely remove them at a later point.

maugustosilva: fair point, will make the necessary changes

maugustosilva: I just realized that I pushed the new code with a bug. I am out of the office today, will push a fix soon

keylime/db/keylime_db.py
```diff
@@ -26,45 +27,65 @@ def make_engine(self, service):
...
+
+                url = URL(
+                    drivername=drivername,
+                    username=None,
+                    password=None,
+                    host=None,
+                    database=(database)
+                )
+                engine_args['connect_args'] = {'check_same_thread': False}
+
```
stefanberger, line 74: Does this case not work with the new parameters?

maugustosilva: Yes. This is the one covered on the automated tests, actually (valid only for `sqlite`)

keylime/db/keylime_db.py
```diff
@@ -26,45 +27,66 @@ def make_engine(self, service):
...
+            engine_args['max_overflow'] = int(m_ovfl)
+
+        else :
+            logger.info('database_url is not set, using multi-parameter database configuration options')
+
+            # This code shall be removed once we fully deprecate the old format
+            try :
+                drivername = config.get(service, 'drivername')
+                database = config.get(service, 'database')
+                logger.warning('Deprecation reminder: please add the suffix "database_" to all database-related parameters on your keylime.conf.')
```
mpeters, line 52: Minor nit: Change "on your keylime.conf" to "in your keylime.conf"

keylime/db/keylime_db.py
```diff
@@ -26,45 +27,66 @@ def make_engine(self, service):
...
+                p_n_prefix = ''
```
mpeters, line 53: This isn't necessarily a change request, just a clarification: What does "p_n_prefix" mean?

keylime/db/keylime_db.py
```diff
@@ -26,45 +27,66 @@ def make_engine(self, service):
...
+        except NoOptionError :
+            p_sz = 5
+            m_ovfl = 10
 
-        return engine
+        engine_args = {}
+
+        url = config.get(service, 'database_url')
+        if url:
+            logger.info('database_url is set, using it for establishing database connection')
```
mpeters, line 41: Minor nit: Change "for establishing" to "to establish"

keylime/db/keylime_db.py
```diff
@@ -26,45 +27,66 @@ def make_engine(self, service):
...
+            # This code shall be removed once we fully deprecate the old format
+            try :
+                drivername = config.get(service, 'drivername')
+                database = config.get(service, 'database')
+                logger.warning('Deprecation reminder: please add the suffix "database_" to all database-related parameters on your keylime.conf.')
+                p_n_prefix = ''
+            except NoOptionError :
+                drivername = config.get(service, 'database_drivername')
+                p_n_prefix = "database_"
+                database = config.get(service, p_n_prefix + 'name')
```
mpeters, line 57: This can just be `'database_name` right? We don't need the prefix var since it will never be anything else.

keylime/db/keylime_db.py
```diff
@@ -26,45 +27,66 @@ def make_engine(self, service):
...
 
+            if drivername == 'sqlite':
+                database = "%s/%s" % (config.WORK_DIR, database)
```
mpeters, line 60: Maybe this new assignment should be to a new variable named something like `database_file` instead of overwriting the original `database` var. Or the other var should be called `database_name` instead of just `database`.

mpeters (changes requested):

keylime/db/keylime_db.py
maugustosilva: p = parameter, n = name

keylime/db/keylime_db.py
maugustosilva: right, we do not need to prefix here, but since we are defining `p_n_prefix` already, I decided to illustrate its use there.

mpeters (approved):

Issue #358, opened 2020-09-21
Add support for multiple *Verifiers* in KeyLime


maugustosilva:
We have a scenario where we need to deploy a KeyLime cluster serving to relative large number of nodes, somewhere in the vicinity of 10E4. In this cluster, the *Verifier* is expected to run periodic attestation of nodes' IMA logs (in addition to the "regular" attestation). Together with a colleague, I did an initial evaluation in an "emulated" testbed, where 2,500 vTPM-equipped VMs played the role of nodes and uncovered problems with the *Verifier*, which went from TCP connection timeout to excessive CPU consumption. While some of these problems are solvable eventually, it is my argument here that "scaling out" the number of *Verifiers* will have multiple benefits, not only in scalability, but also on availability (eventually).

To demonstrate the viability of a "multi-*Verifier*" approach, We've deployed a PoC where the 2.5k where distributed across 5 *Verifiers* (and only one *Registrar*), and all the deployment problems experienced before (e.g., TCP timeouts and high CPU consumption) were no longer detected.

Based on that piece of evidence, my proposal is as a follows.

a) Allow the attribute `cloudverifier_ip`, under the section `[cloud_verifier]` to become, **optionally** a comma-separated list of IPs.
b) Provide a new "hash function" which allow an *Agent*, based on its IP to pick and communicate with one (and only one) of the *Verifiers*
c) Allow a slight change on the `keylime_tenant` CLI tool, were the explicit option `-v` (indicating the *Verifier*'s IP) can be replaced by "auto", or actually omitted (in both cases, this would mean "read the value from `keylime.conf`" file), where the same "hash function" - picking one of the *Verifiers* from the list based on the *Agent*'s IP (CLI option `-t`) - could be applied.

This enhancement would not impact the regular users/deployers of KeyLime. If the "comma-separated list" has a single element, then the "usual behavior" would follow.

Please note that I am well aware of the fact this is a fairly limited improvement, which **does not address** any Availability/Resiliency concerns: I firmly believe this will be part of a much larger discussion, but one can see immediately that the question "well, what happens when a *Verifier* dies?" is: a) exactly the same that we have prior this change and b) our new scenario makes the loss of one *Verifier* (out of "N") less catastrophic in fact.

maugustosilva:
@lukehinds @mpeters please take a look and tell me if it makes sense. I am more than happy to include it in the discussion about a full-fledged (and properly thought-through :-) ) “Scalability and Availability Enhancement”

[...]

PR #657, opened 2021-04-30
Support for multiple named verifiers sharing a single database backend.

apoorvemohan:
To improve the scalability of the Keylime verifier towards the following:
 
1. [Enhancement-22: Enhanced Scalability and High Availability for Cloud Verifier and Registrar](https://github.com/keylime/enhancements/blob/master/22-enhanced-scalability.md)
2. [Issue 358: Add support for multiple *Verifiers* in KeyLime](https://github.com/keylime/keylime/issues/358)

PR Overview:

Scalability improvement: allow multiple verifiers to share a single (external, MySQL/PostgreSQL) database

- In this PR we introduce the (optional) concept of `verifier` identifier, an arbitrary string associated to a verifier process.
- A unique identifier can be assigned to each `verifier` instance at the time of its startup through a new parameter - `cloudverifier_id` - on `keylime.conf`
- Thanks to the per-`verifier` identifier, multiple verifier instances can then share not only a single database, but even a single table (`VerifierMain`)
- Each agent is bound (i.e., "assigned") to a `verifier` process for continuous attestation
- The `agent`-`verifier` affinity is determined by the `tenant` (i.e., `keylime_tenant` CLI)
- The `tenant` can add `agents` by directly talking to a `verifier` instance through its IP address or via a load balancer that further re-directs the request to a `verifier` instance
- An `agent` is tied to a `verifier`'s unique identifier rather than its `<ip>:<port>`. This means that one can simply restart a `verifier` instance, with the same identifier, on a node with a different `<ip>:<port>`. This design decision eases the migration or re-creation instances in case of failures

- A `tenant` can request for the list of `agents`, or a specific `agent`'s status, from any `verifier` instance
- A `tenant` can also request for a "filtered" (i.e., per-`verifier`) `agent` list by specifying a `verifier` identifier
- To delete or reactivate an `agent`, the `tenant` should direct a request to the correct `verifier` instance by specifying its unique `verifier` id 
- New functionality to delete and reactivate `agent` is added to find the correct `agent`-`verifier` mapping and re-direct the delete or reactivate request accordingly. 

Signed-off-by: Apoorve Mohan


stefanberger:
Did you forget to add a database migration script?

apoorvemohan:
> Did you forget to add a database migration script?
@stefanberger Indeed I did forget to add the migration script and the test cases. Thank you for bringing it up. I will add them ASAP.

mpeters:
It's still failing the tests. We can't really invest time in doing a code review if the tests are still failing

apoorvemohan:
@mpeters As I am seeing the errors in the failing tests, I am wondering why those errors were not reported on the initial commit.  Any idea what may have caused such a behavior?

apoorvemohan:
@mpeters The tests are passing now. Apparently, it was an error in the database migration script. 
I am still puzzled though -- why the error was not reported upon the initial commit.

kaifeng:
A common practice is to use hostname as the service identifier, and this is how we are using now.
In addition to that, we have to handle the synchronization between multiple instances to avoid db contention.

keylime/tenant.py
```diff
@@ -951,7 +976,7 @@ def main(argv=sys.argv):
     """
     parser = argparse.ArgumentParser(argv[0])
     parser.add_argument('-c', '--command', action='store', dest='command', default='add',
-                        help="valid commands are add,delete,update,status,list,reactivate,regdelete. defaults to add")
+                        help="valid commands are add,delete,smartdelete,update,status,list,reactivate,smartreactivate,regdelete. defaults to add")
```
mpeters, line 979: What are these new commands "smartdelete" and "smartreactivate" do that's different from regular "delete" and "reactivate"? It seems like they are smart enough to look up the agent first to determine it's verifier's ip/port combo, but why not have the regular delete/reactivate do that? Why do we need a separate command?

keylime.conf
```diff
@@ -128,6 +128,9 @@ ek_handle = generate
 [cloud_verifier]
 #=============================================================================
 
+# Unique identifier for the each verifier instances.
+cloudverifier_id = default
```
mpeters, line 132: Is this configuration variable used by the agents or just the verifier? I know that `cloudverifier_ip` and `cloudverifier_port` are. And what happens if the value in here is different from the one that's passed to keylime_tenant?

keylime/cloud_verifier_tornado.py
```diff
@@ -289,6 +301,9 @@ def post(self):
                     agent_data['enc_alg'] = ""
                     agent_data['sign_alg'] = ""
                     agent_data['agent_id'] = agent_id
+                    agent_data['verifier_id'] = config.get('cloud_verifier', 'cloudverifier_id')
+                    agent_data['verifier_ip'] = config.get('cloud_verifier', 'cloudverifier_ip')
+                    agent_data['verifier_port'] = config.get('cloud_verifier', 'cloudverifier_port')
```
stefanberger, line 306: If someone has an RPM of keylime installed the config file that the user may have modified may not get updated and then won't have these variables. So, I think we need a variant of `config.get()` that returns a given default value if the parameter is not in the config file...

mpeters: I agree, this has come up a couple of times now.

stefanberger: I can send a patch with following the following pattern:

```
def get(section, option):
+def get(section, option, fallback=None):
+    if fallback is not None:
+        return get_config().get(section, option, fallback=fallback)
```

stefanberger: PR https://github.com/keylime/keylime/pull/674

keylime.conf
apoorvemohan: The `cloudverifier_id` variable is only required to configure a verifier instance. 
From the tenants POV, it is used as a parameter to request a filtered list of agents from any verifier instance.

keylime/tenant.py
apoorvemohan: Both `smartdelete` and `smartreactivate` require additional REST call to find the correct verifier instances` information in the case when it is unknown. In the case when the verifier's information is already available, adding this functionality to the existing `delete` and `reactivate` would have introduced unnecessary overhead.

mpeters:
> A common practice is to use hostname as the service identifier, and this is how we are using now.
> In addition to that, we have to handle the synchronization between multiple instances to avoid db contention.

That works unless all of the verifiers are behind a load-balancer, then you need some extra identifier. This PR is to support the deployment where all interactions with the REST API can be load balanced and don't really care about which verifier handles it, but when the verifiers do their asynchronous polling loops, those naturally do care about which verifier is looping over which agents.

I think this is a good short term solution. Longer term I think the verifiers can treat the DB like a simple distributed queue where a batch of agents are pulled at a time, marked as "in-progress" and polled then updated in the DB. It's definitely doable with an SQL database, but the locking semantics might be slightly different based on different engine's implementation.

keylime/cloud_verifier_tornado.py
mpeters: @apoorvemohan can you rebase your work on this (now merged) change so that it's backwards compatible for any existing keylime.conf

keylime/tenant.py
mpeters: If the information is already known, couldn't that be used as the conditional to decide if the extra "smart" API calls are needed?

keylime.conf
mpeters: So the tenant reads this variable from keylime.conf? That would be confusing as it's pretty common to run tenant commands on a machine that is also running a verifier and if you start adding more verifiers it would be strange that the tenant is only talking to one of them.

keylime/tenant.py
apoorvemohan: In that case how would the tenant know, if the provided information is actually correct i.e. how would the tenant know if it should just trust the provided information or make the "smart" API calls?

keylime.conf
apoorvemohan: No, only the verifier reads the variable from the keylime.conf as it is defined under the `[cloud_verifier]`.

maugustosilva:
> > A common practice is to use hostname as the service identifier, and this is how we are using now.
> > In addition to that, we have to handle the synchronization between multiple instances to avoid db contention.
> 
> That works unless all of the verifiers are behind a load-balancer, then you need some extra identifier. This PR is to support the deployment where all interactions with the REST API can be load balanced and don't really care about which verifier handles it, but when the verifiers do their asynchronous polling loops, those naturally do care about which verifier is looping over which agents.
> 
> I think this is a good short term solution. Longer term I think the verifiers can treat the DB like a simple distributed queue where a batch of agents are pulled at a time, marked as "in-progress" and polled then updated in the DB. It's definitely doable with an SQL database, but the locking semantics might be slightly different based on different engine's implementation.

@kaifeng This is a very important point by @mpeters. This is in line with the gradual approach we all have been taking to make `keylime` both scalable and highly available. With this PR as a (pretty decent) short term solution, we can have multiple processes sharing a single database (in our environments, we have a pool of several tens of `registrars` and `verifiers` across multiple nodes sharing the same database) and, through the use of the "verifier id", quickly restart a `verifier` process on different node to (thanks to your PRs) quickly resume the attestation work in case of failure.

keylime/cloud_verifier_tornado.py
apoorvemohan: Rebased the work with the merged changes.

maugustosilva:
@mpeters @apoorvemohan This PR passes successfully on our "private" end-to-end CI/CD
+1 LGTM

keylime/tenant.py
mpeters: I'm still a bit fuzzy on why these extra smart APIs need to exist vs just enhance the existing APIs. Can you walk me through the use case and API calls (not just the individual calls, but the whole sequence that leads to them)?

apoorvemohan: As discussed, the `smartdelete` and `smartreactivate` will be converted to the default `delete` and `reactivate`, and the original non-smart functionality will be available through a flag.

maugustosilva:
@mpeters @apoorvemohan This PR passes successfully on our "private" end-to-end CI/CD
+1 LGTM (after the latest commit to the same PR)

mpeters (approved):

keylime.conf
```diff
@@ -335,7 +338,7 @@ vtpm_policy = {"23":["ffffffffffffffffffffffffffffffffffffffff","000000000000000
 # this file is used if tenant provides "default" as the allowlist file
 # this file should be in the form
 # SHA1sum path_to_file
-allowlist = allowlist.txt
+ima_allowlist = allowlist.txt
```
mpeters, line 341: Something seems off with your git merges. This change (and others) are part of already merged changes in keylime.

mpeters:
@apoorvemohan Something seems off with the git merges as part of this. There are a lot of commits that were previous merged commits in keylime (maybe you were merging stuff from master into your branch rather than rebasing?). I've identified these commits that are wrapped up in your commit as well: ed99c0f401096d69bae3759ac568086b5a36275d, 5a613051a0ca5a8b8138f53f14a6967acfa48265, e3a7da8dbfbc4b209a4f15191b232fdda59ef4b8, f3c31b411dd3dd971fd9d614a39a150655c6797c, eaa242d66c0a937a54cadbdad0b627d51cbbb017

Would it be possible to clean up your PR so it only includes the changes you are submitting? Eg, your changes are rebased off of the current HEAD?

apoorvemohan:
@mpeters Please consider PR #677 as the cleaned up alternative.

PR #659, opened 2021-05-06
webapp: fix tls certs paths

aplanas:
`get_tls_context` search for the TLS certificates and TLS private key.
To do that uses a simple "%s/%s" to build the full path.

The problem is that if the certificates contains already the full path,
the final path will be incorrect.

This patch replace it with `os.path.join`, that respect the full path to
generate the final path.

### DIFF
--- a/keylime/tenant_webapp.py
+++ b/keylime/tenant_webapp.py
@@ -648,9 +648,9 @@ def get_tls_context():
 
     logger.info("Setting up client TLS in %s", tls_dir)
 
-    ca_path = "%s/%s" % (tls_dir, ca_cert)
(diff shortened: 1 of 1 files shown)


maugustosilva:
This PR needs to be rebased with the most current `master` (after the merging of PR #669). I was unable to properly test it due to errors caused by that (already fixed) bug.

aplanas:
rebased

maugustosilva:
This PR (after rebase) passes successfully on our "private" end-to-end CI/CD
+1 LGTM

mpeters (approved):

Issue #664, opened 2021-05-10
agent_uuid set to dmidecode crashes Keylime


jronnblom:
# Environment
 * OS / version: Ubuntu 18.04
 * Processor architecture: amd64
 * TPM Manufacturer: swtpm
 * Keylime version: 6.1.0 from 

# Description 
Change agent_uuid to dmidecode to automatically read the UUID from the system crashes the agent:

May 10 16:28:28 u18can20 keylime_agent[4687]: Traceback (most recent call last):
May 10 16:28:28 u18can20 keylime_agent[4687]:   File "/usr/bin/keylime_agent", line 11, in <module>
May 10 16:28:28 u18can20 keylime_agent[4687]:     load_entry_point('keylime==0.0.0', 'console_scripts', 'keylime_agent')()
May 10 16:28:28 u18can20 keylime_agent[4687]:   File "/usr/lib/python3/dist-packages/keylime/cmd/agent.py", line 15, in main
May 10 16:28:28 u18can20 keylime_agent[4687]:     keylime_agent.main()
May 10 16:28:28 u18can20 keylime_agent[4687]:   File "/usr/lib/python3/dist-packages/keylime/keylime_agent.py", line 507, in main
May 10 16:28:28 u18can20 keylime_agent[4687]:     sys_uuid = ret['retout'].decode('utf-8')
May 10 16:28:28 u18can20 keylime_agent[4687]: AttributeError: 'list' object has no attribute 'decode'

## Expected behavior vs. actual behavior 
To work.

## Steps to reproduce problem
1. Change agent_uuid to dmidecode in /etc/keylime.conf
2. Restart the agent

# Relevant logs

Attach any relevant log files that can help to debug your issue.

jronnblom:
I also think the sys_uuid should be check for validity before continuing.

stefanberger:
Can you test it with PR #672 ?

Current patch: https://github.com/keylime/keylime/commit/8da24928bb743794bc75b0a7bffc1235a601272d.patch

jronnblom:
That works.

> Can you test it with PR #672 ?

stefanberger:
Can you close this issue?

PR #672, opened 2021-05-12
agent: Fix the usage of dmidecode during the agent startup (issue #664)

stefanberger:
The field ret['retout'] returned from cmd_exec.run(cmd) is a list,
so we have to select the first entry of the list to get to the
value returned by dmidecode.

This patch fixes the following error as pointed out in issue #664:

File "/usr/bin/keylime_agent", line 11, in
File "/usr/lib/python3/dist-packages/keylime/cmd/agent.py", line 15, in main
keylime_agent.main()
File "/usr/lib/python3/dist-packages/keylime/keylime_agent.py", line 507, in main
sys_uuid = ret['retout'].decode('utf-8')
AttributeError: 'list' object has no attribute 'decode'

### DIFF
--- a/keylime/keylime_agent.py
+++ b/keylime/keylime_agent.py
@@ -502,8 +502,12 @@ def main():
     elif agent_uuid == 'dmidecode':
         cmd = ['dmidecode', '-s', 'system-uuid']
         ret = cmd_exec.run(cmd)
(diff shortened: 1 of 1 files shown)


kaifeng (approved):
Thanks for fixing this, we use modified version of execute by directly invoking subprocess here, so the issue doesn't exposed in our environment.

PR #673, opened 2021-05-12
corrects markdown to have proper rendering

samyak-jn:
Hi, 
while I was trying to configure keylime in my system, I was going through readme and found some minor rendering issues, have fixed in this commit. Please take a look at it, hope it helps somehow.

Thanks for working on keylime! :D

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,10 +5,10 @@ Status](https://travis-ci.org/keylime/keylime.svg?branch=master)](https://travis
 
 ![keylime](doc/keylime.png?raw=true "Title")
 
-Keylime is an open source scalable trust system harnessing TPM Technology.
+Keylime is an open-source scalable trust system harnessing TPM Technology.
 
 Keylime provides an end-to-end solution for bootstrapping hardware rooted
-cryptographic trust for remote machines, the provisioning of encrypted payloads
+cryptographic trust for remote machines, the provisioning of encrypted payloads, 
 and run-time system integrity monitoring. It also provides a flexible
 framework for the remote attestation of any given `PCR` (Platform Configuration
 Register). Users can create their own customized actions that will trigger when
@@ -18,10 +18,10 @@ Keylime's mission is to make TPM Technology easily accessible to developers and
 users alike, without the need for a deep understanding of the lower levels of a
 TPM's operations. Amongst many scenarios, it well suited to tenants who need to
 remotely attest machines not under their own full control (such as a consumer of
-hybrid cloud or a remote Edge / IoT device in a insecure physical tamper prone
+hybrid cloud or a remote Edge / IoT device in an insecure physical tamper prone
 location.)
 
-Keylime can be driven with a CLI application, web front end and a set of
+Keylime can be driven with a CLI application, web front end, and a set of
 RESTful APIs.
 
 Keylime consists of three main components; The Verifier, Registrar and the
@@ -33,12 +33,12 @@ the agent is running on.
 * The Registrar is a database of all agents registered
 with Keylime and hosts the public keys of the TPM vendors.
 
-* The Agent is deployed to the remote machine that is to be measured or provisoned
+* The Agent is deployed to the remote machine that is to be measured or provisioned
 with secrets stored within an encrypted payload released once trust is established.
 
 ### Rust based Keylime Agent
 
-The verifier, registrar and agent are all developed in Python and situated
+The verifier, registrar, and agent are all developed in Python and situated
 in this repository `keylime`. The agent is currently undergoing a port to the
 [Rust programming language](https://www.rust-lang.org), with this work taking
 place in the [rust-keylime repository](https://github.com/keylime/rust-keylime).
@@ -54,7 +54,7 @@ with the keylime verifier and registrar.
 
 ### TPM Support
 
-Keylime supports TPM version 2.0.
+Keylime supports TPM version *2.0*.
 
 Keylime can be used with a hardware TPM, or a software TPM emulator for
 development, testing, or demonstration purposes.  However, DO NOT USE keylime in
@@ -86,7 +86,7 @@ A hardware TPM should always be used when real secrets and trust is required.
 
 ### Installer script
 
-Keylimes installer requires Python 3.6 or greater.
+Keylime installer requires Python version(>=3.6).
 
  The following command line options are available using `installer.sh` script:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -86,7 +86,7 @@ A hardware TPM should always be used when real secrets and trust is required.
 
 ### Installer script
 
-Keylimes installer requires Python 3.6 or greater.
+Keylime installer requires Python version(>=3.6).
```
mpeters, line 89: I prefer the original wording here. Your suggestion puts the version number inside parenthesis, but the word "version" outside of it.

README.md
```diff
@@ -134,25 +134,26 @@ keylime and related emulators can also be deployed using Docker.
...
-[here](https://github.com/keylime/keylime/blob/master/.ci/run_local.sh)
+Please see either the 
+[Dockerfiles](https://github.com/keylime/keylime/tree/master/docker) or our
+[local CI script](https://github.com/keylime/keylime/blob/master/.ci/run_local.sh)
 which will automate the build and pull of keylime on TPM 1.2 or 2.0.
 
 ### Manual
 
-Keylime requires Python 3.6 & above
+Keylime requires Python version (>=3.6).
```
mpeters, line 144: Again, the original wording is clearer and shorter here.

README.md
```diff
@@ -163,18 +164,17 @@ Keylime uses the Intel TPM2 software set to provide TPM 2.0 support.
...
+* On Ubuntu 20 LTS:
 
-`sudo apt-get install tpm2-tss tpm2-tools tpm2-abrmd'
+`sudo apt-get install tpm2-tss tpm2-tools tpm2-abrmd`
 
-You can also build the tpm2-tss software stack (available at
-https://github.com/tpm2-software/tpm2-tss) as well as the
-tpm2-tools utilities available at https://github.com/tpm2-software/tpm2-tools. See
-README.md in these projects for detailed instructions on how to build and install.
+You can also build the tpm2-tss software stack (available at [tpm2-tss](https://github.com/tpm2-software/tpm2-tss))
```
mpeters, line 175: If you're going to rewrite this paragraph, I think it would be better to just remove the parentheical, and just make the first usage of "tpm2-tss" and "tpm2-tools" be the links to their repositories.

README.md
```diff
@@ -201,13 +201,13 @@ the `tpm2_checkquote` utility in your path.
 ###### TPM 2.0 Resource Manager
 
 Note that it is recommended that you use the tpm2-abrmd resource manager
-(available at https://github.com/tpm2-software/tpm2-abrmd) as well instead of
-communicating directly with the TPM.  See README.md at that project for
-detailed instructions on how to build and install.
+(available at [tpm2-abrmd](https://github.com/tpm2-software/tpm2-abrmd)) as well instead of
```
mpeters, line 204: Same here: remote the parenthetical and just make "tpm2-abrmd" link to the repo.

README.md
```diff
@@ -283,20 +290,19 @@ sudo -E python setup.py install build_ext --openssl=/usr/local/opt/openssl/
 #### Optional Requirements
 
 If you want to support revocation, you also need to have cfssl installed and in your
-path on the tenant agent.  It can be obtained from https://github.com/cloudflare/cfssl.  You
-will also need to set ca_implementation to "cfssl" instead of "openssl" in `/etc/keylime.conf`.
+path on the tenant agent.  It can be obtained from [cloudfare/cfssl](https://github.com/cloudflare/cfssl). You will also need to set ca_implementation to "cfssl" instead of "openssl" in `/etc/keylime.conf`.
```
mpeters, line 293: This reads kind of strange. When rendered, although "cloudflare/cfssl" is a link, the sentence doesn't actually make sense since "can be obtained from X" means X is usually a location and "cloudflare/cfssl" is not a location but it's name.

mpeters (changes requested):
Overall these are good improvements and clarifications. However, some small changes are requested.

README.md
samyak-jn: Ok, makes sense I'll revert this to the original wording.

README.md
samyak-jn: Yes as you pointed it out, I think so too, i'll revert this back to it's original form. Thanks for all the suggestions!

samyak-jn:
Hi @mpeters,
Thank you for suggestions, have made the changes that were requested!

README.md
```diff
@@ -163,18 +164,17 @@ Keylime uses the Intel TPM2 software set to provide TPM 2.0 support.
...
+* On Ubuntu 20 LTS:
 
-`sudo apt-get install tpm2-tss tpm2-tools tpm2-abrmd'
+`sudo apt-get install tpm2-tss tpm2-tools tpm2-abrmd`
 
-You can also build the tpm2-tss software stack (available at
-https://github.com/tpm2-software/tpm2-tss) as well as the
-tpm2-tools utilities available at https://github.com/tpm2-software/tpm2-tools. See
-README.md in these projects for detailed instructions on how to build and install.
+You can also build the tpm2-tss software stack at [tpm2-tss](https://github.com/tpm2-software/tpm2-tss)
```
mpeters, line 175: I think what I said might not have been clear. I think a better way to rework this section would be:

```
You can also build the [tpm2-tss](https://github.com/tpm2-software/tpm2-tss) software stack as well as
[tpm2-tools](https://github.com/tpm2-software/tpm2-tools). See the
README.md in these projects for detailed instructions on how to build and install.
```

This way the URLs aren't printed out and it reads naturally but still gives you the expected links to the projects. I also don't think we need to link to the README for each one since the sentence explanation and links are good enough

README.md
```diff
@@ -200,14 +200,12 @@ the `tpm2_checkquote` utility in your path.
 
 ###### TPM 2.0 Resource Manager
 
-Note that it is recommended that you use the tpm2-abrmd resource manager
-(available at https://github.com/tpm2-software/tpm2-abrmd) as well instead of
-communicating directly with the TPM.  See README.md at that project for
-detailed instructions on how to build and install.
+Note that it is recommended that you use the tpm2-abrmd resource manager at [tpm2-abrmd](https://github.com/tpm2-software/tpm2-abrmd) as well instead of communicating directly with the TPM.
```
mpeters, line 203: This is similar to the above: There's no reason to say "tpm2-abrmd" twice and only have the 2nd one be the link. Just use something like this:

```
Note that it is recommended that you use the [tpm2-abrmd](https://github.com/tpm2-software/tpm2-abrmd) resource manager as well instead of communicating directly with the TPM. 
```

mpeters (changes requested):

README.md
samyak-jn: Yea it would make more sense, I'll update both the paras.

samyak-jn:
Thanks for all your help @mpeters, and for all your work! :taco:

mpeters (approved):

PR #676, opened 2021-05-13
Incremental attestation for IMA

stefanberger:
This PR introduces incremental attestation for IMA by having the verifier maintain state about the last-requested measurement list entry and the state of the 'running hash' for PCR 10 at the last invocation for each agent separately.

We introduce a new parameter 'ima_ml_entry' in the agent's `quote/integrity` API so that the verifier can request the IMA measurement list at a certain entry number, starting with '0' (zero). An old agent will not look at this parameter and will just return the whole IMA measurement list and a new one may return it from that offset unless it was restarted or the request from the verifier is far beyond the number of measurement list entries (due to reboot maybe). A new agent will also return the indicator of the first measurement list entry it returns and and old agent won't respond with any such parameter. The verifier will assume that the whole list was returned by an old agent then (missing parameter) and otherwise work with the entry number that the agent return, which better be the requested number since otherwise the agent will fail attestation.

This series assumes affinity to an instance of a verifier since all state is maintained in memory only and the verifier initiates the request for the quotes. A restarting verifier will therefore request the entire IMA measurement log from all agents, which is as expensive as it is now, but then only request the new IMA measurement log entries afterwards, which should make the verifier by far less busy than it could be now. In fact, in most cases it will be more or less idle since it will not receive any new measurement log entries but only compare the last known PCR 10 state against the one reported by the agent.

### DIFF
--- a/installer.sh
+++ b/installer.sh
@@ -64,7 +64,7 @@ case "$ID" in
         echo "${ID} selected."
         PACKAGE_MGR=$(command -v apt-get)
         PYTHON_PREIN="git patch"
-        PYTHON_DEPS="python3 python3-pip python3-dev python3-setuptools python3-zmq python3-tornado python3-cryptography python3-simplejson python3-requests gcc g++ libssl-dev swig python3-yaml python3-gnupg wget"
+        PYTHON_DEPS="python3 python3-pip python3-dev python3-setuptools python3-zmq python3-tornado python3-cryptography python3-simplejson python3-requests python3-psutil gcc g++ libssl-dev swig python3-yaml python3-gnupg wget"
         if [ "$(uname -m)" = "x86_64" ]; then
             PYTHON_DEPS+=" libefivar-dev"
         fi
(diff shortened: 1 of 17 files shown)


maugustosilva:
Not ready to fully comment on the particulars of the PR, but I can at least say that in its current form, it passes the "end-to-end" private testing.

stefanberger:
I rebased this now on the latest master...

stefanberger:
FYI: I will be posting an enhancement request covering this PR.

stefanberger:
I *think* it is in good shape now.

maugustosilva (approved):
At this point, I believe we should merge this code as is, and let it be used/tested at scale (which I already did, to a point). If there are issues in production we will fix it over time. The gains in functionality justify the eventual risk, in my evaluation.

stefanberger:
@maugustosilva Thanks for testing on a larger setup. I don't have that myself...

Issue #656, opened 2021-04-29
Multiple types of failure if keylime_agent is restarted mid-attestation.


galmasi:
# Environment
 * OS / version: Fedora/32 *and* Ubuntu 18.04 (observed on both) -- we theorize there is no dependency
 * Processor architecture: x86_64 (likely irrelevant)
 * TPM Manufacturer: Infineon _and_ swtpm (Stefan)
 * Keylime version: 6.1.0+, but we think the problem is pervasive
 
# Disclaimer

I do not claim perfect understanding of the causes for the failures under discussion. Feel free to comment and contribute test outcomes.

# Description 
We have observed two separate problems if keylime agent is restarted while attestation is ON. Apologies for the long message; I am attempting to describe two problems with different root causes and likely requite different fixes; but they are triggered by the same actions. Both are heisenbugs -- they manifest occasionally.

Problem 1: attestation fails on the node -- the verifier is unable to decode the quote sent back by the keylime agent. This can be fixed by disregarding the attestation failure and re-adding the keylime agent by means of a tenant "add" command.

Problem 2: an extra authentication session is left in the TPM device. Depending on the value of the device property TPM2_PT_CONTEXT_GAP_MAX, a failure will happen 64 to 64k quotes *later* (on real systems this usually means a day and a half of delay before the error). After that tpm2_quote fails consistently.

In this situation the only fix we know of is to issue a `tpm2_flushcontext -s` command on the machine.

## Expected behavior vs. actual behavior 

* Expected behavior: attestation *always* continues correctly even if the keylime agent was restarted.
* Actual behavior: attestation fails sporadically after keylime agent restarts in one of two different ways. 

## Steps to reproduce problem
1. Set up a keylime system with at least a verifier, registrar and an agent.
2. Turn on attestation for the agent by means of a `keylime_tenant add .... ` command.
3. Wait for the situation to stabilize while watching the keylime agent's log and the keylime verifier's log.
4. Restart the keylime agent (in production they usually issue a systemd command to do it, but @maugustosilva @apoorvemohan observed this by running the agent manually and interrupting it with ctrl+c)
5. In some cases attestation fails immediately (problem 1)
6. In some cases the system appears to perform normally after keylime agent restart, but fails minutes or hours later without any additional action taken (problem 2)

# Relevant logs

*   **Failure mode 1** (verifier cannot decrypt received quote and attestation fails immediately)
``` 
[...]

PR #682, opened 2021-05-26
Flush all sessions from TPM device

maugustosilva:
This commit is a quick (and admittedly ugly) hack to fix a problem that
has been affecting our production environment. While it partially
addresses #656, we have laid out a 3-phase plan to deal with the problem
in a more consistent and sustainable manner.

### DIFF
--- a/keylime/tpm/tpm_main.py
+++ b/keylime/tpm/tpm_main.py
@@ -505,6 +505,12 @@ class tpm(tpm_abstract.AbstractTPM):
             logger.info("Taking ownership with config provided TPM owner password")
             owner_pw = config_pw
 
+        logger.debug("Removing all saved sessions from TPM")
+        if self.tools_version == "3.2":
+            retDict = self.__run(["tpm2_flushcontext", "-s"], raiseOnError=False)
+        elif self.tools_version in ["4.0", "4.2"]:
+            retDict = self.__run(["tpm2_flushcontext", "-s"], raiseOnError=False)
+
         if self.tools_version == "3.2":
             retDict = self.__run(["tpm2_takeownership", "-c"], raiseOnError=False)
             retDict = self.__run(["tpm2_takeownership", "-o", owner_pw, "-e", owner_pw],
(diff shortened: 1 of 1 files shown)


keylime/tpm/tpm_main.py
```diff
@@ -505,6 +505,12 @@ def __take_ownership(self, config_pw):
             logger.info("Taking ownership with config provided TPM owner password")
             owner_pw = config_pw
 
+        logger.debug("Removing all saved sessions from TPM")
+        if self.tools_version == "3.2":
+            retDict = self.__run(["tpm2_flushcontext", "-s"], raiseOnError=False)
+        elif self.tools_version in ["4.0", "4.2"]:
```
axelsimon, line 511: Is there a reason to have an if/elif rather than to have the three versions in a single list, considering both cases result in exactly the same action from what i can tell?

axelsimon (commented):
Barring a small question, i don't see any issue here.

Could you maybe however detail the 3 step plan you had in mind, even if it's been discussed elsewhere? thanks!

keylime/tpm/tpm_main.py
maugustosilva: you are right, there is not...I was under the impression that `tpm2_flushcontext` had a slight variation in parameter between 3.2 and 4.X, but that was not the case.

mpeters: lgtm after removing the conditional

maugustosilva:
> Barring a small question, i don't see any issue here.
> 
> Could you maybe however detail the 3 step plan you had in mind, even if it's been discussed elsewhere? thanks!

Sure, I can give you as much detail as I have, from the Slack channel (spoiler, it is gonna be underwhelming :-) )

Step 1: (ugly hack, alleviates our daily attestation failures): add flushcontext to tpm_init
Step 2: remove passwords from AIK, vacating the need for authentication sessions
Step 3: Ditch tpm2_tools for the tss library

Now, it seems clear to me (and the ones present at the discussion) that Step 3 will require the creation of a new enhancement document.

keylime/tpm/tpm_main.py
axelsimon: agreed

axelsimon (approved):
Looks ok to me :)

mpeters (approved):

Issue #30, opened 2019-01-23
Support more IMA hashing algorithms (TPM 2.0)


jetwhiz:
Keylime currently only supports SHA1 with IMA, particularly since IMA's template_hash field only supports SHA1 (though the Linux kernel extends this SHA1 value into all available hashing algorithms).  Add support for additional algorithms for IMA. 

See (for example): 
* https://github.com/mit-ll/python-keylime/blob/master/keylime/tpm_abstract.py#L345
* https://github.com/mit-ll/python-keylime/blob/master/keylime/ima.py
* https://github.com/mit-ll/python-keylime/blob/master/keylime/ima_emulator_adapter.py

lukehinds:
Just been looking into this @jetwhiz 

On the face of it, it makes sense (sha1 is not collision proof), but I am also aware sha1 is default and a lot of Linux Distributions are running with sha1 as the kernel compiled option:

```
cat /boot/config-$(uname -r) |grep CONFIG_IMA
CONFIG_IMA=y
CONFIG_IMA_MEASURE_PCR_IDX=10
CONFIG_IMA_LSM_RULES=y
# CONFIG_IMA_TEMPLATE is not set
CONFIG_IMA_NG_TEMPLATE=y
# CONFIG_IMA_SIG_TEMPLATE is not set
CONFIG_IMA_DEFAULT_TEMPLATE="ima-ng"
CONFIG_IMA_DEFAULT_HASH_SHA1=y
# CONFIG_IMA_DEFAULT_HASH_SHA256 is not set
CONFIG_IMA_DEFAULT_HASH="sha1"
CONFIG_IMA_WRITE_POLICY=y
CONFIG_IMA_READ_POLICY=y
# CONFIG_IMA_APPRAISE is not set
CONFIG_IMA_KEYRINGS_PERMIT_SIGNED_BY_BUILTIN_OR_SECONDARY=y
```

Is it a case of SHA1 is not "insecure" per say with IMA (perhaps because its an extension of lots of hashses).

Put this way, if someone makes an issue or pulls us up at a conference and says 'you're using SHA1 and that's dangerous" can we say 'yes we agree' or is it a case of no risk is there as collisions are not a factor in sha1's use in IMA and tpm quotes?

jetwhiz:
I don't think setting these variables to use `sha256` would fix the problem that's discussed in this issue.  

There are actually two separate hash algorithms / digests used in IMA.  The template hash/digest and the file hash/digest.
[...]

PR #687, opened 2021-06-14
Increase flexibilty of IMA entry parsing and validation

THS-on:
This PR introduces a flexible way for parsing and validating IMA entries. 
A entry is parsed into an AST with basic input validation and provides functions to reconstruct and validate the template hash. For an entry a validator can be specified which can be specialized based on the IMA template.

Making the parsing more flexible allows Keylime to support more use cases of IMA than files which is introduced with kernel [5.12](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=e58bb688f2e44237990dfb6). Those use mostly `ima-buf` as their template format which is currently not supported in Keylime.

For https://github.com/keylime/enhancements/pull/47 this makes it easier to add granular error context because the validation is unified into a single interface to use.

We are currently testing the experimental IMA support for device-mapper targets and this allows us to easily parse and validate those entries without duplicating the parsing and template hash validation work.

The old code skipped the validation of the template hash if it none or all bits set.  Why is that the case? 
I couldn't find a good explanation by looking around the kernel IMA implementation and Keylime git history.
If this behavior is necessary it should be documented before merging.

### DIFF
--- a/keylime/ima_ast.py
+++ b/keylime/ima_ast.py
@@ -1,0 +1,318 @@
+'''
+SPDX-License-Identifier: Apache-2.0
+Copyright 2021 Thore Sommer
+
+AST with parser and validator for IMA ASCII entries.
+Implements the templates (modes) and types as defined in:
+- https://elixir.bootlin.com/linux/latest/source/security/integrity/ima/ima_template.c
+- https://www.kernel.org/doc/html/v5.12/security/IMA-templates.html
+'''
+
+import binascii
+import codecs
+import hashlib
+import struct
+import abc
+import dataclasses
+
+from typing import Dict, Callable, Any, Optional
+from keylime import keylime_logging
+
+logger = keylime_logging.init_logging("ima")
+
+TCG_EVENT_NAME_LEN_MAX = 255
+SHA_DIGEST_LEN = 20
+MD5_DIGEST_LEN = 16
+
+START_HASH = (codecs.decode(b'0000000000000000000000000000000000000000', 'hex'))
+FF_HASH = (codecs.decode(b'ffffffffffffffffffffffffffffffffffffffff', 'hex'))
+
+NULL_BYTE = ord('\0')
+COLON_BYTE = ord(':')
+
+
+@dataclasses.dataclass
+class Validator:
+    functions: Dict[Any, Callable]
+
+    def get_validator(self, class_type) -> Callable:
+        validator = self.functions.get(class_type, None)
+        if validator is None:
+            logger.warning(f"No validator was implemented for: {class_type} . Using always false validator!")
+            return lambda *_: False
+        return validator
+
+
+class ParserError(TypeError):
+    """Is thrown when a type could not be constructed successfully."""
+
+
+class Mode(abc.ABC):
+    @abc.abstractmethod
+    def is_data_valid(self, validator: Validator):
+        pass
+
+    @abc.abstractmethod
+    def hash(self) -> bytes:
+        pass
+
+
+class Type(abc.ABC):
+    @abc.abstractmethod
+    def struct(self):
+        pass
+
+
+class HexData(Type):
+    data: bytes
+
+    def __init__(self, data: str):
+        try:
+            self.data = codecs.decode(data.encode("utf-8"), "hex")
+        except binascii.Error as e:
+            raise ParserError(f"Provided data was not valid hex: {data}") from e
+
+    def __str__(self):
+        return self.data.decode("utf-8")
+
+    def struct(self):
+        return struct.pack(f"<I{len(self.data)}s",
+                           len(self.data),
+                           self.data)
+
+
+class Signature(HexData):
+    """
+    Class for type "sig".
+    """
+
+
+class Buffer(HexData):
+    """
+    Class for type "buf".
+    """
+
+
+class Name(Type):
+    """
+    Class for type "n" and "n-ng".
+    """
+    name: str
+    legacy: bool = False
+
+    def __init__(self, name: str, legacy=False):
+        self.name = name
+        self.legacy = legacy
+
+    def __str__(self):
+        return self.name
+
+    def struct(self):
+        # The old "n" option is fixed length.
+        if self.legacy:
+            return struct.pack(f"{len(self.name)}sB{TCG_EVENT_NAME_LEN_MAX - len(self.name)}s",
+                               self.name.encode("utf-8"),
+                               NULL_BYTE,
+                               bytearray(TCG_EVENT_NAME_LEN_MAX - len(self.name)))
+
+        return struct.pack(f"<I{len(self.name)}sB",
+                           len(self.name) + 1,
+                           self.name.encode("utf-8"),
+                           NULL_BYTE)
+
+
+class Digest:
+    """
+    Class for types "d" and "d-ng" with and without algorithm
+    """
+    hash: bytes
+    algorithm: Optional[str] = None
+    legacy: bool = False
+
+    def __init__(self, digest: str, legacy=False):
+        self.legacy = legacy
+        tokens = digest.split(":")
+        if len(tokens) == 1:
+            try:
+                self.hash = codecs.decode(tokens[0].encode("utf-8"), "hex")
+            except binascii.Error as e:
+                raise ParserError(f"Digest hash is not valid hex. Got: {tokens[0]}") from e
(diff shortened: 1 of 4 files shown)


keylime/ima_ast.py
```diff
@@ -0,0 +1,320 @@
...
+    def __init__(self, data: str):
+        tokens = data.split()
+        if len(tokens) != 2:
+            raise ParserError()
+        self.digest = Digest(tokens[0], legacy=True)
+        self.path = Name(tokens[1], legacy=True)
+
+    def hash(self):
+        tohash = self.digest.struct() + self.path.struct()
+        return hashlib.sha1(tohash).digest()
```
stefanberger, line 189: What is it hashing in this case with a sha1?

THS-on: The hash function reconstructs the template hash that is in the IMA log.
In this case the template `ima` is build from a `d` (digest) and  `n` (name). 
To match the template hash, get the byte representation of the digest and name with `struct()`, concat them and then hash the result.

stefanberger:
So this PR is definitely a step into the right (o-o) direction. It looks good to me and I think the existing IMA test cases also should cover various aspects of it. 

The one thing I was looking at is whether it changes behavior depending on availability of allow list and exclude list and so on. https://github.com/keylime/keylime/blob/16a9cfd31b02f5c60b1ccc667627fac6144f82d1/keylime/ima.py#L270-L305
I think with the current code does nothing when there's no allow list.

Yours would require and allow list as it seems.
```
def _validate_ima_ng(exclude_regex, allowlist, digest: ima_ast.Digest, path: ima_ast.Name):
    if allowlist is None:
        logger.debug("No allowlist is there for validation so it cannot be in t    if LLOhere")
        return False

    if exclude_regex is not None and exclude_regex.match(path.name):
        logger.debug("IMA: ignoring excluded path %s" % path)
        return True
[...]
```

Maybe your code should also do what the current code does like this:

```
    if allowlist is not None:
        if exclude_regex is not None and exclude_regex.match(path.name):
            logger.debug("IMA: ignoring excluded path %s" % path)
            return True
        [...]
```

THS-on:
Thanks for looking at it.  
>  I think with the current code does nothing when there's no allow list.

Yes that seems to be a change in behavior which I have missed. I've now hopefully fixed this.
The current unit tests didn't catch that. I've extended the unit tests to now check for that use case.

Currently the unit tests only check for `ima-sig` entries should I extend that to cover the other formats?
Does anyone have IMA logs that we could use that are more interesting for testing than just generated by booting a system with the different template formats enabled? In particular for `FF_HASH` case if it happens in on real systems.

maugustosilva (approved):

stefanberger:
LGTM, too. Maybe someone wants to kick off the workflows... ?

keylime/ima_ast.py
```diff
@@ -0,0 +1,316 @@
...
+
+    def valid(self):
+        # TODO why do we need this case?
+        if self.template_hash == FF_HASH:
+            logger.warning("Skipped template_hash validation entry with FF_HASH")
+            return self.mode.is_data_valid(self.validator)
+        if self.template_hash != self.mode.hash():
+            return False
+        if self.validator is None:
+            return False
```
THS-on, line 314: Before this gets merged I would like add an explanation why the exception for `START_HASH` and  `FF_HASH` is made and what in IMA causes this.
In an ideal case we could also provide a test IMA log that has this case.

mpeters: We might have to reach out to folks on IRC about this as it seems to be something from the past. It seems it was added in revision 5bcd88dd by @nabilschear in 2017.

mimizohar: As there is no locking when opening a file for read and subsequently for write, or the reverse, the file measurement has little value.   IMA defines this as a measurement violation and invalidates the measurement list, by extending the TPM with 0xFF's instead of the 0x00's.  Instead of ignoring these measurement violations, each violation should be reviewed and accounted for.  The most common violations are log files, which should not be included in the measurement list.  Other examples are applications which unnecessarily open files rw.  These applications should be fixed.

If you really want to gloss over these measurement violations, at least don't hard code it.  ima-evm-utils defines the option "--ignore-violations".

THS-on: @mimizohar thank you for your insight. 

@mpeters I'll remove the special handling from this PR and if necessary we can introduce a special option to ignore this violations later.

mimizohar: https://wiki.gentoo.org/wiki/Integrity_Measurement_Architecture/Recipes has a nice write up on writing custom policies and excluding log files.

stefanberger: And if we remove the special handling now does it change behavior or change behavior for some people?

THS-on: The behavior should only change for people that rely on that those errors are ignored instead of adding those files to the exclude list. The default IMA configuration given in our documentation should only measure executable files if I saw that correctly and this issue should not occur with that configuration.

I can add that feature back with a configuration option which is enabled by default to keep backwards compatibility, but this behavior was not documented/really known before and I would prefer to disable it by default and users can manually enable it if they actually rely on it.

THS-on:
I've cleaned up my commits and I kept the old functionality for handling time of measure, time of use (ToMToU) errors, but added some more details.

I think in the future we should make this configurable as discussed above but this then will be a new PR.

keylime/ima_ast.py
stefanberger: I don't think this PR should change the behavior of how keylime works. If we removed handling of the 00- and FF-hashes then not even quote verification would work if those types of entries occurred in the log.

THS-on: You are correct and I've removed the behavior change from this PR.

mpeters (approved):
Thanks for including tests for this on the first submission :+1:

kgold2:
> Does anyone have IMA logs that we could use that are more interesting for testing than just generated by booting a system with the different template formats enabled?

I have a set of logs, one composite log that I use for regression testing, and one with various attacks (bad digest, no signature, no public key, bad signature.
If anyone wants them, let me know.

THS-on:
> I have a set of logs, one composite log that I use for regression testing, and one with various attacks (bad digest, no signature, no public key, bad signature.
> If anyone wants them, let me know.

Yes I'm interested to use them for adding better test for the current IMA parsing and validation.

kgold2:
Where would I put them; how should I send them to you?

keylime/ima_ast.py
mimizohar: Please 'd' and 'n' fields should really be limited to the original "ima" template.   At minimum use the ima-ng template, which are based on the 'd-ng' and 'n-ng' fields.  Using the 'ima-sig' template would be better.

mimizohar: "The old code skipped the validation of the template hash if it none or all bits set. Why is that the case?
I couldn't find a good explanation by looking around the kernel IMA implementation and Keylime git history.
If this behavior is necessary it should be documented before merging."

IMA support for updating multiple TPM 2.0 banks based on the per bank crypto algorithm was relatively recently upstreamed.   The second field of the IMA measurement list still contains a SHA1 hash of the template data.  To validate the other TPM banks (e.g. sha256), requires walking the IMA measurement verifying the sha1 template data hash matches the template data, before re-calculating and extending the template data hash for each TPM bank enabled.

THS-on: The `d` and `n` field are limited to the `ima` template. We treat `d`/`d-ng` or `n`/`n-ng` as the same object for validators to consume but reconstructing their hashes differently. 

In the future we should add support for also returning the sh256 etc. hashes and check them also. The issue that is here: https://github.com/keylime/keylime/issues/30.
Does IMA recalculate the hash for each algorithm or does it hash with sha1 and than adds padding for the other algorithms?

mimizohar: > The `d` and `n` field are limited to the `ima` template. We treat `d`/`d-ng` or `n`/`n-ng` as the same object for validators to consume but reconstructing their hashes differently.
> 
> In the future we should add support for also returning the sh256 etc. hashes and check them also. The issue that is here: #30.
> Does IMA recalculate the hash for each algorithm or does it hash with sha1 and than adds padding for the other algorithms?

Support for calculating and extending per TPM bank digests was upstreamed in Linux 5.8.  Refer to the upstream kernel commit 1ea973df6e21 ("ima: Calculate and extend PCR with digests in ima_template_entry"). - Mimi

PR #692, opened 2021-07-07
ci: make run_local.sh work with newer docker versions

THS-on:
The old configuration would result in the following error:

Initializing machine ID from random generator.
Failed to create /init.scope control group: Read-only file system
Failed to allocate manager object: Read-only file system
[!!!!!!] Failed to allocate manager object.
Exiting PID 1...

Tested with docker version 20.10.7.

Signed-off-by: Thore Sommer <mail@thson.de>

### DIFF
--- a/.ci/run_local.sh
+++ b/.ci/run_local.sh
@@ -17,7 +17,8 @@ function tpm2 {
     container_id=$(mktemp)
     docker run --detach --privileged \
         -v $REPO:/root/keylime \
-        -v /sys/fs/cgroup:/sys/fs/cgroup:ro \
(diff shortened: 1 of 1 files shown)


maugustosilva:
Please fix the pylint errors and will go ahead a do a functional end-to-end test.

THS-on:
@maugustosilva I've only changed `.ci/run_local.sh` which shouldn't cause pylint to fail.

The pylint version in the CI went from 2.8.3 to 2.9.3 which I think caused this and the error is in `keylime/tpm_bootlog_enrich.py:73`.

mpeters (approved):

PR #695, opened 2021-07-08
check_pcrs: match PCR if no mb_refstate is provided

### LINKED ISSUE #693: No documentation about `--mb_refstate`
# Description 

After #568, and in relation with #491 a new feature has been added, that will help to do a replay of the measured boot tracked by the event log.  There is no documentation about how this parameter can be used. Seems that is a JSON document but there is no description about the fields and how will be used.

### LINKED ISSUE #694: If the PCR belong to a measured boot list, it is not validated
# Description 

If the PCR belongs to a hard-coded list of PCRs under `config.MEASUREDBOOT_PCRS` and `--mb_refstate` is not used, the check of match do not happens, so the tpm-policy parameter gets ignored.

If we pass a `--tpm-policy` with a PCR that belong to the list, like PCR 0 for example, the validation of this register never happens in the verifier unless we some value to `--mb_refstate` is passed.

I would expect that if this second parameter is not passed, a plain validation still occurs.

aplanas:
If the values for mb_refstate are empty, the PCRs that are present in
the tpm_policy and belong to the measured boot process are ignored.

This patch check that the PCRs form the policy matches the ones that
comes from the quote when there is no measured boot data.

Fix #694

### DIFF
--- a/keylime/tpm/tpm_abstract.py
+++ b/keylime/tpm/tpm_abstract.py
@@ -283,6 +283,9 @@ class AbstractTPM(metaclass=ABCMeta):
                     if val_from_log_hex_stripped != pcrval_stripped:
                         logger.error("For PCR %d and hash SHA256 the boot event log has value %r but the agent returned %r", pcrnum, val_from_log_hex, pcrval)
                         return False
+                elif pcrnum in pcr_allowlist and pcrval not in pcr_allowlist[pcrnum] and not config.STUB_TPM:
+                    logger.error("%sPCR #%s: %s from quote does not match expected value %s", ("", "v")[virtual], pcrnum, pcrval, pcr_allowlist[pcrnum])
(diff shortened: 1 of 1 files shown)


aplanas:
The fails in the lint do not come from the PR. Is it expected to be fixed here or in a different PR?

THS-on:
They should be already fixed in the master branch with https://github.com/keylime/keylime/commit/cc36a908ab87bafeb88318ec851d0e0e4de06009. So just do a rebase on master.

mpeters:
This looks good, but wanted to get another look from @maugustosilva since he knows the mb_refstate lines the best.

maugustosilva (approved):

maugustosilva:
This PR passes successfully on our "private" end-to-end CI/CD.

Issue #691, opened 2021-07-06
UUID case


lukehinds:
So it turns out UUID's should be lower case ("a" through "f" are in fact  hexadecimal values)

https://datatracker.ietf.org/doc/html/rfc4122#section-3

we hit this issue in the rust keylime agent, where the UUID crate generates a UUID lowercase, which does not map to the entry in keylime.conf

So this to me suggests we should user lower in the python side. Where lower is the default and everything is canonicalised to `.lower()`

jronnblom:
Also note that different version of dmidecode can return UPPER or lower. Like in Ubuntu 18.04 vs Ubuntu 20.04. 

This causes a problem in the old python agent.

PR #699, opened 2021-07-13
Set default UUID as lowercase

lukehinds:
This change was a lot less disruptive then I thought

It turns out the python UUID lib generates lower case as default:

```
>>> import uuid
>>> uuid.uuid4()
UUID('fb00e830-7a99-4761-8523-63e494431d7f')
```

So all higher case instances were as a result of the UUID entry
in `keylime.conf`

I have now changed all these to lowercase.

With the code as it it, there will be no need to patch existing UUIDs
in a users database. Anything using `generate` will already be lower
case. Anything that already exists in the db using highercase can
remain that way.

Resolves: #691

### DIFF
--- a/test/test_registrar_db.py
+++ b/test/test_registrar_db.py
@@ -13,7 +13,7 @@ from keylime.db.keylime_db import SessionManager
 # BEGIN TEST DATA
 
 test_data = {
-    'agent_id': '4d32fbb3-d2f1-4a97-9ef7-75bd81c00000',
+    'agent_id': 'd432fbb3-d2f1-4a97-9ef7-75bd81c00000',
     'ek_tpm': """-----BEGIN PUBLIC KEY-----
 MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEA0dLxdAABVJO6qxamjCMh
 yhWZgiFHZHnPEe0tMFyK3fNVr/w8lX9r+QOLxLmkT0IdgsEYtGZGefbD+qQl4O1s
--- a/test/test_registrar_db.py
+++ b/test/test_registrar_db.py
@@ -13,7 +13,7 @@ from keylime.db.keylime_db import SessionManager
 # BEGIN TEST DATA
 
 test_data = {
-    'agent_id': 'D432FBB3-D2F1-4A97-9EF7-75BD81C00000',
+    'agent_id': '4d32fbb3-d2f1-4a97-9ef7-75bd81c00000',
     'ek_tpm': """-----BEGIN PUBLIC KEY-----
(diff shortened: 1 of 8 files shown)


mpeters:
Should we change where the value is read from keylime.conf and lowercase it after reading? Just trying to make it as easy as possible to configure and would hate to have case sensitivity in a config file be a reason something is broken.

lukehinds:
> Should we change where the value is read from keylime.conf and lowercase it after reading? Just trying to make it as easy as possible to configure and would hate to have case sensitivity in a config file be a reason something is broken.

I am honestly not sure that is best approach. The config value is the broken aspect as its using the wrong case. I would have thought its better to not change code to fix a incorrect config. I expect most users will be using `uuid = generate`, which automatically defaults to lower. The current value in keylime.conf is also  only really a static UUID to assist testing by having a pre-known non arbitrary value.

mpeters:
> I am honestly not sure that is best approach. The config value is the broken aspect as its using the wrong case. I would have thought its better to not change code to fix a incorrect config. 

That's fair. It could also be a source of bugs if we change a client's config value to try to "fix" it for them.

lukehinds:
> > I am honestly not sure that is best approach. The config value is the broken aspect as its using the wrong case. I would have thought its better to not change code to fix a incorrect config.
> 
> That's fair. It could also be a source of bugs if we change a client's config value to try to "fix" it for them.

That's also a good point. A user might generate their own UUIDs using another service that utilises mixed case

mpeters:
This looks good but it's failing a test where it adds an agent and can't get it back out again with the same id.

maugustosilva:

maugustosilva:
This PR passes successfully on our "private" end-to-end CI/CD. FYI @lukehinds in our environment, UUIDs (all uppercase) are generated externally and the added to keylime.conf

test/test_registrar_db.py
```diff
@@ -13,7 +13,7 @@
 # BEGIN TEST DATA
 
 test_data = {
-    'agent_id': 'D432FBB3-D2F1-4A97-9EF7-75BD81C00000',
+    'agent_id': '4d32fbb3-d2f1-4a97-9ef7-75bd81c00000',
```
mpeters, line 16: This isn't just a lower-case, it actually changes the first 2 characters from "D4" to "4d". @lukehinds This is the cause behind the failure in the `test/test_registrar_db.py` CI test

mpeters (changes requested):

test/test_registrar_db.py
lukehinds: Ah, well spotted, my bad, no idea how I managed that used the vscode lower case function.

lukehinds:
oof ! DCO!

mpeters:
> oof ! DCO!

Fixed in a squash merge

PR #700, opened 2021-07-13
measured_boot: drop process_refstate

aplanas:
The funcion "process_refstate" check if the parameter is a dict, and if
so returns the same parameter, and if not returns the same parameter.

This function is a noop.  This patch drop the function and the only
place where it is called.


maugustosilva:
This PR passes successfully on our "private" end-to-end CI/CD.

maugustosilva (approved):

PR #704, opened 2021-07-15
tenant: add the option to specify the registrar ip and port

THS-on:
As discussed on slack the tenant now has the option for specifying the registrars ip and port.


mpeters (approved):
lgtm, thanks for picking this up!

PR #706, opened 2021-07-15
Fix some issues introduced during the move to 'ast'

stefanberger:
The 2 patches in this PR fix a runtime failure and a behavior change. Both are due to the move to 'ast'.


mpeters (approved):
lgtm

THS-on:
Thank you for finding the change in behaviour. I'll write some tests to cover that so that future changes won't break that again.

PR #708, opened 2021-07-15
tests: extend ima tests for signature validation and exclude lists

THS-on:
This adds the tests for the IMA behavior regression that were fixed in #706.

I've also added tests to check if the exclude lists actually work.


mpeters (approved):
Thanks for boosting the robustness of the test suite!

PR #709, opened 2021-07-16
ima: Convert pcrval to bytes to increase efficiency

stefanberger:
The pcrval in process_measurement_list originates from the IMA log and is
a hex number in string form. Convert it to its bytes representation once
rather than converting the running_hash every time we do a comparison.


maugustosilva:
LGTM. This PR passes successfully on our "private" end-to-end CI/CD.

maugustosilva (approved):

stefanberger:
Just pushed a slightly updated commit text that reflects that pcrval may not be set.

PR #713, opened 2021-07-20
Automatically calculates the boot_aggregate from the measured boot log.

maugustosilva:
This PR uses the tsseventextend tool (found on package tss2), which
reads the binary measured boot log (sent from the `agent` to the
`verifier`), and then uses it to calculate the boot_aggregate.
This is done by using the (also calculated) values for PCRs 0-7(9)
also from the measured boot log (the need for contemplate both
PCRs 0-7 and 0-9 comes from the combination of older kernels and
older versions of grub).

### DIFF
--- a/installer.sh
+++ b/installer.sh
@@ -121,7 +121,7 @@ case "$ID" in
         GOPKG="golang"
         if [[ ${VERSION_ID} -ge 30 ]] ; then
         # if fedora 30 or greater, then using TPM2 tool packages
-            TPM2_TOOLS_PKGS="tpm2-tools tpm2-tss tpm2-abrmd"
+            TPM2_TOOLS_PKGS="tpm2-tools tpm2-tss tpm2-abrmd tss2"
             NEED_BUILD_TOOLS=0
             HAS_GO_PKG=1
         else
--- a/keylime/ima.py
+++ b/keylime/ima.py
@@ -82,7 +82,7 @@ def _validate_ima_sig(exclude_regex, ima_keyring, allowlist, digest: ima_ast.Dig
     return valid_signature
 
 
-def process_measurement_list(lines, lists=None, m2w=None, pcrval=None, ima_keyring=None):
+def process_measurement_list(lines, lists=None, m2w=None, pcrval=None, ima_keyring=None, boot_aggregates=None):
     running_hash = ima_ast.START_HASH
     found_pcr = (pcrval is None)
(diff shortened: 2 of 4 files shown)


stefanberger:
The bootaggregate can only be calculated in case PCRs 0-7 (or 9?) are also transferred, right? I am wondering whether it wouldn't be possible to calculate that as part of evaluating the measured boot log or the state of those pcrs in native python code rather than calling out to another command line tool?

maugustosilva:
> The bootaggregate can only be calculated in case PCRs 0-7 (or 9?) are also transferred, right? I am wondering whether it wouldn't be possible to calculate that as part of evaluating the measured boot log or the state of those pcrs in native python code rather than calling out to another command line tool?

Correct, we calculate the boot aggregate in both cases, PCRs 0-7 and 0-9, both sha1 and sha256 and add it to the allowlist (in case it is not there). 

I certainly would have preferred calculating it during the already existing invocation of `tpm2_eventlog`, but it looks like this is not possible. Since we are working towards eliminating all invocations of `tpm2-tools` from the python agent, there will be certainly a point in the future where we will have to rewrite that "critical section" of the boot log evaluation in native python.

mpeters:
Just so I'm clear, this is code running on the verifier. So it's not just about removing tpm2-tools from the python agent, but from the whole code base correct? I know we have that as an end goal. Just curious now that Stefan brings it up, how difficult would it be for that to be done in pure python here?

THS-on:
If we are rewriting some of the tpm2_tools functionality in Python we maybe can cooperate with https://github.com/tpm2-software/tpm2-pytss to make those features available outside of Keylime.

maugustosilva:
> Just so I'm clear, this is code running on the verifier. So it's not just about removing tpm2-tools from the python agent, but from the whole code base correct? I know we have that as an end goal. Just curious now that Stefan brings it up, how difficult would it be for that to be done in pure python here?

running on the verifier, only, correct

maugustosilva:
> If we are rewriting some of the tpm2_tools functionality in Python we maybe can cooperate with https://github.com/tpm2-software/tpm2-pytss to make those features available outside of Keylime.

One of my colleagues at IBM (with my help) is working on exactly this. Our current goal is to do exactly what you described, re-implement all "system" calls in keylime as `pytss` invocations.

mpeters (approved):

PR #715, opened 2021-07-23
RFC: ima: Validate ima-buf like ima-ng with exclude and allowlist

stefanberger:
Validate ima-buf like we validate ima-ng with exclude and allowlist.

### DIFF
--- a/keylime/ima.py
+++ b/keylime/ima.py
@@ -82,6 +82,12 @@ def _validate_ima_sig(exclude_regex, ima_keyring, allowlist, digest: ima_ast.Dig
     return valid_signature
 
 
+def _validate_ima_buf(exclude_regex, allowlist, digest: ima_ast.Digest, path: ima_ast.Name, data: ima_ast.Buffer):
+    # disregard data for now
+    del data
+    return _validate_ima_ng(exclude_regex, allowlist, digest, path)
+
+
 def process_measurement_list(lines, lists=None, m2w=None, pcrval=None, ima_keyring=None):
     running_hash = ima_ast.START_HASH
     found_pcr = (pcrval is None)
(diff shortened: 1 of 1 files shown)


stefanberger:
@THS-on It looks like this was missing in your PR...

THS-on:
To my knowledge only the loaded SELinux policy is measured with an `ima-buf` template and using the `ima-ng` validator for that didn't make sense for me. If other parts use that template your change is a simple way to support it.

I'm currently testing it with the upcoming support for device mapper (mainly dm-verity) in IMA but haven't gotten far with it.
My plan is writing a proposal for adding validation for device mapper targets once I got a proof of concept working. Adding validation for that will require some changes to the allowlist format.

stefanberger:
I primarily tested this with keyrings (`measure func=keycheck`) where one now has to specify the expected measurement(s) of keys and keyrings in the allowlist or list the keyrings in the exclude list.

An entry in the allowlist may look like this then:
```
a969030afba9e059ca0208ed0770392e2b398de077583c0b849659e321e6fbe9 .ima
```

This may be ambigous with a filename '.ima', but the chance of collision with the hashes should be low.

We could also require that an allowlist entry look like this:
```
a969030afba9e059ca0208ed0770392e2b398de077583c0b849659e321e6fbe9 %keyring:.ima
```

Of course this could also be ambigous with filenames as well if one was to 'force it'. The thing is just that the verifier will see <foobar> (e.g., .ima) as part of ima-buf and then could prepend '%keying:' to then search for an allowlist entry for '%keyring:<foobar>'. But then the name in the ima-buf may not always represent a keyring, so preprending '%keyring' wouldn't always work.

I'd also rather not introduce a --keyring-allowlist to separate it from today's allowlists.

THS-on:
Ok, I didn't know about that feature.

I would propose that we extend the allowlist format to include different sections. We already support JSON formatted allowlists and there we could specify an extra section for keyrings.

It would be nice if the measurement was similar to the proposed dm additions that use the path field just to indicate what kind of data is that but not the actual data it self.  We will get namespace collisions if it's not clear how that path field should be used. `HASH kind-of-measurement data` would be probably the ideal format where data then contains the name of the keyring.

keylime/ima.py
```diff
@@ -82,6 +82,12 @@ def _validate_ima_sig(exclude_regex, ima_keyring, allowlist, digest: ima_ast.Dig
     return valid_signature
 
 
+def _validate_ima_buf(exclude_regex, allowlist, digest: ima_ast.Digest, path: ima_ast.Name, data: ima_ast.Buffer):
+    # disregard data for now
+    del data
```
aplanas, line 87: This is not needed IMHO. Is OK to ignore it in the call.

stefanberger: pylint requires it...

aplanas: Aha ... I think that there are maybe better options, like renaming to `_data` or disabling the pylint complain.

stefanberger: `_data` would also work.

stefanberger:
Closing this PR in favor of #718

PR #716, opened 2021-07-26
Refacator of check_pcrs to parse then validate

THS-on:
The old code walked through the PCR list and checked if the PCR belongs to some kind of verification. This is changed so that PCR list is parsed with basic validation to a dictionary and then each type of validation is checked.

I've also moved the `config.STUB_TPM`  check before the validation step because nearly all conditions checked for it manually.

This is in preparation for https://github.com/keylime/enhancements/issues/48 to make it easier to continue validation if one check failed.

### DIFF
--- a/keylime/tpm/tpm_abstract.py
+++ b/keylime/tpm/tpm_abstract.py
@@ -8,6 +8,7 @@ import ast
 import hashlib
 import os
 import string
+import typing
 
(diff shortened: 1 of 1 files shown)


stefanberger:
lgtm

mpeters (approved):

Issue #717, opened 2021-07-26
Discussion: Learning the keys from the IMA log and using them for signature verification


stefanberger:
With an IMA policy rule like `measure func=KEY_CHECK keyrings=.ima|.builtin_trusted_keys` the IMA log starts reporting the keys that were loaded onto the respective keyrings. An allowlist could now be used by the verifier to approve the hashes of keys for the various keyrings and the verifier could then learn which keys to use for signature verification since the raw key bytes of the x509 cert in DER format are in the last column of the `ima-buf` entries in the log. Currently all the keys to be used by the verifier have to be passed via the tenant tool alongside the allowlist and exclude list. If one was to strongly manage the hashes in the allowlist then one could rely on the learning of the keys from the IMA log.

```
10 113dd04eb569de6bd1ed9007f5c0da18dd8673dc ima-sig sha256:bc70d24e63213dfd51024fc6b7b0f65c558ce799362b56c9ceef5009e8cc0625 boot_aggregate 
[...]

PR #718, opened 2021-07-27
RFC: Learn the keys to user for signature verification from the IMA's ima-buf log entries

stefanberger:
This PR creates an ima-buf validator and hooks into it to learn the keys that are used on the monitored system and populates the validator's virtual keyrings after the entry has been accepted via either the allowlist or the exclude list as evaluated by the existing ima-ng validator.

The IMA policy on the attested system requires a rule like `measure func=KEY_CHECK keyrings=.ima|.builtin_trusted_keys` to produce ima-ng entries for keys that were loaded onto the given keyrings. Once the validator receives such an entry the administrator is required to either add entries for accepted keys' hashes to an allowlist or to the excludelist and thus has to consider which keys to approve or whether to care at all. Once keys are approved through the allowlist we can then also use them to populate virtual keyrings and use them for signature verification. This then does not require an admin to provide explicit keys via the keylime_tenant tool anymore since keys are approved via the allowlist and learned from the IMA log.


stefanberger:
This PR addresses the proposal/discussion in issue #717

stefanberger:
I have now updated this PR to also persist the learned keys and their keyrings so that iterative attestation can function properly. We can merge it when I am back...

mpeters:
@stefanberger is this still an RFC/WIP? Or is it ready for full review/merge?

stefanberger:
Thinking out loud so that this kind of extension doesn't put us into a kernel where we don't want to be:

Per the previous addition of allowlist support for keys on keyrings (PR for docs here: https://github.com/keylime/keylime-docs/pull/65) the user is forced to add measurement of keys into the allowlist if the IMA policy on the host contains a rule like `measure func=KEY_CHECK keyrings=.ima|.evm` that ends up reporting keyrings and the measurements of keys on them via an ima_buf entry. That said, management of keys via the allowlist is going to be required if such an IMA policy rule is used on the system. So we should be able to pick up the reported keys and learn which ones are used and use them for signature verification, which is what this patch does. I don't think that this should be a problem...

THS-on:
The concept in general seems sound for me. 
The only thing that I can think of that you want to measure other keyrings like the builtin for kernel modules and want to ensure that other files are only signed by another key special for IMA, not your kernel module signing key. If that includes the `.secondary_trusted_keys` keyring which I think can keys from UEFI depending on the configuration I would propose that we expose in option in the allowlist JSON to disable that feature. 

How are `.builtin_trusted_keys` and `.secondary_trusted_keys` handled with the current implementation?

THS-on:
Ok that was basically my question if we have special handling for the those keyrings. 

I would just add the option to either disable it fully or exclude certain keyrings from that feature.

stefanberger:
> I would just add the option to either disable it fully or exclude certain keyrings from that feature.

Add an option to the keylime.conf config file? And you mean certain keyrings should be excluded from the 'learning' also  via the config file?

THS-on:
I would put in the allowlist JSON format with two options:

* Disable the learning completely. Only manually added keys are used
* Either include or exclude keyrings from the learning process.

stefanberger:
> I would put in the allowlist JSON format with two options:

Yes, the config file would be too global, affecting all systems then.

keylime/ima.py
```diff
@@ -349,6 +355,11 @@ def read_allowlist(al_path=None, checksum="", gpg_sig_file=None, gpg_key_file=No
         # version 2 added 'keyrings'
         if "keyrings" not in alist:
             alist["keyrings"] = {}
+        # version 3 added 'ima"
+        if "ima" not in alist:
+            alist["ima"] = {
+                "ignored_keyrings": []
+            }
```
stefanberger, line 362: I added this now because I had added it for v2 as well. However, I am not sure whether this is correct or necessary because if the policy comes in via `process_allowlist`, which is the case with some tests, then this type of extension is not done (a bug?) and then it's easier to do what I did above with `allowlist.get('ima', {}).get('ignored_keyrings', [])` to handle optional fields added only later on.

stefanberger:
> I would put in the allowlist JSON format with two options:
> 
>     * Disable the learning completely. Only manually added keys are used
> 
>     * Either include or exclude keyrings from the learning process.

The allowlist is become more of a policy now... I extended it now with a map for an entry for 'ima' where one can specify the ignored_keyrings. If a `'*'` is in that list all keyrings are ignored, nothing is learned, and also not checked against the keyrings values also specified in that policy.

```
empty_allowlist = {
    "meta": {
        },
    "release": 0,
    "hashes": {},
    "keyrings": {},
    "ima" : {
        "ignored_keyrings": []
    }
}
```

THS-on:
> The allowlist is become more of a policy now...

We should probably merge the exclude list also in that data structure and rename it to policy because I think now that we implement more features only allow and exclude list are not flexible enough.

> If a '*' is in that list all keyrings are ignored, nothing is learned, and also not checked against the keyrings values also specified in that policy.

I like that solution. This allows us to still verify if a keyring has the expected entry without using it for IMA signature verification.

keylime/ima.py
THS-on: The tenant first runs `read_allowlist` and then with that output `process_allowlist` which not all tests do (that could be a bug). The verifier just uses the JSON send via the API. 

I would add a function that adds necessary fields with sane defaults and run it on the JSON provided via the API.

stefanberger:
Ok, so I have confidence now that this is working as it is since I now tested it out on a real system. We have the flexibility to configure it via the allowlist, aka future policy.

stefanberger:
FYI: This is an experimental  policy/allowlist that I have been using now for an system that has all executables and libraries signed and reports the keys loaded onto keyrings using this rule here:

`measure func=KEY_CHECK keyrings=.ima|.builtin_trusted_keys|.evm`

```
{
    "meta" : {
       "version": 3
    },
    "release": 0,
    "hashes": {
    },
    "keyrings": {
        ".ima": [
            "a7d52aaa18c23d2d9bb2abb4308c0eeee67387a42259f4a6b1a42257065f3d5a",
            "687a22762555781d8ce82e285e71c3dce7bdcf0fb8cdd68ae172c9144267555f",
            "447960716965568f381cf303149c5650821ff8b25b953b9ec983af51fcd0fe3c"
        ],
        ".evm": [
            "687a22762555781d8ce82e285e71c3dce7bdcf0fb8cdd68ae172c9144267555f",
            "a7d52aaa18c23d2d9bb2abb4308c0eeee67387a42259f4a6b1a42257065f3d5a"
        ],
        ".builtin_trusted_keys": []
    },
    "ima": {
        "ignored_keyrings": [".builtin_trusted_keys"]
    }
}
```

stefanberger:
I added an enhancement proposal for this: https://github.com/keylime/enhancements/pull/52

PR #719, opened 2021-07-27
config: merge convert and list_convert into the same function

THS-on:
Looking through the config module I found that there are two functions for converting bytes in a nested object structure to strings. I removed list_convert and the code now uses convert.


stefanberger:
Presumably our tests are covering the parsing of config files and would run over the code changes because I was wondering where `convert()` is...

THS-on:
>  I was wondering where convert() is...

It's at the beginning of `config.py`.
https://github.com/keylime/keylime/blob/95ef0415f890b525ab277c37705aba9c22600dff/keylime/config.py#L25-L34

mpeters:
Weird that they were identical... Good catch.

mpeters (approved):

PR #721, opened 2021-07-28
Use tox for pylint

elfosardo:
Using tox automatically gives a reproduceable environment no matter
where the test is executed.
Also move pylint config to pylintrc.


Makefile
```diff
@@ -1,4 +0,0 @@
-
-.PHONY: check
-check:
-	scripts/check_codestyle.sh
```
stefanberger, line 4: I often ran `make` on my system to have the codestyle checked...

stefanberger:
Apart from that one comment I have revealing my obsession with `make`: LGTM

maugustosilva (approved):

Makefile
mpeters: I too would like to keep the make target. It's useful for local dev and they are just 2 entry points to the same code.

elfosardo: the problem with using make to run the codestyle check as it is, is that the environment is not isolated
the check will run using the python libraries installed in the operating system used, while tox creates a virtual environment installing the correct dependencies and running the test using only the virtualenv entries
what if we change the check entry in the Makefile so that it runs tox?
so the check command becomes:
tox -vv epylint
It will be transparent to the end user, but the result will be better IMHO

stefanberger: Sounds good to me.

mpeters (approved):

PR #726, opened 2021-07-30
This single-line commit fixes an error when checking the remaining non validated PCRs

maugustosilva:

### DIFF
--- a/keylime/tpm/tpm_abstract.py
+++ b/keylime/tpm/tpm_abstract.py
@@ -307,7 +307,7 @@ class AbstractTPM(metaclass=ABCMeta):
                     return False
                 pcrs_in_quote.add(pcr_num)
 
-        tpm_policy=ast.literal_eval(tpm_policy)
+        tpm_policy = ast.literal_eval(tpm_policy)
 
         # Check the remaining non validated PCRs
         for pcr_num in pcr_nums - pcrs_in_quote:
--- a/keylime/tpm/tpm_abstract.py
+++ b/keylime/tpm/tpm_abstract.py
@@ -307,12 +307,10 @@ class AbstractTPM(metaclass=ABCMeta):
                     return False
                 pcrs_in_quote.add(pcr_num)
 
-        tpm_policy = ast.literal_eval(tpm_policy)
-
         # Check the remaining non validated PCRs
         for pcr_num in pcr_nums - pcrs_in_quote:
             if pcr_num not in list(pcr_allowlist.keys()):
-                if len(list(tpm_policy.keys())) > 0:
+                if len(list(tpm_policy_.keys())) > 0:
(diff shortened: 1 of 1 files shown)


keylime/tpm/tpm_abstract.py
```diff
@@ -307,6 +307,8 @@ def check_pcrs(self, agentAttestState, tpm_policy, pcrs, data, virtual, ima_meas
                     return False
                 pcrs_in_quote.add(pcr_num)
 
+        tpm_policy=ast.literal_eval(tpm_policy)
```
stefanberger, line 310: spaces before and after `=` ...

keylime/tpm/tpm_abstract.py
```diff
@@ -307,6 +307,8 @@ def check_pcrs(self, agentAttestState, tpm_policy, pcrs, data, virtual, ima_meas
                     return False
                 pcrs_in_quote.add(pcr_num)
 
+        tpm_policy = ast.literal_eval(tpm_policy)
```
stefanberger, line 310: I think this is actually wrong. Look further above and you will find this here:

```
        try:
            tpm_policy_ = ast.literal_eval(tpm_policy)
        except ValueError:
            tpm_policy_ = {}
```

I think the correct solution is to use `tpm_policy_` below in `if len(list(tpm_policy.keys())) > 0:`. @THS-on ?

THS-on: Using `tpm_policy_` should work but I don't know if it is the actual fix here.
I've looked at the code before the refactoring and there also `tpm_policy` was used which shouldn't work when `tpm_policy` is a string, so this code was never executed before.
https://github.com/keylime/keylime/blob/95ef0415f890b525ab277c37705aba9c22600dff/keylime/tpm/tpm_abstract.py#L292-L298

`len(tpm_policy_.keys()) > 0` will always hold when the policy could be successfully evaluated and contains a mask, so why do we have this check here? I think we can remove it and always warn if we have a PCR in our quote that we haven't used in some way.

The second question what PCR is not in `pcr_in_quote` that causes this now code to executed? Because that might be a regression.

THS-on: Also looking at the failing test we currently assume `tpm_policy`  is always a dict encoded as a string. This seems not always to be the case so we should try to eval it only if it is a string in this part:

https://github.com/keylime/keylime/blob/35d0ae1df571e2719e9338f8ea406d0ed2b4cc6e/keylime/tpm/tpm_abstract.py#L236-L239

Another thing is that the tenant uses `json.dumps(..)` but in the verifier `ast.literal_eval(..)` is used. This might cause problems if the format is not identical.

stefanberger:
@maugustosilva I think the ultimately correct fix is in #727

PR #727, opened 2021-08-02
verifier: convert (v)tpm_policy in DB from string to JSONPickleType

THS-on:
The `tpm_policy` might a dict or a string. Now the `tpm_policy` is converted to a dict only if it is a string.

This also fixes the issue were `tpm_policy.keys()` was called even when `tpm_policy` was a string.

@maugustosilva @stefanberger this should fix https://github.com/keylime/keylime/pull/726 hopefully correctly.


keylime/tpm/tpm_abstract.py
```diff
@@ -232,12 +232,15 @@ def __parse_pcrs(self, pcrs, virtual) -> typing.Dict[int, str]:
...
 
-    def check_pcrs(self, agentAttestState, tpm_policy, pcrs, data, virtual, ima_measurement_list, allowlist, ima_keyring, mb_measurement_list, mb_refstate_str):
-        try:
-            tpm_policy_ = ast.literal_eval(tpm_policy)
-        except ValueError:
-            tpm_policy_ = {}
-        pcr_allowlist = tpm_policy_.copy()
+    def check_pcrs(self, agentAttestState, tpm_policy: typing.Union[str, dict], pcrs, data, virtual,
+                   ima_measurement_list, allowlist, ima_keyring, mb_measurement_list, mb_refstate_str):
+        if isinstance(tpm_policy, str):
```
stefanberger, line 237: My guess is the str originates from `agent['tpm_policy']` all the way down in cloud_verifier_common.py, which is an unconverted string read from the DB table. Maybe that would be the better place to convert, because function on the path up to this function have tpm_policy={} as parameter in the function signature, leaving the impression that it's always a dict. IF a dict shows up here it's probably always empty?

THS-on: We are doing this also for the ima allowlist and exclude list, so implemented it in a similar fashion. 
https://github.com/keylime/keylime/blob/35d0ae1df571e2719e9338f8ea406d0ed2b4cc6e/keylime/ima.py#L138-L142

But I agree with you that this is not the ideal place to convert it from a string to a dictionary. We should probably let sqlalchemy do the conversion for us in the future.

THS-on: I've now moved the DB entries for `tpm_policy` and `vtpm_policy`  to JSONPickleType. Now  `tpm_policy` should always be a dict.

stefanberger: Ok, so even if we aren't doing this now, but `_from_db_obj` seems to be a candidate for a place to convert: https://github.com/keylime/keylime/blob/master/keylime/cloud_verifier_tornado.py#L67

stefanberger:
I think this is the best solution to let the DB driver handle it.

mpeters (approved):

PR #728, opened 2021-08-02
add infrastructure to tag and collect revocation events in Keylime

THS-on:
I've implemented the general tagging infrastructure and converted the IMA checks to demonstrate it's usage.
Before I now convert the rest of the code base I would like to get some feedback for the current implementation.

## Design
Currently Keylime operates in a binary state when it comes failures that cause revocations and does not collect information where and why that revocation happened.

This commit introduces the tagging and collection infrastructure and configuration for revocation events and adding context to them.

SeverityLabel

 * Has a name that can be set in the keylime.conf and a severity that is dynamically calculated based on the order in the configuration.

Components

 * Is a enumeration that contains the main components of Keylime that can cause events. Must be specified when creating a Failure object.

Event

 * Holds a revocation event. An event can be identified by their event_id which has the format: "component.[sub_component].event"
 * The severity is automatically assigned based on the event_id.
 * The event contains a context which is string encoded JSON object.
 * An event must be marked irrecoverable if other validation steps are skipped by early returning.

Failure

 * Holds a collection of events and provides add_event(...) for adding new events to it and merge(...) to merge it with another Failure object.
 * Is False if no events are currently in the Failure object and is otherwise True.

Future changes that cause a revocation must extend and return a Failure object instead of returning a boolean value.

Part of enhancement proposal keylime/enhancements#48

### DIFF
--- a/keylime.conf
+++ b/keylime.conf
@@ -278,6 +278,17 @@ measured_boot_policy_name = accept-all
 # The default value for this config item is the empty string.
 # measured_boot_imports = keylime.elchecking
 
+# Severity labels for revocation events strictly ordered from least severity to
+# highest severtiy.
+severity_labels = ["info", "notice", "warning", "error", "critical", "alert", "emergency"]
+
+# Severity policy that matches different event_ids to the severity label.
+# The rules are evaluated from the beginning of the list and the first match is
+# used. The event_id can also be a regex. Default policy assigns the highest
+# severity to all events.
+severity_policy = [{"event_id": ".*", "severity_label": "emergency"}]
+
+
 #=============================================================================
 [tenant]
 #=============================================================================
--- a/keylime/cloud_verifier_common.py
+++ b/keylime/cloud_verifier_common.py
@@ -19,6 +19,7 @@ from keylime import crypto
 from keylime import ca_util
 from keylime import revocation_notifier
 from keylime.agentstates import AgentAttestStates
+from keylime.failure import Failure, Component
 from keylime.tpm.tpm_main import tpm
 from keylime.tpm.tpm_abstract import TPM_Utilities
 from keylime.common import algorithms
@@ -133,11 +134,12 @@ def init_mtls(section='cloud_verifier', generatedir='cv_ca'):
     return context
 
 
-def process_quote_response(agent, json_response, agentAttestState):
+def process_quote_response(agent, json_response, agentAttestState) -> Failure:
     """Validates the response from the Cloud agent.
 
     This method invokes an Registrar Server call to register, and then check the quote.
     """
+    failure = Failure(Component.QUOTE_VALIDATION)
     received_public_key = None
     quote = None
     # in case of failure in response content do not continue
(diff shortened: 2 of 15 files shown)


stefanberger:
I had to read the enhancement proposal before. What does "Agent B removes Agents from a system if any revocation message with the severity level err is generated" mean? What is Agent B here, is it a validator?

THS-on:
> I had to read the enhancement proposal before. What does "Agent B removes Agents from a system if any revocation message with the severity level err is generated" mean? What is Agent B here, is it a validator?

Agent B is a Keylime agent that is deployed to handle revocation events. So if it gets a revocation message from Keylime about Agent A and it contains the severity level err it will remove Agent A from the system because Agent A is now untrusted.

stefanberger:
> Agent B is a Keylime agent that is deployed to handle revocation events. So if it gets a revocation message from Keylime about Agent A and it contains the severity level err it will remove Agent A from the system because Agent A is now untrusted.

So the assumption is that Agent A has a different code base than Agent B? And Agent B cannot be a rogue agent just taking over for Agent A?

THS-on:
> So the assumption is that Agent A has a different code base than Agent B?

Not necessarily but in our case revocation events are handled by a central software (which contains basically Agent B) that then notifies other systems. The agents like Agent A deployed on the devices that are monitored by Keylime.

>  And Agent B cannot be a rogue agent just taking over for Agent A?

Yes, Agent B cannot be a rogue agent in that model because it is basically outside of the system that is monitored by Keylime.

Those user stories are just an example of that agents can react differently based on the severity level and there are also other use cases where different reactions based on severity are useful.

THS-on:
Agent A is a Keylime agent. Agent B is something custom in our case.

keylime.conf
```diff
@@ -258,6 +258,16 @@ revocation_notifier_port = 8992
...
 max_upload_size = 104857600
 
+# Severity labels for revocation events strictly ordered from least severity to
+# highest severtiy.
+severity_labels = ["info", "notice", "warning", "err", "crti", "alert", "emerg"]
+
+# Severity policy that matches different event_ids to the severity label.
+# The rules are evaluated from the beginning of the list and the first match is
+# used. The event_id can also be a regex.
+severity_policy = [{"event_id": "ima.*", "severity_label": "alert"}, \
```
mpeters, line 268: Is this just a sample severity_policy or the default/recommended one?

keylime/tagging.py
```diff
@@ -0,0 +1,196 @@
+'''
```
mpeters, line 1: This is a bit nit picky (naming things usually is), but I think this module should be named something like `keylime.failure` rather than `keylime.tagging`. Yes we are tagging failures, but it's mainly about failures. Or do you envision tagging other events besides just failures?

mpeters:
Overall, this seems like a good approach to me

keylime.conf
THS-on: Just a sample one for testing. I'll provide one that maps all events to the highest severity in the final PR.

keylime/tagging.py
THS-on: Makes sense. I'll rename it. I think that all events that we tag are failures. The non security relevant parts are part of the normal logging.

THS-on:
> Overall, this seems like a good approach to me

Ok. I'll then start to convert other parts of Keylime to also use this feature.

THS-on:
@maugustosilva Have you had a change to look at this PR? Does it pass your CI?

keylime.conf
```diff
@@ -278,6 +278,17 @@ measured_boot_policy_name = accept-all
 # The default value for this config item is the empty string.
 # measured_boot_imports = keylime.elchecking
 
+# Severity labels for revocation events strictly ordered from least severity to
+# highest severtiy.
+severity_labels = ["info", "notice", "warning", "err", "crit", "alert", "emerg"]
```
stefanberger, line 283: crit -> critical , err -> error , emerg -> emergency ?

THS-on: Makes sense I changed that.

THS-on:
I've rebased on master and included the change for the label names to `keylime.conf`. All other parts haven't changed.

maugustosilva (approved):

THS-on:
@maugustosilva now rebased on latest master.

PR #731, opened 2021-08-05
Refactor keylime_logging module

elfosardo:


mpeters:
Is there a reason for this refactor? Is it just a general cleanup or in preparation for something else?

elfosardo:
@mpeters general cleanup

keylime/keylime_logging.py
```diff
@@ -10,26 +10,38 @@
...
+if not config.REQUIRE_ROOT:
+    LOGSTREAM = './keylime-stream.log'
+else:
+    LOGSTREAM = LOGDIR + '/keylime-stream.log'
+
+logging.config.fileConfig(config.CONFIG_FILE)
+
+
+def set_log_func(loglevel, logger):
+    log_levels = {
```
mpeters, line 26: It's probably not a huge performance difference, but it would actually be better if this `log_levels` structure was created outside of this function instead of every time the function is called as a local structure that then also has to be torn down.

mpeters (changes requested):

keylime/keylime_logging.py
elfosardo: makes sense

elfosardo: @mpeters I just have one doubt on this:
considering that the structure depends on logger, this is not a static structure, so it needs to be generated at least once
I'm not sure I see a different way to do it, unless changing other modules

mpeters: Ah you're correct. So going back to the original code, I know it might not look as nice, but if/then statements are really fast to execute. Either way it's not a big deal performance wise, but it should be faster to just take the old code and factor it out into a function vs creating a onetime use structure and throwing it away on every call.

mpeters (changes requested):

keylime/keylime_logging.py
elfosardo: yeah, I was afraid of this, I did a quick test and the dictionary approach is way slower
I'm going to revert to if/else and move that to a function

mpeters (approved):

Issue #703, opened 2021-07-15
[enhancement] No useful default class to convert mb-refstate into a valid Test


aplanas:
Is your an issue a feature request? If so, please raise it as an [enhancement](https://github.com/keylime/enhancements)

# Environment
Any

# Description 
After researching https://github.com/keylime/keylime/issues/693 is clear that there is no mechanism in Keylime to transform a `mb-refstate` JSON document into a valid Test class.

We should implement one and make it the default, instead of AcceptAll.

maugustosilva:
It is on my to-do list :-)

aplanas:
I am giving a try for the simplest possible parser, using directly the combinators already implemented. Lets see.

aplanas:
Yes, thanks, lets close it!

PR #734, opened 2021-08-19
Solves #703 by adding a "non-trivial" example of a "measured boot policy"

maugustosilva:
The original code for it was provided by mspreitz@us.ibm.com, and it was
slightly adapted by me for testing on a cluster of "generic" Ubuntu Focal
 VMs running with Secure Boot enabled. In addition to the example policy,
there is also an example "measured boot reference state", and a couple of
 comments on `/etc/keylime.conf` for two new `verifier`-specific attributes
 (`measured_boot_imports` and `measured_boot_policy_name`), both disabled
 by default.

### DIFF
--- a/keylime.conf
+++ b/keylime.conf
@@ -258,6 +258,26 @@ revocation_notifier_port = 8992
 # size of the actual payloads
 max_upload_size = 104857600
 
+# Change it to specify the directory (e.g., <KEYLIME_DIR>/keylime/elcheking)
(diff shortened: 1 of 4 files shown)


THS-on:
Thank you for adding the example code. It makes it a lot clearer on how to actually use this feature. 
I'll try to make a simple policy for our systems with that example as a base probably next week.

mpeters (approved):

PR #737, opened 2021-08-24
verifier: add option to send revocation messages via webhook

THS-on:
This introduces the option to also send revocation messages via webhook instead of 0mq. 
It is useful for systems that need to act on revocation messages but are not running the Keylime agent.

### DIFF
--- a/keylime/cloud_verifier_common.py
+++ b/keylime/cloud_verifier_common.py
@@ -365,7 +365,9 @@ def process_get_status(agent):
 
 
 def notify_error(agent, msgtype='revocation', event=None):
-    if not config.getboolean('cloud_verifier', 'revocation_notifier'):
+    send_mq = config.getboolean('cloud_verifier', 'revocation_notifier')
+    send_webhook = config.getboolean('cloud_verifier', 'revocation_notifier_webhook', False)
+    if not (send_mq or send_webhook):
         return
 
     # prepare the revocation message:
@@ -391,7 +393,10 @@ def notify_error(agent, msgtype='revocation', event=None):
 
     else:
         tosend['signature'] = "none"
-    revocation_notifier.notify(tosend)
+    if send_mq:
+        revocation_notifier.notify(tosend)
+    if send_webhook:
+        revocation_notifier.notify_webhook(tosend)
 
 
 def validate_agent_data(agent_data):
--- a/keylime/revocation_notifier.py
+++ b/keylime/revocation_notifier.py
@@ -11,6 +11,7 @@ import os
 import sys
 import signal
 
+import requests
 import simplejson as json
 import zmq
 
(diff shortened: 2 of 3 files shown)


THS-on:
The failing linter action is due to a pylint update. I'll make a separate PR that fixes those.

maugustosilva:
This is very good and welcome. I've been meaning to so something about it myself. Would mind holding it until we discuss some sort of "multi-backend" for notifications? My initial idea was to have as options, `nothing` (i.e., disable notifications altogether), `0mq`, `webhook`.

THS-on:
Some generic way of implementing and managing those notifications backends is a good idea. 
I think we can discuss how we handle information output from Keylime to other systems in general once I have fully implemented the revocation context changes.

We currently just needed an easy way to get those messages into our system.

keylime/cloud_verifier_common.py
```diff
@@ -341,7 +341,9 @@ def process_get_status(agent):
 
 
 def notify_error(agent, msgtype='revocation'):
-    if not config.getboolean('cloud_verifier', 'revocation_notifier'):
+    send_mq = config.getboolean('cloud_verifier', 'revocation_notifier')
+    send_webhook = config.getboolean('cloud_verifier', 'revocation_notifier_webhook')
```
stefanberger, line 345: I think we need a 3rd parameter here, which is the default, in case someone has an old keylime.conf after package upgrade....

stefanberger: The default should be 'False'.

keylime/revocation_notifier.py
```diff
@@ -79,6 +80,36 @@ def worker(tosend):
     t.start()
 
 
+def notify_webhook(tosend):
+    url = config.get('cloud_verifier', 'webhook_url')
```
stefanberger, line 84: Also here, in case one has an older keylime.conf, this probably throws an error. We need the third parameter providing the default ('').

keylime/cloud_verifier_common.py
maugustosilva: @THS-on please address @stefanberger comment, so I can test and merge

keylime/revocation_notifier.py
maugustosilva: @THS-on please address @stefanberger comment, so I can test and merge

THS-on: I'll do that. Do we want to merge the feature as is and replace it in the future with something more generic 
or should we keep this just open as a reference?

keylime/cloud_verifier_common.py
THS-on: done

keylime/revocation_notifier.py
maugustosilva: Let's merge this one, and re-work later.

THS-on: Ok, all requested changes should now be implemented.

maugustosilva (approved):

keylime/revocation_notifier.py
```diff
@@ -79,6 +80,36 @@ def worker(tosend):
...
+                    break
+
+                logger.debug(f"Unable to publish revocation message {i} times via webhook, "
+                             f"trying again in {retry_interval} seconds. "
+                             f"Server returned status code: {response.status_code}")
+            except requests.exceptions.RequestException as e:
+                logger.debug(f"Unable to publish revocation message {i} times via webhook, "
+                             f"trying again in {retry_interval} seconds: {e} ")
+
+            time.sleep(config.getfloat('cloud_verifier', 'retry_interval'))
```
stefanberger, line 106: use `retry_interval` from above?

THS-on: Yes. Thanks for catching that.

PR #738, opened 2021-08-24
Fixing pylint erros

THS-on:
This fixes W1514, W0612 and R1735 pylint errors.

W1514 enforces that an encoding is specified when open is used. 

I defaulted to "utf-8" and  used "ascii" or byte encoding where I was relative sure that this is correct. 
Someone else should also look at those changes to make sure that I haven't made wrong assumptions.


maugustosilva (approved):
This passes my most comprehensive end-to-end testing (in addition to the regular unit test). I see no reason to avoid merging it. LGTM

mpeters (approved):

PR #742, opened 2021-08-26
policies: add Simple policy

### LINKED ISSUE #703: [enhancement] No useful default class to convert mb-refstate into a valid Test

aplanas:
The Simple policy is a direct implementation of the different test
combinators already present in the "tests" modules.

This policy is serialized in a JSON file, that is passed via the
constructor of the class, and parsed into the different tests
combinators via the "refstate_to_test" method.

Fix #703

### DIFF
--- a/keylime/elchecking/policies.py
+++ b/keylime/elchecking/policies.py
@@ -1,5 +1,6 @@
 import abc
 import importlib
+import json
 import typing
 
 from keylime import config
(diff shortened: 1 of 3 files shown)


maugustosilva:
@aplanas any updates on this, after the merging of #734?

aplanas:
> @aplanas any updates on this, after the merging of #734?

Actually yes. This PR needs to be changed as the example policy cannot be expressed, with the current parser, as a simple JSON document. I am still figuring out the changes required in the new JSON format, so the policy in the example can be fully expressed.

aplanas:
After further inspection I think that this approach is not good enough.

Closing!

PR #747, opened 2021-09-02
Replace m2crypto with python-cryptography 

sergio-correia:
python-cryptography is already used in parts of the project, so it would
make sense to not use multiple libraries that can do the same thing.

It is also better maintained than M2Crypto and having one less dependency
is always a good thing.

Test coverage slightly improved as well.


keylime/ca_impl_cfssl.py
```diff
@@ -200,18 +232,25 @@ def gencrl(serials, cert, ca_pk):
     try:
         # need to temporarily write out the private key with no password
         # to tmpfs
-        priv_key = os.path.abspath("%s/ca-key.pem" % secdir)
-        with open(priv_key, 'w', encoding="utf-8") as f:
+        privkey_path = os.path.abspath("%s/ca-key.pem" % secdir)
```
stefanberger, line 235: use `{secdir}` to be consistent?

keylime/ca_impl_cfssl.py
```diff
@@ -200,18 +232,25 @@ def gencrl(serials, cert, ca_pk):
...
+        with open(privkey_path, 'w', encoding="utf-8") as f:
             f.write(ca_pk)
-        cmdline = " -ca-key %s -ca cacert.crt" % (priv_key)
+
+        cacert_path = os.path.abspath("%s/cacert.crt" % secdir)
```
stefanberger, line 239: same here?

stefanberger (changes requested):

keylime/ca_impl_cfssl.py
sergio-correia: Thanks, this was updated.

stefanberger:
> Test coverage slightly improved as well.

Would be good if we could show the test coverage via coveralls.io some day...

Otherwise the PR looks good to me.

maugustosilva (approved):

PR #751, opened 2021-09-10
By default use the kernel resource manager instead of tpm2-abrmd

THS-on:
The linux kernel starting with 4.12 ships with its own resource manager  which makes tpm2-abrmd not needed anymore for most use cases. This changes the installer script to not install tpm2-adrmd and changes the default to `device:/dev/tpmrm0`.

We already dropped the tpm2-abrmd dependency for the Debian package by setting that value manually.


stefanberger:
I previously had tried to also run things inside a container without the tpm2-abrmd but had the impression that tools around quoting did require the tpm2-abrmd. So you are sure this works without it?

THS-on:
If you are running a TPM emulator on the same system as the agent is running it might be still useful because there the kernel doesn't provide a resource manager.
Otherwise I've been testing it with libvirt + swtpm and it works as intended.

stefanberger:
Patch LGTM.

mpeters (approved):
lgtm

PR #752, opened 2021-09-13
Simulating use of SSLContext.minimum_version on ssl v3.6

mirespace:
Hi all,

keylime code is using SSL library version 3.7+ but we have 3.6 in Ubuntu Focal. So, the idea is to adapt the needed calls or use of functions and constants to 3.6 with the same functionality. Following this https://docs.python.org/3/library/ssl.html#ssl.OP_NO_TLSv1 and https://docs.python.org/3/library/ssl.html#ssl.create_default_context,  here is a proposal.

We tested that services are up and running after the change.

Any suggestions or concerns are welcome... Thanks in advance.

### DIFF
--- a/keylime/cloud_verifier_common.py
+++ b/keylime/cloud_verifier_common.py
@@ -9,6 +9,7 @@ import os
 import ssl
 import socket
 import time
+import sys
 
 import simplejson as json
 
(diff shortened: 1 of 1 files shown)


keylime/cloud_verifier_common.py
```diff
@@ -114,7 +115,10 @@ def init_mtls(section='cloud_verifier', generatedir='cv_ca'):
 
     try:
         context = ssl.create_default_context(ssl.Purpose.CLIENT_AUTH)
-        context.minimum_version = ssl.TLSVersion.TLSv1_2
+        if sys.version_info >= (3,7):
+            context.minimum_version = ssl.TLSVersion.TLSv1_2
+        else:
+            context.options &= ~ssl.OP_NO_TLSv1_2
```
utkarsh2102, line 121: For keylime maintainers: I am wondering if we cannot default/target to Python 3.6 so we don't have these inconsistencies?

THS-on:
Validated `nmap --script ssl-enum-ciphers -p 8881 VERIFIER_IP` that the patch indeed disables TLS1.1.

keylime/cloud_verifier_common.py
maugustosilva: Not sure if it is a viable strategy, giving the uses in production.

maugustosilva (approved):

keylime/cloud_verifier_common.py
mpeters: Yeah I agree with @maugustosilva we can't restrict it to sub versions of python. We can do things like requiring a minimum version, but not a specific version.

Issue #757, opened 2021-09-15
Revocation actions fail because the cert is read as binary but encoded as text


mpeters:
# Description 

Revocation actions fail on local nodes because it isn't properly handling the cert. A recent change (https://github.com/keylime/keylime/commit/2ba03e9a3daf10fb4b39a68f2c9b556cc99d20d4#diff-bf2d7716986f3aca3950920e063c519a9306b090d95d900bfad8bd5f4c7e6799R111) made reading the cert as binary but some code later one tries to encode it as text.

PR #758, opened 2021-09-15
Fix bug #757 where revoc cert was treated as text

mpeters:
Recent changes to make opens more explicit now open
the revocation cert as binary, so no need to encode it as text.

### DIFF
--- a/keylime/crypto.py
+++ b/keylime/crypto.py
@@ -44,8 +44,7 @@ def rsa_import_privkey(privkey):
 
 
 def x509_import_pubkey(pubkey):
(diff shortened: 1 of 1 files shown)


mpeters:
@THS-on this fixed the issue for me from your encoding changes and it's the only place this code is called. Just want you to give it a lookover to make sure it's good.

THS-on:
Do we want to do the same as with `rsa_import_privkey` where we first try with bytes and if that fails trying it at as a string?
We should probably add some type annotations if we expect the certificates to be bytes.

maugustosilva (approved):

PR #763, opened 2021-09-23
Tenant commands return json

ljmcgann:
Updates tenant commands so they return a json
response, as opposed to only writing to a log


keylime/tenant.py
```diff
@@ -624,15 +624,17 @@ def do_status(self, listing=False, returnresponse=False, bulk=False):
...
+        if listing:
+            verifier_id = ""
+            if self.verifier_id is not None:
+                verifier_id = self.verifier_id
             response = do_cvstatus.get(
                 (f'/v{self.api_version}/agents/?verifier={verifier_id}'),
                 cert=self.cert,
                 verify=False
             )
+            response = response.json()
```
mpeters, line 637: I'm not sure we can discard the original `response` object for it's JSON payload so soon. It will work in the cases where everything is up and running correctly, but in failure modes we have to be able to check the HTTP response object before we can rely on there even being a JSON payload. We can pull the JSON out after we've checked the return codes, or at least with some try/catches in case the methods like `json()` fail. And we shouldn't override the `response` object since that's the only way we can guarantee to check return codes.

keylime/tenant.py
```diff
@@ -642,66 +644,77 @@ def do_status(self, listing=False, returnresponse=False, bulk=False):
...
         else:
             response = do_cvstatus.get(
                 (f'/v{self.api_version}/agents/{agent_uuid}'),
                 cert=self.cert,
                 verify=False
             )
+            response = response.json()
```
mpeters, line 654: This part should happen after we check the return codes. It's ok to create the object we will eventually return here and then augment it in the following section.

keylime/tenant.py
```diff
@@ -642,66 +644,77 @@ def do_status(self, listing=False, returnresponse=False, bulk=False):
...
+            res = response.pop('results')
+            response['results'] = {agent_uuid: res}
 
-        if response.status_code == 503:
```
mpeters, line 652: We can't always swap the HTTP status code with the JSON's `code` value because there might not be a JSON payload in certain failures and we still need to handle those.

keylime/tenant.py
```diff
@@ -642,66 +644,77 @@ def do_status(self, listing=False, returnresponse=False, bulk=False):
...
+                states.state_to_str(states.REGISTERED)
+            return response
+        elif response['code'] == 200:
+            if not listing:
+                if not bulk:
+                    operational_state = states.state_to_str(
+                        response['results'][agent_uuid]['operational_state'])
+                    response['results'][agent_uuid]['operational_state'] = operational_state
+
+                    logger.info("Agent Info:\n%s" % json.dumps(response["results"]))
```
mpeters, line 691: Why not return the response at the end of the 200 code block?

keylime/tenant.py
```diff
@@ -642,66 +644,77 @@ def do_status(self, listing=False, returnresponse=False, bulk=False):
...
+            logger.error("Status command response: %s. Unexpected response from Cloud Verifier.", response['code'])
+            return response
 
     def do_cvdelete(self, verifier_check=True):
         """Delete agent from Verifier."""
         if verifier_check:
-            agent_json = self.do_status(listing=False, returnresponse=True)
+            agent_json = self.do_status()
+            if agent_json['code'] != 200:
+                logger.error("Could not get status of agent %s from "
```
mpeters, line 713: If the `"verifier."` string is hardcoded, why not put it into the original string? So it would have something like `verifier.%s`.

keylime/tenant.py
```diff
@@ -1252,19 +1270,19 @@ def main(argv=sys.argv):
...
-        mytenant.do_status()
+        return mytenant.do_status()
     elif args.command == 'bulkinfo':
-        mytenant.do_status(bulk=True)
+        return mytenant.do_status(bulk=True)
     elif args.command == 'list':
-        mytenant.do_status(listing=True)
+        return mytenant.do_status(listing=True)
     elif args.command == 'reactivate':
         mytenant.do_cvreactivate(args.verifier_check)
```
mpeters, line 1281: Why not return the JSON here as well?

mpeters (changes requested):
Thanks for your fixes ljmcgann. I think it's a good start. Just some minor changes requested.

keylime/tenant.py
maugustosilva: I agree, we should deal with `response` object internally as is, and only printout `json` at the end.

PR #768, opened 2021-10-08
ima: Use seek() to get to the last known last entry

stefanberger:
Use seek() to get to the last known last entry in the IMA measurement
list. Track the size of the IMA measurement list two last known number
of entries in a singleton class that helps us seek to the last entry
and return the tail of the measurement list at the n-th entry as re-
quested by the caller.

We have to remember the last two known number of entries and file sizes
since the validator may not get a quote containing the last entry but
that of the previous last entry and will on the next attestation request
again ask for the same entry.

### DIFF
--- a/keylime/ima.py
+++ b/keylime/ima.py
@@ -30,23 +30,46 @@ logger = keylime_logging.init_logging('ima')
 ALLOWLIST_CURRENT_VERSION = 3
 
 
-def get_from_nth_entry(filedata, nth_entry):
-    """ Get the measurement list starting at the n-th entry, starting with 0.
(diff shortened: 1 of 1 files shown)


stefanberger:
The reasons given in #769 drove the changes to this PR after some more testing. The 'key' in testing was to trigger the situation where the IMA log has n entries but the TPM hasn't been extended with the n-th entry yes, so only has n-1 PCR extensions. In this case the verifier will ask for the (n-1)-th entry again and we need to know the offset in the IMA measurement log where that was, so need to remember that offset along the n-th entry's offset.

PR #775, opened 2021-10-14
scripts: Implement create_policy to create the JSON allowlist from files

stefanberger:
Using either the local IMA ascii_runtime_measurements file or a file with
the same contents, extract keyring and critical data from it and add key
value pairs to their respective maps in the JSON-formatted allowlist.
Similarly, either read a flat-file allowlist and add its entries to the
JSON-formatted allowlist hashes map or read entries from the local IMA
ascii_runtime_measurement, or a file with the same contents, and add
filename - hash pairs to the hashes map.

### DIFF
--- a/scripts/create_policy
+++ b/scripts/create_policy
@@ -1,0 +1,197 @@
+#!/usr/bin/env python3
+
+""" Create a JSON allowlist/policy from given files as input """
+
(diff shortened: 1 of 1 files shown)


stefanberger:
This should go into after #767

PR #785, opened 2021-11-08
Clean up CI scripts and rewrite run_local.sh

mbestavros:
run_local.sh has been setting up its `swtpm` differently from how the Keylime test suite works on Github Actions for a long while now.

Notably, the old approach required privileged containers, while the newer GHA approach does not. This PR rewrites run_local.sh to mirror the GHA setup.


.ci/test_wrapper.sh
```diff
@@ -1,38 +1,38 @@
...
-sed -i 's/.*ExecStart.*/ExecStart=\/usr\/sbin\/tpm2-abrmd --tcti=mssim/' /usr/lib/systemd/system/tpm2-abrmd.service
-systemctl daemon-reload
-systemctl enable tpm2-abrmd
-systemctl start tpm2-abrmd
-# Check that tpm2-abrmd is actually running as starting it won't report failures
-if ! (systemctl -q is-active tpm2-abrmd); then
-    echo "tpm2-abrmd failed to start, exiting"
-    exit 1
-fi
+tpm2-abrmd \
```
mpeters, line 24: @lkatalin or @maugustosilva Do you know if the tpm2-abrmd stuff is needed? I thought newer kernels have their own internal resource manager for this kind of stuff.

lkatalin: @mpeters I don't believe it's necessary unless you are using swtpm with the `socket` option. I've been using it with `swtpm chardev` instead and with the kernel resource manager `/dev/tpmrm0`:

```
modprobe tpm_vtpm_proxy
swtpm_setup --tpm2 \
          --tpmstate /var/vtpm \
          --createek --decryption --create-ek-cert \
          --create-platform-cert \
          --display \
swtpm chardev --tpm2 --vtpm-proxy --tpmstate dir=/var/vtpm -d
export TPM2TOOLS_TCTI="device:/dev/tpmrm0"
export TCTI="device:/dev/tpmrm0"
```

If you do it this way, you also don't need the `dbus-daemon`.

Note: this assumes it is running in a VM without a hardware tpm attached (otherwise I think the device defaults to `/dev/tpmrm1` with the swtpm since the hardware tpm is always tpm0, and this doesn't work well for reasons that I forget).

lkatalin: As for whether swtpm socket or swtpm chardev will work better for testing, that I am not sure.

mpeters: Ok, we'll just go with socket and abrmd for now

mpeters (approved):

PR #789, opened 2021-11-10
Replace simplejson with built-in json module

sergio-correia:
Scipts and docs also updated to reflect that `simplejson` is no longer required.

### DIFF
--- a/auto-ipsec/libreswan/src/local_action_crashsa.py
+++ b/auto-ipsec/libreswan/src/local_action_crashsa.py
@@ -11,8 +11,7 @@ import keylime.secure_mount as secure_mount
 import keylime.config as common
 import keylime.keylime_logging as keylime_logging
 import keylime.cmd_exec as cmd_exec
-
-import simplejson as json
+from keylime import json
 
 # read the config file
 config = common.get_config()
--- a/auto-ipsec/racoon/src/local_action_deletesa.py
+++ b/auto-ipsec/racoon/src/local_action_deletesa.py
@@ -17,8 +17,7 @@ import keylime.config as common
 import keylime.keylime_logging as keylime_logging
 import keylime.cmd_exec as cmd_exec
 import keylime.ca_util as ca_util
-
-import simplejson as json
+from keylime import json
 
 # read the config file
 config = common.get_config()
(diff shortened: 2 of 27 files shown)


README.md
```diff
@@ -273,17 +273,6 @@ You're finally ready to install keylime!
 sudo python3 -m pip install . -r requirements.txt
 ```
 
-#### To run on OSX 10.11+
```
mpeters, line 276: Did you remove this section because it's no longer necessary or because it doesn't work any more?

doc/cloud-agent-tarball-notes.md
```diff
@@ -19,7 +19,7 @@ You can build a single binary for the keylime_agent service.  It uses http://www
 Make sure that you have UPX for binary compression.  On ubuntu: `apt-get install upx-ucl`.
 
 Ensure that you have the tools needed to install keylime normally (see section above). On Ubuntu:
-`apt-get install -y python-dev python-setuptools python-tornado python-m2crypto python-zmq`.  Now pull in the rest of the python dependencies with `sudo python setup.py install`
```
mpeters, line 22: This seems like things that were left over from https://github.com/keylime/keylime/pull/747 is that correct? If so, can you add a note to your commit message that you're also removing older references to m2crypto that are no longer accurate?

mpeters (changes requested):
Small changes to the commit message to mention m2crypto

mpeters:
Overall looks good, just a small request to the commit message

doc/cloud-agent-tarball-notes.md
sergio-correia: Yeah, this is in the 2nd commit, that updates scripts and docs, and it has a note that it's also removing that leftover. Let me know if you want me to expand on the note, or even do it in separate commits.

README.md
sergio-correia: The whole section was about building m2crypto in OSX, so it looked like it could be removed, since that was not required anymore.

mpeters:
Separate commits are best if it's not too much trouble.

sergio-correia:
> Separate commits are best if it's not too much trouble.

I updated the second commit to only deal with `simplejson`. I will submit a follow-up PR cleaning up the leftovers from #747.

PR #795, opened 2021-11-18
Implement agent-local revocation mechanism

ueno:
This adds support for the agent-local revocation mechanism as described in the [enhancements#53](https://github.com/keylime/enhancements/blob/master/53_agent_local_revocation.md). It's currently draft state, as it lacks tests and protocol version bump.

### DIFF
--- a/keylime/cloud_verifier_tornado.py
+++ b/keylime/cloud_verifier_tornado.py
@@ -4,6 +4,7 @@ import os
 import signal
 import sys
 import traceback
+from concurrent.futures import ThreadPoolExecutor
 from multiprocessing import Process
 
 import tornado.ioloop
@@ -29,6 +30,9 @@ from keylime.elchecking import policies
 from keylime.failure import MAX_SEVERITY_LABEL, Component, Failure
 
 logger = keylime_logging.init_logging("cloudverifier")
+
+# mTLS configuration to connect to the agent
+mtls_options = None
 
 
 try:
@@ -227,11 +231,6 @@ class VersionHandler(BaseHandler):
 
 
 class AgentsHandler(BaseHandler):
-    mtls_options = None  # Stores the cert, key and password used by the verifier for mTLS connections
-
-    def initialize(self, mtls_options):
-        self.mtls_options = mtls_options
-
     def head(self):
         """HEAD not supported"""
         web_util.echo_json_response(self, 405, "HEAD not supported")
@@ -491,9 +490,7 @@ class AgentsHandler(BaseHandler):
                         mtls_cert = agent_data["mtls_cert"]
                         agent_data["ssl_context"] = None
                         if agent_mtls_cert_enabled and mtls_cert:
-                            agent_data["ssl_context"] = web_util.generate_agent_mtls_context(
-                                mtls_cert, self.mtls_options
-                            )
+                            agent_data["ssl_context"] = web_util.generate_agent_mtls_context(mtls_cert, mtls_options)
 
                         if agent_data["ssl_context"] is None:
                             logger.warning("Connecting to agent without mTLS: %s", agent_id)
@@ -561,7 +558,7 @@ class AgentsHandler(BaseHandler):
                 if not isinstance(agent, dict):
                     agent = _from_db_obj(agent)
                 if agent["mtls_cert"]:
-                    agent["ssl_context"] = web_util.generate_agent_mtls_context(agent["mtls_cert"], self.mtls_options)
+                    agent["ssl_context"] = web_util.generate_agent_mtls_context(agent["mtls_cert"], mtls_options)
                 agent["operational_state"] = states.START
                 asyncio.ensure_future(process_agent(agent, states.GET_QUOTE))
                 web_util.echo_json_response(self, 200, "Success")
@@ -855,6 +852,70 @@ async def invoke_provide_v(agent):
         asyncio.ensure_future(process_agent(agent, states.GET_QUOTE))
 
 
+async def invoke_notify_error(agent, tosend):
+    if agent is None:
+        logger.warning("Agent deleted while being processed")
+        return
+    kwargs = {
+        "data": tosend,
+    }
+    if agent["ssl_context"]:
+        kwargs["context"] = agent["ssl_context"]
+    res = tornado_requests.request(
+        "POST",
+        f"http://{agent['ip']}:{agent['port']}/v{agent['supported_version']}/notifications/revocation",
+        **kwargs,
+    )
+    response = await res
+
+    if response is None:
+        logger.warning(
+            "Empty Notify Revocation response from cloud agent %s",
+            agent["agent_id"],
+        )
+    elif response.status_code != 200:
+        logger.warning(
+            "Unexpected Notify Revocation response error for cloud agent %s, Error: %s",
+            agent["agent_id"],
+            response.status_code,
+        )
+
+
+async def notify_error(agent, msgtype="revocation", event=None):
+    notifiers = revocation_notifier.get_notifiers()
+    if len(notifiers) == 0:
+        return
+
+    tosend = cloud_verifier_common.prepare_error(agent, msgtype, event)
+    if "webhook" in notifiers:
+        revocation_notifier.notify_webhook(tosend)
+    if "zeromq" in notifiers:
+        revocation_notifier.notify(tosend)
+    if "agent" in notifiers:
+        verifier_id = config.get(
+            "cloud_verifier", "cloudverifier_id", fallback=cloud_verifier_common.DEFAULT_VERIFIER_ID
+        )
+        session = get_session()
+        agents = session.query(VerfierMain).filter_by(verifier_id=verifier_id).all()
+        futures = []
+        loop = asyncio.get_event_loop()
+        # Notify all agents asynchronously through a thread pool
+        with ThreadPoolExecutor() as pool:
+            for agent_db_obj in agents:
+                if agent_db_obj.agent_id != agent["agent_id"]:
+                    agent = _from_db_obj(agent_db_obj)
+                    if agent["mtls_cert"]:
+                        agent["ssl_context"] = web_util.generate_agent_mtls_context(agent["mtls_cert"], mtls_options)
+                func = functools.partial(invoke_notify_error, agent, tosend)
+                futures.append(await loop.run_in_executor(pool, func))
+            # Wait for all tasks complete in 60 seconds
+            try:
+                for f in asyncio.as_completed(futures, timeout=60):
+                    await f
+            except asyncio.TimeoutError as e:
+                logger.error("Timeout during notifying error to agents: %s", e)
+
+
 async def process_agent(agent, new_operational_state, failure=Failure(Component.INTERNAL, ["verifier"])):
     # Convert to dict if the agent arg is a db object
     if not isinstance(agent, dict):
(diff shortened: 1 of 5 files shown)


kkaarreell:
Hi @ueno , I think it would be good to have e2e test for this feature. I can create one but just want to confirm. Is changing the settings in `revocation_notifiers` the only thing to be done?

ueno:
> Is changing the settings in revocation_notifiers the only thing to be done?

Yes, I think so. Let me finalize this PR in this week so it can actually be tested.

ansasaki (changes requested):

keylime/cloud_verifier_tornado.py
```diff
@@ -884,6 +884,32 @@ async def invoke_provide_v(agent):
...
+    if len(notifiers) == 0:
+        return
+
+    tosend = cloud_verifier_common.prepare_error(agent, msgtype, event)
+    if 'zeromq' in notifiers:
+        revocation_notifier.notify(tosend)
+    if 'webhook' in notifiers:
+        revocation_notifier.notify_webhook(tosend)
+    if 'agent' in notifiers:
+        await invoke_notify_error(tosend)
```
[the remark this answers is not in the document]
ueno: Thanks, fixed.

ueno:
I have no idea why this breaks webhook integration tests, does anyone have any clue?
Also we might need to suppress code scanning warnings on logging certificates.

kkaarreell:
> I have no idea why this breaks webhook integration tests, does anyone have any clue? Also we might need to suppress code scanning warnings on logging certificates.

webhook notifier is not set for some reason. I have added the debug output and only zeromq notifier is in the list.

kkaarreell:
So this is because you have removed `revocation_notifier_webhook` from `keylime.conf` and the function `limeUpdateConf` is only changing existing option, not adding a missing one. I think we need to update the test to set `revocation_notifiers` accordingly.

ueno:
Thanks, but this PR [keeps](https://github.com/keylime/keylime/pull/795/files#diff-bf2d7716986f3aca3950920e063c519a9306b090d95d900bfad8bd5f4c7e6799R32) `revocation_notifier_webhook` for backward compatibility. So if the fallback doesn't take place, I guess that would be an issue in this PR.

ueno:
Ah! That explains it, thanks a lot Karel!

ueno:
@THS-on any suggestion on how to overcome [this](https://github.com/keylime/keylime/pull/795/checks?check_run_id=6687939352) in CodeQL scanning? I don't think revealing certificate path through logging is a problem, but it seems I can't suppress it through `# lgtm` comment (as pre-commit prohibit too long lines...)

THS-on (approved):
LGTM. The only worry that I have is that an rouge agent can block the async io loop by not responding. Can you test what happens if a agent just never terminates the connection?

ueno:
> The only worry that I have is that an rouge agent can block the async io loop by not responding. Can you test what happens if a agent just never terminates the connection?

That's a good point, thanks for raising it; I'll test it in modified agent with a new e2e test plan. Given that the existing notification mechanism (ZeroMQ) does not guarantee consistency, perhaps we could wrap the `invoke_notify_error` (async) call in a [task](https://docs.python.org/3/library/asyncio-task.html#creating-tasks) to prioritize other awaitables?

THS-on:
> Given that the existing notification mechanism (ZeroMQ) does not guarantee consistency, perhaps we could wrap the invoke_notify_error (async) call in a [task](https://docs.python.org/3/library/asyncio-task.html#creating-tasks) to prioritize other awaitables?

Ideally we would run them in another thread.

THS-on:
> @THS-on, ok I've [wrapped](https://github.com/keylime/keylime/pull/795/commits/0ae060d470b8d9e453b86fca658feee7859e1d9d#diff-67033196f7dec303b3ea55793c6c13fb4b03a72a5515239475a871c3d638dfe4R894-R904) it with threading.Thread (let me know if there is a better approach).

I think it is fine for now.

With zeroMQ all agents are notified when another agent fails. In the current implementation only the failing agent gets message, correct? Is that what we want? I would expect all agents to get the message.

keylime/cloud_verifier_tornado.py
```diff
@@ -856,6 +857,53 @@ async def invoke_provide_v(agent):
...
+        raise Exception("Agent deleted while being processed")
+    res = tornado_requests.request(
+        "POST",
+        f"http://{agent['ip']}:{agent['port']}/v{agent['supported_version']}/notifications/revocation",
+        data=tosend,
+        context=agent["ssl_context"],
+    )
+    response = await res
+
+    if response.status_code != 200:
```
THS-on, line 871: response might be `None`.

keylime/cloud_verifier_tornado.py
```diff
@@ -856,6 +857,53 @@ async def invoke_provide_v(agent):
         asyncio.ensure_future(process_agent(agent, states.GET_QUOTE))
 
 
+async def invoke_notify_error(agent, tosend):
+    if agent is None:
+        raise Exception("Agent deleted while being processed")
```
THS-on, line 862: Which part should handle this exception?

keylime/cloud_verifier_tornado.py
```diff
@@ -856,6 +857,53 @@ async def invoke_provide_v(agent):
...
+    if len(notifiers) == 0:
+        return
+
+    tosend = cloud_verifier_common.prepare_error(agent, msgtype, event)
+    if "webhook" in notifiers:
+        revocation_notifier.notify_webhook(tosend)
+    if "zeromq" in notifiers:
+        revocation_notifier.notify(tosend)
+    if "agent" in notifiers:
+        loop = asyncio.new_event_loop()
```
THS-on, line 895: Why is the loop created here and not in the worker?

ueno:
> With zeroMQ all agents are notified when another agent fails. In the current implementation only the failing agent gets message, correct? Is that what we want? I would expect all agents to get the message.

That's an interesting point I haven't thought about. Perhaps this use-case could be implemented by enabling both "agent" and "webhook" notifiers?

THS-on:
> Perhaps this use-case could be implemented by enabling both "agent" and "webhook" notifiers?

Not really. "webhook" is intended for relaying the information to a single source outside of the Keylime system. Revocation actions are useful for other agents to act on the state change e.g. dropping all connections to that agent. The current implementation just sends this information to the compromised agent which is only useful in some cases.

ueno:
I see; the latest commits should address that (i.e., all the registered agents are notified). I've also changed the usage of threads from manually creating it to use `ThreadPoolExecutor`.

@THS-on could you take a look again?

keylime/cloud_verifier_tornado.py
ueno: Added a check

keylime/cloud_verifier_tornado.py
ueno: Yeah, just converted it to a warning.

keylime/cloud_verifier_tornado.py
ueno: Fixed.

THS-on:
@ueno in general LGTM. Thank you for the changes.

I'll try to test it till the end of the week manually.

THS-on:
During testing I ran into this error message:

```
Jun 18 15:24:49 fedora keylime_verifier[946]: /usr/lib64/python3.10/asyncio/base_events.py:1881: RuntimeWarning: coroutine 'invoke_notify_error' was never awaited
Jun 18 15:24:49 fedora keylime_verifier[946]:   handle = self._ready.popleft()
Jun 18 15:24:49 fedora keylime_verifier[946]: RuntimeWarning: Enable tracemalloc to get the object allocation traceback
```

Also can you check on how timeouts are handled if the agent is not available anymore?

ueno:
> During testing I ran into this error message:
> 
> ```
> Jun 18 15:24:49 fedora keylime_verifier[946]: /usr/lib64/python3.10/asyncio/base_events.py:1881: RuntimeWarning: coroutine 'invoke_notify_error' was never awaited
> Jun 18 15:24:49 fedora keylime_verifier[946]:   handle = self._ready.popleft()
> Jun 18 15:24:49 fedora keylime_verifier[946]: RuntimeWarning: Enable tracemalloc to get the object allocation traceback
> ```

Thanks for checking. I was confused with the `loop.run_in_executor()` usage: the result of this function needs to be awaited, as well as `asyncio.as_complete()` has to to be called to wait with timeout (and the results should also be awaited). Those should be fixed now; I have tested with the 4-agent setup described in [keylime-tests](https://github.com/RedHat-SP-Security/keylime-tests/blob/main/TESTING.md#running-multi-host-tests).

> Also can you check on how timeouts are handled if the agent is not available anymore?

Currently, all notification tasks share the same timeout (60 sec).

keylime/cloud_verifier_tornado.py
```diff
@@ -855,6 +866,66 @@ async def invoke_provide_v(agent):
...
 
 
+async def invoke_notify_error(agent, tosend):
+    if agent is None:
+        logger.warning("Agent deleted while being processed")
+        return
+    kwargs = {
+        "data": tosend,
+    }
+    if agent["ssl_context"]:
```
THS-on, line 876: This works only if all the agents event loops are running on a single process. If not you will need to recreate the contexts using `web_util.generate_agent_mtls_context(...)`.

keylime/cloud_verifier_tornado.py
```diff
@@ -62,10 +63,14 @@ def get_AgentAttestStates():
     "pcr10": "",
     "next_ima_ml_entry": 0,
     "learned_ima_keyrings": {},
+    # filled later from ssl_contexts
     "ssl_context": None,
 }
 
 
+ssl_contexts = {}
```
THS-on, line 71: This does not work when Keylime is running on multiple processes. See other comment.

THS-on (changes requested):
There is still one issue with the ssl_context, but the rest of the mechanism works nicely.

keylime/cloud_verifier_tornado.py
ueno: Thanks for pointing that out. I've removed the caching but added context recreation for connecting to other agents.

THS-on (changes requested):
One small change then it fully works.

keylime/keylime_agent.py
```diff
@@ -223,6 +219,25 @@ def do_POST(self):
...
+        except Exception as e:
+            logger.warning("POST returning 400 response, could not parse body data: %s", e)
+            web_util.echo_json_response(self, 400, "content is invalid")
+            return
+
+        if "notifications" in rest_params:
+            if rest_params["notifications"] == "revocation":
+                revocation_notifier.process_revocation(
+                    json_body, perform_actions, cert_path=self.server.revocation_cert_path
+                )
```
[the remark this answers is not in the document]
ueno: Good catch, fixed

THS-on (approved):

ueno:
Let me remove the last commit ("Add new test plan that exercises agent-local revocation") which refers to my fork of keylime-tests for testing. Once this PR is merged we can update the upstream keylime-tests repo and then add the test plan back.

kkaarreell:
Hello, is there some work remaining blocking the merge? Since the zeromq removal is important for RHEL-9 packaging and testing, I am trying to find out whether we can expect this change to land within days or later. Thank you.

THS-on:
@kkaarreell there should be nothing blocking this PR. Ideally someone besides me would also review it. @mpeters, @maugustosilva? If it is time sensitive I can merge it, but @mpeters needs to create a new release.

kkaarreell:
It is not time sensitive. I can "workaround it", it just won't be worth if if this PR would be merge in few days.

PR #799, opened 2021-11-30
agent: drop openstack uuid option

THS-on:
The current code is not generic enough to be useful.


mpeters:
I'm not sure why we would drop this. In fact, I was thinking of expanding it to other "cloudish" providers like AWS, GCE and Azure. It can be useful is the agent UUID is also tied to the broader inventory of virtual machines you have and the unique ID from your cloud is a good one to use.

THS-on:
The issue is not the functionality itself, but the implementation which has no configuration options to make it useful.
I think providing a way to fetch the UUID from a third party is a good idea.

If we implement this it should be done such that is easy implementable also in the rust agent.

mpeters:
> The issue is not the functionality itself, but the implementation which has no configuration options to make it useful.

What do you mean by this? What configuration options should exist?

> If we implement this it should be done such that is easy implementable also in the rust agent.

How would that work? A separate binary that fetched the UUID somehow?

THS-on:
> What do you mean by this? What configuration options should exist?

The current code just hard codes the metadata service. This should be at least configurable.

https://github.com/keylime/keylime/blob/53b47c5cfa29023138abe24e5464a3a7e24089d6/keylime/openstack.py#L14-L16

> How would that work? A separate binary that fetched the UUID somehow?

Either that or is there a reason why the UUID cannot be configured using environment variable during startup?

mpeters:
> Either that or is there a reason why the UUID cannot be configured using environment variable during startup?

I like this option. So maybe we deprecate `openstack` and add a new `environment` which uses the `KEYLIME_AGENT_UUID` environment variable which can be set in whatever way the user wants.

THS-on:
@mpeters that sound like a good idea. I'll add that to this PR.

THS-on:
I've now added to option to set the UUID via the `KEYLIME_AGENT_UUID` environment variable.

mpeters (approved):

PR #803, opened 2021-12-02
Remove the SHA1 requirement for IMA

### LINKED ISSUE #30: Support more IMA hashing algorithms (TPM 2.0)

### LINKED ISSUE #801: IMA emulator should use SHA256 instead of SHA1

# Environment
 * OS / version: Fedora35
 * TPM Manufacturer: swtpm
 * Keylime version: latest upstream (a712b8393c75f6db0d29f95e57c268fc56450d7d)

# Description 
swtpm starting version 0.7.0 stopped using SHA1 for pcs, uses SHA256 instead.
Keylime IMA emulator should be also rather using SHA256 as it would allow to test e.g. system configured in FIPS.

For more details see https://bugzilla.redhat.com/show_bug.cgi?id=2028023

THS-on:
Modern kernel versions (>= 5.10) hash the IMA entries with all available PCR bank algorithms, but the entry in the IMA log is still the SHA1 entry in most distributions. 
This PR removes the special handling for PCR10 and IMA and just uses the default hash algorithm of the agent.

This also fixes an issue that the hash algorithm for the data PCR might not match between agent and verifier and introduces a new IMA emulator adapter that supports multiple hash algorithms.

Fixes #30, fixes #801

### DIFF
--- a/keylime/agentstates.py
+++ b/keylime/agentstates.py
@@ -6,27 +6,46 @@ Copyright 2021 IBM Corporation
 
 import threading
 
-from keylime.ima_ast import START_HASH, FF_HASH
+from keylime.ima_ast import get_FF_HASH, get_START_HASH
+from keylime.common.algorithms import Hash
 from keylime.ima_file_signatures import ImaKeyrings
 
 class TPMState():
     """ TPMState models the state of the TPM's PCRs """
+
     def __init__(self):
         """ constructor """
         self.pcrs = {}
+        self.hash_alg = {}  # Record the hash algorithm that a given PCR uses
         for pcr_num in range(0, 24):
             self.reset_pcr(pcr_num)
 
+    def init_pcr(self, pcr_num, hash_alg):
+        """" Initializes a PCR """
+        if pcr_num not in self.hash_alg:
+            self.hash_alg[pcr_num] = hash_alg
+
+        if self.pcrs[pcr_num] is None:
+            if 17 <= pcr_num <= 23:
+                self.pcrs[pcr_num] = get_FF_HASH(self.hash_alg[pcr_num])
+            else:
+                self.pcrs[pcr_num] = get_START_HASH(self.hash_alg[pcr_num])
+
     def reset_pcr(self, pcr_num):
         """ Reset a specific PCR """
-        if 17 <= pcr_num <= 23:
-            self.pcrs[pcr_num] = FF_HASH
-        else:
-            self.pcrs[pcr_num] = START_HASH
+        self.pcrs[pcr_num] = None
 
     def get_pcr(self, pcr_num):
-        """ Get the state of a PCR """
-        return self.pcrs[pcr_num]
+        """
+        Get the state of a PCR.
+
+        :returns: PCR value or None if not initialized
+        """
+        return self.pcrs.get(pcr_num, None)
+
+    def used_pcr(self, pcr_num):
+        """Check if a PCR was actually requested at least once"""
+        return (pcr_num in self.hash_alg) and (self.pcrs[pcr_num] is not None)
 
     def set_pcr(self, pcr_num, pcr_value):
         """ Set the value of a PCR """
@@ -73,7 +92,9 @@ class AgentAttestState():
         """ Return a dict with the IMA pcrs """
         ima_pcrs_dict = {}
         for pcr_num in self.ima_pcrs:
-            ima_pcrs_dict[pcr_num] = self.tpm_state.get_pcr(pcr_num)
+            # Only output IMA PCRs that were used at least once
+            if self.tpm_state.used_pcr(pcr_num):
+                ima_pcrs_dict[pcr_num] = self.tpm_state.get_pcr(pcr_num)
         return ima_pcrs_dict
 
     def set_ima_pcrs(self, ima_pcrs_dict):
@@ -90,8 +111,10 @@ class AgentAttestState():
         """ Set the value of the next_ima_ml_entry field """
         self.next_ima_ml_entry = next_ima_ml_entry
 
-    def get_pcr_state(self, pcr_num):
+    def get_pcr_state(self, pcr_num, hash_alg=Hash.SHA1):
         """ Return the PCR state of the given PCR """
+        if not self.tpm_state.used_pcr(pcr_num):
+            self.tpm_state.init_pcr(pcr_num, hash_alg)
         return self.tpm_state.get_pcr(pcr_num)
 
     def get_boottime(self):
--- a/keylime/common/algorithms.py
+++ b/keylime/common/algorithms.py
@@ -2,6 +2,10 @@
 SPDX-License-Identifier: Apache-2.0
 Copyright 2020 Kaifeng Wang
 """
+import enum
+import hashlib
+
+from typing import Optional
 
 
 def is_accepted(algorithm, accepted):
@@ -13,17 +17,44 @@ def is_accepted(algorithm, accepted):
     return algorithm in accepted
 
 
-class Hash:
+class Hash(str, enum.Enum):
     SHA1 = 'sha1'
     SHA256 = 'sha256'
     SHA384 = 'sha384'
     SHA512 = 'sha512'
     SM3_256 = 'sm3_256'
-    supported_algorithms = (SHA1, SHA256, SHA384, SHA512, SM3_256)
 
     @staticmethod
     def is_recognized(algorithm):
-        return algorithm in Hash.supported_algorithms
+        try:
+            Hash(algorithm)
+            return True
+        except ValueError:
+            return False
+
+    def hash(self, data: bytes) -> Optional[bytes]:
+        if self == Hash.SHA1:
+            return hashlib.sha1(data).digest()
+        if self == Hash.SHA256:
+            return hashlib.sha256(data).digest()
+        if self == Hash.SHA384:
+            return hashlib.sha384(data).digest()
+        if self == Hash.SHA512:
+            return hashlib.sha512(data).digest()
+        if self == Hash.SM3_256:
+            # SM3 might not be guaranteed to be there
+            try:
+                return hashlib.new("sm3", data).digest()
+            except ValueError:
+                return None
+
+        return None
+
+    def get_size(self):
+        return _HASH_SIZE.get(self)
+
+    def __str__(self):
+        return self.value
 
 
 class Encrypt:
@@ -54,8 +85,5 @@ _HASH_SIZE = {
     Hash.SHA256: 256,
     Hash.SHA384: 384,
     Hash.SHA512: 512,
+    Hash.SM3_256: 256,
 }
-
-
-def get_hash_size(algorithm):
-    return _HASH_SIZE.get(algorithm, 0)
--- a/keylime/ima.py
+++ b/keylime/ima.py
@@ -20,6 +20,7 @@ from keylime import keylime_logging
 from keylime import ima_ast
 from keylime.agentstates import AgentAttestState
 from keylime import ima_file_signatures
+from keylime.common.algorithms import Hash
 from keylime.failure import Failure, Component
 
 
@@ -204,9 +205,10 @@ def _validate_ima_buf(exclude_regex, allowlist, ima_keyrings: ima_file_signature
     return failure
 
 
-def _process_measurement_list(agentAttestState, lines, lists=None, m2w=None, pcrval=None, ima_keyrings=None, boot_aggregates=None):
+def _process_measurement_list(agentAttestState, lines, hash_alg, lists=None, m2w=None, pcrval=None, ima_keyrings=None,
+                              boot_aggregates=None):
     failure = Failure(Component.IMA)
-    running_hash = agentAttestState.get_pcr_state(config.IMA_PCR)
+    running_hash = agentAttestState.get_pcr_state(config.IMA_PCR, hash_alg)
     found_pcr = (pcrval is None)
     errors = {}
     pcrval_bytes = b''
@@ -258,10 +260,10 @@ def _process_measurement_list(agentAttestState, lines, lists=None, m2w=None, pcr
             continue
 
         try:
-            entry = ima_ast.Entry(line, ima_validator)
+            entry = ima_ast.Entry(line, ima_validator, pcr_hash_alg=hash_alg)
 
             # update hash
-            running_hash = hashlib.sha1(running_hash + entry.template_hash).digest()
+            running_hash = hash_alg.hash(running_hash + entry.pcr_template_hash)
 
             validation_failure = entry.invalid()
 
@@ -280,7 +282,7 @@ def _process_measurement_list(agentAttestState, lines, lists=None, m2w=None, pcr
 
             # Keep old functionality for writing the parsed files with hashes into a file
             if m2w is not None and (type(entry.mode) in [ima_ast.Ima, ima_ast.ImaNg, ima_ast.ImaSig]):
-                hash_value = codecs.encode(entry.mode.digest.hash, "hex")
+                hash_value = codecs.encode(entry.mode.digest.bytes, "hex")
                 path = entry.mode.path.name
                 m2w.write(f"{hash_value} {path}\n")
         except ima_ast.ParserError:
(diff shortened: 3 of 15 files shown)


keylime/agentstates.py
```diff
@@ -6,11 +6,14 @@
...
 import threading
 
-from keylime.ima_ast import START_HASH, FF_HASH
+from keylime.ima_ast import get_FF_HASH, get_START_HASH
+from keylime.common.algorithms import Hash
 from keylime.ima_file_signatures import ImaKeyrings
 
 class TPMState():
     """ TPMState models the state of the TPM's PCRs """
+    hash_alg = Hash.SHA256  # TODO: This should be configurable
```
stefanberger, line 15: I think we need to be able to configure this in AgentAttestStates.add() when calling set_ima_pcrs(). The mess is with overlap of sm3 versus sha256 via the size because at the moment we can only derive usage of the PCR bank/hash via the size of the hash, which is a bit ambiguous. The agent would have to tell use what it is using...

THS-on: >  The agent would have to tell use what it is using...

The problem is that we don't actually know the algorithm before the first attestation round.

What  do you think about just keeping a PCR for every possible hash type that we support and when we update (`update_ima_attestation(..)`) or get PCR values (`get_pcr_state(...)`) you'll have to specify a hash algorithm?

THS-on:
Changing this would make Keylime with SHA256 as default TPM hash algorithm only work on kernel versions starting with 5.10.
When switching to SHA1 it should still work with older versions.
@mpeters, @maugustosilva do you run Keylime with kernel versions less than 5.10?

If avoidable I would skip writing a workaround to support the older kernel versions.

keylime/common/algorithms.py
```diff
@@ -13,17 +17,44 @@ def is_accepted(algorithm, accepted):
...
+            # SM3 might not be guaranteed to be there
+            try:
+                return hashlib.new("sm3", data).digest()
+            except ValueError:
+                return None
+
+        return None
+
+    def get_size(self):
+        return _HASH_SIZE.get(self)
```
stefanberger, line 54: SM3 is not listed below in _HASH_SIZE...

keylime/agentstates.py
stefanberger: > What do you think about just keeping a PCR for every possible hash type that we support and when we update (update_ima_attestation(..)) or get PCR values (get_pcr_state(...)) you'll have to specify a hash algorithm?

I would try to not keep it for ever possible hash. The get_pcr_state() in _process_measurement_list could request PCR 10 for sha1 or sha256 (as examples) by inferring from the size of the pcrval_bytes() which algo is used (let's toss out sm3 support due to ambiguity to infer from size or have the agent say what exactly it uses). But then get_pcr_state() should return any existing value that it has been given (restored from DB) and only ever do a reset of a PCR value if the PCR holds the START_HASH value and adjust it for example from sha1 to sha256. Does this make sense?

THS-on: When we call `get_pcr_state()` in `_process_measurement_list` we actually know the hashing algorithm that was used, so there is no need to infer them. 
We then just need to lazily initialize the `START_FF` hashes because during object creation we don't actually now the hash algorithm.

THS-on: @stefanberger I've now added a implementation for that.

keylime/common/algorithms.py
THS-on: fixed

keylime/agentstates.py
```diff
@@ -6,28 +6,49 @@
...
+
+        :returns: PCR value or None if not initialized and no hash_alg was specified
+        """
+        if pcr_num not in self.hash_alg:
+            if hash_alg:
+                self.hash_alg[pcr_num] = hash_alg
+            else:
+                return None
+
+        if self.pcrs[pcr_num] is None:
```
stefanberger, line 41: Create new `init_pcr(self, pcr_num)` for this initialization?

THS-on: This is a good idea. Done.

keylime/agentstates.py
```diff
@@ -6,27 +6,46 @@
...
+from keylime.common.algorithms import Hash
 from keylime.ima_file_signatures import ImaKeyrings
 
 class TPMState():
     """ TPMState models the state of the TPM's PCRs """
+
     def __init__(self):
         """ constructor """
         self.pcrs = {}
+        self.hash_alg = {}
```
stefanberger, line 19: Add this comment (or something like it):
# Record the hash_alg that a given PCR is using

I hope they will all be using the same hash algo...

keylime/agentstates.py
```diff
@@ -90,8 +111,10 @@ def set_next_ima_ml_entry(self, next_ima_ml_entry):
         """ Set the value of the next_ima_ml_entry field """
         self.next_ima_ml_entry = next_ima_ml_entry
 
-    def get_pcr_state(self, pcr_num):
+    def get_pcr_state(self, pcr_num, hash_alg=Hash.SHA1):
         """ Return the PCR state of the given PCR """
+        if not self.tpm_state.used_pcr(pcr_num):
+            self.tpm_state.init_pcr(pcr_num, hash_alg)
```
stefanberger, line 117: Great!

keylime/agentstates.py
THS-on: Done. Yeah we should enforce that the agent cannot switch algorithm after initial attestation.

THS-on:
I think this PR is now ready for testing, but there are still some questions that need to be discussed.

* Kernel versions older that 5.10 do not extend the PCRs as expected (just rehash the data with the correct algorithm). Do we want to implement extra handling for that or not? Downgrading the general `tpm_hash_alg` to `sha1` should do the trick on those installations.
* Older agent versions will send us only the SHA1 hash of the IMA PCR, should add special handling for that?
* Does the REST API version need to be increased?

keylime/ima.py
```diff
@@ -204,9 +205,10 @@ def _validate_ima_buf(exclude_regex, allowlist, ima_keyrings: ima_file_signature
     return failure
 
 
-def _process_measurement_list(agentAttestState, lines, lists=None, m2w=None, pcrval=None, ima_keyrings=None, boot_aggregates=None):
+def _process_measurement_list(agentAttestState, lines, lists=None, m2w=None, pcrval=None, ima_keyrings=None,
+                              boot_aggregates=None, hash_alg=Hash.SHA1):
```
stefanberger, line 209: Nit: Should we make this parameter mandatory, meaning no fallback value, and put it after `lists` ?

mpeters:
> do you run Keylime with kernel versions less than 5.10

Yes. The latest releases of RHEL all use kernels older than 5.10 (RHEL 8 uses 4.18). It's not packaged for RHEL 8 and I believe RHEL 9 will use 5.11, but we still want it to be possible to monitor systems running RHEL 8.

mpeters:
> do you run Keylime with kernel versions less than 5.10

If someone tries to use an unsupported hash algo, what happens now? As long as there's a good error message in the log (agent at least, but also verifier if possible) that would be ok.

mpeters:
> I think this PR is now ready for testing, but there are still some questions that need to be discussed.
> 
> * Kernel versions older that 5.10 do not extend the PCRs as expected (just rehash the data with the correct algorithm). Do we want to implement extra handling for that or not? Downgrading the general `tpm_hash_alg` to `sha1` should do the trick on those installations.

I'm fine with that as long as the error messages are clear.

> * Older agent versions will send us only the SHA1 hash of the IMA PCR, should add special handling for that?

What special handling are you thinking about?

> * Does the REST API version need to be increased?

I don't think we need an API version increment for this change. It added something to the allowlist format, but I'm not sure that's enough to qualify for a new API version. I could be convinced otherwise though if people felt strongly about it.

stefanberger:
I wanted to test this and had to setup a new environment wit Fedora 34 on the agent side and Fedora 35 on the server side. But now I cannot even get beyond the error message:

keylime.tenant.UserError: TPM Quote from cloud agent is invalid for nonce: bpom...

THS-on:
@mpeters 
> If someone tries to use an unsupported hash algo, what happens now? As long as there's a good error message in the log (agent at least, but also verifier if possible) that would be ok.

IMA verification will fail. I'll add a better error message if another hash algorithm then SHA1 is used on kernel versions <5.10.

> Older agent versions will send us only the SHA1 hash of the IMA PCR, should add special handling for that?
>What special handling are you thinking about?

Downgrading to SHA1 if it is allowed and we get only a SHA1 hash for PCR 10.

>I don't think we need an API version increment for this change. It added something to the allowlist format, but I'm not sure that's > enough to qualify for a new API version. I could be convinced otherwise though if people felt strongly about it.

I agree.

THS-on:
@stefanberger 

Running both this PR or is one running an older version?

stefanberger:
> Running both this PR or is one running an older version?

Oh dear... I am using VMs with swtpm and forgot to copy the issuercert.pem from the host issuing the EK cert into `/var/lib/keylime/tpm_cert_store` and then it fails as if something was wrong with the quote itself.

stefanberger:
I now switched the vTPM from sha1 + sha256 bank support to only support the sha256 bank. Current keylime master cannot handle it. This PR then fixes the issue and it works as before when sha1 was used. My testing was with file signatures being verified and here I think it's primarily the quote that has to pass.

stefanberger:
I also tried it with just a SHA-384 PCR bank enabled but that then fails due to the agent I think...

THS-on:
> I also tried it with just a SHA-384 PCR bank enabled but that then fails due to the agent I think...

Thanks for finding this. `sim_extend(..)` would use the default hash algorithm which is different if the agent and verifier don't share the same configuration.

keylime/ima.py
THS-on: Yeah fixed.

maugustosilva:
I just ran a quick test with this PR, and it is breaking from me. Using an allowlist on the old format, I see the following error `IMA measurement list does not match TPM PCR c048eccecfc4ca23422d9944ad8fee0ba56c0914520e0022b3497d724053e4c2`. 

Will investigate further, but please hold the merge for now.

Here is an example of my allowlist, (all `sha1`):

```
b57eed1320e37009eba35efb495823d281249ec6  boot_aggregate

5e0d723f4af771ff1ab0d3cb89af8a15f8bc9071  /usr/sbin/fstab-decode
46f0eda094e2d58c364c5833be5e6714a6ff03ea  /usr/sbin/update-rc.d
19ff1a2a7d5bdca0eb002cd97d3937e7965ce76f  /usr/sbin/groupadd
6c87f56ca56c54531766afa757631fce02aed5cc  /usr/sbin/chcpu
dff321cf5f8797cbdde935e862d41ca3d5cdaf07  /usr/sbin/pam_extrausers_chkpwd
00c01c4fdb74f4863be6305cd31020d5bc06971d  /usr/sbin/slattach
b45b054cbb450ad4ba3c5d85ca58a64282d625c9  /usr/sbin/pwck
a4f4492fc7918caf163ea73b23655d76390dd3a0  /usr/sbin/blkdeactivate
442ba2dad8d3acbd48ed226215791fb8a0707141  /usr/sbin/tarcat
cf8a92ba84f0f180a7d442ad544927e004b04d49  /usr/sbin/crm_error
c55277ef48cde28b44f0de17ba561ccf6300ad5c  /usr/sbin/cryptdisks_stop
bce28209f9e164f4f1339423e86fc4de93a2d66a  /usr/sbin/virtlockd
f46ada5edd0ba3a95d948ea46925bddbcf1458cc  /usr/sbin/dpkg-reconfigure
feecd84949dc89090ce2e4226145dd8d776c8819  /usr/sbin/xtables-legacy-multi
50901f91a8ae7f7c1e83d16943eb015b757c2b73  /usr/sbin/start-stop-daemon
57679ba23812ea8ebdb5f257df5bdb7f599e87f8  /usr/sbin/pam-auth-update
6a8339be3cb913463ec33c48d388c5898f594c5d  /usr/sbin/cryptsetup
```

THS-on:
@maugustosilva If you only provide SHA1 hashes did you change your `tpm_hash_alg` on the agent side to `sha1`?

maugustosilva:
I was looking through the code. You're right, of course. Will let you know as soon as I finish the test.

maugustosilva:
Now I see the error in the measured boot log

```
[root@node4 /]# cat /etc/keylime.conf | grep hash
tpm_hash_alg = sha1
accept_tpm_hash_algs = sha512,sha384,sha256,sha1
tpm_hash_alg = sha1
```

and on the `verifier`

```
2021-12-14 21:01:13.746 - keylime.tpm - ERROR - For PCR 0 and hash SHA256 the boot event log has value '2f197c3b472dddd6fb7f9606ae9123f4a186c3a283c9b12a28a605bb5f81df0e' but the agent returned '284d1138c16af8f0be30a48a59835acb69eb09ac'
2021-12-14 21:01:13.746 - keylime.tpm - ERROR - For PCR 1 and hash SHA256 the boot event log has value '2f882dd70543f88e4c377fe03b7a870f460774940fb0ac02145f0f4f3592fa39' but the agent returned '25434a48e97d7cc3f8c0b22e87a039af8cef8625'
2021-12-14 21:01:13.747 - keylime.tpm - ERROR - For PCR 2 and hash SHA256 the boot event log has value '4fa069d286b2edfb365213f20faaaf513545bdb2524054ea39ceef22e1c29ae5' but the agent returned '807c32de5b85d7d104b039f986d510dbde633f0d'
2021-12-14 21:01:13.747 - keylime.tpm - ERROR - For PCR 3 and hash SHA256 the boot event log has value '3d458cfe55cc03ea1f443f1562beec8df51c75e14a9fcf9a7234a13f198e7969' but the agent returned 'b2a83b0ebf2f8374299a5b2bdfc31ea955ad7236'
2021-12-14 21:01:13.747 - keylime.tpm - ERROR - For PCR 4 and hash SHA256 the boot event log has value '7cb5826fcb0438241b22850c34e9f1691298ce02f8cc94815445fea6595bd1a' but the agent returned 'c0150dad582a6f5669d725b2af46b53df6b9d49f'
2021-12-14 21:01:13.747 - keylime.tpm - ERROR - For PCR 5 and hash SHA256 the boot event log has value 'b191309f12890cd219c5e043cfa796d1a5075eb837800cb22f42259438a8aed6' but the agent returned '583d0fe407d13ec85c4c9a760c8265531cd69a11'
2021-12-14 21:01:13.747 - keylime.tpm - ERROR - For PCR 6 and hash SHA256 the boot event log has value '3d458cfe55cc03ea1f443f1562beec8df51c75e14a9fcf9a7234a13f198e7969' but the agent returned 'b2a83b0ebf2f8374299a5b2bdfc31ea955ad7236'
2021-12-14 21:01:13.748 - keylime.tpm - ERROR - For PCR 7 and hash SHA256 the boot event log has value 'b926225ac488e9c50ef2fa815aa7104b385a06907093bfb1dc62eeb7abecddf1' but the agent returned '7679047866e616450760ce711ea91126eb45bef6'
2021-12-14 21:01:13.748 - keylime.tpm - ERROR - For PCR 8 and hash SHA256 the boot event log has value '5c51521e50258ad78cda19004c6e38546659d3d15b328bd06992873a34070d5d' but the agent returned '7913fa4c9231aedaa427e1affb099f057e0ad1ed'
2021-12-14 21:01:13.748 - keylime.tpm - ERROR - For PCR 9 and hash SHA256 the boot event log has value '9547ed04feb1387152361bae5fde503b8f92b840f0835e062d63a6db145e115e' but the agent returned 'a9d4e41ec51fc55398954e42a2dec2241f889484'
2021-12-14 21:01:13.748 - keylime.tpm - ERROR - For PCR 14 and hash SHA256 the boot event log has value 'e3991b7ddd47be7e92726a832d6874c5349b52b789fa0db8b558c69fea29574e' but the agent returned '77db66d60aa0c2cd1cea6c34fed4f2aa014bf285'
```

THS-on:
@maugustosilva thanks for the report. 

The `parse_mb_bootlog(..)` always used SHA256. The latest commit should make this depended on the algorithm the agent uses.

THS-on:
@maugustosilva on which kernel version are you? If you try to use sha256 with a kernel version lower than 5.10 you should get an warning that IMA might not work correctly.

@mpeters should we change the default `tpm_hash_alg` to `sha1`, so that it works on all systems by default. Once RHEL uses a kernel version greater than 5.10 we can switch to `sha256` by default then.

maugustosilva:
> @maugustosilva on which kernel version are you? If you try to use sha256 with a kernel version lower than 5.10 you should get an warning that IMA might not work correctly.

Kernel 5.4 (Ubuntu Focal)

> 
> @mpeters should we change the default `tpm_hash_alg` to `sha1`, so that it works on all systems by default. Once RHEL uses a kernel version greater than 5.10 we can switch to `sha256` by default then.

Right, we need to ensure it is not gonna be broken for users out of the box. Focal, for instance, is LTS.

maugustosilva:
@THS-on It now works properly with your latest commits, thanks.

maugustosilva (approved):

THS-on:
I've now changed the default `tpm_hash_alg` to `sha1` to prevent Keylime being broken by default on older kernel versions.

mpeters:
> @mpeters should we change the default `tpm_hash_alg` to `sha1`, so that it works on all systems by default. Once RHEL uses a kernel version greater than 5.10 we can switch to `sha256` by default then.

Yeah, I think we can go with `sha1` as the default for now as the most popular enterprise linux distro doesn't support 256 yet.

mpeters (approved):

PR #808, opened 2021-12-15
Add generator for measured boot refence state for example policy

THS-on:
This adds a script `create_mbrefstate` which converts a UEFI eventlog into a valid mb refstate for the example policy.

Also some changes to the example policy are done to make it work out of the box with my Fedora 34 installation.

@ljmcgann can you please verify if this works also for your setup?


maugustosilva:
I have fully tested it. It is extremely useful, and just be merged immediately. We can polish the "rough edges" on the "generator" in subsequent PRs.

maugustosilva (approved):

PR #809, opened 2021-12-16
ima_emulator: specify sys.argv as the named parameter argv in main()

sergio-correia:
Fixes the following issue, which followed-up from b0dcaa8:

```
Traceback (most recent call last):
  File "/usr/local/bin/keylime_ima_emulator", line 33, in <module>
    sys.exit(load_entry_point('keylime==0.0.0', 'console_scripts', 'keylime_ima_emulator')())
TypeError: main() missing 1 required positional argument: 'argv'
```

Other utilities under `cmd/` also use this approach.

Signed-off-by: Sergio Correia <scorreia@redhat.com>

PR #810, opened 2021-12-16
Fix argv handling in main()

kkaarreell:
argv is not passed to main from entry_point which leads
to the following traceback:

  File "/usr/local/bin/keylime_ima_emulator", line 33, in <module>
    sys.exit(load_entry_point('keylime==0.0.0', 'console_scripts', 'keylime_ima_emulator')())
TypeError: main() missing 1 required positional argument: 'argv'

argv argument is not necessary as we can access sys.argv
directly.


THS-on:
@kkaarreell #809 already fixes that.

kkaarreell:
I see. I am closing it then. Thank you.

PR #811, opened 2021-12-16
Enable functional testing with Packit CI

kkaarreell:
Enables functional testing using Packit/TFT CI on tests from
https://github.com/RedHat-SP-Security/keylime-tests


kkaarreell:
@mpeters @sergio-correia @mbestavros FYI

kkaarreell:
Tests seems to be running properly. 

Currently, there is just one functional test enabled doing the basic attestation on localhost using emulated TPM and keylime IMA emulator. I plan to enable additional tests once they are implemented and stable.

Also, please be aware that due to a bug in the Packit/TFT service tests are listed in a wrong order in the log (execution follows the proper order as timestamps suggest).

sopos:
nice

kkaarreell:
I don't know why test is failing on Rawhide but I can see it has newer tpm2-tss version (tpm2-tss-3.1.0-4.fc36). Maybe there is another regression in that package.

kkaarreell:
Still don't know what is the problem with Rawhide but after downgrading tpm2-tss to a package from F35 the test passes.

mpeters (approved):

PR #812, opened 2021-12-16
Run existing agent tests against the rust-keylime agent

mbestavros:
The current `test_restful.py` test suite runs a number of tests against the Python agent. This pull request adds a number of helper functions and a new set of tests that:

- shut down the Python agent
- start up the Rust agent
- run agent-specific tests against the Rust agent

Currently, one of the tests (`agent_quotes_identity_get`) passes, while others fail (mostly due to missing APIs in `rust-keylime`). I've added skips to the broken tests with documentation on the relevant `rust-keylime` issues blocking them, so it should still pass without fail.

### DIFF
--- a/test/run_tests.sh
+++ b/test/run_tests.sh
@@ -20,6 +20,7 @@ UMODE_OPT=""
 COVERAGE=0
 COVERAGE_DIR=`pwd`
 unset COVERAGE_FILE
+RUST_TEST="${RUST_TEST:-0}"
 
 for opt in "$@"; do
   shift
@@ -163,6 +164,11 @@ echo "==========================================================================
 echo $'\t\t\tInstalling test requirements'
 echo "=================================================================================="
 pip3 install $UMODE_OPT -r $KEYLIME_DIR/test/test-requirements.txt
+if [ "$RUST_TEST" == 1 ]
+then
+    git clone https://github.com/keylime/rust-keylime.git $KEYLIME_DIR/../rust-keylime
+    cargo build --manifest-path $KEYLIME_DIR/../rust-keylime/Cargo.toml --bin keylime_agent
+fi
 
 # Install Keylime
 echo
--- a/test/test_restful.py
+++ b/test/test_restful.py
@@ -100,6 +100,7 @@ cv_process = None
 reg_process = None
 agent_process = None
 tenant_templ = None
+SKIP_RUST_TEST = not bool(os.getenv("SKIP_RUST_TEST"))
 
 # Class-level components that are not static (so can't be added to test class)
 public_key = None
@@ -252,11 +253,17 @@ def launch_registrar():
     return True
 
 
-def launch_cloudagent():
+def launch_cloudagent(agent="python"):
     """Start up the cloud agent"""
     global agent_process, script_env, FORK_ARGS
+    if agent == "python":
+        agent_path = "keylime_agent"
+    elif agent == "rust":
+        agent_path = f"{PACKAGE_ROOT}/../rust-keylime/target/debug/keylime_agent"
+    else:
+        agent_path = "echo"
     if agent_process is None:
-        agent_process = subprocess.Popen("keylime_agent",
+        agent_process = subprocess.Popen(agent_path,
                                          shell=False,
                                          preexec_fn=os.setsid,
                                          stdout=subprocess.PIPE,
@@ -275,7 +282,13 @@ def launch_cloudagent():
                 sys.stdout.write('\n\033[94m' + line + '\033[0m')
         t = threading.Thread(target=initthread)
         t.start()
-        time.sleep(10)
+        for retry in range(10):
+            r = subprocess.run(["ss", "-l", "-t", "-n", "-p",  "( sport = :9002 )"], stdout=subprocess.PIPE, check=False)
+            if b":9002" in r.stdout:
+                break
+            if retry == 10:
+                raise Exception(f"{agent} keylime_agent failed to launch")
+            time.sleep(1)
     return True
 
 
@@ -307,6 +320,9 @@ def kill_cloudagent():
     os.killpg(os.getpgid(agent_process.pid), signal.SIGINT)
     agent_process.wait()
     agent_process = None
+
+    # Run tpm2_clear to allow other processes to use the TPM
+    subprocess.run("tpm2_clear", stdout=subprocess.PIPE, check=False)
 
 
 def services_running():
@@ -517,7 +533,6 @@ class TestRestful(unittest.TestCase):
 
         # We want a real cloud agent to communicate with!
         launch_cloudagent()
-        time.sleep(10)
         test_020_agent_keys_pubkey_get = RequestsClient(tenant_templ.agent_base_url, tls_enabled=True, ignore_hostname=True)
         response = test_020_agent_keys_pubkey_get.get(
             f'/v{self.api_version}/keys/pubkey',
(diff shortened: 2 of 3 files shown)


aplanas:
This will block the Python code changes on issues on the rust side.

mpeters:
> This will block the Python code changes on issues on the rust side.

This is true and one of the reasons this won't be merged as-is currently. But eventually the rust-agent will become the official keylime agent and should block changes to the python side that break the rust side.

mbestavros:
@mpeters I'm marking this as ready. I've tidied everything up so there should be minimal code duplication, made sure it passes lint, added skips (with notes and links) to the broken tests, and fixed a bug with Rust agent invocation on the Github runners. Looks like all tests pass now.

test/run_tests.sh
```diff
@@ -163,6 +163,8 @@ echo "==========================================================================
 echo $'\t\t\tInstalling test requirements'
 echo "=================================================================================="
 pip3 install $UMODE_OPT -r $KEYLIME_DIR/test/test-requirements.txt
+git clone https://github.com/keylime/rust-keylime.git $KEYLIME_DIR/../rust-keylime
```
aplanas, line 166: Can we use a variable, that if is set will do the clone and the cargo build? Something line `RUST_TEST` set as 0 in the script but overwritten from the value in the environment, as `RUST_TEST="${RUST_TEST:-0}"`

This way by default we do not clone and call `cargo build` when running the Python tests.  There is no guarantee that the environment will even have Rust installed when building and testing a Python program.

test/test_restful.py
```diff
@@ -276,6 +276,33 @@ def initthread():
     return True
 
 
+def launch_rust_cloudagent():
```
aplanas, line 279: Can we do a `@unittest.skipIf`? I do no think that is OK for a Python test to unconditionally run a test that depends on cargo and the Rust toolchain.

test/test_restful.py
```diff
@@ -276,6 +276,33 @@ def initthread():
...
+    """Start up the cloud agent"""
+    global agent_process, script_env, FORK_ARGS
+    if agent_process is None:
+        agent_process = subprocess.Popen(["cargo", "run", "--manifest-path", f"{PACKAGE_ROOT}/../rust-keylime/Cargo.toml", "--bin", "keylime_agent"],
```
aplanas, line 283: Why do not call directly `keylime_agent`, instead of running it from `cargo run`?

test/test_restful.py
```diff
@@ -915,6 +945,42 @@ def test_060_cv_version_get(self):
         self.assertEqual(results['current_version'], api_version.current_version())
         self.assertEqual(results['supported_versions'], api_version.all_versions())
 
+    # Rust agent testset
+    def test_070_rust_agent_setup(self):
```
aplanas, line 949: `skipIf`?

aplanas (changes requested):

test/run_tests.sh
mbestavros: Very true - I've added this in my latest push, and gated the Rust tests and build behind it.

test/test_restful.py
mbestavros: Done! See above.

test/test_restful.py
mbestavros: Good point - I just didn't know how to call it directly, but it's in the `.cargo/bin` directory. Added.

test/test_restful.py
mbestavros: Added!

test/test_restful.py
```diff
@@ -276,6 +277,33 @@ def initthread():
     return True
 
 
+def launch_rust_cloudagent():
+    """Start up the cloud agent"""
+    global agent_process, script_env, FORK_ARGS
+    if agent_process is None:
+        agent_process = subprocess.Popen("/.cargo/bin/keylime_agent",
```
aplanas, line 284: Do you think that the rust agent will be in the root? Is because is inside a container?

test/test_restful.py
```diff
@@ -276,6 +277,33 @@ def initthread():
...
+                line = agent_process.stdout.readline()
+                if line == b'':
+                    break
+                # line = line.rstrip(os.linesep)
+                line = line.decode('utf-8')
+                sys.stdout.flush()
+                sys.stdout.write('\n\033[94m' + line + '\033[0m')
+        t = threading.Thread(target=initthread)
+        t.start()
+        time.sleep(10)
```
aplanas, line 303: I guess that this sleep is here to wait for the agent to do the bind. This can produce flaky tests (tests that sometimes fails).

Maybe a better option is to wait (not tested):

```python
for retry in range(10):
    r = subprocess.run(["ss", "-l", "-t", "-n", "-p",  "( sport = :9002 )"], stdout=subprocess.PIPE)
    if b':9002" in r.stdout:
        break
    time.sleep(1)
else:
   raise Exception("Rust keylime_agent failed to run")
```

test/test_restful.py
```diff
@@ -915,6 +946,49 @@ def test_060_cv_version_get(self):
...
 
+    # Rust agent testset
+    @unittest.skipIf(RUST_TEST, "Testing against rust-keylime is disabled!")
+    def test_070_rust_agent_setup(self):
+        """Set up the Rust agent"""
+
+        # Kill the Python agent and launch the Rust agent!
+        kill_cloudagent()
+        launch_rust_cloudagent()
+        time.sleep(10)
```
aplanas, line 957: this second sleep can be skipped, if you use the "wait" loop. In any case there were a sleep inside the launch command.

test/test_restful.py
```diff
@@ -915,6 +946,49 @@ def test_060_cv_version_get(self):
...
         self.assertEqual(results['supported_versions'], api_version.all_versions())
 
+    # Rust agent testset
+    @unittest.skipIf(RUST_TEST, "Testing against rust-keylime is disabled!")
+    def test_070_rust_agent_setup(self):
+        """Set up the Rust agent"""
+
+        # Kill the Python agent and launch the Rust agent!
+        kill_cloudagent()
+        launch_rust_cloudagent()
```
aplanas, line 956: you launch the agent as a side effect of this test, you can consider to put it in `setUp`, and the kill in `tearDown`

test/test_restful.py
mbestavros: I realized I pushed a broken path, which I've now fixed. It should be correct now.

test/test_restful.py
mbestavros: This helper was copy-pasted from the `launch_cloudagent` function, which also does it this way. Likewise for the 10-second wait after initialization.

I'm not opposed to changing it, but it would also make sense to do it for the Python launch code too.

test/test_restful.py
mbestavros: See above

test/test_restful.py
mbestavros: Putting the invocation in `setUp` and `tearDown` means we'd have to restart the agent every single test run, which would be extremely slow. Plus, some tests require the agent to be running between them, so that wouldn't work anyway.

It _might_ make sense to do it in `setUpClass`, but since we have to run separate versions of the agent at different times for different tests, that doesn't work at the moment either.

It's in a separate test so we know it's invoked before the rest of the Rust test suite. This is necessary because one of the reused tests (020) includes its own call to `launch_cloudagent`. If the Rust agent isn't running when that call happens, the Python agent will be booted up instead.

mpeters: Even if we can't put it in `setUp()` and `tearDown()` it might make sense to pull it out into it's own function that checks to see if the agent is running and if not, starts the agent (if it's not running, and does a loop to make sure it's up before returning). Then this function can be included in every test that needs that agent. This would allow developers to be able to run single test methods as they are developing (a common pattern in most projects).

Then we could handle stopping the agents in the class teardown if they are running.

I know this is a change to the way things were for the other agent (the existing tests were fragile in this way), but it's becoming more necessary to fix if we copy that fragility into more places.

mbestavros: @mpeters Okay, I'll give that a try. I think the cleanest way might actually just be to improve the existing `launch_cloudagent` function - it already checks if there's an agent running. If we provide a parameter to toggle between launching the Python and Rust agents, and incorporate @aplanas' suggestion around waiting until it's up, we could just include it in front of all tests that need an agent, as you say. Then, for the Rust tests, we just have one extra that stops it and restarts the Rust agent - that should default to Python for now and allow Rust when needed.

mpeters (changes requested):

test/test_restful.py
```diff
@@ -100,6 +100,7 @@ def cmp(a, b):
 reg_process = None
 agent_process = None
 tenant_templ = None
+RUST_TEST = not bool(os.getenv("RUST_TEST"))
```
mpeters, line 103: Why is this `not`? If you're negating the env var value shouldn't this var be called something like `SKIP_RUST_TEST`?

test/test_restful.py
```diff
@@ -275,7 +282,14 @@ def initthread():
...
         t.start()
-        time.sleep(10)
+        for retry in range(10):
+            r = subprocess.run(["ss", "-l", "-t", "-n", "-p",  "( sport = :9002 )"], stdout=subprocess.PIPE)
+            if b":9002" in r.stdout:
+                break
+            elif retry == 10:
+                raise Exception(f"{agent} keylime_agent failed to launch")
+            time.sleep(1)
+        #time.sleep(10)
```
mpeters, line 292: You can just remove this commented out code. It's safe in the git history.

mpeters (changes requested):
Looking good. Some small changes/clarifications requested.

mbestavros:
Current status is that some cross-agent tests fail with this error message:

```
Error in test.test_restful.TestRestful.test_071_agent_keys_pubkey_get
  File "/usr/lib/python3.9/site-packages/urllib3/connectionpool.py", line 670, in urlopen
    httplib_response = self._make_request(
  File "/usr/lib/python3.9/site-packages/urllib3/connectionpool.py", line 381, in _make_request
    self._validate_conn(conn)
  File "/usr/lib/python3.9/site-packages/urllib3/connectionpool.py", line 978, in _validate_conn
    conn.connect()
  File "/usr/lib/python3.9/site-packages/urllib3/connection.py", line 362, in connect
    self.sock = ssl_wrap_socket(
  File "/usr/lib/python3.9/site-packages/urllib3/util/ssl_.py", line 384, in ssl_wrap_socket
    return context.wrap_socket(sock, server_hostname=server_hostname)
  File "/usr/lib64/python3.9/ssl.py", line 500, in wrap_socket
    return self.sslsocket_class._create(
  File "/usr/lib64/python3.9/ssl.py", line 1040, in _create
    self.do_handshake()
  File "/usr/lib64/python3.9/ssl.py", line 1309, in do_handshake
    self._sslobj.do_handshake()
ssl.SSLError: [SSL: WRONG_VERSION_NUMBER] wrong version number (_ssl.c:1129)
```

I've talked about this offline with @lkatalin, and she suggests that it may be a bug introduced by `rust-keylime` introducing mTLS, and might be fixed by this upstream PR: https://github.com/keylime/rust-keylime/pull/316

In any case, it seems like the test failure is down to upstream bugs and not due to these test changes - and, in any case, the Rust tests will not be run as part of the test suite yet.

test/test_restful.py
mbestavros: Done!

mbestavros:
A quick summary of the changes I've made:

- Rewrote `launch_cloudagent()` as detailed [here](https://github.com/keylime/keylime/pull/812#discussion_r793987819); now, it will invoke the agent if it isn't already running, and can start either the Python or Rust agent. It also is a lot smarter about waiting for the agent to boot (instead of waiting 10 seconds, it checks if the agent's socket is open). This means we don't need a separate `launch_rust_cloudagent()` function anymore, and individual tests should be a bit more stable.
- Renamed `RUST_TEST` to `SKIP_RUST_TEST` in `test_restful` to better reflect what it does
- Added `launch_cloudagent()` invocations in front of all the Rust tests
- Addressed various suggestions/fixes from @mpeters

aplanas:
@mbestavros very cool changes! Thanks a lot!

mbestavros:
@aplanas Thanks for your reviews! They were very helpful 😄

mpeters (approved):

PR #816, opened 2021-12-23
Mitigations against event type manipulation in UEFI eventlog

THS-on:
This provides the initial hardening of the example policy against event type manipulation. 
This not directly an issue with Keylime, because the UEFI eventlog does not include the metadata in the hash that is used to extend the PCR. The original issue can be found here: https://github.com/google/go-attestation/blob/master/docs/event-log-disclosure.md

The most effective way to mitigate against that is to reduce the usage of `acceptAll ` and limit how often they can be used.

As discussed with @mpeters we will first update our example policy and then provide documentation on how to write custom policies.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -3,6 +3,6 @@ jobs:
   trigger: pull_request
   metadata:
     targets:
-    - fedora-all
+    - fedora-stable
     - centos-stream-9-x86_64
     skip_build: true
--- a/keylime/elchecking/example.py
+++ b/keylime/elchecking/example.py
@@ -100,7 +100,8 @@ class Example(policies.Policy):
                 raise Exception(f'refstate lacks {req}')
 
         dispatcher = tests.Dispatcher(('PCRIndex', 'EventType'))
-        vd = tests.VariableDispatch()
+        vd_driver_config = tests.VariableDispatch()
+        vd_authority = tests.VariableDispatch()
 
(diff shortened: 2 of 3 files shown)


THS-on:
@kkaarreell the Fedora rawhide tests are currently failing. Is there something that we can fix them or can we just disable them until the issue in rawhide itself is solved?

kkaarreell:
According to @sergio-correia same patches as in
 https://bugzilla.redhat.com/show_bug.cgi?id=1984634
are missing in rawhide.
Sergio, did you open a rawhide bug regarding it?

Re tests, do you mean disable them all or just on Rawhide?

sergio-correia:
@kkaarreell: there was a bug open for it already [1], so I just added a comment there some days ago. The maintainer of `tpm2-tss` in Fedora is currently waiting for a newer release -- to be released soon(tm) --, which should support OpenSSL 3.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=2008179

kkaarreell:
@THS-on Would you mind adding a commit that replaces `fedora-all` with `fedora-stable` in `. packit. yaml`?

THS-on:
@kkaarreell done. The tests now pass.

mpeters:
@maugustosilva any chance you could take a look at these changes?

maugustosilva (approved):

maugustosilva:
Passes all my tests, let's merge

PR #818, opened 2022-01-03
Various code improvements for mostly the agent

THS-on:
This fixes various validation issues with the agent and drops `/usr/local` as an included path by default because we do not use a custom patched version of tpm2-tools anymore.

### DIFF
--- a/keylime/keylime_agent.py
+++ b/keylime/keylime_agent.py
@@ -211,6 +211,10 @@ class Handler(BaseHTTPRequestHandler):
             config.echo_json_response(self, 400, "API Version not supported")
             return
 
+        if rest_params.get("keys", None) not in ["ukey", "vkey"]:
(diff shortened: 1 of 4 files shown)


keylime/keylime_agent.py
```diff
@@ -114,14 +119,20 @@ def do_GET(self):
 
             # Sanitization assurance (for tpm.run() tasks below)
             if not (nonce.isalnum() and
-                    (pcrmask is None or pcrmask.isalnum()) and
-                    (vpcrmask is None or vpcrmask.isalnum() and
-                    ima_ml_entry.isalnum())):
+                    (pcrmask is None or config.valid_hex(pcrmask)) and
```
mpeters, line 122: why drop the check for `vpcrmask`?

THS-on: Because we don't have vtpm support in the Python agent. I think it does not make sense to validate data that we do not use.

mpeters (approved):

PR #823, opened 2022-01-06
ima_emulator: fix IMA hash validation and add more options

THS-on:
ToMToU errors are handled by the IMA hash, so the correct hash algorithm
needs to be specified there (by default SHA1).

The option for measuring other files than config.IMA_ML was added.

### DIFF
--- a/keylime/cmd/ima_emulator_adapter.py
+++ b/keylime/cmd/ima_emulator_adapter.py
@@ -20,11 +20,11 @@ from keylime import ima_ast
 tpm_instance = tpm(need_hw_tpm=True)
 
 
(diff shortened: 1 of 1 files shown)


THS-on:
@kkaarreell now it makes sense that the IMA emulator worked but on actual hardware it failed.

kkaarreell:
It would be great to have such a HW available for testing via Packit.

THS-on:
We could try to verify your parser against Ken's in his ibmtss package, but the issue is that we use the ASCII log and ibmtss uses the binary format.

mpeters (approved):

PR #825, opened 2022-01-07
Allow graceful shutdown of the verifier

THS-on:
@kkaarreell noticed that the shutdown of the verifier results in an error. This should fix that.

### DIFF
--- a/keylime/cloud_verifier_tornado.py
+++ b/keylime/cloud_verifier_tornado.py
@@ -3,7 +3,7 @@
 SPDX-License-Identifier: Apache-2.0
 Copyright 2017 Massachusetts Institute of Technology.
 '''
-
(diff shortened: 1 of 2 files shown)


mpeters (approved):

THS-on:
The Fedora 34 test is failing and cannot see why this change should have an impact on the agent quote status and it passes on centos 9. @kkaarreell maybe a timing issue between adding the agent and checking the agent state?

kkaarreell:
@THS-on could you reschedule tests by adding the above comment?

kkaarreell:
That last failure is due to occasional error during TPM setup. I think you can ignore it, usually test rerun passes. I am not sure what is the culprit but it is not related to this PR.

PR #826, opened 2022-01-10
Support for devicer mapper IMA validation

THS-on:
Some systems depend on that files are protected by either encryption or something like dm-verity, but before there was no way to confirm from the outside if those were setup correctly.  Now we can validate that through IMA and this PR adds support for validating the different device mapper target and device events introduced in kernel version 5.15 [1].

The IMA allowlist format is extended to include a `dm_policy` for configuring rules to validate those entries [2]. 
The rules can be matched either on the `uuid` or `name` of the device using a regex. Entries can be validated against  constants or regex expressions. `None` can be used to ignore a entry. Please check if this works most use cases.

Because the format for dm-ima has a higher complexity than the base IMA format the lexer and parser is generated with lark [3].
This has the advantage that the overall code is more readable, less error prone to parsing errors and simpler to extend.

Currently with some of the attribute values change in target table no new event is generated. I've proposed a new event that handles that case [4]. If you have the time have a look at this change.

Parsing and tests are currently implemented for following device mapper targets: verity, cache, crypt, integrity, linear, mirror and snapshot. More will be added after initial release of this feature.

Lark versions from 0.8.1 to the latest version (1.1.1) should work, which covers the most of the versions packaged today.
Note that on older versions we fallback to a slower parser.

The transformation of the lark tokens into python objects  heavily relies on dataclasses, so Python 3.6 users will have to install the dataclasses compatibility package (https://pypi.org/project/dataclasses/).

### Notes
- Kernel documentation for cache needs to be updated
- Parsing rules for mirror, multipath, raid and striped are more complex
- end-to-end tests are failing because lark is not installed

[1] https://www.kernel.org/doc/html/latest/admin-guide/device-mapper/dm-ima.html
[3] https://github.com/lark-parser/lark
[4] https://lore.kernel.org/dm-devel/20220106203436.281629-1-public@thson.de/
[2] See tests for example policies


keylime/agentstates.py
```diff
@@ -6,9 +6,9 @@
 
 import threading
 
-from keylime.ima_ast import get_FF_HASH, get_START_HASH
+from keylime.ima.ima_ast import get_FF_HASH, get_START_HASH
```
aplanas, line 9: nit: rename `ima_ast` to `ast`?

keylime/ima/ima_dm.py
```diff
@@ -0,0 +1,800 @@
...
+    Variant of ChainMap that handles updates for nested ChainMaps
+    Based on the example in https://docs.python.org/3/library/collections.html#collections.ChainMap
+    """
+
+    def __setitem__(self, key, value):
+        for mapping in self.maps:
+            if key in mapping:
+                mapping[key] = value
+                return
+        self.maps[0][key] = value
```
aplanas, line 38: nit for clarity: `super().__setitem(key, value)`

THS-on: This code is directly taken from the Python documentation, but I agree with you that this is more clear.

THS-on:
@aplanas renaming the `ima_ast` to `ast` and `ima_dm` to `dm` makes sense and I'll change that but in what would you rename the original `ima.py` file?

aplanas:
> @aplanas renaming the `ima_ast` to `ast` and `ima_dm` to `dm` makes sense and I'll change that but in what would you rename the original `ima.py` file?

I think that we can do that later. I am still digesting the PR!

THS-on:
@mpeters I still need to include a patch for dataclasses support for Python 3.6. I'll rebase it tomorrow on master, add a warning that it is still experimental and add that patch for dataclasses.

THS-on:
@mpeters should I import the fallback code for dataclasses directly into Keylime or should people install it from pypi?

mpeters:
> @mpeters should I import the fallback code for dataclasses directly into Keylime or should people install it from pypi?

What does the fallback code look like? Is is commonly packaged by distros (RH, Fedora, SUSE, Ubuntu are our biggest concerns)?

THS-on:
> What does the fallback code look like?

We would just import the official dataclasses backport into Keylime (basically this patch: https://github.com/utkarsh2102/python-keylime/blob/1ea72cf75309f606ccfcd9015e29dcd4be79c654/debian/patches/embed-dataclasses-module.patch) and load it conditionally based on the Python version. 

The dataclasses package  is not really packaged for any distribution (it is only in EPEL 8).

aplanas:
> The dataclasses package is not really packaged for any distribution (it is only in EPEL 8).

Just to confirm that this is the case for SUSE (SLE*) : ((

THS-on:
I now added the backport for dataclasses. The centos stream 8 failure is not related to the PR.

THS-on:
@mpeters I think this is ready for going into the 6.4 release as an experimental feature. I've added a warning that is generated when a policy contains a device mapper policy.

mpeters (approved):

PR #827, opened 2022-01-11
Drop dataclasses module usage

aplanas:
Dataclasses module is not present in Python 3.6, an interpreter still
used in some distributions.

Even there is a 3rd party compatibility module, the current usage seems
not properly justified.  For example, in one of the dataclasses there is
an user-provided __init__ constructor.

### DIFF
--- a/keylime/failure.py
+++ b/keylime/failure.py
@@ -5,7 +5,6 @@ Copyright 2021 Thore Sommer
 Tagging of failure events that might cause revocation in Keylime.
 '''
 import ast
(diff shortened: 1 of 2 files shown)


THS-on (approved):
LGTM. I couldn't find any occurrence where the where code depends on those classes actually being dataclasses.

THS-on (approved):
LGTM. I couldn't find any occurrence where the where code depends on those classes actually being dataclasses.

PR #835, opened 2022-01-14
Move all Keylime web components to Tornado

THS-on:
We are currently using Tornado for the verifier, but not for the agent and registrar. 
This PR moves the agent and registrar to Tornado and unifies the utilities used for Tornado in the `tornado_helpers` module.

### DIFF
--- a/demo/agent_monitor/tenant_agent_monitor.py
+++ b/demo/agent_monitor/tenant_agent_monitor.py
@@ -24,6 +24,8 @@ import argparse
 import keylime_logging
 import common
 import sys
+
+import keylime.tornado_helpers
 from keylime import json
 sys.path.insert(0, '../../keylime/')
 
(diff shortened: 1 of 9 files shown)


THS-on:
@kkaarreell the test fails, but all individual steps have a PASS and 0 bad are reported, do you know why?

THS-on:
Ok it seems that the tests reach a timeout of 15min, but I don't know why.

kkaarreell:
@THS-on I can reproduce the problem locally. keylime_agent hung and didn't react on kill signal. However, I can reproduce it using this branch but not using the keylime master branch. Is there something in this PR that would explain it?

kkaarreell:
I have update the test suite. It won't timeout but better check the test log to see that there hasn't been any traceback when the process has been terminated.

THS-on:
@kkaarreell It is fine for me if the CI fails when stopping the services fails for some reason, but it should be visible in the logs. 

Moving the registrar and agent to Tornado seems to cause some timeouts, that I cannot get rid off.
I'll revert those changes and just keep the general improvements in this PR.

kkaarreell:
OK, I will partially revert the test update on Monday. I have realized that the former version could have masked some issues so it is possible those will become visible.

THS-on:
Superseded by #837

kkaarreell:
@THS-on I have updated tests and also provided basic instructions on how to run them locally on a test system.
https://github.com/RedHat-SP-Security/keylime-tests/blob/main/TESTING.md

PR #837, opened 2022-01-16
Various fixes and improvements to the web components

THS-on:
Initially the idea was to convert all the components to the Tornado framework with #835, but moving the current threaded code to async did cause more problems than it solved.

This PR takes the general improvements from #835 without changing the components to Tornado.

### DIFF
--- a/demo/agent_monitor/tenant_agent_monitor.py
+++ b/demo/agent_monitor/tenant_agent_monitor.py
@@ -95,7 +95,7 @@ class AgentsHandler(BaseHandler):
         """
         logger.info('Agent Monitor POST')
         try:
-            rest_params = common.get_restful_params(self.request.path)
+            rest_params = keylime.web_util.get_restful_params(self.request.path)
 
             if "agents" not in rest_params:
                 keylime.web_util.echo_json_response(self, 400, "uri not supported")
--- a/demo/agent_monitor/tenant_agent_monitor.py
+++ b/demo/agent_monitor/tenant_agent_monitor.py
@@ -24,6 +24,8 @@ import argparse
 import keylime_logging
 import common
 import sys
+
+import keylime.web_util
 from keylime import json
 sys.path.insert(0, '../../keylime/')
 
@@ -57,14 +59,14 @@ class BaseHandler(tornado.web.RequestHandler):
 
 class MainHandler(tornado.web.RequestHandler):
     def get(self):
-        common.echo_json_response(
+        keylime.web_util.echo_json_response(
             self, 405, "Not Implemented: Use /agents/ interface instead")
 
 
 class AgentsHandler(BaseHandler):
     def head(self):
         """HEAD not supported"""
-        common.echo_json_response(self, 405, "HEAD not supported")
+        keylime.web_util.echo_json_response(self, 405, "HEAD not supported")
 
     def get(self):
         """This method handles the GET requests to retrieve status on agents from the Agent Monitor.
@@ -75,7 +77,7 @@ class AgentsHandler(BaseHandler):
         was not found, it either completed successfully, or failed.  If found, the agent_id is still polling
         to contact the Cloud Agent.
         """
-        common.echo_json_response(self, 405, "GET not supported")
+        keylime.web_util.echo_json_response(self, 405, "GET not supported")
 
     def delete(self):
         """This method handles the DELETE requests to remove agents from the Agent Monitor.
@@ -83,7 +85,7 @@ class AgentsHandler(BaseHandler):
         Currently, only agents resources are available for DELETEing, i.e. /agents. All other DELETE uri's will return errors.
         agents requests require a single agent_id parameter which identifies the agent to be deleted.
         """
-        common.echo_json_response(self, 405, "DELETE not supported")
+        keylime.web_util.echo_json_response(self, 405, "DELETE not supported")
 
     def post(self):
         """This method handles the POST requests to add agents to the Agent Monitor.
@@ -96,7 +98,7 @@ class AgentsHandler(BaseHandler):
             rest_params = common.get_restful_params(self.request.path)
 
             if "agents" not in rest_params:
-                common.echo_json_response(self, 400, "uri not supported")
+                keylime.web_util.echo_json_response(self, 400, "uri not supported")
                 logger.warning(
                     'POST returning 400 response. uri not supported: ' + self.request.path)
                 return
@@ -106,7 +108,7 @@ class AgentsHandler(BaseHandler):
             if agent_id is not None:  # we have to know who phoned home
                 content_length = len(self.request.body)
                 if content_length == 0:
-                    common.echo_json_response(
+                    keylime.web_util.echo_json_response(
                         self, 400, "Expected non zero content length")
                     logger.warning(
                         'POST returning 400 response. Expected non zero content length.')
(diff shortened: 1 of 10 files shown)


THS-on:
@aplanas in this PR the web related functions are moved from `config.py` to `web_util.py`, so you don't have to move them for #831.

aplanas:
> @aplanas in this PR the web related functions are moved from `config.py` to `web_util.py`, so you don't have to move them for #831.

Oh very cool. I paused #831 for now, and I will rebase it on top of this one.

aplanas:
@THS-on BTW, what are your thoughts of taking advantage of the work done during kind of changes, and provide unit tests for them?

Is a bit more more, but also those changes can be split in smaller commits.

THS-on:
@aplanas adding more unit tests is a good idea. Especially now that we refactor some parts of Keylime it is nice to have some confidence that the behaviour did not change.

aplanas:
> @aplanas adding more unit tests is a good idea. Especially now that we refactor some parts of Keylime it is nice to have some confidence that the behaviour did not change.

I am of the same opinion! I added the tests in my last PR that changed the `config`, and I think that a good candidate to test here are the moved `echo_json_response`, `list_to_dict` or `get_restful_params`, and maybe bits of `revocation_listener` (like extracting the inner function `perform_actions` and testing it independently) or `init_mtls`.

mpeters:
I agree with @aplanas. This change looks good (better organization and less duplication) and will be approved after tests are added.

maugustosilva (approved):

THS-on:
@mpeters @aplanas I've now added tests for `echo_json_response` and `get_restful_params`.

kkaarreell:
Hello @THS-on , regarding the following test failure:
```
:: [ 19:42:36 ] :: [  BEGIN   ] :: Running 'limeStopAgent'
Process wasn't terminated, sending SIGKILL signal...
:: [ 19:42:39 ] :: [   FAIL   ] :: Command 'limeStopAgent' (Expected 0, got 9)
/usr/share/beakerlib/journal.sh: line 1: 16580 Killed                  keylime_${NAME} >> ${LOGFILE} 2>&1
```
This looks similar to your former PR. In this case agent didn't stop in 3 seconds after receiving SIGTERM so it has been killed with SIGKILL. 

In your former PR there was a traceback after SIGINT (not SIGTERM) but it was slightly different situation. For few days I was using in tests SIGINT to stop the process (since SIGINT has to be used to terminate a process when executed via the coverage script) but later I have changed it to use SIGINT only for coverage script and SIGTERM directly otherwise. 

Just now I have checked how does your new code behave when SIGINT is sent and the output was:
```
                out: Sending SIGINT
                out: :: [ 16:23:25 ] :: [  BEGIN   ] :: Running 'limeStopAgent'
                out: 2022-01-18 16:23:24.591 - keylime.local_action_rm_ssh - INFO - my serial: 2, cert serial: 2
                out: 2022-01-18 16:23:25.861 - keylime.cloudagent - INFO - Stopping revocation listener...
                out: 2022-01-18 16:23:25.863 - keylime.cloudagent - INFO - TERM Signal received, shutting down...
                out: 2022-01-18 16:23:28.892 - keylime.cloudagent - INFO - TERM Signal received, shutting down...
                out: 2022-01-18 16:23:28.920 - keylime.tpm - WARNING - Could not read YAML output of tpm2_getcap.
                out: Exception ignored in: <module 'threading' from '/usr/lib64/python3.10/threading.py'>
                out: Traceback (most recent call last):
                out:   File "/usr/lib64/python3.10/threading.py", line 1551, in _shutdown
                out:     lock.acquire()
                out:   File "/usr/local/lib/python3.10/site-packages/keylime-0.0.0-py3.10.egg/keylime/keylime_agent.py", line 710, in shutdown_handler
                out:     sys.exit(0)
                out: SystemExit: 0
                out: :: [ 16:23:31 ] :: [   PASS   ] :: Command 'limeStopAgent' (Expected 0, got 0)
```
Not sure if that helps.

THS-on:
@kkaarreell thank you for the detailed information. Can I get this output log also from the CI or do I have to run the tests locally for that.

It seems that there is still some thread running. I join all the threads/processes that I started, so I don't know what thread might cause this.

Do you have a simple way to reproduce it outside the tests because if I just start the agent and stop it with SIGINT it works on my machine?

kkaarreell:
Hello @THS-on , unfortunately agent.log is not available through CI, Packit interface doesn't allow to upload attachments. Also, I believe that agent "freeze" has something to do the with the tested scenario (maybe something about revocation?) since the 2nd test is not failing this way and agent stops properly. I will do more tests today in order to identify the relevant test scenario that makes such a difference.

Btw, would it help to print e.g. last 20 lines from service log when service stop fails?

THS-on:
>  I will do more tests today in order to identify the relevant test scenario that makes such a difference.

Awesome, thank you.

> Btw, would it help to print e.g. last 20 lines from service log when service stop fails?

That might be helpful to at least get an idea where in the process it failed.

kkaarreell:
Hi @THS-on , according to my testing it is `autorun.sh` script in payload that makes a difference. 
In my test it looks like this
https://github.com/RedHat-SP-Security/keylime-tests/blob/main/functional/basic-attestation-on-localhost/payload/autorun.sh
I have confirmed that agent stops properly when this file is excluded from the payload.

THS-on:
@kkaarreell thank you for finding this.

I found out that the payload thread was not terminating correctly. This is now fixed by marking this thread as a daemon and fixing the infinity loop.

keylime/keylime_agent.py
```diff
@@ -338,7 +337,7 @@ def initthread():
                         logger.info("No payload script %s found in %s/unzipped", initscript, secdir)
                     else:
                         logger.info("Executing payload script: %s/unzipped/%s", secdir, initscript)
-                        payload_thread = threading.Thread(target=initthread)
+                        payload_thread = threading.Thread(target=initthread, daemon=True)
```
aplanas, line 340: I am not sure here. Maybe this is needed and correct, but I feel that making it possible that the Python code can ends and exit before the thread is done, can cause problems.

For example, I am not sure what will happen if the agent throws the payload script in a thread, that is executed inside `{secdir}/zypper`, and we end the agent that maybe will try to umount `{secdir}`. This can make the umount to fail and leave the `{secdir}` present in the system, or cause problems in the payload script later.

keylime/keylime_agent.py
```diff
@@ -620,74 +686,33 @@ def main():
...
+        instance_tpm.flush_keys()
+        logger.debug("Flushed keys successfully")
+        sys.exit(0)
+
+    signal.signal(signal.SIGTERM, shutdown_handler)
+    signal.signal(signal.SIGQUIT, shutdown_handler)
+    signal.signal(signal.SIGINT, shutdown_handler)
+
+    # Keep the main thread alive by waiting for the server thread
+    serverthread.join()
```
aplanas, line 718: do we need to do `revocation_process.join()`?

maybe we can register the payload script thread too, and join it here too?

demo/agent_monitor/tenant_agent_monitor.py
```diff
@@ -24,6 +24,8 @@
 import keylime_logging
 import common
 import sys
+
+import keylime.web_util
```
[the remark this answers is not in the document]
THS-on: Because I don't know if this demo code actually works this only changed by the automatic refactoring.

keylime/cloud_verifier_tornado.py
```diff
@@ -723,7 +724,7 @@ async def invoke_get_quote(agent, need_pubkey):
 
     if response.status_code != 200:
         # this is a connection error, retry get quote
-        if response.status_code == 599:
+        if response.status_code in [500, 599]:
```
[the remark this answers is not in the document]
THS-on: The issue is that socket timeouts are treated as `500` status codes. If we find another way to distinguish them from other `500` issues I'm happy to change this.

keylime/keylime_agent.py
THS-on: Yeah you are right. This fix is not totally optimal, but do we want the agent to hang if the payload script does not terminate?

keylime/keylime_agent.py
THS-on: Not really because this will already keep the main thread active until we call the shutdown handler.

keylime/cloud_verifier_tornado.py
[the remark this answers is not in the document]
maugustosilva: I have been seeing a growing number of transient `500` status codes which are exactly sockets timeouts, at scale (with hundreds to thousands of `agents`). It is imperative we retry in case we get the aforementioned status code.

THS-on:
@mpeters I think this is now ready to merge.

mpeters (approved):

PR #842, opened 2022-01-20
verifier: do not call finish() twice

THS-on:
web_util.echo_json_response(..) already calls finish. Calling it twice results in an exception.


maugustosilva (approved):

maugustosilva:
Noticed the same problem while running "chaos monkey" tests (killing processes), a simple and necessary fix

PR #846, opened 2022-01-24
verifier: use json.loads(..) instead of ast.literal_eval(..)

THS-on:
The data sent via the API is JSON not Python dicts.

This failed once I build more complex policies for IMA or measured boot.

@maugustosilva can you confirm that you also expect it to be JSON and not Python dicts?


maugustosilva:
Yes, LGTM

maugustosilva (approved):

Issue #840, opened 2022-01-19
keylime agent fails to run on systems with certain Nuvoton 75x chips.


galmasi:
Is your an issue a feature request? If so, please raise it as an [enhancement](https://github.com/keylime/enhancements)

# Environment
 * OS / version: Ubuntu 18.04 with custom-build backport kernel 
 * Processor architecture: AMD64 with an actual AMD processor
 * TPM Manufacturer: Nuvoton 75x with a funny vendor string
 * Keylime version: 6.2.0

# Description 

Keylime agent fails to start on a Lenovo 635 rack server with the above mentioned Nuvoton TPM chip. The failure is caused by the fact that the output from `tpm2_getcap properties-fixed` is not valid YAML. Keylime calls "getcap" to get the manufacturer ID from the TPM device (this is used to determine whether the TPM is "real" or "simulated").

For reference, here is the actual output from invoking the tpm2_getcap command:

```
TPM2_PT_VENDOR_STRING_1:
  raw: 0x4E504354
  value: "NPCT"
TPM2_PT_VENDOR_STRING_2:
  raw: 0x37357800
  value: "75x"
TPM2_PT_VENDOR_STRING_3:
  raw: 0x22212134
  value: ""!!4"             <----------- quote character inside string causes failure
TPM2_PT_VENDOR_STRING_4:
  raw: 0x726C7300
  value: "rls"
```

The YAML parser fails because the output above is not valid YAML. The keylime agent does not catch the YAML processing failure in any way, causing the agent to stop.

## Expected behavior vs. actual behavior 

1. tpm2-tools should have produced valid YAML in any situation. Issue opened https://github.com/tpm2-software/tpm2-tools/issues/2921
2. keylime_agent should catch YAML parsing failures.
3. Possibly keylime_agent should *not* rely on the output of tpm2-tools being valid YAML. It would be possible, for example, to disregard the "value" entries from the output and parse the raw values instead.

## Steps to reproduce problem
1. Run `tpm2_getcap properties-fixed` on any system with the above named Nuvoton chip
2. Confirm that the output from the call is not valid YAML
3. Run keylime_agent in that same setup, and watch it crash.

# Relevant logs

Not attaching the log because in this case it actually is detrimental to understanding the issue. All it shows is a YAML parsing failure, and the root cause for said failure can be visualizer more easily by calling `tpm2_getcap`.

mpeters:
@galmasi just verifying, but this issue should be closed with your fix, correct?

PR #848, opened 2022-01-25
Fix for Issue #840

galmasi:
This is a small fix to proactively fix Issue #840 by identifying non-escaped double quotes in the tpm2-tools output.

The fix is tricky because it is very difficult to test -- I have access to exactly 1 device that shows this error.
In addition, the tpm2-tools have already fixed the defect, so must make sure *escaped* double quotes don't get escaped again in error.


mpeters:
Can we add unit tests for this?

galmasi:
Had a discussion with Marcio about just how to do this. I will submit a commit in a few that adds an extra optional parameter to the function we are testing. The default None value for the param would let the system go on as before -- call tpm2-tools. Otherwise we are free to unit test the function by providing fake output from tpm2-tools as a string.

maugustosilva (approved):

PR #852, opened 2022-01-26
Move umask calls only on entry points

aplanas:
`umask` is set several times in the code, causing sinde effects.
    
This patch move out the calls to `umask` from the utility code into the entry points of the different services (agent, verifier, registrar and ca).


mpeters (approved):

keylime/tpm/tpm_abstract.py
```diff
@@ -133,10 +133,9 @@ def __read_tpm_data(self):
             return {}
 
     def __write_tpm_data(self):
-        os.umask(0o077)
         if os.geteuid() != 0 and config.REQUIRE_ROOT:
             logger.warning("Creating tpm metadata file without root. Sensitive trust roots may be at risk!")
-        with open('tpmdata.yml', 'w', encoding="utf-8") as f:
+        with os.fdopen(os.open("tpmdata.yml", os.O_WRONLY | os.O_CREAT, 0o600), "w", encoding="utf-8") as f:
```
THS-on, line 138: Can you also change this to not depend on the CWD?

aplanas: Indeed, I was planing to do this in a different refactoring PR. For example, ca_util if full of this problem.

You want me to fix this issue in tpm_abstract here?

THS-on: No, then put it in the separate PR.

THS-on (approved):

PR #853, opened 2022-01-27
Add exponential backoff on failing for tpm, tenant and verifier.

ruocco:
This small patchset adds the option on `keylime.conf`, enabled by default, to use exponential backoff when communication fails (with tpm, tenant and verifier).

Formula is `wait time = base ^ ntries`. With default base value `2`, so that wait times are `1, 2, 4, 8, 16, [...]`, still stopping on max_ntries.

### DIFF
--- a/keylime/common/retry.py
+++ b/keylime/common/retry.py
@@ -1,0 +1,13 @@
+"""
+SPDX-License-Identifier: Apache-2.0
+Copyright 2021 Angelo Ruocco - IBM Research Lab Zurich
+"""
+
+def retry_time(exponential, base, ntries, logger):
+    if exponential:
+        if base > 1:
+            return base**ntries
+        elif logger:
+            logger.warning("Base %f incompatible with exponential backoff", base)
+
+    return abs(base)
--- a/keylime.conf
+++ b/keylime.conf
@@ -438,12 +438,16 @@ accept_tpm_hash_algs = sha512,sha384,sha256,sha1
 accept_tpm_encryption_algs = ecc,rsa
 accept_tpm_signing_algs = ecschnorr,rsassa
 
-# How long to wait between failed attempts to connect to a cloud agent, in
-# seconds.  Floating point values accepted here.
-retry_interval = 1
+# Wether or not to use an exponantial backoff algorithm for retries.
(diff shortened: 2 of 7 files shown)


keylime/common/retry.py
```diff
@@ -0,0 +1,11 @@
+"""
+SPDX-License-Identifier: Apache-2.0
+Copyright 2021 Angelo Ruocco - IBM Research Lab Zurich
+"""
+
+def retry_time(exponential, base, ntries):
+    if base < 1 and exponential:
+        print("Base value %f incompatible with exponantial backoff" % base)
```
THS-on, line 8: Instead of using print, pass the logger as an argument.

Can you add a test for this function?

THS-on (commented):
Exponential backoff is a nice feature, but with 10 retries the agent has 2046 seconds to retry before it fails which is over 30 minutes. We the should change it to something like 4 for the TPM and 5 for connections to the agent.

keylime/common/retry.py
```diff
@@ -0,0 +1,11 @@
...
+        return base
+    else:
+        return base**(int(not exponential) + ntries * exponential)
```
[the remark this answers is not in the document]
ruocco: I have updated the PR, added unit tests, changed the print for a log, changed default retries values, made the retry algorithm more clear.

keylime/common/retry.py
```diff
@@ -0,0 +1,16 @@
...
+    if exponential:
+        if base > 1:
+            return base**ntries
+        else:
+            try:
+                logger.warning("Base %f incompatible with exponential backoff", base)
+            except:
+                None
+
+    return abs(base)
```
THS-on, line 16: Instead of catching all exceptions you can also just check if the logger is there:
```suggestion
def retry_time(exponential, base, ntries, logger=None):
    if exponential:
        if base > 1:
            return base**ntries
        elif logger:
            logger.warning("Base %f incompatible with exponential backoff", base)

    return abs(base)
```

ruocco: Ah yeah of course, fixed!

THS-on (approved):

maugustosilva (approved):

PR #865, opened 2022-01-31
registrar, tpm_main: ensure that correct types are commited to DB.

THS-on:
bytes do not work well with postgres and bool is stored as int, so we convert them.

@kkaarreell can you check if that fixes your Postgres issue?

### DIFF
--- a/keylime/registrar_common.py
+++ b/keylime/registrar_common.py
@@ -94,7 +94,7 @@ class ProtectedHandler(BaseHTTPRequestHandler, SessionManager):
                 logger.warning('GET returning 404 response. agent_id %s not found.', agent_id)
                 return
 
-            if not agent.active:
+            if not bool(agent.active):
                 web_util.echo_json_response(self, 404, "agent_id not yet active")
                 logger.warning('GET returning 404 response. agent_id %s not yet active.', agent_id)
                 return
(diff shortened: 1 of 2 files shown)


THS-on:
This is more of a hotfix than solving the underling issue.

Supporting multiple databases (in our case sqlite, mysql and postgres) is tricky even with SQLAlchemy, because not every DB supports every data type and has limitations in their own way. Ideally we should have a Python object that we convert to and from the DB object, but the current code does not do that. Also we currently have no type annotations and checks in place which are enforced mypy and pylint, which makes it a lot harder to reason about what type a variable has.

I really like what [pydantic](https://pydantic-docs.helpmanual.io/usage/models/#orm-mode-aka-arbitrary-class-instances) is doing for type safety and ORM, but I don't know if we can use it in Keylime.

maugustosilva:
I cannot quite test it fully (i.e., all combinations with, "identity only", "identity + mb refstate" and "indentity + mb_refstate + IMA allowlist v1") with MySQL due to the problem with (the absence of) LONGTEXT in certain columns. However, if it works for both sqlite and postgresql, we should merge and selectively fix it on MySQL if needed.

mpeters (approved):

PR #815, opened 2021-12-22
db: only set length on Text type if supported

THS-on:
PostgreSQL does not support Text with a length MySQL does.

Should fix #814

PR #866, opened 2022-01-31
Fix LONGTEXT column type migration for MySQL

kkaarreell:
Without this fix LONGTEXT column type is converted to TEXT type
which isn't big enough to store large data like allowlist.

### DIFF
--- a/keylime/migrations/versions/eeb702f77d7d_allowlist_rename.py
+++ b/keylime/migrations/versions/eeb702f77d7d_allowlist_rename.py
@@ -35,10 +35,10 @@ def downgrade_registrar():
 def upgrade_cloud_verifier():
     # need to use batch_alter_table since SQLite <3.25.0 doesn't do RENAME COLUMN
     with op.batch_alter_table('verifiermain') as batch_op:
-        batch_op.alter_column('ima_whitelist', new_column_name='allowlist', existing_type=sa.Text(), existing_nullable=True)
+        batch_op.alter_column('ima_whitelist', new_column_name='allowlist', existing_type=sa.Text().with_variant(sa.Text(429400000), "mysql"), existing_nullable=True)
 
 
 def downgrade_cloud_verifier():
     # need to use batch_alter_table since SQLite <3.25.0 doesn't do RENAME COLUMN
     with op.batch_alter_table('verifiermain') as batch_op:
-        batch_op.alter_column('allowlist', new_column_name='ima_whitelist', existing_type=sa.Text(), existing_nullable=True)
+        batch_op.alter_column('allowlist', new_column_name='ima_whitelist', existing_type=sa.Text().with_variant(sa.Text(429400000), "mysql"), existing_nullable=True)


THS-on (approved):
LGTM

mpeters (changes requested):
You can't update an existing migration with changes because it won't work on existing installs. You must create a new migration for that.

THS-on:
> You can't update an existing migration with changes because it won't work on existing installs. You must create a new migration for that.

I think it is fine because it is currently broken, so nobody could have a working install with that. We already did something similar here: https://github.com/keylime/keylime/pull/815

The test is failing because of an upstream issue: https://github.com/sqlalchemy/alembic/issues/982

maugustosilva:
I can confirm it is currently broken (for a MySQL test installation).

mpeters:
It's only broken if they try to use large allowlists. Maybe they were doing signatures before, or had small allowlists and large exclude lists (focusing only on the kernel or something else). Doing the fix like this instead of as a new migration will mean they will have to destroy their database, then upgrade and let keylime recreate the schema. I'm against any fix that requires something like this which can just be a new migration and upgrade like any other.

THS-on:
> It's only broken if they try to use large allowlists.

Oh, I thought it was fully broken for MySQL. If not then I agree that we need to add an migration on top of this fix to also change it current installations.

kkaarreell:
Would that be just another file with different name? Or is it more complicated?

THS-on:
> Would that be just another file with different name? Or is it more complicated?

You need to create a new migration file with alembic:
```
alembic -c keylime/migrations/alembic.ini revision -m "Added account table"
```

mpeters:
Alembic can be a tricky tool to use as it's not in the normal development flow. I don't mind taking a crack at it if you'd like @kkaarreell

maugustosilva:
@mpeters I have some free time now and can do that... whomever does it, please keep in mind that the `mb_refstate` column needs to be fixed too

kkaarreell:
> Alembic can be a tricky tool to use as it's not in the normal development flow. I don't mind taking a crack at it if you'd like @kkaarreell

Please do @mpeters or @maugustosilva . Thank you.

mpeters:
> @mpeters I have some free time now and can do that... whomever does it, please keep in mind that the `mb_refstate` column needs to be fixed too

@maugustosilva would you mind taking a crack it since you want this other change in there as well?

maugustosilva:
Certainly, on it

kkaarreell:
Should I close this PR?

kkaarreell:
I am sorry, I am already OOO, will check it tomorrow. Eventually, you can check it using tests from pgsql branch of keylime-tests. Either on Fedora system with tmt or by modifying the test to run additional SQL command and creating PR for keylime-tests repo.

kkaarreell:
> @kkaarreell while attempting to deploy with `MySQL`, I see the `allowlist` column is already of type `LONGTEXT`, but `mb_restate` is not. Are you seeing differently in `PostgreSQL`?

To clarify, the problem I was observing was with mariadb on F35. PostgreSQL didn't manifest that problem with insufficient field size for allow list.

mpeters:
New migration was added to fix this in https://github.com/keylime/keylime/pull/871

Issue #867, opened 2022-01-31
Typo in configuration file: listen_notfications instead of listen_notifications


ansasaki:
Is your an issue a feature request? If so, please raise it as an [enhancement](https://github.com/keylime/enhancements)

# Environment
 * OS / version: N/A
 * Processor architecture: N/A 
 * TPM Manufacturer:  N/A
 * Keylime version: Current master (1ebc78bb)

# Description 
The configuration ``listen_notfications`` in ``cloud_agent`` section has a typo which is propagated throughout the code and also to the rust agent:

https://github.com/keylime/keylime/blob/1ebc78bb8a8e6884d34a33f763a94f874574dbd1/keylime.conf#L87

## Expected behavior vs. actual behavior 
Ideally the configuration keywords are spelled correctly to avoid confusion. Currently a user could unintentionally disable the service listening for revocation notifications by editing the configuration file manually.

## Steps to reproduce problem
1. Check the relevant line from the default configuration file:
https://github.com/keylime/keylime/blob/1ebc78bb8a8e6884d34a33f763a94f874574dbd1/keylime.conf#L87

# Relevant logs

N/A

PR #861, opened 2022-01-28
Handle missing PCR banks dependent on the hash algorithm correctly

THS-on:
The hash algorithm itself is there as a key in the output, but the value is None.

This happened on a Debian + swtpm system were the `tpm2_pcrread` output looks like the following:
```
sha1:
sha256:
  0 : 0xE21B703EE69C77476BCCB43EC0336A9A1B2914B378944F7B00A10214CA8FEA93
  1 : 0xCB0FD2FC0B87CD3676C1E370DE8446AB4967CF92356B6FCE001CC4F05932FCD6
  2 : 0xE8D71943876AA0BE05CF35E32E451C0435388D3B44FA4586843E4B5227033C87
  3 : 0xE21B703EE69C77476BCCB43EC0336A9A1B2914B378944F7B00A10214CA8FEA93
  4 : 0x5A1C3A55A9E0A037F9FAB97B15C7A71F30DFE362C477957F26671C1165104766
  5 : 0x4D7544A02B1EEB2D34DD93A453EF23385EB378F527B21AB094E78E7BD2F160DC
  6 : 0xE21B703EE69C77476BCCB43EC0336A9A1B2914B378944F7B00A10214CA8FEA93
  7 : 0xE21B703EE69C77476BCCB43EC0336A9A1B2914B378944F7B00A10214CA8FEA93
  8 : 0x0000000000000000000000000000000000000000000000000000000000000000
  9 : 0x0000000000000000000000000000000000000000000000000000000000000000
  10: 0x64BAC4DD70BD36018665AB297E89D8CAF98A0A83E44CCD33800EB18A2ED80405
  11: 0x0000000000000000000000000000000000000000000000000000000000000000
  12: 0x0000000000000000000000000000000000000000000000000000000000000000
  13: 0x0000000000000000000000000000000000000000000000000000000000000000
  14: 0x0000000000000000000000000000000000000000000000000000000000000000
  15: 0x0000000000000000000000000000000000000000000000000000000000000000
  16: 0x0000000000000000000000000000000000000000000000000000000000000000
  17: 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF
  18: 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF
[...]

PR #868, opened 2022-01-31
Fix typo for config param listen_notifications

mpeters:
Fixes issue #867

### DIFF
--- a/keylime.conf
+++ b/keylime.conf
@@ -84,7 +84,7 @@ extract_payload_zip = True
 agent_uuid = d432fbb3-d2f1-4a97-9ef7-75bd81c00000
 
 # Whether to listen for revocation notifications from the verifier or not.
-listen_notfications = True
+listen_notifications = True
 
 # The path to the certificate to verify revocation messages received from the
 # verifier.  The path is relative to /var/lib/keylime.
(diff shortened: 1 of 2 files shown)


THS-on (approved):
The CI failure seems not to be related to this change because it only happens for one system, LGTM.

@kkaarreell the revocation action test seems flaky. Do you know if this is a timing issue or something related to zeromq?

kkaarreell:
> @kkaarreell the revocation action test seems flaky. Do you know if this is a timing issue or something related to zeromq?

It is possible that it is a timing issue. I have never observed it before on a localhost test but I saw something like that on multi-host test. Did it failed just once? I think it should be possible to reschedule just that one failing test eventually from the Details page.

THS-on:
> Did it failed just once? 

No it happens quite frequently, but goes away after a few retries. For example in https://github.com/keylime/keylime/pull/861 there were multiple tries necessary to get them all to pass.

kkaarreell:
> > Did it failed just once?
> 
> No it happens quite frequently, but goes away after a few retries. For example in #861 there were multiple tries necessary to get them all to pass.

Should be addressed by https://github.com/RedHat-SP-Security/keylime-tests/pull/46

PR #873, opened 2022-02-01
setup.py and config improvements

### LINKED ISSUE #548: Python 3.9 is out

# Environment
 * OS / version: any
 * Processor architecture: any
 * TPM Manufacturer: any
 * Keylime version: `master` branch

# Description 
I just noticed there is a list of Python minor versions, at https://github.com/keylime/keylime/blob/1d0f695c0848b001acb0ab01de46094494884506/setup.py#L53-L55 , and it does not include the latest (3.9).

## Steps to reproduce problem
1. 

# Relevant logs

THS-on:
The keylime.conf is copied into the package path instead of trying to install it globally into /etc which fails if the user is non root.
We will fallback on this configuration the other potential ones do not exist.
This makes it easier to install Keylime in virtual environments.

Also new Python versions are added and a version for the package is set. 
(I've used 6.4.0 because this will probably the next version.)

Fixes #548

### DIFF
--- a/.ci/test_wrapper.sh
+++ b/.ci/test_wrapper.sh
@@ -34,5 +34,13 @@ then
 else
     REPO_DIR="/root/keylime"
 fi
+
+# Move /etc/keylime.conf because there might a old one distributed with the container.
+if [ -f "/etc/keylime.conf" ]
+then
+    echo "Moving /etc/keylime.conf from the container to /etc/keylime.conf.orig"
+    mv /etc/keylime.conf /etc/keylime.conf.orig
+fi
+
 chmod +x $REPO_DIR/test/run_tests.sh
 $REPO_DIR/test/run_tests.sh -s openssl
--- a/setup.py
+++ b/setup.py
@@ -12,6 +12,7 @@ with open('README.md', encoding="utf-8") as fh:
 
 setuptools.setup(
     name='keylime',
+    version='6.3.1',
     description=(
         'TPM-based key bootstrapping and system '
         'integrity measurement system for cloud'),
@@ -36,6 +37,8 @@ setuptools.setup(
         'Programming Language :: Python :: 3.6',
         'Programming Language :: Python :: 3.7',
         'Programming Language :: Python :: 3.8',
+        'Programming Language :: Python :: 3.9',
+        'Programming Language :: Python :: 3.10',
         'Topic :: System :: Hardware',
     ],
     entry_points={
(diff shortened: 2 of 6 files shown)


THS-on:
The CI tests are failing because the `keylime.conf` is no longer installed to `/etc/keylime.conf` by the python package which is exactly what this intend behind this change is.

.ci/test_wrapper.sh
```diff
@@ -34,5 +34,13 @@ then
 else
     REPO_DIR="/root/keylime"
 fi
+
+# Remove /etc/keylime.conf because there might a old one distributed with the container.
+if [ -f "/etc/keylime.conf" ]
+then
+    echo "Removing old /etc/keylime.conf"
+    rm /etc/keylime.conf
```
mpeters, line 42: Instead of removing it, what do you think about copying it to say `/etc/keylime.conf.bak` or something. This is just for running the tests, but sometimes as a developer I run the ci wrapper just to make sure everything is working, and I'd hate to lose anything I'd done to the keylime.conf just in case.

setup.py
```diff
@@ -2,14 +2,22 @@
...
+import shutil
+
 import setuptools
 
+
 with open('README.md', encoding="utf-8") as fh:
     long_description = fh.read()
 
+# Copy keylime.conf temporarily into the keylime folder.
+# This is done in oder to include it as a package resource because data_files is deprecated.
```
mpeters, line 15: Typo in comments, should be "order"

setup.py
```diff
@@ -2,14 +2,22 @@
...
+shutil.copy("keylime.conf", "keylime")
 
 setuptools.setup(
     name='keylime',
+    version='6.4.0',
```
mpeters, line 20: Let's not bump this up just yet. I expect we'll have a 6.3.1 shortly since we have a lot of minor improvements recently.

mpeters (changes requested):

setup.py
maugustosilva: +1

.ci/test_wrapper.sh
THS-on: The config is now moved to `/etc/keylime.conf.orig` instead of deleting it.

THS-on:
@kkaarreell when this gets merged, we will have a fallback configuration in the path of the Python package and now longer force it to be installed to `/etc/keylime.conf`. If you want to modify the config you'll need to copy it to `/etc/keylime.conf` manually.

kkaarreell:
Hi @THS-on , I have pushed small update to keylime-tests, could you try rescheduling CI tests to see if it fixes the failure?
https://github.com/RedHat-SP-Security/keylime-tests/pull/48

THS-on:
@mpeters instead of copying the `keylime.conf` into the `keylime` folder temporary I created now a symlink.
This has the advantage that something like `python -m test.test_web_util` now works in a venv out of the box.

PR #880, opened 2022-02-07
cloud_verifier_tornado: use fork_processes

aplanas:
If the cloud_verifier/multiprocessing_pool_num_workers is different from
1, the call to the `.start()` process will fails, as previous call to
`.add_stockets()` is already initializing the internal ioloop.

The raised exception will be:

  File "/usr/bin/keylime_verifier", line 11, in <module>
    load_entry_point('keylime==6.3.0', 'console_scripts', 'keylime_verifier')()
  File "/usr/lib/python3.6/site-packages/keylime/cmd/verifier.py", line 21, in main
    cloud_verifier_tornado.main()
  File "/usr/lib/python3.6/site-packages/keylime/cloud_verifier_tornado.py", line 1122, in main
    server.start(config.getint('cloud_verifier', 'multiprocessing_pool_num_workers'))
  File "/usr/lib64/python3.6/site-packages/tornado/tcpserver.py", line 220, in start
    process.fork_processes(num_processes)
  File "/usr/lib64/python3.6/site-packages/tornado/process.py", line 129, in fork_processes
    raise RuntimeError("Cannot run in multiple processes: IOLoop instance "
RuntimeError: Cannot run in multiple processes: IOLoop instance has already been initialized. You cannot call IOLoop.instance() before calling start_processes()

This was introduced in https://github.com/keylime/keylime/commit/50661f8b33f6b7335104cd4c0dfff711705ee96e

This patch revert back to call `.process.fork_processes()` after the
`.bind_sockets()` line, that is happening before the `.start()`, and
drop the optional parameter in the last method call.

### DIFF
--- a/keylime/cloud_verifier_tornado.py
+++ b/keylime/cloud_verifier_tornado.py
@@ -1121,13 +1121,16 @@ def main():
     sockets = tornado.netutil.bind_sockets(
         int(cloudverifier_port), address=cloudverifier_host)
 
+    tornado.process.fork_processes(config.getint(
+        'cloud_verifier', 'multiprocessing_pool_num_workers'))
+
     server = tornado.httpserver.HTTPServer(app, ssl_options=context, max_buffer_size=max_upload_size)
     server.add_sockets(sockets)
 
     signal.signal(signal.SIGTERM, lambda *_: sys.exit(0))
 
     try:
-        server.start(config.getint('cloud_verifier', 'multiprocessing_pool_num_workers'))
+        server.start()
         if tornado.process.task_id() == 0:
             # Start the revocation notifier only on one process
             if config.getboolean('cloud_verifier', 'revocation_notifier'):
--- a/keylime/crypto.py
+++ b/keylime/crypto.py
@@ -211,5 +211,5 @@ def generate_selfsigned_cert(name, key, valid_until) -> x509.Certificate:
         .serial_number(x509.random_serial_number())\
         .not_valid_before(datetime.datetime.utcnow())\
         .not_valid_after(valid_until)\
-        .sign(key, hashes.SHA256())
+        .sign(key, hashes.SHA256(), backend=default_backend())
     return cert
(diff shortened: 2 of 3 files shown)


THS-on:
LGTM. The order is the same as described in the Tornado documentation: https://www.tornadoweb.org/en/stable/httpserver.html#http-server

@kkaarreell is it possible to add a version of CentOS that has Python 3.6 to the CI?

THS-on (approved):

maugustosilva (approved):

lkatalin:
Sorry I can't quite tell, would any of these changes need to be included in the Rust agent as there was some code change in `agent.py` that seems related to crypto?

THS-on:
@lkatalin I don't think so.

mpeters (approved):

kkaarreell:
> @kkaarreell is it possible to add a version of CentOS that has Python 3.6 to the CI?

Particular CentOS Stream version contains the same python version as the respective RHEL release. Python 3.6 is AFAIK in RHEL-6 so in theory we could enable CI for CentOS Stream 8 but I didn't check whether keylime would work for some time. I am afraid that we do not have all the build dependencies for keylime available.

aplanas:
@kkaarreell I am testing it on SLE that have Python 3.6. Keylime is working there.

Seems that the problems are with Tornado and the crypto libraries. If they are old the Keylime code needs to be adapted.  So far the adaptations are present until v6.3.0, but I am not testing master.

kkaarreell:
Hi @aplanas, the thing is that in Packit CI we are limited to Fedora and CentOS Stream distributions. @sergio-correia provides required keylime deps for CentOS Stream 9 in a Copr repo, maybe he could prepare similar repo for CS8.

sergio-correia:
I actually have it already, and I have a stale PR enabling it for the tests repository. I will go back to it and see if we can get it merged.

PR #881, opened 2022-02-07
Fix small wording mistakes caught while reading the code

axelsimon:


sergio-correia (approved):

keylime/tenant.py
```diff
@@ -615,7 +615,7 @@ def do_cv(self):
             sys.exit()
 
     def do_cvstatus(self):
-        """Perform operational state look up for agent"""
+        """Perform operational state look up for cloudverifier"""
```
mpeters, line 618: This still isn't correct. It's looking up the agent's status on the verfier. This is in contrast to the agent's status on the registrar (the other status command).

mpeters (changes requested):

keylime/tenant.py
axelsimon: Got you. Correcting this. "Operational state look up for agent on verifier" is a bit of a mouthful, but i'm going with this for now. I'll try to find something better for "keylimectl".

mpeters (approved):

PR #886, opened 2022-02-11
agent, crypto: add localhost, server and contact ip to agent certificate

### LINKED ISSUE #882: Improve CN / SAN of generated certificates
Currently, Keylime generates certificates that are only valid for one common name:
`CN = keylime `
and
```
X509v3 Subject Alternative Name:
    DNS:keylime
```

This causes issues with tools that verify a certificate's validity more strictly, namely Go's standard library's TLS, which expect to see a given IP in the cert's SAN field (subject alt names) when connecting directly to an IP address.

One of the difficulties is that we don't know in advance what IP address a Keylime component will be listening on. However, we should always be able to listen and serve on the different variations of localhost/127.0.0.1.

Therefore, as discussed in the bi-weekly meeting, we should add the following entries to the SAN fields when generating a certificate:
- 127.0.0.1
- ::1
- localhost
- localhost.localdomain
- CONTACT_IP ([see enhancement #47](https://github.com/keylime/enhancements/blob/master/47_agent_ip_port.md))

THS-on:
I've also added the ip where the agent binds to to the certificate.

Fixes #882

### DIFF
--- a/keylime/crypto.py
+++ b/keylime/crypto.py
@@ -10,6 +10,9 @@ import hashlib
 import os
 import secrets
 
+from ipaddress import IPv4Address, IPv6Address, ip_address
+from typing import Optional, List
+
 # Crypto implementation using python cryptography package
 from cryptography import exceptions
 from cryptography import x509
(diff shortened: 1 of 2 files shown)


mpeters (approved):

lkatalin:
@THS-on is this something we should also do on the Rust side?

THS-on:
@lkatalin yes. Keylime ignores this attributes, but @axelsimon run into the issue that it is not that simple to just ignore host names in go. It doesn't hurt adding them.

PR #888, opened 2022-02-11
Add config necessary for Python actions in Rust agent

### LINKED ISSUE #887: Add config options as part of enhancement-55
We need changes to `keylime.conf` in order for the Rust agent to be able to run revocation scripts as outlined in https://github.com/keylime/enhancements/blob/master/55_revocation_actions_without_python.md.

The work to bring the Rust agent in line with the Python agent in terms of revocation scripts is [pending a merge](https://github.com/keylime/rust-keylime/pull/321) into the Rust agent repo, and requires specific changes to `keylime.conf`. Those changes must be made here instead of in the Rust agent repo, since `keylime.conf` from this repo is what actually gets installed with `./installer.sh`. Without these changes, we can't update the way the Rust agent runs revocation scripts (at least not as we currently plan to update it in the [PR](https://github.com/keylime/rust-keylime/pull/321)).

Will attach a PR to make this change shortly.

lkatalin:
Fixes #887 

This is a direct copy of @ansasaki 's keylime.conf change in the
Rust agent repo in https://github.com/keylime/rust-keylime/pull/321.
We need it here so that when Keylime is installed with ./installer.sh,
these config options are available to the Rust agent - and probably
will be used by the Python agent eventually as part of
https://github.com/keylime/enhancements/blob/master/55_revocation_actions_without_python.md

Signed-off-by: Lily Sturmann <lsturman@redhat.com>
Co-authored-by: Anderson Toshiyuki Sasaki <ansasaki@redhat.com>

### DIFF
--- a/keylime.conf
+++ b/keylime.conf
@@ -110,6 +110,16 @@ revocation_actions=
 # with a working directory of /var/lib/keylime/secure/unzipped.
 payload_script=autorun.sh
 
+# The path to the directory containing the pre-installed revocation action
+# scripts.  Ideally should point to an fixed/immutable location subject to
+# attestation.  The default is /usr/libexec/keylime.
+revocation_actions_dir = /usr/libexec/keylime
+
+# Whether to allow running revocation actions sent as part of the payload.  The
+# default is True and setting as False will limit the revocation actions to the
+# pre-installed ones.
+allow_payload_revocation_actions = True
+
 # Jason @henn made be do it! He wanted a way for Keylime to measure the
 # delivered payload into a pcr of choice.
 # Specify a PCR number to turn it on.


lkatalin:
I'm not sure how to fix this Greetings error:
`Error: Resource not accessible by integration`

keylime.conf
```diff
@@ -110,6 +110,16 @@ revocation_actions=
 # with a working directory of /var/lib/keylime/secure/unzipped.
 payload_script=autorun.sh
 
+# The path to the directory containing the pre-installed revocation action
+# scripts.  Ideally should point to an fixed/immutable location subject to
+# attestation.  The default is /usr/libexec/keylime.
+revocation_actions_dir = /usr/libexec/keylime
```
mpeters, line 116: I'm curious why these are needed. It looks like it's just setting them as the default. Does the rust agent not use the same defaults?

lkatalin: @mpeters It's based on [enhancement 55](https://github.com/keylime/enhancements/blob/master/55_revocation_actions_without_python.md#design-details), if you look at the design details. Not sure if there is already an issue open to implement all of the changes in that doc, if so then I could link it. The other parts of the enhancement are still to-do, but we can't get this [PR](https://github.com/keylime/rust-keylime/pull/321) through on the Rust agent until these config options are added because it expects them and throws an error otherwise.

This is actually part of a larger meta-discussion that could be had on how to keep the two repos in sync for changes like this.

If it would be too disorganized to add this to `keylime/keylime`, we can probably do a workaround on the Rust agent side instead. But if this change needs to be made to `keylime/keylime` anyway eventually I figured I would go ahead and add it.

ansasaki: @mpeters When I wrote the changes to the rust agent I did not add hardcoded defaults and expected the values to always be present in the configuration file, which now seems to me a naive approach.

I think the best here is to add hardcoded default values so that the rust agent is able to run even when the values are not present in the configuration file. I'll make this change to the [PR](https://github.com/keylime/rust-keylime/pull/321).

This doesn't solve the problem of having the configuration files out of sync, though.

Issue #896, opened 2022-02-21
revocation_notifier won't connect to webhook without simplejson


kkaarreell:
Is your an issue a feature request? If so, please raise it as an [enhancement](https://github.com/keylime/enhancements)

# Environment
 * OS / version: Fedora36
 * Processor architecture: x86_64 
 * TPM Manufacturer: swtpm
 * Keylime version: current upstream (3590e21)

# Description 

Without python3-simplejson installed revocation_notifier won't connect to the webhook URL but fail with
```
2022-02-21 15:07:34.264 - keylime.revocation_notifier - INFO - Sending revocation event via webhook...
Exception in thread Thread-2:
Traceback (most recent call last):
  File "/usr/lib64/python3.10/threading.py", line 1009, in _bootstrap_inner
    self.run()
  File "/usr/lib64/python3.10/threading.py", line 946, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/local/lib/python3.10/site-packages/keylime-6.3.1-py3.10.egg/keylime/revocation_notifier.py", line 123, in wor
ker_webhook
    response = session.post(url, json=tosend, timeout=5)
  File "/usr/lib/python3.10/site-packages/requests/sessions.py", line 577, in post
    return self.request('POST', url, data=data, json=json, **kwargs)
  File "/usr/lib/python3.10/site-packages/requests/sessions.py", line 515, in request
    prep = self.prepare_request(req)
  File "/usr/lib/python3.10/site-packages/requests/sessions.py", line 443, in prepare_request
    p.prepare(
  File "/usr/lib/python3.10/site-packages/requests/models.py", line 321, in prepare
    self.prepare_body(data, files, json)
  File "/usr/lib/python3.10/site-packages/requests/models.py", line 473, in prepare_body
    body = complexjson.dumps(json, allow_nan=False)
  File "/usr/lib64/python3.10/json/__init__.py", line 238, in dumps
    **kw).encode(obj)
  File "/usr/lib64/python3.10/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib64/python3.10/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/usr/lib64/python3.10/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type bytes is not JSON serializable
```

This can be currently seen in 
[...]

PR #897, opened 2022-02-23
revocation_notifier: convert the data to str in the notifiers

sergio-correia:
python-requests will internally use either simplejson (preferred)
or the built-in the json module; when it uses the latter, it may
encounter issues handling bytes instead of strings.

Let's convert the data to str beforand, to avoid this issue.

Resolves: #896

### DIFF
--- a/keylime/revocation_notifier.py
+++ b/keylime/revocation_notifier.py
@@ -80,6 +80,12 @@ def stop_broker():
 
 
 def notify(tosend):
+    # python-requests internally uses either simplejson (preferred) or
+    # the built-in json module, and when it is using the built-in one,
+    # it may encounter difficulties handling bytes instead of strings.
+    # To avoid such issues, let's convert `tosend' to str beforehand.
+    tosend = json.bytes_to_str(tosend)
+
     def worker(tosend):
         context = zmq.Context()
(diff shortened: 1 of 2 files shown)


keylime/revocation_notifier.py
```diff
@@ -80,6 +80,12 @@ def stop_broker():
 
 
 def notify(tosend):
+    # python-requests internally uses either simplejson (preferred) or
+    # the built-in json module, and when it is using the built-in one,
+    # it may encounter difficulties handling bytes instead of strings.
+    # To avoid such issues, let's convert `tosend' to str beforehand.
+    tosend = json.__bytes_to_str(tosend)
```
THS-on, line 87: `__bytes_to_str` is marked as a private function with the `__` prefix. Can you change the name of the function to `bytes_to_str` if we are now using it outside the `json.py` file? 

Otherwise LGTM.

THS-on (changes requested):

keylime/revocation_notifier.py
sergio-correia: Good point, I have updated it.

THS-on (approved):

maugustosilva (approved):

mpeters (approved):

PR #900, opened 2022-02-23
Drop privileges of agent process after startup

stefanberger:
This PR modifies the keylime agent in such a way that root-owned files are opened near process start and then the agent process drops privileges by switching to another user specified by the run_as variable in keylime.conf. By default no privileges are dropped since the root user now needs to prepare the filesystem beforehand and give the run_as user access to the WORK_DIR, which is typically /var/lib/keylime.

### DIFF
--- a/keylime/keylime_agent.py
+++ b/keylime/keylime_agent.py
@@ -43,6 +43,7 @@ from keylime import json
 from keylime import revocation_notifier
 from keylime import registrar_client
 from keylime import secure_mount
+from keylime import user_utils
 from keylime import web_util
 from keylime import api_version as keylime_api_version
 from keylime.common import algorithms, validators
@@ -637,7 +638,18 @@ def main():
                                'but it\'s is not found on the system.')
 
     # initialize the tmpfs partition to store keys if it isn't already available
-    secure_mount.mount()
+    secdir = secure_mount.mount()
+
+    # Now that operations requiring root privileges are done, drop privileges
+    # if 'run_as' is available in the configuration.
+    if os.getuid() == 0:
+        run_as = config.get('cloud_agent', 'run_as', fallback='')
+        if run_as != '':
+            user_utils.chown(secdir, run_as)
+            user_utils.change_uidgid(run_as)
+            logger.info(f"Dropped privileges to {run_as}")
+        else:
+            logger.warning("Cannot drop privileges since 'run_as' is empty or missing in keylime.conf agent section.")
 
     # Instanitate TPM class
 
--- a/keylime/keylime_agent.py
+++ b/keylime/keylime_agent.py
@@ -636,6 +636,9 @@ def main():
             raise RuntimeError('agent_uuid is configured to use dmidecode, '
                                'but it\'s is not found on the system.')
 
+    # initialize the tmpfs partition to store keys if it isn't already available
+    secure_mount.mount()
+
     # Instanitate TPM class
 
     instance_tpm = tpm()
@@ -650,9 +653,6 @@ def main():
     contact_port = os.getenv("KEYLIME_AGENT_CONTACT_PORT", None)
     if contact_port is None and config.has_option('cloud_agent', 'agent_contact_port'):
         contact_port = config.get('cloud_agent', 'agent_contact_port', fallback="invalid")
-
-    # initialize the tmpfs partition to store keys if it isn't already available
-    secure_mount.mount()
 
     # change dir to working dir
     fs_util.ch_dir(config.WORK_DIR)
--- a/keylime/keylime_agent.py
+++ b/keylime/keylime_agent.py
@@ -181,12 +181,7 @@ class Handler(BaseHTTPRequestHandler):
                 ima_ml_entry = int(ima_ml_entry)
                 if ima_ml_entry > self.server.next_ima_ml_entry:
                     ima_ml_entry = 0
-                ima_log_file = None
-                if os.path.exists(config.IMA_ML):
-                    ima_log_file = open(config.IMA_ML, 'r', encoding="utf-8")
-                ml, nth_entry, num_entries = ima.read_measurement_list(ima_log_file, ima_ml_entry)
-                if ima_log_file:
-                    ima_log_file.close()
+                ml, nth_entry, num_entries = ima.read_measurement_list(self.server.ima_log_file, ima_ml_entry)
                 if num_entries > 0:
                     response['ima_measurement_list'] = ml
                     response['ima_measurement_list_entry'] = nth_entry
@@ -403,7 +398,7 @@ class CloudAgentHTTPServer(ThreadingMixIn, HTTPServer):
     next_ima_ml_entry = 0 # The next IMA log offset the verifier may ask for.
     boottime = int(psutil.boot_time())
 
-    def __init__(self, server_address, RequestHandlerClass, agent_uuid, contact_ip):
+    def __init__(self, server_address, RequestHandlerClass, agent_uuid, contact_ip, ima_log_file):
         """Constructor overridden to provide ability to pass configuration arguments to the server"""
         # Find the locations for the U/V transport and mTLS key and certificate.
         # They are either relative to secdir (/var/lib/keylime/secure) or absolute paths.
@@ -459,6 +454,7 @@ class CloudAgentHTTPServer(ThreadingMixIn, HTTPServer):
             self, server_address, RequestHandlerClass)
         self.enc_keyname = config.get('cloud_agent', 'enc_keyname')
         self.agent_uuid = agent_uuid
+        self.ima_log_file = ima_log_file
 
     def add_U(self, u):
         """Threadsafe method for adding a U value received from the Tenant
@@ -622,6 +618,10 @@ def main():
         if not os.access(ML, os.F_OK):
             logger.warning("Measurement list path %s not accessible by agent. Any attempt to instruct it to access this path - via \"keylime_tenant\" CLI - will result in agent process dying", ML)
 
+    ima_log_file = None
+    if os.path.exists(config.IMA_ML):
+        ima_log_file = open(config.IMA_ML, 'r', encoding="utf-8")
+
     if config.get('cloud_agent', 'agent_uuid') == 'dmidecode':
         if os.getuid() != 0:
             raise RuntimeError('agent_uuid is configured to use dmidecode, '
@@ -728,7 +728,7 @@ def main():
     if keylime_ca == "default":
         keylime_ca = os.path.join(config.WORK_DIR, 'cv_ca', 'cacert.crt')
 
-    server = CloudAgentHTTPServer(serveraddr, Handler, agent_uuid, contact_ip)
+    server = CloudAgentHTTPServer(serveraddr, Handler, agent_uuid, contact_ip, ima_log_file)
     context = web_util.generate_mtls_context(server.mtls_cert_path, server.rsakey_path, keylime_ca, logger=logger)
     server.socket = context.wrap_socket(server.socket, server_side=True)
     serverthread = threading.Thread(target=server.serve_forever, daemon=True)
--- a/keylime/keylime_agent.py
+++ b/keylime/keylime_agent.py
@@ -192,12 +192,10 @@ class Handler(BaseHTTPRequestHandler):
             # generally speaking, retrieving the 15Kbytes of a boot log does not seem significant compared to the
             # potential Mbytes of an IMA measurement list.
             if TPM_Utilities.check_mask(imaMask, config.MEASUREDBOOT_PCRS[0]):
-                if not os.path.exists(config.MEASUREDBOOT_ML):
-                    logger.warning("TPM2 event log not available: %s", config.MEASUREDBOOT_ML)
+                if not self.server.tpm_log_file_data:
+                    logger.warning(f"TPM2 event log not available: {config.MEASUREDBOOT_ML}")
                 else:
-                    with open(config.MEASUREDBOOT_ML, 'rb') as f:
-                        el = base64.b64encode(f.read())
-                    response['mb_measurement_list'] = el
+                    response['mb_measurement_list'] = self.server.tpm_log_file_data
 
             web_util.echo_json_response(self, 200, "Success", response)
             logger.info('GET %s quote returning 200 response.', rest_params["quotes"])
@@ -398,7 +396,7 @@ class CloudAgentHTTPServer(ThreadingMixIn, HTTPServer):
     next_ima_ml_entry = 0 # The next IMA log offset the verifier may ask for.
     boottime = int(psutil.boot_time())
 
-    def __init__(self, server_address, RequestHandlerClass, agent_uuid, contact_ip, ima_log_file):
+    def __init__(self, server_address, RequestHandlerClass, agent_uuid, contact_ip, ima_log_file, tpm_log_file_data):
         """Constructor overridden to provide ability to pass configuration arguments to the server"""
         # Find the locations for the U/V transport and mTLS key and certificate.
         # They are either relative to secdir (/var/lib/keylime/secure) or absolute paths.
@@ -455,6 +453,7 @@ class CloudAgentHTTPServer(ThreadingMixIn, HTTPServer):
         self.enc_keyname = config.get('cloud_agent', 'enc_keyname')
         self.agent_uuid = agent_uuid
         self.ima_log_file = ima_log_file
+        self.tpm_log_file_data = tpm_log_file_data
 
     def add_U(self, u):
         """Threadsafe method for adding a U value received from the Tenant
@@ -622,6 +621,11 @@ def main():
     if os.path.exists(config.IMA_ML):
         ima_log_file = open(config.IMA_ML, 'r', encoding="utf-8")
 
+    tpm_log_file_data = None
+    if os.path.exists(config.MEASUREDBOOT_ML):
+        with open(config.MEASUREDBOOT_ML, 'rb') as tpm_log_file:
+            tpm_log_file_data = base64.b64encode(tpm_log_file.read())
+
     if config.get('cloud_agent', 'agent_uuid') == 'dmidecode':
         if os.getuid() != 0:
             raise RuntimeError('agent_uuid is configured to use dmidecode, '
@@ -728,7 +732,7 @@ def main():
     if keylime_ca == "default":
         keylime_ca = os.path.join(config.WORK_DIR, 'cv_ca', 'cacert.crt')
 
-    server = CloudAgentHTTPServer(serveraddr, Handler, agent_uuid, contact_ip, ima_log_file)
+    server = CloudAgentHTTPServer(serveraddr, Handler, agent_uuid, contact_ip, ima_log_file, tpm_log_file_data)
     context = web_util.generate_mtls_context(server.mtls_cert_path, server.rsakey_path, keylime_ca, logger=logger)
     server.socket = context.wrap_socket(server.socket, server_side=True)
     serverthread = threading.Thread(target=server.serve_forever, daemon=True)
--- a/keylime/keylime_agent.py
+++ b/keylime/keylime_agent.py
@@ -181,7 +181,12 @@ class Handler(BaseHTTPRequestHandler):
                 ima_ml_entry = int(ima_ml_entry)
                 if ima_ml_entry > self.server.next_ima_ml_entry:
                     ima_ml_entry = 0
-                ml, nth_entry, num_entries = ima.read_measurement_list(config.IMA_ML, ima_ml_entry)
+                ima_log_file = None
+                if os.path.exists(config.IMA_ML):
+                    ima_log_file = open(config.IMA_ML, 'r', encoding="utf-8")
+                ml, nth_entry, num_entries = ima.read_measurement_list(ima_log_file, ima_ml_entry)
+                if ima_log_file:
+                    ima_log_file.close()
                 if num_entries > 0:
                     response['ima_measurement_list'] = ml
                     response['ima_measurement_list_entry'] = nth_entry
--- a/keylime/secure_mount.py
+++ b/keylime/secure_mount.py
@@ -4,6 +4,7 @@ Copyright 2017 Massachusetts Institute of Technology.
 '''
 
 import os
+import shutil
 
 from keylime import keylime_logging
 from keylime import cmd_exec
@@ -85,6 +86,12 @@ def mount():
 
 def umount():
     """Umount all the devices mounted by Keylime."""
+
+    # Make sure we leave tmpfs dir empty even if we did not mount it or
+    # if we cannot unmount it. Ignore errors while deleting. The deletion
+    # of the 'secure' directory will result in an error since it's a mount point.
+    shutil.rmtree(mount(), ignore_errors=True)
+
     while _MOUNTED:
         directory = _MOUNTED.pop()
         logger.info("Unmounting %s", directory)
--- a/keylime/user_utils.py
+++ b/keylime/user_utils.py
@@ -1,0 +1,100 @@
+#!/usr/bin/python3
+
+'''
+SPDX-License-Identifier: Apache-2.0
+Copyright 2022 IBM Corporation
+'''
+
+import os
+import grp
+import pwd
+
+from keylime import keylime_logging
+
+# Configure logger
+logger = keylime_logging.init_logging('privileges')
+
+
+def string_to_uidgid(user_and_group):
+    """ Translate the user_and_group string to uid and gid.
+        The userandgroup parameter must be a string of the format '<user>[:[<group>]]'
+        or '[<user>]:gid'
+        User and group can be strings or integers. If no group is given, -1 will be
+        returned for gid.
+        :raises ValueError: if user or group could not be resolved
+    """
+    params = user_and_group.split(':')
+    if len(params) > 2:
+        raise ValueError(f"User and group {user_and_group} are in wrong format. Expected <user>[:[<group>]] or [<user>]:group")
+
+    gid = None
+    if len(params) == 2 and len(params[1]) > 0:
+        if params[1].isnumeric():
+            gid = int(params[1])
+            if gid < 0:
+                raise ValueError(f"User and group {user_and_group} contains an illegal value")
+        else:
+            try:
+                gr = grp.getgrnam(params[1])
+                gid = gr.gr_gid
+            except KeyError as e:
+                raise ValueError(f'Could not resolve group {params[1]}: {e}') from e
+
+    uid = None
+    if len(params[0]) > 0:
+        if params[0].isnumeric():
+            uid = int(params[0])
+            if uid < 0:
+                raise ValueError(f"User and group {user_and_group} contains an illegal value")
+        else:
+            try:
+                passwd = pwd.getpwnam(params[0])
+                uid = passwd.pw_uid
+            except KeyError as e:
+                raise ValueError(f'Could not resolve user {params[0]}: {e}') from e
+
+    if uid is None and gid is None:
+        raise ValueError(f"User and group {user_and_group} are in wrong format. Expected <user>[:[<group>]] or [<user>]:group")
+
+    return uid, gid
+
+
+def change_uidgid(user_and_group):
+    """ Change uid and gid of the current process.
+        The user_and_group parameter must be a string of the format
+        '<user>[:[<group>]]' or [<user>]:<group>. User and group can
+        be strings or integers.
+        :raises ValueError: if user or group could not be resolved
+        :raises RuntimeError: if setting gid or uid did not succeed
+    """
+    uid, gid = string_to_uidgid(user_and_group)
+
+    # First change group and then user
+    if gid is not None:
+        try:
+            os.setgid(gid)
+        except (OSError, PermissionError) as e:
+            raise RuntimeError(f'Could not set gid to {gid}: {e}') from e
+
+    if uid is not None:
+        try:
+            os.setuid(uid)
+        except (OSError, PermissionError) as e:
+            raise RuntimeError(f'Could not set uid to {uid}: {e}') from e
+
+
+def chown(path, user_and_group):
+    """ Change ownership on a given file to the new owner described in
+        user_and_group string.
+        The user_and_group parameter must be a string of the format
+        '<user>[:[<group>]]' or [<user>]:<group>. User and group can
+        be strings or integers.
+        :raises ValueError: if user or group could not be resolved
+    """
+    uid, gid = string_to_uidgid(user_and_group)
+    if uid is None:
+        uid = -1
+    if gid is None:
+        gid = -1
+
+    os.chown(path, uid, gid)
--- a/keylime.conf
+++ b/keylime.conf
@@ -146,6 +146,13 @@ tpm_signing_alg = rsassa
 # create a new EK upon startup, and neither will it flush the EK upon exit.
 ek_handle = generate
 
+# The user account to switch to to drop privileges when started as root
+# If left empty, the agent will keep running with high privileges.
+# The user and group specified here must allow the user to access the
+# WORK_DIR (typically /var/lib/keylime) and /dev/tpmrm0. Therefore,
+# suggested value for the run_as parameter is keylime:tss.
+run_as =
+
 #=============================================================================
 [cloud_verifier]
 #=============================================================================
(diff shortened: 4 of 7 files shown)


kkaarreell:
Hi @stefanberger , out of curiosity I have tried this change and my agent fails to access `/var/lib/keylime/secure` and `/var/lib/keylime/secure/tmpi8raqg76`. Both the mount point and the file are created (I believe by agent itself) with ownership `root:root` despite my configuration expects `kagent:tss`. I think agent should change the ownership accordingly before dropping root privileges.
Pasting agent log below
```
2022-02-24 04:27:01.277 - keylime.secure_mount - INFO - mounting secure storage location /var/lib/keylime/secure on tmpfs
2022-02-24 04:27:01.339 - keylime.tpm - WARNING - INSECURE: Keylime is using a software TPM emulator rather than a real hardware TPM.
2022-02-24 04:27:01.339 - keylime.tpm - WARNING - INSECURE: The security of Keylime is currently NOT linked to a hardware root of trust.
2022-02-24 04:27:01.339 - keylime.tpm - WARNING - INSECURE: Only use Keylime in this mode for testing or debugging purposes.
2022-02-24 04:27:01.339 - keylime.tpm - INFO - Taking ownership with config provided TPM owner password
2022-02-24 04:27:01.545 - keylime.tpm - INFO - TPM Owner password confirmed: keylime
2022-02-24 04:27:02.342 - keylime.cloudagent - INFO - Agent UUID: d432fbb3-d2f1-4a97-9ef7-75bd81c00000
2022-02-24 04:27:02.344 - keylime.cloudagent - INFO - Dropped privileges to kagent:tss
2022-02-24 04:27:02.346 - keylime.cloudagent - INFO - Key for U/V transport and mTLS certificate not found, generating a new one
  File "/usr/local/bin/keylime_agent", line 33, in <module>
    sys.exit(load_entry_point('keylime==6.3.1', 'console_scripts', 'keylime_agent')())
  File "/usr/local/lib/python3.10/site-packages/keylime-6.3.1-py3.10.egg/keylime/cmd/agent.py", line 15, in main
    keylime_agent.main()
  File "/usr/local/lib/python3.10/site-packages/keylime-6.3.1-py3.10.egg/keylime/keylime_agent.py", line 749, in main
    server = CloudAgentHTTPServer(serveraddr, Handler, agent_uuid, contact_ip, ima_log_file, tpm_log_file)
  File "/usr/local/lib/python3.10/site-packages/keylime-6.3.1-py3.10.egg/keylime/keylime_agent.py", line 423, in __init__
    with open(keyname, "wb") as f:
PermissionError: [Errno 13] Permission denied: '/var/lib/keylime/secure/tci_rsa_key'
```

keylime/keylime_agent.py
```diff
@@ -723,7 +727,26 @@ def main():
...
         keylime_ca = os.path.join(config.WORK_DIR, 'cv_ca', 'cacert.crt')
 
-    server = CloudAgentHTTPServer(serveraddr, Handler, agent_uuid, contact_ip)
+    ima_log_file = None
+    if os.path.exists(config.IMA_ML):
+        ima_log_file = open(config.IMA_ML, 'r', encoding="utf-8")
+
+    tpm_log_file = None
+    if os.path.exists(config.MEASUREDBOOT_ML):
+        tpm_log_file = open(config.MEASUREDBOOT_ML, 'rb')
```
THS-on, line 736: The UEFI event log can be loaded into memory instead of reading it every time, because it doesn't change after boot.

keylime/user_utils.py
```diff
@@ -0,0 +1,71 @@
...
+
+def string_to_uidgid(userandgroup):
+    """ Translate the userandgroup string to uid and gid.
+        The userandgroup parameter must be a string of the format '<user>[:<group>]'.
+        User and group can be strings or integers. If no group is given, -1 will be
+        returned for gid.
+    """
+    params = userandgroup.split(':')
+    if len(params) > 2:
+        return None, None, "User and group '%s' are in wrong format. Expected <user>[:group]" % userandgroup
```
THS-on, line 26: Use the new Python 3 format strings. 
```suggestion
        return None, None, f"User and group '{userandgroup}' are in wrong format. Expected <user>[:group]"
```

keylime/user_utils.py
```diff
@@ -0,0 +1,71 @@
...
+
+    gid = -1
+    if len(params) == 2:
+        if params[1].isnumeric():
+            gid = int(params[1])
+        else:
+            try:
+                gr = grp.getgrnam(params[1])
+                gid = gr.gr_gid
+            except Exception as e:
```
THS-on, line 36: Try to catch only exceptions that are expected. In this example:
```suggestion
            except KeyError as e:
```

THS-on (changes requested):
If possible use Python 3 format strings.

THS-on:
Do we want to drop to `tss:tss` or something  like `keylime:keylime` by default?

keylime/user_utils.py
```diff
@@ -0,0 +1,71 @@
...
+    """ Change uid and gid of the current process.
+        The userandgroup parameter must be a string of the format
+        '<user>[:<group>]'. User and group can be strings or integers.
+    """
+    uid, gid, err = string_to_uidgid(userandgroup)
+    if err:
+        return err
+
+    # First change group and then user
+    if gid >= 0:
```
aplanas, line 61: I can see how `gid == -1` is to maintain the same user, but for me is a more natural way that `if uid is not None` and `if gid is not None` imply to maintain the same `uid` or `gid` independently.  This open the door to have `uid`, `uid:`, `:gid`, `uid:gid` combinations in the `cloud_agent.run_as` parameter.

All this can be covered updating the parser, and here treated as a two independent choices.

Also a test for those situation will make easy the refactoring of the code.

keylime/user_utils.py
```diff
@@ -0,0 +1,71 @@
...
+        try:
+            os.setgid(gid)
+        except Exception as e:
+            return 'Could not set gid to %d: %s' % (gid, str(e))
```
aplanas, line 65: I think that is OK to kill the process here? An error in the active `uid` or `gid` should be fatal, shouldn't?

aplanas (commented):
Just some comments. Thanks!!

stefanberger:
> I think agent should change the ownership accordingly before dropping root privileges.

I actually had code that did that but the way config.WORK_DIR is set it seemed more like a potential security risk. When I ran the keylime test from the privileged container (`cd .ci; run_local.sh`) it ended up changing ownership of all files and directories under the keylime git checkout. So maybe what it could do instead is try to recursively access() all files under config.WORK_DIR and error out with a message when a file isn't readable so that root knows what to do.

aplanas:
> > I think agent should change the ownership accordingly before dropping root privileges.
> 
> I actually had code that did that but the way config.WORK_DIR is set it seemed more like a potential security risk.

Oh I missed this. I think that the directory should be there with the correct user/group beforehand, and should not be created by keylime agent in any case. I mean the `WORK_DIR` alone (`/var/lib/keylime`).  The subtree from there it should be created by the agent after the change of user.

I am testing the patch to see that this is what we are doing.

keylime/user_utils.py
```diff
@@ -0,0 +1,71 @@
...
+    if params[0].isnumeric():
+        uid = int(params[0])
+    else:
+        try:
+            passwd = pwd.getpwnam(params[0])
+            uid = passwd.pw_uid
+        except Exception as e:
+            return None, None, 'Could not resolve user %s: %s' % (params[0], str(e))
+
+    return uid, gid, None
```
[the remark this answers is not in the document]
stefanberger: Hm ... So, yes, it's a go pattern.  You mean it should throw a RuntimeError()? What I don't like about python is that sometimes you don't know what all exceptions can surface from functions. I would at least not kill the process from within this library function here but let the caller handle the error...

kkaarreell:
> > I think agent should change the ownership accordingly before dropping root privileges.
> 
> I actually had code that did that but the way config.WORK_DIR is set it seemed more like a potential security risk. When I ran the keylime test from the privileged container (`cd .ci; run_local.sh`) it ended up changing ownership of all files and directories under the keylime git checkout. So maybe what it could do instead is try to recursively access() all files under config.WORK_DIR and error out with a message when a file isn't readable so that root knows what to do.

I meant changing the ownership specifically for files that agent is creating. I believe this is necessary for the agent to successfully  start right on the first attempt. 

Btw, it would be handy if WORKDIR could be configured in keylime.conf itself. Exporting variables requires more changes when starting services using systemd unit file.

FTR, I am developing a test using this feature in 
https://github.com/RedHat-SP-Security/keylime-tests/pull/61

keylime/user_utils.py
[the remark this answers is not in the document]
THS-on: Yeah it is not ideal how exceptions in Python work. I would raise in this case a ValueError because it could not resolve the user/group. We can start to add comments for the functions with something like `:raises ValueError: if user or group could not be resolved`, so that we can more easily see which of our functions raise exceptions.

keylime/user_utils.py
stefanberger: Done now and raising exception with this string.

keylime/user_utils.py
stefanberger: Fixed.

stefanberger: Fixed.

keylime/user_utils.py
```diff
@@ -0,0 +1,71 @@
...
+import grp
+import pwd
+
+from keylime import keylime_logging
+
+# Configure logger
+logger = keylime_logging.init_logging('privileges')
+
+
+def string_to_uidgid(userandgroup):
```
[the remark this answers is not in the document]
stefanberger: Done.

keylime/user_utils.py
stefanberger: Added unittest and swtiched to use `None`.

keylime/keylime_agent.py
stefanberger: Yes, reading it only once.

stefanberger:
@kkaarreell  Would your tests work if you ran `chown -R user:group /var/lib/keylime` before starting the daemon?

kkaarreell:
> @kkaarreell Would your tests work if you ran `chown -R user:group /var/lib/keylime` before starting the daemon?

@stefanberger I am configuring agent to use `/var/lib/keylime-agent` instead and do `chown` before starting it.
See https://github.com/RedHat-SP-Security/keylime-tests/pull/61/files#diff-830c32e28b1c8b0d067879a7d638e827ad4b8c4ff0a6773ce1bd5d5cb6c839e3R45

THS-on:
The drop of privileges has to occur before the tpm or the webserver object are created so that they already create their files with the correct user. @kkaarreell is is the cause behind our found issues.

How do we want to handle the tpmfs creation and removal? Because we cannot call umount as a non root user.

The agent shutdown handler should probably close the the IMA log.

kkaarreell:
Could we delegate this work on a separate thread that won't drop its privileges? Or what would be the setup that would ensure that agent starts properly on the first attempt?

THS-on:
> Could we delegate this work on a separate thread that won't drop its privileges?

Yes but with shutdown handling this looks for a easy source to create bugs, so I would avoid it if possible.

> Or what would be the setup that would ensure that agent starts properly on the first attempt?

I'm fine with throwing an error if Keylime should not run as root, but should create it's own tpmfs. For packaging I think that the best way is to just ship a mount point that is created by systemd (https://github.com/utkarsh2102/python-keylime/blob/master/debian/python3-keylime-agent.var-lib-keylime-secure.mount).

kkaarreell:
But even if the mount point exists it would error out due to the file mentioned in case 3. How to address that one?

THS-on:
> But even if the mount point exists it would error out due to the file mentioned in case 3. How to address that one?

`tpmdata.yml` and `tmp8eh2kxg7` are created by creating the tpm object in the Python code before we drop the privileges. We just need to move the code for dropping privileges a few lines up.

stefanberger:
I moved the opening of files to an earlier position and now also added a patch that modifies file ownership of a few directories and files before dropping privileges. Unfortunately I have some issues with the hardware TPM 2 in a system (seems broken) I was testing and cannot currently go further than this. But maybe someone can give it a try again.

kkaarreell:
I have just pushed test update that creates the mount point prior starting the agent. Test results should be available in ~10 minutes at https://github.com/RedHat-SP-Security/keylime-tests/pull/61/

kkaarreell:
> I have just pushed test update that creates the mount point prior starting the agent. Test results should be available in ~10 minutes at [RedHat-SP-Security/keylime-tests#61](https://github.com/RedHat-SP-Security/keylime-tests/pull/61)

And the updated test `/functional/basic-attestation-on-localhost` has passed.

stefanberger:
> I have just pushed test update that creates the mount point prior starting the agent. Test results should be available in ~10 minutes at [RedHat-SP-Security/keylime-tests#61](https://github.com/RedHat-SP-Security/keylime-tests/pull/61)

Looks good...

keylime/keylime_agent.py
```diff
@@ -723,7 +733,24 @@ def main():
...
-    server = CloudAgentHTTPServer(serveraddr, Handler, agent_uuid, contact_ip)
+    # Now that root-owned files are opened drop privileged
+    # If 'run_as' is not available keep running with high privileges
+    if os.getuid() == 0:
+        run_as = config.get('cloud_agent', 'run_as', fallback='')
+        if run_as != '':
+            for p, recurse in [("", False),
+                               ("tpmdata.yml", False),
+                               ("secure", True),
+                               ("cv_ca", True)]:
```
kkaarreell, line 744: This might makes sense since `cv_ca` is the default location when searching for certiticates but it is also possible that this directory won't exist. Is this really intentional? I am asking because the other two locations are created by the agent itself but that does not apply to `cv_ca`.

stefanberger: I wasn't sure what creates this directory. How should it be handled?

kkaarreell:
It is created by verifier which uses it to store generated certificates, as described here.
https://github.com/keylime/keylime/blob/3590e21b7e4b48a2023aadf2486b116ef5be2375/keylime.conf#L169
While agent is using cacert and client-cert from this directory (not in my case as I am copying them to other location) I think it should not own that whole directory. Personally, I would not list it as these certificates should exist before agent starts and therefore it is up to admin to configure permissions and agent accordingly.
@THS-on thoughts?

keylime/keylime_agent.py
kkaarreell: It is created by verifier which uses it to store generated certificates, as described here.
https://github.com/keylime/keylime/blob/3590e21b7e4b48a2023aadf2486b116ef5be2375/keylime.conf#L169
While agent is using cacert and client-cert from this directory (not in my case as I am copying them to other location) I think it should not own that whole directory. Personally, I would not list it as these certificates should exist before agent starts and therefore it is up to admin to configure permissions and agent accordingly.
@THS-on thoughts?

stefanberger:
> It is created by verifier which uses it to store generated certificates, as described here.

I didn't know that. I have the situation where the agent needs it and that would mean agent and verifier need to be co-located or the verifier would have to be started once to create this directory and files.

I now modified the code to change ownership on this directory and its files as well if it exists, doing nothing otherwise.

```
2022-02-24 18:14:54.252 - keylime.cloudagent - INFO - Using existing mTLS cert in /var/lib/keylime/secure/tci_mtls_cert
  File "/usr/local/bin/keylime_agent", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.9/site-packages/keylime/cmd/agent.py", line 15, in main
    keylime_agent.main()
  File "/usr/local/lib/python3.9/site-packages/keylime/keylime_agent.py", line 754, in main
    context = web_util.generate_mtls_context(server.mtls_cert_path, server.rsakey_path, keylime_ca, logger=logger)
  File "/usr/local/lib/python3.9/site-packages/keylime/web_util.py", line 110, in generate_mtls_context
    context.load_verify_locations(cafile=ca_path)
PermissionError: [Errno 13] Permission denied
```

kkaarreell:
> > It is created by verifier which uses it to store generated certificates, as described here.
> 
> I didn't know that. I have the situation where the agent needs it and that would mean agent and verifier need to be co-located or the verifier would have to be started once to create this directory and files.

Not necessarily. Admin can generate those certificates in advance (e.g. using the `keylime_ca` utility) and either copy them in place or point agent to the respective location and ensure that agent will be able to read them. What I am trying to say is that if an admin wants agent running as non-root one has to configure system accordingly in advance. Changing ownership of files and directories that are possibly used by other keylime services and thus making them unaccessible (e.g. after service restart) doesn't seem right.

keylime/keylime_agent.py
aplanas: Is there a possibility of not doing this?

IMHO this is a security hole. If some user manage to put something there, it will be now automatically owned by the agent.  Is very good that the current code is picky and is selecting by name the elements that want to change, but still we can crate link and start making the agent to chown stuff outside the WORK_DIR.

Also, as commented here, this makes the verifier and the agent not co-installable. For me this is OK, but is a signal of that there is something wrong in the design.

Ideally the agent should not take any action, here nor outside, as the less code the more secure will be (and dumb). It should check if some action can be done, and if no die!

We need to document all the preparations required to work the agent in different scenario. For me is totally OK the idea suggesting that if the agent cannot mount the tmpfs, is game over.  If it is not root, the admin needs to prepare the system so the agent can mount the tmpfs (maybe adjusting fstab, or better with a systemd mount service).  The same for the certificates: if the agent cannot read them, something did go wrong during the installation, so die.

Does this makes sense? Am I missing something important that is against this idea?

THS-on: I agree with @aplanas we should not change permissions of arbitrary files. My proposed changes are:
1. move the privilege drop before  `# Instanitate TPM class` (line 640) (Note that dmidecode should run before because I think it requires root)
2. Fail if tpmfs is not there. Do not try to create it if the agent does not run as root.
3. Fail if the Keylime CA certificate is not readable

stefanberger: Why do we have tmpfs there? Is this to be able to unmount and forget the files under this directory, so an equivalent of an `rm -rf <dir>` at shutdown that wouldn't need root rights and therefore be much more flexible?

> but still we can crate link and start making the agent to chown stuff outside the WORK_DIR.

I tried to prevent exactly this by using `follow_symlinks = False`.
```
os.chown(rootpath, uid, gid, follow_symlinks = False)
```

os.walk doesn't follow links by default.
```
os.walk(top[, topdown=True[, onerror=None[, followlinks=False]]])
```

aplanas: > Why do we have tmpfs there?

Do not take my word, but I think that is optional. The admin can create a `secure` directory beforehand, and will be used by the agent. Only if is not present will create one as a `tmpfs`. The advantage is that is not persistent, so will be gone always, as you commented.

> I tried to prevent exactly this by using `follow_symlinks = False`.

Yes, you are being super careful here. I liked a lot that you are enumerating the elements and not doing a wild `chown`.

But the basic idea remains, I think. If somehow a file is dropped here under with my "aplanas:user", it will be `chown`ned by the agent. This is the case for `cv_ca`, that can be there because was created by the verifier!

My suggestion is more like a principle: the agent should do the least amount of work possible, and IMHO the `chown` is something that we can skip here with preparation and documentation. I am not a security knowledge developer, but my intuition tells me that a `chown` is something that a hacker will find interesting.

stefanberger:
Update now moved the dropping of privileges to line 640, so pretty early. I can now run it with these preparatory steps:
```
$ chown -R tss /var/lib/keylime/cv_ca/
$ chown tss /var/lib/keylime
$ rm -rf /var/lib/keylime/secure/ /var/lib/keylime/tpmdata.yml  /var/lib/keylime/tmpfs-dev/
$ ls -la /var/lib/keylime
total 12
drwxr-xr-x.  4 tss  root   41 Feb 25 08:58 .
drwxr-xr-x. 72 root root 4096 Feb 25 08:38 ..
drwx------.  2 tss  root 4096 Apr 26  2021 cv_ca
drwxr-xr-x.  2 root root 4096 Feb 25 08:39 tpm_cert_store
```

keylime/secure_mount.py
```diff
@@ -63,7 +63,7 @@ def check_mounted(secdir):
 def mount():
     secdir = os.path.join(config.WORK_DIR, "secure")
 
-    if not config.MOUNT_SECURE:
+    if not config.MOUNT_SECURE or os.getuid != ():
```
aplanas, line 66: You mean `or not os.getuid()`? But I am not sure that this is correct. It can be possible that WORK_DIR is already there, `MOUNT_SECURE` as `False`, and after the downgrade the `secure` dir can be created.

If we are creating it as root, that means that we should downgrade early.

stefanberger:
I think this now it is getting close to being mergeable. The tmpfs is mounted early now and its ownership changed. At the end it's emptied before trying to unmount it.

keylime/keylime_agent.py
```diff
@@ -627,6 +637,20 @@ def main():
             raise RuntimeError('agent_uuid is configured to use dmidecode, '
                                'but it\'s is not found on the system.')
 
+    # initialize the tmpfs partition to store keys if it isn't already available
+    secdir = secure_mount.mount()
```
aplanas, line 641: I do not dislike this compromise here, but at this point I will fully go to this path that @THS-on pointed:

Copy https://github.com/utkarsh2102/python-keylime/blob/master/debian/python3-keylime-agent.var-lib-keylime-secure.mount into https://github.com/keylime/keylime/tree/master/services, and document it (maybe in the keylime.conf)

That way the package can easily downgrade the user changing the conf and enabling the mount point.

Also, you mean that the secure_mount.mount() can be in the same place and you can skip the last `chown`.

In any case, as commented, this is a nice compromise.

keylime/secure_mount.py
```diff
@@ -85,6 +86,13 @@ def mount():
 
 def umount():
     """Umount all the devices mounted by Keylime."""
+
+    # Make sure we leave tmpfs dir empty even if we did not mount it
+    try:
+        shutil.rmtree(mount())
+    except OSError:
+        pass
```
aplanas, line 94: I am not sure that this is OK. For one, if the user decided not to use `tmpfs` maybe really do not want to remove the payload (to debug, inspect or audit)

Also, we are calling `mount`, with is totally confusing because is a function that have a side effect (I know that in the happy path is idempotent, but is scary).

And finally we should at least `log.error()`.

But IMHO all this change should not be present?

aplanas:
I am very happy with this PR. I am changing the openSUSE package to include this, as I cannot wait! : )))

Also I am interested in trying to replicate this logic in the Rust agent, it no one is already doing it.

keylime/keylime_agent.py
stefanberger: > In any case, as commented, this is a nice compromise.

What does this mean now? Do you want me to keep the compromise or undo it and require that everyone use the systemd config file.

The thing is also the WORK_DIR ( secdir = $WORK_DIR/secure) can be set like this (config.py):

```
if not REQUIRE_ROOT:
    DEFAULT_WORK_DIR = os.path.abspath(".")
else:
    DEFAULT_WORK_DIR = '/var/lib/keylime'
WORK_DIR = os.getenv('KEYLIME_DIR', DEFAULT_WORK_DIR)
```

I am afraid that removing the compromise considering this me** wouldn't work for someone who set KEYLIME_DIR.

keylime/keylime_agent.py
```diff
@@ -627,6 +637,20 @@ def main():
...
+
+    # Now that root-owned files are opened drop privileged
+    # If 'run_as' is not available keep running with high privileges
+    if os.getuid() == 0:
```
sergio-correia, line 645: Just a thought: I wonder if we should create an additional setting, e.g `run_as_root` to be false by default, and only run the agent as root if this is explicitly enabled/asked for by the configuration.

stefanberger: I think that's the same as having the configuration `run_as = foo:bar` set to some explicit values, such as `foo:bar`. I left it blank in the default config file for backwards compatibility purposes.

keylime/keylime_agent.py
aplanas: > > In any case, as commented, this is a nice compromise.
> 
> What does this mean now? Do you want me to keep the compromise or undo it and require that everyone use the systemd config file.

Is your call. I personally prefer the systemd idea (but that is me). I am totally fine if remains as it is. I can add the mount service as a patch in the distribution, or rethink this later.

> 
> The thing is also the WORK_DIR ( secdir = $WORK_DIR/secure) can be set like this (config.py):
> 
> ```
> if not REQUIRE_ROOT:
>     DEFAULT_WORK_DIR = os.path.abspath(".")
> else:
>     DEFAULT_WORK_DIR = '/var/lib/keylime'
> WORK_DIR = os.getenv('KEYLIME_DIR', DEFAULT_WORK_DIR)
> ```

IMHO `REQUIRE_ROOT` is totally wrong. Seems to me a flag that is only used for testing.

The right flag is the one that you introduced: the user:group make `REQUIRE_ROOT` meaningless (except for old scripts, CI and the test suite)

Also `REQUIRE_ROOT` and `KEYLIME_DIR` are totally orthogonal, I can execute root  or non-root in `/var/lib/keylime`, and root or non-root in `what-ever-KEYLIME_DIR-says`.

Should be there later a different PR to address this.

> I am afraid that removing the compromise considering this me** wouldn't work for someone who set KEYLIME_DIR.

Not sure to understand this. If we are afraid of breaking the agent when KEYLIME_DIR is set to `/var/lib/keylime` and all that is inside is owned by `root:root`, and we now upgrade the agent and change the conf to use `keylime:tss`, and now the agent fails to run, that is exactly what should happen!

The migration process should be documented and executed by the package or by the admin, but not by the agent.

But if you mean that there is some incompatibility of setting `KEYLIME_DIR` and this PR, I do not see it.

With this PR it is now expected that the installation will set `KEYLIME_DIR` for whatever `user:group` is in the config.

keylime/secure_mount.py
stefanberger: I was wondering why we bother with mounting tmpfs at all when one can just remove a whole directory upon shutdown. Because if the shutdown doesn't happen (kill -9) then we are in the same situation as not removing the directory and files still lay around. So I am still wondering about the tmpfs.

For simplicity reasons I chose `shutil.rmtree()` to remove everything under and including the `secure` directory. We will get an error when this function tries to remove the mount point `secure` since tmpfs is still mounted. I don't think we should log this. Now using this (best-effort):

```
    # Make sure we leave tmpfs dir empty even if we did not mount it or
    # if we cannot unmount it. Ignore errors while deleting. The deletion
    # of the 'secure' directory will result in an error since it's a mount point.
    shutil.rmtree(mount(), ignore_errors=True)
```

aplanas: I see, but I still will like to drop this change and postpone it to a different PR, and focus this one only in the downgrade case.

I think that the `tmpfs` analysis can be more complicated.

keylime/keylime_agent.py
stefanberger: > But if you mean that there is some incompatibility of setting KEYLIME_DIR and this PR, I do not see it.

I do not see an incompatibility with setting KEYLIME_DIR, actually the opposite. If we were to rely on the systemd script for mounting we wouldn't be able to drop privileges since the secure directory may still have root ownership then  unless we chown() it as this PR does.

keylime/secure_mount.py
stefanberger: But this is part of downgrading since we cannot unmount anymore but have to delete the contents of this directory to achieve a similar effect. I think it would be a bug in this PR not to do this.

stefanberger: > I think that the `tmpfs` analysis can be more complicated.

Not sure. It's mounted at the beginning, filled with files, unmounted at the end and files are gone.

THS-on: if `config.SECURE_MOUNT` is true then we should remove the files otherwise we can keep them for debugging purposes.

maugustosilva (approved):

keylime/keylime_agent.py
THS-on: We should switch to `keylime:tss` in a future release, but only with a major version change. The current default is backwards compatible to current installs.

kkaarreell: While running as non-root by default would be better from a security perspective, I am afraid that such settings would not allow fluent upgrade for existing users. We should think through and document what would be the recommended setup and whether it could be automated (including the migration). 
One thing that comes to my mind is using certificates provided by CV (cacert+cline-cert) which is the current default. Either we change the default or find a way how to make these certificates accessible to the agent. Also, agent should not be able write to `/var/lib/keylime` directly (I am looking at you `tpmdata.yml`) as there is also the content of CV and R and enabling write also means we are enabling removal. Therefore agent should have a dedicated directory for all its content owned by the agent. If CV is configured to create these certificates it could prepare everything for the agent. There could be a dedicated script doing that so an administrator could eventually reuse it.

kkaarreell: Also, I think that this topic would probably deserve dedicated discussion on the keylime channel.

keylime/keylime_agent.py
aplanas: I have the feeling that I am misunderstanding something. There are like several directories here. One in KEYLIME_DIR, that will be created under installation, with the correct user.  There is the secure directory that as today can be pre-created by the user, or if is missing, it will be created as a mount with tmpfs.

In this scenario, if the secure dir was not created, it will not be cleaned by the agent, but if was created by the agent as a tmpfs, will be umounted.

So the agent sometimes have responsibility over secure and sometimes don't, depending of some config vars and preconditions (like if it was already present). I agree that this is a bit of a mess now and should be fixed.

With systemd is like the same. It should be expected that systemd will mount the secure in, maybe tmpfs (or in some weird encrypted fs, we do not know) and under the correct user permissions (if not, the agent will fail to write the payload and will die). It will be already present before the agent is executed, and will be umounted later after the agent is gone.

Basically we do not make managing secure a responsibility of the agent.

stefanberger:
I pushed some minor update with fixes to error strings and Copyright strings.

keylime/secure_mount.py
aplanas: > But this is part of downgrading since we cannot unmount anymore but have to delete the contents of this directory to achieve a similar effect. I think it would be a bug in this PR not to do this.

Exactly. Downgrading privileges makes umount impossible if secure is not part of fstab (with the user parameter). That is exactly we should make the mount / umount outside the agent.

If we insist of doing the mount / umount in the agent, we should the do mount after the drop of privileges (like was before). If the mount can be done so can be umounted.

stefanberger: > If we insist of doing the mount / umount in the agent, we should the do mount after the drop of privileges (like was before). If the mount can be done so can be umounted.

That's not possible:
```
$ whoami
stefanb
$ mkdir foo
$ mount -t tmpfs ./foo ./foo/
mount: /home/stefanb/dev/keylime/foo: must be superuser to use mount.
```

kkaarreell: Btw, Linux kernel should be able to do a lazy unmount.

aplanas: > > If we insist of doing the mount / umount in the agent, we should the do mount after the drop of privileges (like was before). If the mount can be done so can be umounted.
> 
> That's not possible:

It is possible. From `man 8 mount`:

```
   Non-superuser mounts
       Normally, only the superuser can mount filesystems. However, when fstab contains the user option on
       a line, anybody can mount the corresponding filesystem.

       Thus, given a line

          /dev/cdrom /cd iso9660 ro,user,noauto,unhide

       any user can mount the iso9660 filesystem found on an inserted CDROM using the command:

          mount /cd
```

kkaarreell: > Btw, Linux kernel should be able to do a lazy unmount.

Reading more about it now, most likely it won't be the safest approach.

aplanas:
I added the mount service here: https://github.com/keylime/keylime/pull/903

stefanberger:
> I added the mount service here: #903

Thanks for the information. Does this have an impact on this PR?

aplanas:
> Thanks for the information. Does this have an impact on this PR?

@stefanberger no, just indirectly in the sense that is another step in this direction.

I am still testing this change but with the exception that the mount is done as root and the removal during the umount, this patch LGTM so far.

If we do not agree, I think that we can merge this and work later on those issues

aplanas (approved):

stefanberger:
>> Thanks for the information. Does this have an impact on this PR?
> @stefanberger no, just indirectly in the sense that is another step in this direction.
> 
> I am still testing this change but with the exception that the mount is done as root and the removal during the umount, this patch LGTM so far.
> 
> If we do not agree, I think that we can merge this and work later on those issues

@aplanas, I do agree, though.

THS-on (approved):

PR #902, opened 2022-02-28
config: fix config search order

aplanas:
The configuration search order for RawConfigParser is from left (less
priority) to right (high priority).

This patch set the correct search order and adjust the test.

### DIFF
--- a/keylime/config.py
+++ b/keylime/config.py
@@ -7,7 +7,8 @@ import os.path
 import configparser
 from typing import Optional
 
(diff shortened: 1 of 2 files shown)


THS-on (approved):
Oh, I missed that. Do you have a link to the documentation, where the order is defined?

aplanas:
> Oh, I missed that.

Actually this is on me : (

> Do you have a link to the documentation, where the order is defined?

https://docs.python.org/3/library/configparser.html#configparser.ConfigParser.read

maugustosilva (approved):

Issue #603, opened 2021-03-29
keylime_verifier = etting state to FAILED


kgoldman:
Is your an issue a feature request? If so, please raise it as an [enhancement](https://github.com/keylime/enhancements)

# Environment
 * OS / version:  Fedora 32
 * Processor architecture: x86
 * TPM Manufacturer: na
 * Keylime version: today's master

# Description 

The readme says to start the verifier first, which is intuitive.  When I try that with today's master, I get the error in the log below.  

1. Is this a regression?  I think it used to work.
2. What is the work around?
3. It wuold be better, instead of just saying refused, to say what it was trying to connect to, what was unreachable, what 'first verified' means, etc.

## Expected behavior vs. actual behavior 

Expected verifier to start with no error.

## Steps to reproduce problem

1. Type keylime_verifier

# Relevant logs

021-03-29 15:54:45.949 - keylime.cloudverifier - INFO - Connection to 127.0.0.1 refused after 10/10 tries, trying again in 1.000000 seconds
2021-03-29 15:54:47.041 - keylime.cloudverifier - WARNING - Agent D432FBB3-D2F1-4A97-9EF7-75BD81C00000 was not reachable for quote in 10 tries, setting state to FAILED
2021-03-29 15:54:47.041 - keylime.cloudverifier - ERROR - Polling thread error: 'first_verified'
2021-03-29 15:54:47.042 - keylime.cloudverifier - ERROR - 'first_verified'
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/keylime/cloud_verifier_tornado.py", line 571, in process_agent
    if agent['first_verified']:  # only notify on previously good agents
KeyError: 'first_verified'

jronnblom:
I got the same problem but on 5.8.1. I think it occurs when I already have had a running system and then stop and start everything.

Then when keylime_verifier is started it reads the data from the db and puts in to agent[]. However since we cant talk to the agent (still down) then the first_verified key isn't in agent[].

I did change it to first check if the key is in the agent[]

if 'first_verified' in agent and agent['first_verified']:

and that seems to have fixed it. Then it sets the agent as failed. This also becomes a problem with the:

if agent['pending_event']

that there are a lot of. Or I just covered it up and created another problem :)

-J

lkatalin:
I am also hitting this issue.

PR #905, opened 2022-03-02
Add defaults for ephemeral keys for agent records

mpeters:
We already do this for agent entries coming from POST requests, but
not always when we restart and repopulate from the DB.

This should fix issue #603

### DIFF
--- a/keylime/cloud_verifier_tornado.py
+++ b/keylime/cloud_verifier_tornado.py
@@ -103,6 +103,11 @@ def _from_db_obj(agent_db_obj):
     agent_dict = {}
     for field in fields:
         agent_dict[field] = getattr(agent_db_obj, field, None)
+
+    # add default fields that are ephemeral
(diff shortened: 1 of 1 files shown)


keylime/cloud_verifier_tornado.py
```diff
@@ -103,6 +103,11 @@ def _from_db_obj(agent_db_obj):
     agent_dict = {}
     for field in fields:
         agent_dict[field] = getattr(agent_db_obj, field, None)
+
+    # add default fields that are ephemeral
+    for key in list(exclude_db.keys()):
```
aplanas, line 108: Is the `list()` required (I see that is also later in line 513)?

Otherwise LGTM.

aplanas (approved):

aplanas (approved):

aplanas (approved):
More clear with `items`, true.

PR #907, opened 2022-03-02
Allow minor version increases from Rust agent

### LINKED ISSUE #906: Allow API minor versions greater than latest major version
We made Zmq optional in favor of a REST API endpoint for revocations in the Rust agent, and updated our API version in accordance with [the enhancement](https://github.com/keylime/enhancements/blob/master/53_agent_local_revocation.md#notesconstraintscaveats-optional), from `2.0` to `2.1`. This broke our ability to communicate with the registrar (see [issue](https://github.com/keylime/rust-keylime/issues/340) on Rust Keylime repo).

Since the Rust agent may implement changes that require a minor API version bump sooner than the Python side does, it seems reasonable to expect that these API versions may not always be completely in sync, and to allow communication as long a the major versions do not differ. Although if others disagree then we could solve the problem a different way. This seems the most straightforward to me.

I expect the Rust agent should be able to upgrade a minor version to API `2.1` and still talk to the Python components that are at version `2.0`.

## Steps to reproduce problem
1. 

# Relevant logs
When I add a line to `test/test_api_version.py` to test for version `2.1`, it fails:

```
Failure in test.test_api_version.APIVersion_Test.test_is_supported_version
  File "/usr/lib64/python3.9/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/usr/lib64/python3.9/unittest/case.py", line 592, in run
    self._callTestMethod(testMethod)
  File "/usr/lib64/python3.9/unittest/case.py", line 550, in _callTestMethod
    method()
  File "/root/keylime/test/test_api_version.py", line 35, in test_is_supported_version
    self.assertTrue(api_version.is_supported_version("2.1"), "2.1 is a supported version")
  File "/usr/lib64/python3.9/unittest/case.py", line 688, in assertTrue
    raise self.failureException(msg)
AssertionError: False is not true : 2.1 is a supported version

```

lkatalin:
Happy to take suggestions here.

Fixes #906

### DIFF
--- a/keylime/api_version.py
+++ b/keylime/api_version.py
@@ -3,6 +3,7 @@ SPDX-License-Identifier: Apache-2.0
 Copyright 2021 Red Hat, Inc
 '''
 import re
+import math
 
 from packaging import version
 
@@ -15,6 +16,8 @@ LATEST_VERSIONS = {
     "1": "1.0",
     "2": "2.0"
 }
+
+MINIMUM_VERSION = "1.0"
 
 def current_version():
     return CURRENT_VERSION
@@ -34,7 +37,12 @@ def all_versions():
 
 def is_supported_version(v):
     v_obj = version.parse(str(v))
-    return v_obj.base_version in VERSIONS
+    next_major = math.floor(float(CURRENT_VERSION)) + 1
+    try:
+        float(str(v_obj))
+    except:
+        return False
+    return float(MINIMUM_VERSION) <= float(str(v_obj)) < next_major
 
 def normalize_version(v):
     v = str(v)
--- a/test/test_api_version.py
+++ b/test/test_api_version.py
@@ -32,6 +32,7 @@ class APIVersion_Test(unittest.TestCase):
 
     def test_is_supported_version(self):
         self.assertTrue(api_version.is_supported_version("2.0"), "2.0 is a supported version")
+        self.assertTrue(api_version.is_supported_version("2.1"), "2.1 is a supported version")
         self.assertTrue(api_version.is_supported_version("v2.0"), "v2.0 is a supported version")
         self.assertTrue(api_version.is_supported_version("1.0"), "1.0 is a supported version")
         self.assertTrue(api_version.is_supported_version("v1.0"), "v1.0 is a supported version")


mpeters:
We had a lot of discussions about this when defining the API versioning enhancement and we came to the conclusion that it's just not possible to have the verifier and registrar support future versions. The most common scenario was likely that agents would be older and live longer in production systems, and the server side components would have more freedom to upgrade (not embedded systems, smaller number of nodes to upgrade, etc) that the best way was to have the server-side components accept older API versions and not the other way around. From the enhancement doc https://github.com/keylime/enhancements/blob/master/45_api_versioning.md

> This proposal does not guarantee the ability for a newer client (using say v3.4) to communicate with an older server-side component (using say v3.3). Backwards compatible changes are within scope, not future-compatible ones. But it also doesn't explicitly disallow it. In some cases it might work out in practice for scenarios like the following: The only difference between v3.4 and v3.3 is the addition of a new optional field. The server side would just ignore this new unknown field. But any organization doing this would have no guarantees and would need to do their own extensive vetting.

So while I'm not opposed to allowing it to not explicitly happen (remove the forced error), it's something we have to acknowledge will have no support.

keylime/api_version.py
```diff
@@ -16,6 +17,8 @@
     "2": "2.0"
 }
 
+MINIMUM_VERSION = "1.0"
```
mpeters, line 20: Instead of hardcoding this, can we calculate it based on what is in the `VERSIONS` list?

keylime/api_version.py
```diff
@@ -34,7 +37,12 @@ def all_versions():
 
 def is_supported_version(v):
     v_obj = version.parse(str(v))
-    return v_obj.base_version in VERSIONS
```
mpeters, line 37: I would still prefer we explicitly check if it's in the `VERSIONS` list. If not, we could fall back to a check that it's only a little bit more than the `CURRENT_VERSION`, maybe just 1 or 2 minor versions different. And if that is the case, we issue a warning saying the agent's version is from the future and we can't guarantee compatibility, or something like that.

mpeters: Or maybe we should leave `is_supported_version()` alone and create a new method named something like `is_slightly_future_version()` or something similar. Than we can be very clear that a slight future version is not supported, but we'll allow the request to go through anyway.

test/test_api_version.py
```diff
@@ -32,6 +32,7 @@ def test_normalize_version(self):
 
     def test_is_supported_version(self):
         self.assertTrue(api_version.is_supported_version("2.0"), "2.0 is a supported version")
+        self.assertTrue(api_version.is_supported_version("2.1"), "2.1 is a supported version")
```
mpeters, line 35: Having a separate `is_slightly_future_version()` method would allow us to not formalize support for a version that doesn't exist yet. Future proofing APIs is hard :)

mpeters (changes requested):

lkatalin:
Interesting, thanks @mpeters for the review and discussion. I see now in the API [enhancement doc](https://github.com/keylime/enhancements/blob/master/45_api_versioning.md#notesconstraintscaveats) where it says newer agents are not guaranteed to be able to talk to older server-side. I'd like to make sure the Rust agent is guaranteed to work with server-side components, however that affects the versioning schema.

I'm not strongly opinionated about allowing a `2.1` agent to talk to a `2.0` verifier or registrar, just want to keep everything in sync, esp when the Rust agent incorporates an enhancement proposal first. Is there a more organized way to do this? Should the Rust agent just not have its own API versioning and only take cues from `keylime/keylime`'s API? Should one repo not upgrade until the other does?

mpeters:
> Should the Rust agent just not have its own API versioning and only take cues from `keylime/keylime`'s API? 

It shouldn't have it's own API version, but have the same API version as the current version of keylime if that makes sense. It will be chaos if they evolve independently.1

> Should one repo not upgrade until the other does?

Right now it's tricky since the rust agent is not the official agent. So it will naturally drift, but I would think it preferable that the drift happen then other way. Meaning a feature goes into the python registrar/verifier/agent first and then the rust agent either at the same time or slightly behind. In the future when the rust agent is the official agent, API changes will need to go into the registrar/verifier first and then the rust agent at the same time or slightly behind (commit wise). But releases will be synced so that the registrar/verifier/rust-agent are tagged at the same time with the same API version.

Does that help at all?

Issue #553, opened 2021-02-23
Perform checks for tpm_cert_store


lukehinds:
Currently if the tpm_cert_store is not present, we just fail to find the EK cert

```
2021-02-23 08:21:03.900 - keylime.tpm - ERROR - No Root CA matched EK Certificate
2021-02-23 08:21:03.901 - keylime.tenant - WARNING - Invalid EK certificate
```

We should do some sort of validation that the directory exists and if not warn the user. I expect a lot of users will miss that this needs copying into place first.

PR #912, opened 2022-03-08
tpm_main: Check if tpm_cert_store exists (#553)

ansasaki:
Check if tpm_cert_store exists when searching for EK certificates.

Fixes: #553


keylime/tpm/tpm_main.py
```diff
@@ -890,6 +890,8 @@ def verify_ek(self, ekcert):
         """
         # openssl x509 -inform der -in certificate.cer -out certificate.pem
         try:
+            tpm_ek_ca.check_tpm_cert_store()
+
```
mpeters, line 894: It seems that other than this new call `check_tpm_cert_store()` is unused before this. Strange...

But `check_tpm_cert_store()` calls `sys.exit()` when it fails. Are we sure that's what we want to do? Would it be better to throw an exception?

ansasaki: Yes, it would be better to throw an exception. I'll make the changes.

I was surprised too when I found that check implemented but not called. I thought it was dropped at some point by accident, but it seems it was never used.

mpeters (approved):

Issue #535, opened 2021-02-03
Untrusted data in format string


MikeSpreitzer:
Is your an issue a feature request? If so, please raise it as an [enhancement](https://github.com/keylime/enhancements)

# Environment
 * OS / version: all
 * Processor architecture: all
 * TPM Manufacturer: all
 * Keylime version: current `master` branch (commit d987c71dd878492ac2b59eca781009fe270d219a), and earlier

# Description 
The first argument of `logger.error` &etc is a format string, not the formatted message; the formatting arguments follow.  But many of our logging statements do the formatting before calling the log method, supplying the formatted message as the format string.  This is dangerous in general, and particularly when the format string contains untrusted data (e.g., from the keylime agent).  For an example, see https://github.com/keylime/keylime/blob/d987c71dd878492ac2b59eca781009fe270d219a/keylime/tpm/tpm_abstract.py#L253

## Expected behavior vs. actual behavior 
That example currently reads like this:
```
                logger.error("Invalid %sPCR in quote: %s" % (("", "v")[virtual], pcrs))
```
but to keep data out of the format string we would write
```
                logger.error("Invalid %sPCR in quote: %s", ("", "v")[virtual], pcrs)
```

## Steps to reproduce problem
1. read the code

# Relevant logs

Attach any relevant log files that can help to debug your issue.

lukehinds:
At some point we had a plan to move over entirely to fstrings and most new patches to do this.

I think the following would be good.
1. We move over entirely to [f-strings ](https://security.stackexchange.com/questions/238338/are-there-any-security-concerns-to-using-python-f-strings-with-user-input), if you agree this is suitable @MikeSpreitzer 
2. We plug something into CI check the old string format is not used in any new PRs (and if it is, we fail the CI run)

How do you view that @MikeSpreitzer et al?

MikeSpreitzer:
I am fond of the f-strings, but to pass one where an old-style format string is expected would be a mistake.  Can you expand a little on exactly what you are suggesting?

MikeSpreitzer:
See Log4j

PR #914, opened 2022-03-08
Use lazy string formatting when logging (#535)

ansasaki:
Avoid using formatted message as the format string for logging and use
lazy string formatting instead.  This includes avoiding using f-strings
for logging and prefer the recommended %-string formatting.

The changes include enabling the logging check in pylint to detect
non-recommended strings formatting usage.

Fixes: #535

### DIFF
--- a/keylime/ca_impl_cfssl.py
+++ b/keylime/ca_impl_cfssl.py
@@ -45,8 +45,8 @@ def post_cfssl(params, data):
                 logger.error(
                     "Quiting after max number of retries to connect to cfssl server")
                 raise e
-            logger.info(
-                "Connection to cfssl refused %d/%d times, trying again in %f seconds..." % (numtries, maxr, retry))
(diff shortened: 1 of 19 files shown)


keylime/ca_impl_cfssl.py
```diff
@@ -45,8 +45,8 @@ def post_cfssl(params, data):
                 logger.error(
                     "Quiting after max number of retries to connect to cfssl server")
                 raise e
-            logger.info(
-                "Connection to cfssl refused %d/%d times, trying again in %f seconds..." % (numtries, maxr, retry))
+            logger.info("Connection to cfssl refused %d/%d times, trying again in %f "
+                        "seconds...", numtries, maxr, retry)
```
mpeters, line 49: What is the reason for breaking the log message into multiple strings?

ansasaki: Sorry about that, the reason is to not have too long lines. I will fix these.

PR #920, opened 2022-03-14
Registrar Client and Tenant validate CA certificates

### LINKED ISSUE #681: TLS connections (e.g. tenant->verifier) are established without server certificate validation

# Environment
 * OS / version: 
 * Processor architecture: 
 * TPM Manufacturer: 
 * Keylime version: [d632e2a](https://github.com/keylime/keylime/commit/d632e2a3b510167ba997197f5232bc4456aa0394)

# Description 
TLS connections (e.g. tenant->verifier) are established without server certificate validation. This can lead to MiTM

The expected behaviors is that server certificates are validated properly to mitigate MiTM attacks. The actual behavior skips validating server certificates. See [tenant.py#L557](https://github.com/keylime/keylime/blob/master/keylime/tenant.py#L557)

## Steps to reproduce problem
1. See [tenant.py#L557](https://github.com/keylime/keylime/blob/master/keylime/tenant.py#L557), [tenant.py#L644](https://github.com/keylime/keylime/blob/master/keylime/tenant.py#L644),..

# Relevant logs

THS-on:
This removes all `verify=False` from requests calls outside of the tests and webapp.

Fixes #681

### DIFF
--- a/keylime/registrar_client.py
+++ b/keylime/registrar_client.py
@@ -16,6 +16,7 @@ from keylime import api_version as keylime_api_version
 
 logger = keylime_logging.init_logging('registrar_client')
 tls_cert_info = ()
+ca_cert = False
 tls_enabled = False
 api_version = keylime_api_version.current_version()
 
@@ -23,6 +24,7 @@ api_version = keylime_api_version.current_version()
 def init_client_tls(section):
     global tls_cert_info
     global tls_enabled
+    global ca_cert
 
     # make this reentrant
     if tls_cert_info:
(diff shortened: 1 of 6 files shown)


keylime/registrar_client.py
```diff
@@ -16,13 +16,15 @@
...
 tls_cert_info = ()
+ca_cert = False
 tls_enabled = False
 api_version = keylime_api_version.current_version()
 
 
 def init_client_tls(section):
     global tls_cert_info
     global tls_enabled
+    global ca_cert
```
aplanas, line 27: Not to be resolved in this PR, but we are starting to have a lots of global state via global vars. This will be a pain later, and we should create some issue to try to remove them.

keylime/registrar_client.py
```diff
@@ -75,8 +83,8 @@ def getData(registrar_ip, registrar_port, agent_id):
 
     response = None
     try:
-        client = RequestsClient(f'{registrar_ip}:{registrar_port}', tls_enabled)
-        response = client.get(f'/v{api_version}/agents/{agent_id}', cert=tls_cert_info, verify=False)
+        client = RequestsClient(f'{registrar_ip}:{registrar_port}', tls_enabled, ignore_hostname=True)
```
aplanas, line 86: I think that now most of the `RequestsClient` calls makes `ignore_hostname` to `True`, except for tenant_webapp.

Do you think that can be better to remove it as a default parameter, and make it explicit?

aplanas (approved):
Not very knowledge about some of the bits changed here, but code wise LGTM

keylime/registrar_client.py
THS-on: I agree that we should start to remove global state where not strictly necessary.

keylime/registrar_client.py
THS-on: We could, but I like that you have to explicitly write that hostnames are ignored and are otherwise more secure by default.

Disabling the hostname validation in TLS is a trade off with we currently deploy certificates and security. I think that in the future we can remove most of the `ignore_hostname=True` if certificates contain the correct SAN attributes.

aplanas: I see, thanks!

maugustosilva (approved):

maugustosilva:
LGTM

PR #921, opened 2022-03-15
services: make the services run as keylime user instead of root

THS-on:
The services for the registrar and verifier are changed to run as keylime
user. The agent starts as root but the installer.sh changes the "run_as"
option to drop privileges to keylime:tss.

The installer.sh is changed to create the keylime user, create the correct
directories, change the permissions to the keylime user and to also install
the secure mount point.

I have also removed the `agent_non_root_condition.sh` script because it is no longer necessary.


mpeters (approved):

THS-on:
@aplanas @ueno I think the Fedora and SUSE packaging are using the systemd services shipped by Keylime. 
So you will have to make sure that the services still work, by creating the necessary folders `/var/{lib,log,run}/keylime` with the correct permissions and the `keylime` user in the packaging.

Adding a patch for dropping privileges in the agent to `keylime:tss` by default might also be sensible for packaging.

services/keylime_registrar.service
```diff
@@ -3,7 +3,12 @@ Description=The Keylime registrar service
 After=network.target
 
 [Service]
+Group=keylime
```
aplanas, line 6: tss?

aplanas: Also, is not the idea to be root an later downgrade?

services/keylime_registrar.service.template
```diff
@@ -1,9 +1,16 @@
 [Unit]
 Description=The Keylime registrar service
-After=network.target
+After=network.target keylime_verifier.service
+StartLimitInterval=10s
+StartLimitBurst=5
 
 [Service]
+Group=keylime
```
aplanas, line 8: (same)

aplanas:
I am wondering, is there any access to the TPM in the verifier / registrar? IIRC I saw errors that indicated that is doing that.

services/keylime_registrar.service
THS-on: The registrar should not need TPM access.

services/keylime_registrar.service.template
THS-on: same as above

THS-on:
> I am wondering, is there any access to the TPM in the verifier / registrar? IIRC I saw errors that indicated that is doing that.

No they only need access to tpm-tools. If not this is a bug.

services/keylime_registrar.service
aplanas: but is tss the group, or keylime?

THS-on: keylime is the group of the user. tss of the TPM ownership.

mpeters: Right, so the agent needs to be `keylime:tss` but the verifier/registrar needs to be `keylime:keylime`

THS-on: @mpeters yeah, the agent starts as `root` and drops to `keylime:tss` and the verfier/registrar can just run as `keylime:keylime`.

PR #922, opened 2022-03-16
State in --help that SHA-256 is used for --allowlist-checksum

kkaarreell:
SHA-2 is a set of cryptographic hash functions but specifically
SHA-256 is used for --allowlist-checksum.

### DIFF
--- a/keylime/tenant.py
+++ b/keylime/tenant.py
@@ -1271,7 +1271,7 @@ def main(argv=sys.argv):
     parser.add_argument('--mb_refstate', action='store', dest='mb_refstate',
                         default=None, help="Specify the location of a measure boot reference state (intended state)")
     parser.add_argument('--allowlist-checksum', action='store', dest='allowlist_checksum',
-                        default=None, help="Specify the SHA2 checksum of an allowlist")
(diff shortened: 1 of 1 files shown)


THS-on (approved):

THS-on:
@kkaarreell the centos stream 9 test fails because there is a GPG key mismatch for the beakerlib repo.

kkaarreell:
> @kkaarreell the centos stream 9 test fails because there is a GPG key mismatch for the beakerlib repo.

I think this failure is related to a current effort to disallow SHA1 signatures for RPMs. We can probably expect CentOS Stream 9 tests failing for several days or weeks. Should we disable it temporarily?

THS-on:
> Should we disable it temporarily?

As long as it is just a few days I wouldn't bother. If we can expect it to be broken for a month or two then I would disable it.

kkaarreell:
https://pagure.io/copr/copr/issue/2106 suggests a fix could be deployed next Monday. I would expect at least one more week of failures as copr repos would have to be regenerated.

THS-on:
Ok, then we will just wait.

mpeters (approved):

PR #926, opened 2022-03-16
Restores the option to use keylime agents without mTLS

maugustosilva:
This is a minimalistic set of code changes to allow agents to be
accessed without any TLS. This gives Keylime operators the power to
decide, based on the internal characteristics of their security policies
and processes, if agent identification via TLS certificates is
necessary.

Following the same model used for other features, such as
`revocation_notifier` and `revocation_notifier_webhook`, a couple of new
boolean configuration options, `mtls_cert_enabled` (agent-specific) and
`agent_mtls_cert_enabled` (verifer and tenant) are introduced. While the
use of mTLS for communication with agents is kept as the default
behavior, these three values can be set to "False" to fully disable the
communication with agents via TLS.

### DIFF
--- a/keylime.conf
+++ b/keylime.conf
@@ -40,6 +40,11 @@ registrar_port = 8890
 # If rsa_keyname is not a absolute path it is relative to /var/lib/keylime/secure.
 rsa_keyname = tci_rsa_key
 
+# Enable mTLS communication between agent, verifier and tenant.
+# Details on why setting it to "False" is generally considered insecure can be found
+# on https://github.com/keylime/keylime/security/advisories/GHSA-2m39-75g9-ff5r
+mtls_cert_enabled = True
+
 # The name of the X509 certificate that is used for mTLS. Uses the rsa_keyname as a key.
 # If mtls_cert is not a absolute path it is relative to /var/lib/keylime/secure.
 # This certificate must be self signed.
@@ -109,6 +114,13 @@ revocation_actions=
 # cloud-init lite =)  Keylime will run it with a /bin/sh environment and
 # with a working directory of /var/lib/keylime/secure/unzipped.
 payload_script=autorun.sh
+
+# In case mTLS for the agent is disabled and the use of payloads is still
+# required, this option has to be set to "True" in order to allow the agent 
+# to start. Details on why this configuration (mTLS disabled and payload enabled)
+# is generally considered insecure can be found on 
+# https://github.com/keylime/keylime/security/advisories/GHSA-2m39-75g9-ff5r
+enable_insecure_payload = False
 
 # Jason @henn made be do it! He wanted a way for Keylime to measure the
 # delivered payload into a pcr of choice.
@@ -239,6 +251,9 @@ registrar_private_key = default
 registrar_private_key_pw = default
 
 # mTLS configuration for connecting to the agent.
+# Details on why setting it to "False" is generally considered insecure can be found
+# on https://github.com/keylime/keylime/security/advisories/GHSA-2m39-75g9-ff5r
+agent_mtls_cert_enabled = True
 # Set 'agent_mtls_cert' to 'CV' for using the CV CA for the connections.
 agent_mtls_cert = CV
 agent_mtls_private_key =
@@ -375,8 +390,10 @@ ca_cert = default
 my_cert = default
 private_key = default
 
-
 # mTLS configuration for connecting to the agent.
+# Details on why setting it to "False" is generally considered insecure can be found
+# on https://github.com/keylime/keylime/security/advisories/GHSA-2m39-75g9-ff5r
+agent_mtls_cert_enabled = True
 # Set 'agent_mtls_cert' to 'CV' for using the CV CA for the connections.
 agent_mtls_cert = CV
 agent_mtls_private_key =
--- a/keylime/cloud_verifier_tornado.py
+++ b/keylime/cloud_verifier_tornado.py
@@ -514,12 +514,11 @@ class AgentsHandler(BaseHandler):
                             agent_data[key] = val
 
                         # Prepare SSLContext for mTLS connections
-                        # TODO: drop special handling after initial upgrade
+                        agent_mtls_cert_enabled = config.getboolean('cloud_verifier', 'agent_mtls_cert_enabled', fallback=False)
                         mtls_cert = registrar_data.get('mtls_cert', None)
                         agent_data['ssl_context'] = None
-                        if mtls_cert:
-                            agent_data['ssl_context'] = web_util.generate_agent_mtls_context(mtls_cert,
-                                                                                             self.mtls_options)
+                        if agent_mtls_cert_enabled and mtls_cert:
+                            agent_data['ssl_context'] = web_util.generate_agent_mtls_context(mtls_cert, self.mtls_options)
 
                         if agent_data['ssl_context'] is None:
                             logger.warning('Connecting to agent without mTLS: %s', agent_id)
--- a/keylime/keylime_agent.py
+++ b/keylime/keylime_agent.py
@@ -297,10 +297,15 @@ class Handler(BaseHTTPRequestHandler):
         dec_payload = None
         enc_payload = None
         if self.server.payload is not None:
-            dec_payload = crypto.decrypt(
-                self.server.payload, bytes(self.server.K))
+            if (not self.server.mtls_cert_enabled and
+                not config.getboolean('cloud_agent', 'enable_insecure_payload', fallback=False)):
+                logger.warning('agent mTLS is disabled, and unless "enable_insecure_payload" is set to "True", payloads cannot be deployed')
+                enc_payload = None
+            else:
+                dec_payload = crypto.decrypt(
+                    self.server.payload, bytes(self.server.K))
+                enc_payload = self.server.payload
 
-            enc_payload = self.server.payload
         elif os.path.exists(enc_path):
             # if no payload provided, try to decrypt one from a previous run stored in encrypted_payload
             with open(enc_path, 'rb') as f:
@@ -386,6 +391,7 @@ class CloudAgentHTTPServer(ThreadingMixIn, HTTPServer):
     rsapublickey_exportable = None
     mtls_cert_path = None
     rsakey_path = None
+    mtls_cert_enabled = False
     mtls_cert = None
     done = threading.Event()
     auth_tag = None
@@ -405,9 +411,6 @@ class CloudAgentHTTPServer(ThreadingMixIn, HTTPServer):
         keyname = config.get('cloud_agent', 'rsa_keyname')
         if not os.path.isabs(keyname):
             keyname = os.path.join(secdir, keyname)
-        certname = config.get('cloud_agent', 'mtls_cert')
-        if not os.path.isabs(certname):
-            certname = os.path.join(secdir, certname)
 
         # read or generate the key depending on configuration
         if os.path.isfile(keyname):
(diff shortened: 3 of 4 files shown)


keylime/keylime_agent.py
```diff
@@ -426,23 +424,34 @@ def __init__(self, server_address, RequestHandlerClass, agent_uuid, contact_ip,
...
-            with open(certname, "wb") as f:
-                # By default generate a TLS certificate valid for 5 years
-                valid_util = datetime.datetime.utcnow() + datetime.timedelta(days=(360 * 5))
-                mtls_cert = crypto.generate_selfsigned_cert(agent_uuid, rsa_key, valid_util, agent_ips)
-                f.write(mtls_cert.public_bytes(serialization.Encoding.PEM))
-
-        self.mtls_cert_path = certname
-        self.mtls_cert = mtls_cert
+        self.mtls_cert_enabled = config.getboolean('cloud_agent', 'mtls_cert_enabled', fallback=False)
+        if self.mtls_cert_enabled :
```
THS-on, line 428: It is not common to have a space after `:` in Python can you fix that in your PR?
```suggestion
        if self.mtls_cert_enabled:
```

THS-on (changes requested):
If the mtls_cert_enabled setting is True the entire payload mechanism should be disabled!

mpeters:
> If the mtls_cert_enabled setting is True the entire payload mechanism should be disabled!

Should this be a BIG warning in the log? Or should we actually refuse to start the agent if mtls is disabled and the payload is enabled?

THS-on:
> Should this be a BIG warning in the log? Or should we actually refuse to start the agent if mtls is disabled and the payload is enabled?

We should fore disable the payload mechanism. If you disable mTLS you cannot have the payload mechanism and we **should not** provide a way to override that. There should no way to run into https://github.com/keylime/keylime/security/advisories/GHSA-2m39-75g9-ff5r again.

mpeters:
> We should force disable the payload mechanism. If you disable mTLS you cannot have the payload mechanism and we **should not** provide a way to override that. There should no way to run into [GHSA-2m39-75g9-ff5r](https://github.com/keylime/keylime/security/advisories/GHSA-2m39-75g9-ff5r) again.

I'm not sure I agree. We need to have correct defaults and make it really obvious when someone is doing something insecure, but that doesn't mean users can't deliberately force insecurity. There are a lot of valid reasons to sometimes do insecure things.

keylime/tenant.py
```diff
@@ -302,8 +307,18 @@ def init_add(self, args):
...
                 # Try to connect to the agent to get supported version
-                with RequestsClient(f"{self.agent_ip}:{self.agent_port}", tls_enabled=True, cert=self.agent_cert,
-                                    ignore_hostname=True, verify_custom=self.registrar_data['mtls_cert']) as get_version:
+                if self.registrar_data['mtls_cert']:
+                    self.verify_custom = self.registrar_data['mtls_cert']
+                    if self.registrar_data['mtls_cert'] == "disabled" :
+                        self.tls_agent_enabled = False
+                        self.verify_custom = False
+                        logger.warning(
+                            "Warning: mTLS for agents is disabled: the identity of each node will be based on the properties of the TPM only. "
```
mpeters, line 316: 2 comments on this: can we make the "Warning" part strong by using all caps? And could we have separate error messages depending on whether `payload_script` is set or not?

mpeters (changes requested):

maugustosilva:
There are several tunables in keylime which allows an user to tradeoff security for either ease of deployment, convenience of development or (even) scalability, such as `enable_tls` and `require_ek_cert`. This current PR simply follows the same well-established philosophy, giving an operator the ability to decide by himself, based on its own internal security and compliance policies, how to use the framework. 

The **defaults are as secure as it can be made within the state-of-the-art** and the deviation from this scenario is properly communicated to the user via strong worded warnings.

The fact the agent authentication is not done within Keylime via mTLS does not mean that no form of authentication is carried out - via policies and procedures - outside of Keylime. Again, it is up to the user to decide what is his acceptable level of risk, based, among other factors, on the level of access control that one has over his infrastructure.

Evidently, if someone wants to restrict the use of Keylime, forcing the disabling of the payload in the absence of mTLS, they can patch Keylime on their environment to do so.

THS-on:
`enable_tls` is actually enforced by the code for registrar connections:
https://github.com/keylime/keylime/blob/99f444dc46861e9dc9eadaadd5267f5e14164f3c/keylime/registrar_client.py#L80-L82

Disabling `require_ek_cert` does not expose and contentious attack surface as disabling mTLS does.

> The fact the agent authentication is not done within Keylime via mTLS does not mean that no form of authentication is carried out - via policies and procedures - outside of Keylime.

Can you describe a way how this is done as a reference to point people to?

> Evidently, if someone wants to restrict the use of Keylime, forcing the disabling of the payload in the absence of mTLS, they 
can patch Keylime on their environment to do so.

Yes but we should be somewhat secure by default even if you disable mTLS. As a compromise I would disable payloads if `mtls_cert_enabled` is true and provide an option `enable_insecure_payloads` to explicitly opt-in for enabling payloads if mTLS is disabled.

keylime/keylime_agent.py
```diff
@@ -426,23 +424,34 @@ def __init__(self, server_address, RequestHandlerClass, agent_uuid, contact_ip,
...
+                    agent_ips.append(contact_ip)
+                with open(certname, "wb") as f:
+                    # By default generate a TLS certificate valid for 5 years
+                    valid_util = datetime.datetime.utcnow() + datetime.timedelta(days=(360 * 5))
+                    mtls_cert = crypto.generate_selfsigned_cert(agent_uuid, rsa_key, valid_util, agent_ips)
+                    f.write(mtls_cert.public_bytes(serialization.Encoding.PEM))
+
+            self.mtls_cert_path = certname
+            self.mtls_cert = mtls_cert
+        else :
```
THS-on, line 451:
```suggestion
        else:
```

keylime/keylime_agent.py
```diff
@@ -745,12 +754,16 @@ def main():
         keylime_ca = os.path.join(config.WORK_DIR, 'cv_ca', 'cacert.crt')
 
     server = CloudAgentHTTPServer(serveraddr, Handler, agent_uuid, contact_ip, ima_log_file, tpm_log_file_data)
-    context = web_util.generate_mtls_context(server.mtls_cert_path, server.rsakey_path, keylime_ca, logger=logger)
-    server.socket = context.wrap_socket(server.socket, server_side=True)
+    if server.mtls_cert_enabled :
```
THS-on, line 757:
```suggestion
    if server.mtls_cert_enabled:
```

keylime/keylime_agent.py
```diff
@@ -745,12 +754,16 @@ def main():
...
+        context = web_util.generate_mtls_context(server.mtls_cert_path, server.rsakey_path, keylime_ca, logger=logger)
+        server.socket = context.wrap_socket(server.socket, server_side=True)
     serverthread = threading.Thread(target=server.serve_forever, daemon=True)
 
     # register it and get back a blob
-    mtls_cert = server.mtls_cert.public_bytes(serialization.Encoding.PEM)
+    mtls_cert = "disabled"
+    if server.mtls_cert :
```
THS-on, line 764:
```suggestion
    if server.mtls_cert:
```

keylime/tenant.py
```diff
@@ -153,13 +155,16 @@ def get_tls_context(self):
...
+        agent_mtls_context = (None, None)
         # Check for user defined CA to connect to agent
-        agent_mtls_cert = config.get("cloud_verifier", "agent_mtls_cert", fallback=None)
-        agent_mtls_private_key = config.get("cloud_verifier", "agent_mtls_private_key", fallback=None)
+        agent_mtls_cert_enabled = config.getboolean('tenant', 'agent_mtls_cert_enabled', fallback=False)
 
-        agent_mtls_context = tls_context
-        if agent_mtls_cert != "CV":
-            agent_mtls_context = (agent_mtls_cert, agent_mtls_private_key)
+        if agent_mtls_cert_enabled :
```
THS-on, line 162:
```suggestion
        if agent_mtls_cert_enabled:
```

keylime/tenant.py
```diff
@@ -302,8 +307,18 @@ def init_add(self, args):
                 self.supported_version = "1.0"
             else:
                 # Try to connect to the agent to get supported version
-                with RequestsClient(f"{self.agent_ip}:{self.agent_port}", tls_enabled=True, cert=self.agent_cert,
-                                    ignore_hostname=True, verify_custom=self.registrar_data['mtls_cert']) as get_version:
+                if self.registrar_data['mtls_cert']:
+                    self.verify_custom = self.registrar_data['mtls_cert']
+                    if self.registrar_data['mtls_cert'] == "disabled" :
```
THS-on, line 312:
```suggestion
                    if self.registrar_data['mtls_cert'] == "disabled":
```

THS-on (changes requested):

maugustosilva:
Your proposal for a new `enable_insecure_payloads` (**off by default**) is quite acceptable. I shall implement it promptly, on this very same PR.

keylime.conf
```diff
@@ -110,6 +113,11 @@ revocation_actions=
 # with a working directory of /var/lib/keylime/secure/unzipped.
 payload_script=autorun.sh
 
+# In case mTLS for the agent is disabled and the use of payloads is still
+# required, this option has to be set to "True" in order to allow the agent 
+# to start.
+enable_insecure_payload = False
```
THS-on, line 119: Add a reference to https://github.com/keylime/keylime/security/advisories/GHSA-2m39-75g9-ff5r and explain why this is generally a bad idea to do.

keylime.conf
```diff
@@ -40,6 +40,9 @@ registrar_port = 8890
 # If rsa_keyname is not a absolute path it is relative to /var/lib/keylime/secure.
 rsa_keyname = tci_rsa_key
 
+# Enable mTLS communication between agent, verifier and tenant.
+mtls_cert_enabled = True
```
THS-on, line 44: Add a reference why disabling this is generally a bad idea (https://github.com/keylime/keylime/security/advisories/GHSA-2m39-75g9-ff5r).

keylime/keylime_agent.py
```diff
@@ -745,12 +754,21 @@ def main():
         keylime_ca = os.path.join(config.WORK_DIR, 'cv_ca', 'cacert.crt')
 
     server = CloudAgentHTTPServer(serveraddr, Handler, agent_uuid, contact_ip, ima_log_file, tpm_log_file_data)
-    context = web_util.generate_mtls_context(server.mtls_cert_path, server.rsakey_path, keylime_ca, logger=logger)
-    server.socket = context.wrap_socket(server.socket, server_side=True)
+    if server.mtls_cert_enabled:
+        context = web_util.generate_mtls_context(server.mtls_cert_path, server.rsakey_path, keylime_ca, logger=logger)
+        server.socket = context.wrap_socket(server.socket, server_side=True)
+    else:
+        if not config.getboolean('cloud_agent', 'enable_insecure_payload', fallback=False):
```
THS-on, line 761: Instead of not starting, would just not accept payloads by immediately returning instead of handling U/V key and payload POST request.

Somewhere here:
https://github.com/keylime/keylime/blob/99f444dc46861e9dc9eadaadd5267f5e14164f3c/keylime/keylime_agent.py#L210-L230


Adding a warning during startup is still a good idea.

THS-on (changes requested):

keylime.conf
maugustosilva: done

keylime.conf
maugustosilva: done

keylime/keylime_agent.py
maugustosilva: I am perfectly ok with **not** starting, unless the user demonstrates he knows what he is doing by setting the new parameter explicitly. I firmly believe this to be better tradeoff between our two "competing philosophies" on the matter.

THS-on: > I firmly believe this to be better tradeoff between our two "competing philosophies" on the matter.

You are right.  We still need to not accept the payload in the POST request because otherwise an attacker could still deploy a file on the agent (just not executing it).

mpeters: I agree. It'd rather complain early about incompatible options than deny something at runtime which would be harder for the confused operator to understand.

keylime.conf
mpeters: The wording is a bit confusing. It makes it seem like setting `mtls_cert_enabled` to `True` is considered insecure. Maybe it should say something like "Details on why setting this to False is considered insecure"... or similar.

mpeters (changes requested):

THS-on (commented):
After some small changes the PR is I think ready.
It is clear what happens if some disables mTLS and disabling it does not fully compromise the agent (if no other authentication outside of Keylime is used).

keylime/cloud_verifier_tornado.py
```diff
@@ -514,12 +514,11 @@ def post(self):
...
 
                         # Prepare SSLContext for mTLS connections
-                        # TODO: drop special handling after initial upgrade
+                        agent_mtls_cert_enabled = config.getboolean('cloud_verifier', 'agent_mtls_cert_enabled', fallback=False)
                         mtls_cert = registrar_data.get('mtls_cert', None)
                         agent_data['ssl_context'] = None
-                        if mtls_cert:
-                            agent_data['ssl_context'] = web_util.generate_agent_mtls_context(mtls_cert,
-                                                                                             self.mtls_options)
+                        if agent_mtls_cert_enabled and mtls_cert :
```
THS-on, line 520:
```suggestion
                        if agent_mtls_cert_enabled and mtls_cert:
```

keylime/tenant.py
```diff
@@ -302,8 +307,18 @@ def init_add(self, args):
                 self.supported_version = "1.0"
             else:
                 # Try to connect to the agent to get supported version
-                with RequestsClient(f"{self.agent_ip}:{self.agent_port}", tls_enabled=True, cert=self.agent_cert,
-                                    ignore_hostname=True, verify_custom=self.registrar_data['mtls_cert']) as get_version:
+                if self.registrar_data['mtls_cert'] == "disabled" :
```
THS-on, line 310:
```suggestion
                if self.registrar_data['mtls_cert'] == "disabled":
```

THS-on (approved):
Besides a few style changes this LGTM.

@maugustosilva can you squash the commits into one before this gets merged?

mpeters (approved):

Issue #929, opened 2022-03-21
Code quality improvements 


THS-on:
This issue tracks the different things that can be done to improve the code quality. If you want to fix some one of the issues please comment on this issue, so I can update it.

### pylint
We currently have a large list of warnings that we ignore. Those have to be reviewed and removed where possible.

**Current warnings:**
- [ ] W1509
- [ ] C0103
- [ ] C0115
- [ ] C0116
- [ ] C0201
- [ ] C0209
- [ ] C0301
- [ ] C0302
- [ ] C0111
- [ ] W0102
- [ ] W0511
- [ ] W0602
- [ ] W0603
- [ ] W0703
- [ ] W1201
- [ ] W1203
- [ ] E1120
- [ ] R0801
- [ ] R0902
- [ ] R0903
- [ ] R0904
- [ ] R0912
- [ ] R0913
- [ ] R0914
- [ ] R0915
- [ ] R0201
- [ ] R0911
- [ ] R1729
- [ ] R1732

### MyPy
We currently not enforce types. This makes it harder to reason about some functions and allows for subtile errors.
A strict mypy configuration currently returns about 2000 warnings that can probably removed by going through Keylime and adding type annotations. 

Configuration used:
```
[mypy]
plugins = sqlmypy
strict = True
follow_imports = silent
ignore_missing_imports = True
```

### LGTM, semgrep and Bandit
There are multiple tools for analyzing the AST of the Python code and finding common security and code quality issues.
We should at least integrate one of them into our CI. 

Fuzzing methods that take user controlled input with something like https://github.com/google/atheris can also be useful (mentioned by @kkaarreell).

### Converting dicts to classes
For the agent event loop we use a dict to store all the necessary state of an agent. This is not ideal because it is hard to reason about what is actually in this dict. We should convert this dict to a class that holds all the required state. For this class we can then also provide conversion functions from and to the ORM class.

### Removal of `config.REQUIRE_ROOT`
With #900 merged this flag should be fully removed.

### Removal of vTPM code
Originally Keylime supported vTPMs a feature from Xen. This is no longer the case and the deep quote feature is not implemented in swtpm. Most of the code is already removed but there is still some broken skeleton code in there which should be removed.

I think we can move the issue on how to trust the underlying hypervisor out of Keylime and into the platform that manages the hypervisor (e.g. OpenStack).

### Removal of STUB TPM logic
[...]

PR #930, opened 2022-03-21
Add type hints and initial Mypy configuration

THS-on:
This PR adds full type hints to:
- `ima_ast.py`
- `failure.py`
- `algorithms.py`
- `keylime_loggin.py`
- `json.py`

I also added a Mypy configuration with `strict` set to true which should enable most of the Mypy checks.

Once more parts of Keylime have type hints I will add this to the tox configuration.

Part of #929

### DIFF
--- a/keylime/failure.py
+++ b/keylime/failure.py
@@ -7,15 +7,14 @@ Tagging of failure events that might cause revocation in Keylime.
 import ast
 import enum
 import functools
-import json
 import re
-from typing import List, Optional, Tuple, Callable, Union
+from typing import List, Optional, Tuple, Callable, Union, Dict, Any
 
 from keylime import config
+from keylime import json
 from keylime import keylime_logging
 
 logger = keylime_logging.init_logging("failure")
-
 
 @functools.total_ordering
 class SeverityLabel:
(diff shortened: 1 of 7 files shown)


keylime/failure.py
```diff
@@ -92,18 +95,18 @@ class Failure:
     events: List[Event]
     recoverable: bool
     highest_severity: Optional[SeverityLabel]
-    _component: Optional[Component]
+    _component: Component
     _sub_components: Optional[List[str]]
 
-    def __init__(self, component, sub_components=None):
+    def __init__(self, component: Component, sub_components: Optional[List[str]] = None):
```
mpeters, line 101: Just a question on style here: Does the type hints syntax prefer whitespace around the equals sign?

THS-on: I think so: https://peps.python.org/pep-3107/#parameters.

mpeters:
Overall I like this change as it should prevent some accidental bugs in the future. Being unfamiliar with Mypy, are there any performance considerations?

THS-on:
>  Being unfamiliar with Mypy, are there any performance considerations?

Type hints are ignored at runtime, so there should be no performance impact other than the hints are getting parsed and then ignored.

maugustosilva (approved):

THS-on:
@kkaarreell why did this e2e CI pass?

kkaarreell:
There is some issue with tmt plan on c9s and c8s where prepare steps are replaced, not added. As a result c8s and c9s are run against upstream master. I have just found out few minutes back.

PR #938, opened 2022-03-23
pylint: Fix issues related to R0201 

stefanberger:
This PR fixes the following pylint issues:
- R0201 no-self-use


mpeters (approved):

stefanberger:
@mpeters Thanks but let's run the tests first. Why are they not running?

THS-on:
@stefanberger C0115 & C0116 are not really solved just not enabled in pylint. It seems like that another option is already disabling them. Manually enabling them returns a lot of warnings.

PR #939, opened 2022-03-23
pylint: Fix issues related to W0102 dangerous-default-value

stefanberger:
When main() functions have the parameters argv=sys.argv then simply
add '#pylint: disable=dangerous-default-value' on the line to suppress
the error. The rationale is that main() will only be called once and
therefore the argv should only ever equal sys.argv.


THS-on (approved):
LGTM.

maugustosilva (approved):

PR #940, opened 2022-03-24
general: drop support for API version 1.0

THS-on:
API version 1.0 is considered insecure and should therefore no longer be used.
We included support for it in 6.3.0 to support updating the server side first,
but future versions will no longer include support for 1.0.

Agent mTLS connections can still be manually disabled, but this is highly
discouraged for normal installations. More details:
https://github.com/keylime/keylime/security/advisories/GHSA-2m39-75g9-ff5r

### DIFF
--- a/keylime/cloud_verifier_tornado.py
+++ b/keylime/cloud_verifier_tornado.py
@@ -471,8 +471,13 @@ class AgentsHandler(BaseHandler):
                     agent_data['mtls_cert'] = registrar_data.get('mtls_cert', None)
                     agent_data['ak_tpm'] = registrar_data['aik_tpm']
 
-                    # TODO: Always error for v1.0 version after initial upgrade
-                    if registrar_data.get('mtls_cert', None) is None and agent_data['supported_version'] != "1.0":
+                    if not keylime_api_version.is_supported_version(agent_data['supported_version']):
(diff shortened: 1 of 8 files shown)


keylime/cloud_verifier_tornado.py
```diff
@@ -859,7 +864,7 @@ async def invoke_provide_v(agent):
         pass
     v_json_message = cloud_verifier_common.prepare_v(agent)
 
-    # TODO: remove special handling after initial upgrade
+    # ssl_context is only None with it mTLS verification was disabled in the configuration
```
stefanberger, line 867: parser error in comment near '...None with it mTLS ...'; suggested solution: `... None if mTLS ...`

stefanberger:
Rest LGTM

keylime/cloud_verifier_tornado.py
THS-on: Thanks for catching that. Is now fixed.

THS-on:
> This looks good to me, but I'm going to hold off on approving and merging until we're ready to for the next version bump.

Yeah this should be included just before we release 6.4.0.

THS-on:
Closing this as we only fully remove support for 1.0 once IBM migrates to a new agent.

PR #937, opened 2022-03-22
pylint: Fix R1732 consider-using-with

stefanberger:
Signed-off-by: Stefan Berger <stefanb@linux.ibm.com>

PR #942, opened 2022-03-24
Implement broader key type support for Keylime's signing mechanisms

mbestavros:
Previously, Keylime used GPG for all its signing and verification functionality - most of which was contained within the `gpg.py` library. This PR rewrites `gpg.py` with the goal of making it more modular. Specifically:

- the library now supports additional OpenSSL key types and signatures - it will automatically detect what kind of key it's using and use the appropriate crypto libraries;
- the library has been refactored/renamed to a more generic `signing.py` to reflect the fact it's no longer GPG-specific

Additionally, the `verify_filesignature` function has been split in two: now, instead of requiring keys to be read from disk, the main logic is contained in a new function `verify_signature` that will accept input from memory, and a helper function `verify_signature_from_disk` that interacts with it to read keys from disk and call the main function. This helper function should be a drop-in replacement for the old function. This new flexibility will allow for the verification code to be used in more places within Keylime where keys might not live on disk.

For now, this rewrite retains support for GPG keys as-is, while adding new support for ECDSA keys and signatures. This can easily be expanded to more OpenSSL key types in the future.


mbestavros:
cc @mpeters @lukehinds for review

keylime/signing.py
```diff
@@ -0,0 +1,77 @@
...
+        gpg = gnupg.GPG()
+        logger.debug("Importing GPG key")
+        gpg_imported = gpg.import_keys(key.decode("utf-8"))
+        if gpg_imported.count == 1: # pylint: disable=E1101
+            logger.debug("GPG key successfully imported")
+        else:
+            raise Exception(f"Unable to import GPG key")
+
+        # The Python PGP library won't let you read a signature from memory, hence this hack.
+        with open('tmp.sig', 'wb') as sig_temp:
```
THS-on, line 55: Use a `NamedTemporaryFile` for this.
I implemented a similar hack here for the ssl library (It uses a TemporaryDirectory because i couldn't just load it and immediately remove it. In this case `NamedTemporaryFile` should work):
https://github.com/keylime/keylime/blob/9a03bd405a363546ace05c843ce25cab7f08faa6/keylime/requests_client.py#L33-L38

THS-on (commented):
Can you add tests for the signing module?
Otherwise the changes look sensible to me.

keylime/signing.py
```diff
@@ -0,0 +1,77 @@
...
+from cryptography.hazmat.primitives.asymmetric import ec
+from cryptography.hazmat.primitives.serialization import load_pem_public_key
+
+from keylime import keylime_logging
+
+logger = keylime_logging.init_logging('signing')
+
+def verify_signature_from_disk(key_file, filename, sig_file, file_description):
+    """
+       Read signing materials from disk and verify.
```
stefanberger, line 19: May want to document the parameters like you did below: Verify the file signature (sig) using a public key (key) with the file (file).

mbestavros: Done!

keylime/signing.py
```diff
@@ -0,0 +1,78 @@
...
+from cryptography.hazmat.primitives.asymmetric import ec
+from cryptography.hazmat.primitives.serialization import load_pem_public_key
+
+from keylime import keylime_logging
+
+logger = keylime_logging.init_logging('signing')
+
+def verify_signature_from_disk(key_file, filename, sig_file, file_description):
+    """
+       Verify the file signature on disk (sig) using a public key on disk (key)
```
stefanberger, line 19: sig -> sig_file
key -> key_file etc.

stefanberger:
As for the linter errors 
```
keylime/cmd/ima_emulator_adapter.py:73:12: E4702: Iterated dict 'position' is being modified inside for loop body, iterate through a copy of it instead. (modified-iterating-dict)
************* Module keylime.tenant
keylime/tenant.py:1146:19: E0601: Using variable 'response' before assignment (used-before-assignment)
```

I got those fixed in PR #937

keylime/signing.py
mbestavros: Thanks for the suggestion @THS-on - I've replaced my hack with `NamedTemporaryFile`!

keylime/signing.py
mbestavros: Whoops - fixed, thanks for the catch!

mpeters (changes requested):

stefanberger:
FYI: You can run tests locally:
```
cd .ci
./run_local.sh
```

mbestavros:
Per @THS-on's suggestion, I've written up a basic test suite for the signing library. It tests verification of a GPG signature, an EC signature, and an invalid signature.

Interestingly, it seems like there's already testing in place for signing, though it's not direct - it's via the `read_allowlist` test: https://github.com/keylime/keylime/blob/master/test/test_ima_verification.py#L268

I don't see a reason we can't have both sets of tests.

stefanberger:
```
sudo python3 -m unittest test.test_signing
```

This fix gets it to work:

```
diff --git a/keylime/signing.py b/keylime/signing.py
index a3c10bf..dd1867f 100644
--- a/keylime/signing.py
+++ b/keylime/signing.py
@@ -57,6 +57,7 @@ def verify_signature(key, sig, file):
         # The Python PGP library won't let you read a signature from memory, hence this hack.
         with tempfile.NamedTemporaryFile() as temp_sig:
             temp_sig.write(sig)
+            temp_sig.flush()
             verified = gpg.verify_data(temp_sig.name, file)
 
     # OpenSSL
```

THS-on:
Stefan's solution should work, but the more common approach that I have seen is `temp_sig.seek(0)` instead  of flush.

mbestavros:
@stefanberger @THS-on Thanks, both! `temp_sig.flush()` seems to work, but I can try `seek(0)` as well.

Any other pieces of feedback at the moment?

mbestavros:
@THS-on I caught a GPG string bug in the `read_allowlist` test, so that's fixed as of now. Don't know if there are others causing the `testing-farm` failures, though.

kkaarreell:
> @kkaarreell the e2e tests are failing because the `GPG` is no longer in the error message string, so can you change that?

Done in https://github.com/RedHat-SP-Security/keylime-tests/pull/72
You can rerun packit tests to verify that.

mbestavros:
@mpeters Looks like the Packit tests pass - is this ready for merge?

mpeters (approved):

Issue #680, opened 2021-05-26
The tenant registers agent to the verifier first before handling initial quote provided by the agent. 


igurel:
Is your an issue a feature request? If so, please raise it as an [enhancement](https://github.com/keylime/enhancements)

# Environment
 * OS / version: Ubuntu 20.04
 * Processor architecture: 
 * TPM Manufacturer: 
 * Keylime version: [d632e2a](https://github.com/keylime/keylime/commit/d632e2a3b510167ba997197f5232bc4456aa0394)

# Description 
The tenant registers agent to the verifier first and then it handles initial quote provided by the agent. 

## Expected behavior vs. actual behavior 
The tenant registers agent to the verifier first and then it handles initial quote provided by the agent. This may lead to an issue that the verifier keeps asking quotes from the agent before proper attestation quote validation by the tenant. This may happen by e.g. a malicious agent, another malicious process or http request (sent to the agent by the tenant) which doesn’t timeout

## Steps to reproduce problem
1. See the order in [tenant.py#L1176](https://github.com/keylime/keylime/blob/master/keylime/tenant.py#L1176)

# Relevant logs

Attach any relevant log files that can help to debug your issue.

THS-on:
I've tried switching the order of `do_cv()` and `do_qoute()` and nothing immediately broke. 
On first glance I couldn't find that `do_qoute()` has a dependency on `do_cv()` other than removing the agent from the verifier if it failed.  Is there another reason why we don't just switch the order?

There would be a change in behavior.  The failed agent never gets added to the verifier and is there in a stopped state. If that is desired the order should not be changed and only the EK validation could be moved before the `do_cv()`.

igurel:
> I've tried switching the order of `do_cv()` and `do_qoute()` and nothing immediately broke.

I've tried too and it should work fine.

maugustosilva:
While I am certainly willing to support such change, we would need a new enhancement proposal for this.

ansasaki:
@igurel Hello, do you plan to propose the enhancement? I think this change makes sense (together with #679), and would like to offer help, if needed.

PR #947, opened 2022-03-29
Verify identity quote before adding agent to the verifier

ansasaki:
This implements the enhancement proposed in https://github.com/keylime/enhancements/blob/master/66_quote_before_register.md

Swap the order of `do_cv()` and `do_quote()` calls in the tenant.

Fixes: #680

### DIFF
--- a/keylime/tenant.py
+++ b/keylime/tenant.py
@@ -1032,94 +1032,90 @@ class Tenant():
                 raise e
             break
 
-        try:
-            if response is not None and response.status_code != 200:
-                raise UserError(
-                   f"Status command response: {response.status_code} Unexpected response from Cloud Agent.")
+        if response is not None and response.status_code != 200:
+            raise UserError(
+               f"Status command response: {response.status_code} Unexpected response from Cloud Agent.")
 
-            if "results" not in response_body:
-                raise UserError(
-                    f"Error: unexpected http response body from Cloud Agent: {str(response.status)}")
+        if "results" not in response_body:
+            raise UserError(
+                f"Error: unexpected http response body from Cloud Agent: {str(response.status)}")
 
-            quote = response_body["results"]["quote"]
-            logger.debug("Agent_quote received quote: %s", quote)
+        quote = response_body["results"]["quote"]
+        logger.debug("Agent_quote received quote: %s", quote)
 
-            public_key = response_body["results"]["pubkey"]
-            logger.debug("Agent_quote received public key: %s", public_key)
+        public_key = response_body["results"]["pubkey"]
+        logger.debug("Agent_quote received public key: %s", public_key)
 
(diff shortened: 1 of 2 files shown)


keylime/tenant.py
```diff
@@ -1118,7 +1092,6 @@ def do_quote(self):
                 raise UserError(
                     f"Posting of Encrypted U to the Cloud Agent failed with response code {response.status_code} ({response.text})")
         except Exception as e:
```
THS-on, line 1094: The try-catch block can be fully removed, instead of capturing and re-raising the exception.

THS-on (changes requested):
Can you check if the verification with `do_verify()` still works?

kkaarreell:
Hi @ansasaki , I have found an issue with patch code coverage diff generating that manifests itself in your PR. I have committed a fix in https://github.com/RedHat-SP-Security/keylime-tests/pull/73. Would you mind rescheduling Packit tests in your PR (just F35 would be enough) so I can double-check it works fine? Thank you.

keylime/tenant.py
ansasaki: Thanks for the review, I'll remove it.

ansasaki:
> Can you check if the verification with `do_verify()` still works?

I checked by passing `--verify` to the add command and it worked fine. However, when I tried with the rust agent I got errors on JSON parsing. 

> Hi @ansasaki , I have found an issue with patch code coverage diff generating that manifests itself in your PR. I have committed a fix in [RedHat-SP-Security/keylime-tests#73](https://github.com/RedHat-SP-Security/keylime-tests/pull/73). Would you mind rescheduling Packit tests in your PR (just F35 would be enough) so I can double-check it works fine? Thank you.

I pushed some changes which triggered re-runs for all tests.

THS-on:
> I checked by passing --verify to the add command and it worked fine. However, when I tried with the rust agent I got errors on JSON parsing.

This seems like a bug in the rust agent.

kkaarreell:
> I pushed some changes which triggered re-runs for all tests.

Right, I can see that the patch is correct now. Unfortunately, e2e tests do not utilize the modified code. Any suggestions for test scenarios (or patches) are welcome. :-)

ansasaki:
> > I checked by passing --verify to the add command and it worked fine. However, when I tried with the rust agent I got errors on JSON parsing.
> 
> This seems like a bug in the rust agent.

It seems that when the rust agent returns a `400` error it does not include the `results` field. I need to debug it further.

ansasaki:
> It seems that when the rust agent returns a `400` error it does not include the `results` field. I need to debug it further.

Indeed the responses for error cases were not wrapped in the JSON structure. I created a PR to fix this:

https://github.com/keylime/rust-keylime/pull/356

THS-on (approved):
LGTM.

PR #948, opened 2022-03-30
Implement verifier-side signature check for the addallowlist tenant command

mbestavros:
This PR is a first prototype of verifier-side allowlist signature verification.

Previously, the only time allowlist signatures were validated was when the tenant read an allowlist from disk as part of the `add` or `addallowlist` commands. Ideally, this would be expanded to include a similar check in the verifier once the allowlist had been sent over.

There is a significant complicating factor to this idea, however: once an allowlist is read and verified, the tenant performs significant pre-processing on the payload it eventually sends to the the verifier, to the point where the "allowlist" sent to the verifier could be vastly different from the artifact that was validated on the tenant. (For example, if the allowlist is of an older non-JSON format, the tenant will automatically convert it to JSON.) This makes signature validation impossible. There are a few solutions to this problem, the most obvious of which is to perform any allowlist processing after any signature validation steps have happened.

This PR implements a possible first step towards adding the second, verifier-side signing step by augmenting the `addallowlist` command. As it turns out, the cloud verifier's `addallowlist` handler doesn't actually use anything from the allowlist payload sent over by the tenant - meaning the tenant can send over an unprocessed version of the allowlist, which can then be validated on the verifier. This PR does exactly that by collecting the signing materials on the tenant, sending them over to the verifier, and performing an additional signature check on the verifier with those materials. No other functionality should be impacted by the new validation.

Very open to feedback/discussion on the best way to proceed.

### DIFF
--- a/keylime.conf
+++ b/keylime.conf
@@ -360,6 +360,11 @@ severity_policy = [{"event_id": ".*", "severity_label": "emergency"}]
 # Set to True to treat ToMToU entries as errors.
 tomtou_errors = False
 
+# If set to true, the verifier will require associated signatures and private
+# keys when getting an allowlist, and will check the allowlist against those
+# signatures before returning the requested allowlist.
+enforce_allow_list_signatures = False
+
 #=============================================================================
 [tenant]
 #=============================================================================
--- a/keylime/cloud_verifier_tornado.py
+++ b/keylime/cloud_verifier_tornado.py
@@ -3,6 +3,7 @@
 SPDX-License-Identifier: Apache-2.0
 Copyright 2017 Massachusetts Institute of Technology.
 '''
+import base64
 import signal
 import traceback
 import sys
@@ -28,6 +29,7 @@ from keylime.db.keylime_db import DBEngineManager, SessionManager
 from keylime import keylime_logging
 from keylime import cloud_verifier_common
 from keylime import revocation_notifier
+from keylime import signing
 from keylime import web_util
 from keylime import tornado_requests
 from keylime import api_version as keylime_api_version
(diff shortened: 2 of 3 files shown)


mbestavros:
cc @mpeters @lukehinds @THS-on - would love your feedback.

kkaarreell:
There is a test failure related to this PR
```
keylime.tenant - ERROR - expected str, bytes or os.PathLike object, not NoneType
  File "/usr/local/lib/python3.10/site-packages/keylime-6.3.1-py3.10.egg/keylime/cmd/tenant.py", line 18, in main
    tenant.main()
  File "/usr/local/lib/python3.10/site-packages/keylime-6.3.1-py3.10.egg/keylime/tenant.py", line 1487, in main
    mytenant.do_add_allowlist(vars(args))
  File "/usr/local/lib/python3.10/site-packages/keylime-6.3.1-py3.10.egg/keylime/tenant.py", line 1202, in do_add_allowlist
    with open(args["allowlist"], 'rb') as alist_f:
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

THS-on:
We can make a few changes to make this work:

- Merge the exclude list and allowlist into one single JSON object.
- Only allow server side signature validation against the JSON object, not the legacy formats.
- Run the `update_allowlist` function on the verifier, after signature validation.

We also need to make sure that we don't make implementing https://github.com/keylime/enhancements/pull/69 harder.

mpeters:
> We can make a few changes to make this work:
> 
> * Merge the exclude list and allowlist into one single JSON object.

In general I think this is a good move for the future. An allowlist and an exclude list can be closely tied together and you might want the same protections for the exclude as much as the allow. Maybe call it "ima-policy" or something similar.

> * Only allow server side signature validation against the JSON object, not the legacy formats.

I disagree with this. In fact, I think we need to store the original allowlist as-given by the user, no matter the format. We are always going to have to turn the on-disk format into an in-memory one, so any validation of checksums/sigs will have to happen to the original on-disk format. No reason we can't keep it in whatever format was originally submitted and then do any transformations on it as it's loaded by the verifier, no matter the original format.

> * Run the `update_allowlist` function on the verifier, after signature validation.
> 
> We also need to make sure that we don't make implementing [keylime/enhancements#69](https://github.com/keylime/enhancements/pull/69) harder.

Definitely agreed :)

mbestavros:
Superseded by #980

PR #949, opened 2022-03-31
tpm_emulator: move everything into systemd services

THS-on:
The IMA emulator is now also a systemd service, which depends on the TPM emulator.

I originally wanted to switch to swtpm, but the TCTI interface is not supported on most shipped tss versions.


mpeters (approved):

ima_stub_service/installer.sh
```diff
@@ -9,16 +9,16 @@ if [[ $EUID -ne 0 ]]; then
...
 fi
+
+# create services and socket
+for service in "tpm_emulator.service" "ima_emulator.service"; do
+  cp  ${service} /etc/systemd/system
+  chmod 644 /etc/systemd/system/${service}
+  systemctl enable ${service}
+done
+
+systemctl start ima_emulator.service
```
kkaarreell, line 24: I think there should be `systemctl daemon-reload` call before we start using new unit file.

THS-on: You are right. I'll make a PR that also cleans up some parts from the main installer.sh, that I forgot to remove.

PR #950, opened 2022-04-01
Enable SQL query cache for JSONPickleType

ansasaki:
Add the cache_ok = True flag to the JSONPickleType classes to supress SQLAlchemy warnings and enable query caching for improved performance.

It should be safe to enable caching for the JSONPickleType as it does not include additional state attributes which affects how it renders SQL.

This also adds an option to enable debugging the SQL queries and caching via `DEBUG_DB` environment variable which is only effective when `INSECURE_DEBUG` is also enabled.

See:
- https://docs.sqlalchemy.org/en/14/faq/performance.html
- https://docs.sqlalchemy.org/en/14/core/custom_types.html

### DIFF
--- a/keylime/config.py
+++ b/keylime/config.py
@@ -202,3 +202,7 @@ BOOTSTRAP_KEY_SIZE = 32
 CA_IMPL = get_config().get('general', 'ca_implementation', fallback='openssl')
 
 CRL_PORT = 38080
+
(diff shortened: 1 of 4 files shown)


mpeters (approved):

mpeters:
@ansasaki some lint failures to fix first before merging

sergio-correia (approved):

PR #952, opened 2022-04-01
tpm_emulator: use tpm2-abrmd as resource manager

THS-on:
Talking directly to the TPM causes problems with transient objects, so we use tpm2-abrmd as a resource manager.

I have only validated those changes on Fedora 35.

### DIFF
--- a/ima_stub_service/ima_emulator.service
+++ b/ima_stub_service/ima_emulator.service
@@ -1,10 +1,10 @@
 [Unit]
 Description=IMA emulator for Keylime
-After=network.target tpm_emulator.service
-Requires=tpm_emulator.service
+After=network.target tpm_emulator_rm.service
+Requires=tpm_emulator_rm.service
 
 [Service]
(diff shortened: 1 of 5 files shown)


kgold2:
What's the problem?  Transient objects should be closed when the connection to /dev/tpmrm0 is closed.  I presume Intel's abrmd does the same.

This should only be an issue if the application opens and closes connections during its processing.

THS-on:
> What's the problem? Transient objects should be closed when the connection to /dev/tpmrm0 is closed. I presume Intel's abrmd does the same.

The problem is that for the emulator we directly talk to the IBM TPM emulator without a resource manager. This causes transient objects to be never flushed and after a few runs the TPM is out of memory. This PR adds tpm2-abrmd as a resource manager between tpm2-tools and the TPM emulator.

THS-on:
As discussed on Slack I'll add flushing after each TPM interaction instead of re-introducing tpm2-abrmd.

PR #955, opened 2022-04-04
Fix tpm import in test_tpm.py

kkaarreell:
This fixes a problem with running the test through the
"coverage run". Currently, it fails due to the following error:

  AttributeError: module 'keylime' has no attribute 'tpm'

### DIFF
--- a/test/test_tpm.py
+++ b/test/test_tpm.py
@@ -1,5 +1,5 @@
 import unittest
-import keylime
+from keylime.tpm.tpm_main import tpm
 
 # ############################################################
 # list of input challenges for get_tpm_manufacturer function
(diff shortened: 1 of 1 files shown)


kkaarreell:
@mpeters The current script is broken as this import won't work when running the test as a standalone script, i.e.
`python3 test_tpm.py`
This is what coverage script does as it runs each test individually.
https://github.com/keylime/keylime/blob/master/test/run_tests.sh#L199
OTOH, for non-coverage scenario tests are run altogether via `green` which is IMHO what makes a difference as the required import is most likely done by some of the former tests.
https://github.com/keylime/keylime/blob/master/test/run_tests.sh#L218

mpeters (approved):

PR #957, opened 2022-04-05
Remove stubbing support for TPM, VTPM and IMA

THS-on:
This feature was unused and stubbing was never implemented for newer components in Keylime.


aplanas (approved):
Thanks a lot!

mgerstner (approved):

mpeters:
What is the reason to include the removal of all of the archive stuff in this PR? I agree with removing it (git itself is an archive, no need to keep things around), but would it make more sense as separate PRs?

THS-on:
> What is the reason to include the removal of all of the archive stuff in this PR? I agree with removing it (git itself is an archive, no need to keep things around), but would it make more sense as separate PRs?

I didn't want to touch this code during the removal of the stubs. I can make a separate PR, but I don't think there is much discussion about its removal needed.

mpeters:
I guess it was more about keeping separate things separate, but it's not a big deal either way.

THS-on:
> I guess it was more about keeping separate things separate, but it's not a big deal either way.

You are right, I split this PR into this and #961.

maugustosilva (approved):

PR #959, opened 2022-04-05
Avoid mounting secdir while unmounting it

kkaarreell:
mount() call will mount secdir if not mounted.
As a result we end up removing content of an empty mount point.

At the same time this extra mount() call causes a failure
in test_check_umount_umount() from test_secure_mount.py

### DIFF
--- a/keylime/secure_mount.py
+++ b/keylime/secure_mount.py
@@ -62,11 +62,18 @@ def check_mounted(secdir):
     return False
 
 
-def mount():
+def get_secdir():
     secdir = os.path.join(config.WORK_DIR, "secure")
 
     if not config.MOUNT_SECURE:
         secdir = os.path.join(config.WORK_DIR, "tmpfs-dev")
+
+    return secdir
+
+def mount():
+    secdir = get_secdir()
+
+    if not config.MOUNT_SECURE:
         if not os.path.isdir(secdir):
             os.makedirs(secdir)
         return secdir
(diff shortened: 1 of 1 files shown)


mpeters (approved):

kkaarreell:
@THS-on @aplanas This is my attempt to address the failure in `test_check_umount_umount`. 
There is a question whether we should remove `secdir` only when it is mounted or always delete it. When not mounted, would it be possible that keylime would use it to store data anyway? If yes, we should delete it either way. In this patch however I am preserving former behavior which would (in fact) delete it only if mounted.

kkaarreell:
Hi @mpeters, I have just forced pushed tiny update, please review it once more. Thank you.

keylime/secure_mount.py
```diff
@@ -88,10 +95,14 @@ def mount():
 def umount():
     """Umount all the devices mounted by Keylime."""
 
-    # Make sure we leave tmpfs dir empty even if we did not mount it or
-    # if we cannot unmount it. Ignore errors while deleting. The deletion
-    # of the 'secure' directory will result in an error since it's a mount point.
-    shutil.rmtree(mount(), ignore_errors=True)
+    secdir = get_secdir()
+    if not config.MOUNT_SECURE:
```
aplanas, line 99: Can you join both branches? `if check_mounted(secdir) or not config.MOUNT_SECURE:`

kkaarreell: Will do. I will just switch the order to not to check the mount only if MOUNT_SECURE is not used.

aplanas (approved):

maugustosilva (approved):

PR #967, opened 2022-04-06
Workaround for mock_open not supporting iteration in Python 3.6

kkaarreell:
In Python 3.6 mock_open does not support iteration around text
files. This is causing test failures in test_secure_mount.py
which is using mock_open. We workaround the issue by not using
iterator.

For more details see:
  https://code-examples.net/en/q/17a1c75
  https://bugs.python.org/issue21258

### DIFF
--- a/keylime/secure_mount.py
+++ b/keylime/secure_mount.py
@@ -21,7 +21,11 @@ def check_mounted(secdir):
     """Inspect mountinfo to detect if a directory is mounted."""
     secdir_escaped = secdir.replace(" ", r"\040")
     with open("/proc/self/mountinfo", "r", encoding="utf-8") as f:
-        for line in f:
+        # because of tests we use readlines() and avoid using iterator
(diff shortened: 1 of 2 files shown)


kkaarreell:
There are alternatives:
 * implement workaround described at https://code-examples.net/en/q/17a1c75 into individual test cases
 * do not run failing tests on CentOS Stream 8
 * do not run whole test suite on CentOS Stream 8

I have tried various patches from https://bugs.python.org/issue32933 and https://bugs.python.org/issue21750 
but it didn't fix the problem so fixing the issue in mock directly may not be possible, at least in a short term.

THS-on (approved):
I think this approach is fine, because `/proc/self/mountinfo` is not that large and we can load it into memory before iterating over it. (I don't think that readlines() supports lazy iterators.)

aplanas (approved):

mpeters (approved):

PR #968, opened 2022-04-07
Remove last traces of M2Crypto

sergio-correia:
Since it was removed in #747.


THS-on:
@sergio-correia please add the sign-off message to your commit.

@kkaarreell can you check if `M2Crypto` is not part of the CI?

kkaarreell:
> @kkaarreell can you check if `M2Crypto` is not part of the CI?

We do not have any tests explicitly using it. Unless there is something in keylime upstream tests... But I don't see anything, at least with `grep`.

THS-on (approved):

mpeters (approved):

PR #969, opened 2022-04-07
Deprecate API version 1.0

mpeters:
Adding new capability to mark API versions as "deprecated" which will
be displayed when verifier and registrar start and a warning is issued
anytime a client (agent or tenant) uses a deprecated API.

This will allow us to mark APIs as deprecated before they are fully removed.

### DIFF
--- a/keylime/web_util.py
+++ b/keylime/web_util.py
@@ -190,6 +190,16 @@ def get_restful_params(urlstring):
     return path_params
 
 
+def validate_api_version(handler, version, logger):
+    if not version or not keylime_api_version.is_supported_version(version):
+        echo_json_response(handler, 400, "API Version not supported")
+        return False
+
+    if keylime_api_version.is_deprecated_version(version):
+        logger.warning('Client request to API version %s is deprecated and will be removed in future versions.', version)
+    return True
+
+
 def _list_to_dict(alist):
     """Convert list into dictionary via grouping [k0,v0,k1,v1,...]"""
     params = {}
--- a/keylime/api_version.py
+++ b/keylime/api_version.py
@@ -15,6 +15,7 @@ LATEST_VERSIONS = {
     "1": "1.0",
     "2": "2.0"
 }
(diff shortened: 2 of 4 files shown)


aplanas (approved):

THS-on (changes requested):
@mpeters can you rebase on master and fix the pylint issues?

```
************* Module keylime.web_util
keylime/web_util.py:194:4: R1705: Unnecessary "elif" after "return", remove the leading "el" from "elif" (no-else-return)
************* Module keylime.cloud_verifier_tornado
keylime/cloud_verifier_tornado.py:264:0: W0311: Bad indentation. Found 14 spaces, expected 12 (bad-indentation)
keylime/cloud_verifier_tornado.py:333:0: W0311: Bad indentation. Found 14 spaces, expected 12 (bad-indentation)
```

I support this approach that we mark an API version deprecated for one version.

keylime/web_util.py
```diff
@@ -190,6 +190,16 @@ def get_restful_params(urlstring):
     return path_params
 
 
+def validate_api_version(handler, version, logger):
+    if not version or not keylime_api_version.is_supported_version(version):
+        echo_json_response(handler, 400, "API Version not supported")
+        return False
+
+    if keylime_api_version.is_deprecated_version(version):
+        logger.warning('Client request to API version %s is deprecated and will be removed in future versions.', version)
```
THS-on, line 199: We should sanitize the API version before logging, but otherwise the PR looks fine to me.

THS-on (changes requested):

keylime/web_util.py
mpeters: That's already happening in `get_restful_params()` as part of pulling it from the URL. It's sanitized and normalized.

THS-on: Yes, you are right. I missed that.

THS-on (approved):

PR #971, opened 2022-04-08
Bump required packaging module version to 20.0

kkaarreell:
We are already consuming attributes such as 'minor'
and 'major' introduced by this version in api_version.py.

### DIFF
--- a/packit-ci.fmf
+++ b/packit-ci.fmf
@@ -4,7 +4,9 @@
 
   prepare:
     how: shell
-    script: bash -c 'ln -s $(pwd) /var/tmp/keylime_sources'
+    script:
+     - rm -f /etc/yum.repos.d/tag-repository.repo
+     - ln -s $(pwd) /var/tmp/keylime_sources
 
   discover:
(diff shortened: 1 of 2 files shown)


THS-on (approved):

sergio-correia (approved):

kkaarreell:
test_restful.py is unstable on CS8, seems that TPM restarting fails frequently. Maybe we should exclude this test on CS8.

kkaarreell:
I think there are some packaging related issue with the Rawhide image used in CI. I cannot reproduce it myself though, it must be related to the internal Testing farm repo.

kkaarreell:
JFTR, https://koji.fedoraproject.org/koji/buildinfo?buildID=1944200 is responsible for broken deps in Rawhide.

kkaarreell:
> test_restful.py is unstable on CS8, seems that TPM restarting fails frequently. Maybe we should exclude this test on CS8.

Done.

mpeters (approved):

PR #974, opened 2022-04-11
Full removal of vTPM support

THS-on:
Keylime no longer supports vTPM 1.2 from Xen and swtpm + libvirt never supported it for TPM 2.0, so we now remove it completely.


THS-on:
@kkaarreell this PR removes the `--vtpm_policy` option from the tenant. So once this gets merged the tests should be adjusted to reflect that.

kkaarreell:
> @kkaarreell this PR removes the `--vtpm_policy` option from the tenant. So once this gets merged the tests should be adjusted to reflect that.

I will prepare PR doing that.

kkaarreell:
> > @kkaarreell this PR removes the `--vtpm_policy` option from the tenant. So once this gets merged the tests should be adjusted to reflect that.
> 
> I will prepare PR doing that.

Done, you can rerun your tests.

axelsimon (approved):
Looks good to me. However, given this bundles a destructive action (db migration script dropping columns), we need to issue a clear warning to users of vTPM (are there any?) that this will do that.

THS-on:
> Looks good to me. However, given this bundles a destructive action (db migration script dropping columns), we need to issue a clear warning to users of vTPM (are there any?) that this will do that.

There was never working vTPM support for TPM 2.0, so I am pretty sure that nobody using Keylime version 6.0.0 and up has any use for that feature. I think we can remove it without any extra warnings.

maugustosilva:
@axelsimon I second @THS-on evaluation, even though none of us have hard evidence at the moment. To be best of my knowledge, nobody has ever used vTPM with Xen outside of the team who initially wrote the paper (implementing something akin to "deep quote", requiring both a pTPM and vTPM). On other hand, I have experienced heavy use of vTPM with libvirt, which never required the "vTPM" support on keylime (as it treats the vTPM just like any other pTPM).

maugustosilva (approved):

PR #976, opened 2022-04-12
config, webapp: remove tls_check_hostnames option

THS-on:
This was only used in the tenant webapp code, but not used. No other part of Keylime uses it, so we remove the tls_check_hostnames from the general section of the keylime.conf

### DIFF
--- a/keylime.conf
+++ b/keylime.conf
@@ -5,9 +5,6 @@
 # Turn on or off TLS Keylime wide.
 enable_tls = True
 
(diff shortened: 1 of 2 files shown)


kkaarreell:
Config option usage in keylime-tests will be removed via https://github.com/RedHat-SP-Security/keylime-tests/pull/85

aplanas (approved):

mpeters (approved):

PR #980, opened 2022-04-25
Major overhaul of allowlist functionality

mbestavros:
This PR implements a few significant improvements to Keylime's allowlist functionality, notably:

- The partially-complete Allowlists API on the verifier has been finished and is now fully functional with `create`, `retrieve`, and `delete` actions.
- Allowlists are now stored in their own, separate database - instead of `VerifierMain`, they now live in the as-yet unused `VerifierAllowlist` database.
- There have been major changes to the `add` command for adding agents to accommodate the new database arrangement - now, you can add an agent without providing an allowlist file if an allowlist was previously uploaded using the Allowlists API.
  - At the same time, existing functionality has been preserved: providing an allowlist file with the `add` command will still store the allowlist as before.
- Refactored allowlist processing code to perform disk loading and further processing separately from each other
- Deferred allowlist processing/reformatting to the verifier, which allows for verifier-side signature validation

TODO:

- Ensure namespaces are consistent
- At present, Keylime still stores the edited/processed allowlist artifact, and not the original as provided by the user on the tenant.
- Some further database work is needed, and it's possible some migrations will need to be written.
- Some tests are broken by these changes that will need to be fixed.
- General code cleanup

### DIFF
--- a/keylime/cloud_verifier_common.py
+++ b/keylime/cloud_verifier_common.py
@@ -38,7 +38,7 @@ def get_AgentAttestStates():
     return AgentAttestStates.get_instance()
 
 
-def process_quote_response(agent, json_response, agentAttestState) -> Failure:
+def process_quote_response(agent, allowlist, json_response, agentAttestState) -> Failure:
     """Validates the response from the Cloud agent.
 
     This method invokes an Registrar Server call to register, and then check the quote.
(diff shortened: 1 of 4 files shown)


lukehinds:
nice work Mark. 

What's the thinking behind a separate DB?

mpeters:
A separate DB table is also necessary for the 1-to-many relationship for multi-allowlist. That's a separate feature, but it's something we'll need eventually so if it works as a design for this feature then lets go with it.

THS-on:
@mbestavros thanks for clarifying.

> I haven't implemented any functionality to remove uploaded allowlists when an agent is deleted - in this draft of the code, if you provide an allowlist file, you must also provide an allowlist name, and it will persist after the agent is deleted. This can change, of course.

I would say if there is now name provided generate a automatic one and remove the entry after the agent is removed, otherwise persist it.

lukehinds:
do you want to rebase this @mbestavros ?

PR #982, opened 2022-04-26
Add pre-commit hook to enforce code style with isort and black

THS-on:
This PR reformats the entire code base using isort and black. [pre-commit](https://pre-commit.com/) is used in the CI and can be used locally to enforce the code style before creating a commit.

I set the default line width to 100, but we can change that also to something like 120. I think that smaller than 100 does not make sense for our codebase. 

Please look if the formatting made something not readable, so that we disable it manually for that part.


mpeters:
> I set the default line width to 100, but we can change that also to something like 120. I think that smaller than 100 does not make sense for our codebase.

I agree that 120 makes more sense than 100.

keylime/agentstates.py
```diff
@@ -1,27 +1,28 @@
...
-        """ constructor """
+        """constructor"""
         self.pcrs = {}
         self.hash_alg = {}  # Record the hash algorithm that a given PCR uses
         for pcr_num in range(0, 24):
             self.reset_pcr(pcr_num)
 
     def init_pcr(self, pcr_num, hash_alg):
-        """" Initializes a PCR """
+        """ " Initializes a PCR"""
```
aplanas, line 25: remove extra quote

keylime/backport_dataclasses.py
```diff
@@ -4,13 +4,14 @@
 Source: https://pypi.org/project/dataclasses/
 """
 # pylint: skip-file
+# fmt: off
```
aplanas, line 7: why skip the format here? too big?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,10 @@
+repos:
```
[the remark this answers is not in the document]
THS-on: It can be executed locally and is enforced in the CI. You should run black and isort locally before commiting and I think pre-commit makes it pretty convenient, but you can run it also outside of pre-commit.

keylime/benchmark/utility_plot_time_series.py
```diff
@@ -34,27 +34,25 @@ def main(argv=sys.argv):  #pylint: disable=dangerous-default-value
...
+    #         if each_file is not None:
+    #             with open(each_file) as input_file:
+    #                 content = [x.strip() for x in input_file.readlines()]
+    #                 index = 0
+    #                 for timestamp in content:
+    #                     if index != 0 and index <= (len(content) - 3):
+    #                         dec_secs = float(content[index + 1]) - float(timestamp)
+    #                         msecs = dec_secs * 1000
+    #                         difference_list.append(msecs)
+    #                     index = index + 1
```
aplanas, line 50: maybe is a chance to remove this block?

keylime/ca_util.py
```diff
@@ -295,6 +300,7 @@ def get_crl_distpoint(cert_path):
     logger.info("No CRL distribution points in %s", cert_path)
```
aplanas, line 300: there is something weird here: the extra "pass" and the indentation for sure. But maybe the code scanning complain makes sense too

keylime/cloud_verifier_tornado.py
```diff
@@ -218,23 +219,22 @@ def get(self):
...
-            self, 405, "Not Implemented: Use GET interface instead")
+        web_util.echo_json_response(self, 405, "Not Implemented: Use GET interface instead")
 
     def data_received(self, chunk):
         raise NotImplementedError()
 
 
 class AgentsHandler(BaseHandler):
-    mtls_options = None  # Stores the cert, key and password used by the verifier for mTLS connections
+    mtls_options = (
```
aplanas, line 235: Maybe is better to move the comments on top, so black avoid this weird formatting

keylime/backport_dataclasses.py
aplanas: ouch, of course

keylime/ca_util.py
THS-on: fixed

mpeters (approved):

THS-on:
@aplanas @maugustosilva  can we merge this or are their still issues to address?

aplanas (approved):

PR #987, opened 2022-05-10
Defer loading PyZMQ to avoid optional dependency

ueno:
Importing the Python module zmq brings in quite a few system
dependencies through libzmq, including cryptographic libraries, while
the revocation notification feature in Keylime is not always desired.
This patch changes the top-level import of zmq to local imports at the
actual usage.

Signed-off-by: Daiki Ueno <dueno@redhat.com>

### DIFF
--- a/keylime/revocation_notifier.py
+++ b/keylime/revocation_notifier.py
@@ -12,7 +12,6 @@ from multiprocessing import Process
 from typing import Optional
 
 import requests
-import zmq
 
 from keylime import config, crypto, json, keylime_logging, secure_mount
 from keylime.common import retry
@@ -24,6 +23,12 @@ _SOCKET_PATH = "/var/run/keylime/keylime.verifier.ipc"
 
 
 def start_broker():
+    assert config.getboolean("cloud_verifier", "revocation_notifier")
+    try:
+        import zmq  # pylint: disable=import-outside-toplevel
+    except ImportError as error:
+        raise Exception("install PyZMQ for 'revocation_notifier' option") from error
+
     def worker():
         # do not receive signals form the parent process
         os.setpgrp()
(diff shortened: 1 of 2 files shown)


ueno:
This is a preparation for #795.

ueno:
The pylint errors seem irrelevant to this PR:
```console
************* Module keylime.tpm.tpm2_objects_test
keylime/tpm/tpm2_objects_test.py:195:25: E1101: Instance of 'EllipticCurvePublicNumbers' has no 'e' member; maybe 'x'? (no-member)
keylime/tpm/tpm2_objects_test.py:196:25: E1101: Instance of 'EllipticCurvePublicNumbers' has no 'n' member; maybe 'x'? (no-member)
keylime/tpm/tpm2_objects_test.py:216:25: E1101: Instance of 'EllipticCurvePublicNumbers' has no 'e' member; maybe 'x'? (no-member)
keylime/tpm/tpm2_objects_test.py:218:16: E1101: Instance of 'EllipticCurvePublicNumbers' has no 'n' member; maybe 'x'? (no-member)
```
I tried to add type annotation to the caller, i.e.,
```python
new_rsa_pubkey: RSAPublicKey = pubkey_from_tpm2b_public(correct_rsa_obj)
```
but it didn't help; pylint still thinks `pubkey_from_tpm2b_public` returns `EllipticCurvePublicKey`.

keylime/revocation_notifier.py
```diff
@@ -30,6 +29,15 @@
 
 
 def start_broker():
+    try:
```
mpeters, line 32: Is there any way to only do this if the `revocation_notifier` option is set?

mpeters (changes requested):

THS-on:
@ueno the lint issues should be fixed now in master.

keylime/revocation_notifier.py
ueno: I believe this code is only called from the code path taken when `revocation_notifier` option is set, but to be more explicit about it, I've added `assert`.

Perhaps we could wrap the `zmq` module like [`LazyLoader`](https://docs.python.org/3/library/importlib.html#importlib.util.LazyLoader) and consolidate those checks in a single place, but I haven't looked into it closely.

maugustosilva (approved):

mpeters (approved):

PR #994, opened 2022-05-18
Refactor allowlist reading/sending to enable verifier-side signature checks

mbestavros:
This PR refactors how allowlists are read and processed by the tenant and verifier, enabling verifier-side signature checks.

Now, instead of all allowlist processing being done tenant-side, the tenant sends the raw allowlist artifact along with any provided validation artifacts (checksum, signature, signing key) as a bundle to the verifier. (Note that the existing tenant-side signature check is preserved.)

The verifier will then unbundle everything and perform its own signature validation step before performing the final processing step and storing the formatted allowlist in the database as before.

This enhances security by thwarting compromised tenant attacks - prior to this, Keylime's security model relied on a trusted tenant; however, if a malicious tenant sent a bad payload (ex. one that did not pass signature checks), the verifier would blindly accept it. The new verifier-side signature check makes this attack impossible.

Verifier-side signature validation can be controlled via a new configuration option in `keylime.conf`. By default, signatures are not enforced, but setting `enforce_allow_list_signatures` to `True` will require that check.

### DIFF
--- a/keylime.conf
+++ b/keylime.conf
@@ -107,9 +107,9 @@ revocation_actions=
 payload_script=autorun.sh
 
 # In case mTLS for the agent is disabled and the use of payloads is still
-# required, this option has to be set to "True" in order to allow the agent 
+# required, this option has to be set to "True" in order to allow the agent
 # to start. Details on why this configuration (mTLS disabled and payload enabled)
-# is generally considered insecure can be found on 
+# is generally considered insecure can be found on
 # https://github.com/keylime/keylime/security/advisories/GHSA-2m39-75g9-ff5r
 enable_insecure_payload = False
 
@@ -326,6 +326,11 @@ severity_policy = [{"event_id": ".*", "severity_label": "emergency"}]
 # By default we ignore those entries and only print a warning.
 # Set to True to treat ToMToU entries as errors.
 tomtou_errors = False
+
+# If enabled, the Keylime verifier will require that a file signature and
+# associated key be sent alongside newly-created allowlists, and will perform a
+# signature check before storing them in the database.
+require_allow_list_signatures = False
 
 #=============================================================================
 [tenant]
--- a/keylime/cloud_verifier_tornado.py
+++ b/keylime/cloud_verifier_tornado.py
@@ -27,6 +27,7 @@ from keylime.db.keylime_db import DBEngineManager, SessionManager
 from keylime.db.verifier_db import VerfierMain, VerifierAllowlist
 from keylime.elchecking import policies
 from keylime.failure import MAX_SEVERITY_LABEL, Component, Failure
+from keylime.ima import ima
 
 logger = keylime_logging.init_logging("cloudverifier")
 
@@ -416,6 +417,29 @@ class AgentsHandler(BaseHandler):
                     logger.warning("POST returning 400 response. Expected non zero content length.")
                 else:
                     json_body = json.loads(self.request.body)
+
+                    # Handle allowlists/IMA policies
+                    ima_policy_bundle = json.loads(json_body.get("ima_policy_bundle"))
+                    if ima_policy_bundle:
+                        try:
+                            allowlist = ima.unbundle_ima_policy(
+                                ima_policy_bundle,
+                                verify=config.getboolean(
+                                    "cloud_verifier", "require_allow_list_signatures", fallback=False
+                                ),
+                            )
+                            ima_policy = json.dumps(ima.process_ima_policy(allowlist, ima_policy_bundle["excllist"]))
+                        except ima.SignatureValidationError as e:
+                            web_util.echo_json_response(self, e.code, e.message)
+                            logger.warning(e.message)
+                            return
+                    else:
+                        # Read legacy allowlist if present in payload
+                        ima_policy = json_body.get("allowlist", None)
+                        if not ima_policy:
+                            logger.info("Allowlist data not provided with request! Using default allowlist.")
+                            ima_policy = json.dumps(ima.process_ima_policy(ima.EMPTY_ALLOWLIST, []))
+
                     agent_data = {}
                     agent_data["v"] = json_body["v"]
                     agent_data["ip"] = json_body["cloudagent_ip"]
@@ -424,7 +448,7 @@ class AgentsHandler(BaseHandler):
                     agent_data["public_key"] = ""
                     agent_data["tpm_policy"] = json_body["tpm_policy"]
                     agent_data["meta_data"] = json_body["metadata"]
-                    agent_data["allowlist"] = json_body["allowlist"]
+                    agent_data["allowlist"] = ima_policy
                     agent_data["mb_refstate"] = json_body["mb_refstate"]
                     agent_data["ima_sign_verification_keys"] = json_body["ima_sign_verification_keys"]
                     agent_data["revocation_key"] = json_body["revocation_key"]
--- a/keylime/ima/ima.py
+++ b/keylime/ima/ima.py
@@ -1,3 +1,4 @@
+import base64
 import codecs
 import copy
 import datetime
@@ -19,6 +20,25 @@ logger = keylime_logging.init_logging("ima")
 
 # The version of the allowlist format that is supported by this keylime release
 ALLOWLIST_CURRENT_VERSION = 5
+
+# A correctly-formatted empty allowlist.
+#
+# IMA allowlists of versions older than 5 will not have the "log_hash_alg"
+# parameter. Hard-coding it to "sha1" is perfectly fine, and the fact one
+# specifies a different algorithm on the kernel command line (e.g., ima_hash=sha256)
+# does not affect normal operation of Keylime, since it does not validate the
+# hash algorithm received from agent's IMA runtime measurements.
+# The only situation where this hard-coding would become a problem is if and when
+# the kernel maintainers decide to use a different algorithm for template-hash.
+EMPTY_ALLOWLIST = {
+    "meta": {
+        "version": ALLOWLIST_CURRENT_VERSION,
+    },
+    "release": 0,
+    "hashes": {},
+    "keyrings": {},
+    "ima": {"ignored_keyrings": [], "log_hash_alg": "sha1"},
+}
 
 
 class IMAMeasurementList:
@@ -392,7 +412,7 @@ def update_allowlist(allowlist):
     return allowlist
 
 
-def process_allowlists(allowlist, exclude):
+def process_ima_policy(allowlist, exclude):
     # Pull in default config values if not specified
     if allowlist is None:
         allowlist = read_allowlist()
@@ -420,49 +440,49 @@ def process_allowlists(allowlist, exclude):
     return {"allowlist": allowlist, "exclude": exclude}
 
 
-# IMA allowlists of versions older than 5 will not have the "log_hash_alg"
-# parameter. Hard-coding it to "sha1" is perfectly fine, and the fact one
-# specifies a different algorithm on the kernel command line (e.g., ima_hash=sha256)
-# does not affect normal operation of Keylime, since it does not validate the
-# hash algorithm received from agent's IMA runtime measurements.
-# The only situation where this hard-coding would become a problem is if and when
-# the kernel maintainers decide to use a different algorithm for template-hash.
-empty_allowlist = {
-    "meta": {
-        "version": ALLOWLIST_CURRENT_VERSION,
-    },
-    "release": 0,
-    "hashes": {},
-    "keyrings": {},
-    "ima": {"ignored_keyrings": [], "log_hash_alg": "sha1"},
-}
+# Read allowlist files from disk, validate signatures and checksums, and prepare for sending.
+# Does not process exclusion lists.
+def read_allowlist(alist=None, checksum="", al_sig_file=None, al_key_file=None):
 
-
-def read_allowlist(al_path=None, checksum="", gpg_sig_file=None, gpg_key_file=None):
-    if al_path is None:
-        al_path = config.get("tenant", "ima_allowlist")
+    al_key = b""
+    al_sig = b""
 
     # If user only wants signatures then an allowlist is not required
-    if al_path is None or al_path == "":
-        return copy.deepcopy(empty_allowlist)
+    if alist is None or alist == "":
+        alist_bytes = json.dumps(copy.deepcopy(EMPTY_ALLOWLIST)).encode()
 
-    # Purposefully die if path doesn't exist
-    with open(al_path, "rb") as f:
-        pass
+    elif isinstance(alist, str):
+        if alist == "default":
+            alist = config.get("tenant", "ima_allowlist")
 
-    # verify GPG signature if needed
-    if gpg_sig_file and gpg_key_file:
-        signing.verify_signature_from_file(gpg_key_file, al_path, gpg_sig_file, "allowlist")
+        # Purposefully die if path doesn't exist
+        with open(alist, "rb") as alist_f:
+            logger.debug("Loading allowlist from %s", alist)
+            alist_bytes = alist_f.read()
 
-    # Purposefully die if path doesn't exist
-    with open(al_path, "rb") as f:
-        logger.debug("Loading allowlist from %s", al_path)
-        alist_bytes = f.read()
-        sha256 = hashlib.sha256()
-        sha256.update(alist_bytes)
-        calculated_checksum = sha256.hexdigest()
-        alist_raw = alist_bytes.decode("utf-8")
-        logger.debug("Loaded allowlist from %s with checksum %s", al_path, calculated_checksum)
+    else:
+        raise Exception("Invalid allowlist provided")
+
+    # verify file signature if needed
+    if al_sig_file and al_key_file:
+        logger.debug(
+            "Loading signature (%s) and key (%s) and checking against allowlist (%s)", al_sig_file, al_key_file, alist
+        )
+        with open(al_key_file, "rb") as key_f:
+            al_key = key_f.read()
+        with open(al_sig_file, "rb") as sig_f:
+            al_sig = sig_f.read()
+        if signing.verify_signature(al_key, al_sig, alist_bytes):
+            logger.debug("Allowlist passed signature verification")
+        else:
+            raise Exception(
+                f"Allowlist signature verification failed comparing allowlist ({alist}) against sig_file ({al_sig_file})"
+            )
+
+    sha256 = hashlib.sha256()
+    sha256.update(alist_bytes)
+    calculated_checksum = sha256.hexdigest()
+    logger.debug("Loaded allowlist from %s with checksum %s", alist, calculated_checksum)
 
     if checksum:
         if checksum == calculated_checksum:
@@ -472,7 +492,53 @@ def read_allowlist(al_path=None, checksum="", gpg_sig_file=None, gpg_key_file=No
                 f"Checksum of allowlist does not match! Expected {checksum}, Calculated {calculated_checksum}"
             )
 
+    # Return artifacts as a JSON bundle, ready for sending
+    bundle = {
+        "ima_policy": base64.b64encode(alist_bytes).decode(),
+        "key": base64.b64encode(al_key).decode(),
+        "sig": base64.b64encode(al_sig).decode(),
+        "checksum": checksum,
+    }
+
+    return bundle
+
+
+class SignatureValidationError(Exception):
+    def __init__(self, message, code):
+        self.message = message
+        self.code = code
+        super().__init__(self.message)
+
+
+# Reads allowlist bundles sent to the verifier, validates them, and processes them. Returns an allowlist.
+def unbundle_ima_policy(allowlist_bundle, verify=True):
+    allowlist_raw = base64.b64decode(allowlist_bundle["ima_policy"])
+
+    # Verify allowlist signatures if set to enforce
+    if verify:
+        if not allowlist_bundle.get("sig"):
+            raise SignatureValidationError(
+                message="Verifier is enforcing signature validation, but no signature was provided.", code=405
+            )
+        if not allowlist_bundle.get("key"):
+            raise SignatureValidationError(
+                message="Verifier is enforcing signature validation, but no public key was provided.", code=405
+            )
+
+        allowlist_sig = base64.b64decode(allowlist_bundle.get("sig"))
+        allowlist_sig_key = base64.b64decode(allowlist_bundle.get("key"))
+
+        if not signing.verify_signature(allowlist_sig_key, allowlist_sig, allowlist_raw):
+            raise SignatureValidationError(message="Signature verification for allowlist failed!", code=401)
+        logger.info("Allowlist signature validation passed.")
+
+    return canonicalize_allowlist(allowlist_raw, allowlist_bundle["checksum"])
+
+
+# Manipulates allowlists into database-ready format.
+def canonicalize_allowlist(alist_bytes, checksum=""):
     # if the first non-whitespace character in the file is '{' treat it as the new JSON format
+    alist_raw = alist_bytes.decode("utf-8")
     p = re.compile(r"^\s*{")
     if p.match(alist_raw):
         logger.debug("Reading allow list as JSON format")
@@ -494,7 +560,7 @@ def read_allowlist(al_path=None, checksum="", gpg_sig_file=None, gpg_key_file=No
         # convert legacy format into new structured format
         logger.debug("Converting legacy allowlist format to JSON")
 
-        alist = copy.deepcopy(empty_allowlist)
+        alist = copy.deepcopy(EMPTY_ALLOWLIST)
         alist["meta"]["timestamp"] = str(datetime.datetime.now())
         alist["meta"]["generator"] = "keylime-legacy-format-upgrade"
         if checksum:
@@ -556,7 +622,7 @@ def main():
 
     al_data = read_allowlist(allowlist_path)
     excl_data = read_excllist(exclude_path)
-    lists = process_allowlists(al_data, excl_data)
+    lists = process_ima_policy(al_data, excl_data)
 
     measure_path = config.IMA_ML
     # measure_path='../scripts/ima/ascii_runtime_measurements_ima'
@@ -573,7 +639,7 @@ def main():
 
     al_data = read_allowlist("measure2allow.txt")
     excl_data = read_excllist(exclude_path)
-    lists2 = process_allowlists(al_data, excl_data)
+    lists2 = process_ima_policy(al_data, excl_data)
     process_measurement_list(AgentAttestState("2"), lines, lists2)
 
     print("done")
--- a/keylime/tenant.py
+++ b/keylime/tenant.py
@@ -186,19 +186,12 @@ class Tenant:
             # Auto-enable IMA (or-bit mask)
             self.tpm_policy["mask"] = hex(int(self.tpm_policy["mask"], 0) | (1 << config.IMA_PCR))
 
-            if isinstance(args["allowlist"], str):
-                if args["allowlist"] == "default":
-                    args["allowlist"] = config.get("tenant", "allowlist")
-                try:
-                    al_data = ima.read_allowlist(
-                        args["allowlist"], args["allowlist_checksum"], args["allowlist_sig"], args["allowlist_sig_key"]
-                    )
-                except Exception as ima_e:
-                    raise UserError(str(ima_e)) from ima_e
-            elif isinstance(args["allowlist"], list):
-                al_data = args["allowlist"]
-            else:
-                raise UserError("Invalid allowlist provided")
+            try:
+                al_data = ima.read_allowlist(
+                    args["allowlist"], args["allowlist_checksum"], args["allowlist_sig"], args["allowlist_sig_key"]
+                )
+            except Exception as ima_e:
+                raise UserError(str(ima_e)) from ima_e
 
         # Read command-line path string IMA exclude list
         excl_data = None
@@ -215,7 +208,8 @@ class Tenant:
         # Set up IMA
         if TPM_Utilities.check_mask(self.tpm_policy["mask"], config.IMA_PCR):
             # Process allowlists
-            self.allowlist = ima.process_allowlists(al_data, excl_data)
+            al_data["excllist"] = excl_data
+            self.allowlist = al_data
 
         # Read command-line path string TPM event log (measured boot) reference state
         mb_refstate_data = None
@@ -632,7 +626,7 @@ class Tenant:
             "cloudagent_ip": self.cv_cloudagent_ip,
             "cloudagent_port": self.agent_port,
             "tpm_policy": json.dumps(self.tpm_policy),
-            "allowlist": json.dumps(self.allowlist),
+            "ima_policy_bundle": json.dumps(self.allowlist),
             "mb_refstate": json.dumps(self.mb_refstate),
             "ima_sign_verification_keys": json.dumps(self.ima_sign_verification_keys),
             "metadata": json.dumps(self.metadata),
--- a/test/test_ima_verification.py
+++ b/test/test_ima_verification.py
@@ -3,6 +3,7 @@ SPDX-License-Identifier: Apache-2.0
 Copyright 2020 IBM Corporation
 """
 
+import base64
 import codecs
 import hashlib
 import os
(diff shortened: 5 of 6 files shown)


kkaarreell:
Test failures suggests that it is no longer possible to register an agent using previously added named allowlist.
```
::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
::   Add keylime agent using the named allowlist
::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::

:: [ 18:20:56 ] :: [  BEGIN   ] :: Running 'lime_keylime_tenant -t 127.0.0.1 -u d432fbb3-d2f1-4a97-9ef7-75bd81c00000 --allowlist-name list2 --exclude excludelist.txt -f allowlist.txt -c add'
Reading configuration from ['/etc/keylime.conf']
2022-05-18 18:20:57.389 - keylime.tpm - INFO - TPM2-TOOLS Version: 5.2
2022-05-18 18:20:57.392 - keylime.tenant - INFO - Setting up client TLS in /var/lib/keylime/cv_ca
2022-05-18 18:20:57.392 - keylime.registrar_client - WARNING - TLS is enabled.
2022-05-18 18:20:57.392 - keylime.registrar_client - INFO - Setting up client TLS...
2022-05-18 18:20:57.425 - keylime.tenant - INFO - TPM PCR Mask from policy is 0x408000
<Response [200]>
2022-05-18 18:20:57.547 - keylime.tpm - WARNING - PCR #0 in quote not found in tpm_policy, skipping.
2022-05-18 18:20:57.548 - keylime.tenant - WARNING - DANGER: EK cert checking is disabled and no additional checks on EKs have been specified with ek_check_script option. Keylime is not secure!!
2022-05-18 18:20:57.548 - keylime.tenant - INFO - Quote from 127.0.0.1 validated
2022-05-18 18:20:57.585 - keylime.tenant - ERROR - Response code 400: Allowlist data not provided with request!
2022-05-18 18:20:57.585 - keylime.tenant - ERROR - POST command response: 400 Unexpected response from Cloud Verifier: {"code": 400, "status": "Allowlist data not provided with request!", "results": {}}
:: [ 18:20:57 ] :: [   PASS   ] :: Command 'lime_keylime_tenant -t 127.0.0.1 -u d432fbb3-d2f1-4a97-9ef7-75bd81c00000 --allowlist-name list2 --exclude excludelist.txt -f allowlist.txt -c add' (Expected 0, got 0)
```
Also, keylime_tenant probably should not exit with 0 in such a situation but that is a different issue I guess.

THS-on:
@mbestavros thank you for this PR. I'll try to finish https://github.com/keylime/enhancements/pull/71, so that everything IMA related is actually part of that bundle. 

For the bundle format I would add also a `type` and an `alg` field as described in https://github.com/keylime/enhancements/pull/65#issuecomment-1085843137 and rename `allowlist` to `ima_policy'. This makes it more flexible for us in the future.

Having the key part of the bundle has some consequences (more below). Do you think we need some kind of trusted keyring on the verifier? 

> This enhances security by thwarting potential man-in-the-middle attacks - prior to this, Keylime's security model relied on a trusted tenant; however, if a malicious tenant sent a bad payload (ex. one that did not pass signature checks), the verifier would blindly accept it. The new verifier-side signature check makes this attack impossible.

Not really. In the current design the key is still in the bundle. An attacker can create their own malicious bundle and provide a their own valid key with it. I my opinion if an attacker has API access to the verifier it is mostly game over.

What is solves is that we can provide a known good IMA policy for e.g. Fedora 35 which can be downloaded from the internet signed with a key from Fedora and Keylime validates if that key is the same as we expect. Moving it into the verifier makes it usable also for people that don't use the tenant (like me).

mbestavros:
> Test failures suggests that it is no longer possible to register an agent using previously added named allowlist.
> 
> ```
> ::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
> ::   Add keylime agent using the named allowlist
> ::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
> 
> :: [ 18:20:56 ] :: [  BEGIN   ] :: Running 'lime_keylime_tenant -t 127.0.0.1 -u d432fbb3-d2f1-4a97-9ef7-75bd81c00000 --allowlist-name list2 --exclude excludelist.txt -f allowlist.txt -c add'
> Reading configuration from ['/etc/keylime.conf']
> 2022-05-18 18:20:57.389 - keylime.tpm - INFO - TPM2-TOOLS Version: 5.2
> 2022-05-18 18:20:57.392 - keylime.tenant - INFO - Setting up client TLS in /var/lib/keylime/cv_ca
> 2022-05-18 18:20:57.392 - keylime.registrar_client - WARNING - TLS is enabled.
> 2022-05-18 18:20:57.392 - keylime.registrar_client - INFO - Setting up client TLS...
> 2022-05-18 18:20:57.425 - keylime.tenant - INFO - TPM PCR Mask from policy is 0x408000
> <Response [200]>
> 2022-05-18 18:20:57.547 - keylime.tpm - WARNING - PCR #0 in quote not found in tpm_policy, skipping.
> 2022-05-18 18:20:57.548 - keylime.tenant - WARNING - DANGER: EK cert checking is disabled and no additional checks on EKs have been specified with ek_check_script option. Keylime is not secure!!
> 2022-05-18 18:20:57.548 - keylime.tenant - INFO - Quote from 127.0.0.1 validated
> 2022-05-18 18:20:57.585 - keylime.tenant - ERROR - Response code 400: Allowlist data not provided with request!
> 2022-05-18 18:20:57.585 - keylime.tenant - ERROR - POST command response: 400 Unexpected response from Cloud Verifier: {"code": 400, "status": "Allowlist data not provided with request!", "results": {}}
> :: [ 18:20:57 ] :: [   PASS   ] :: Command 'lime_keylime_tenant -t 127.0.0.1 -u d432fbb3-d2f1-4a97-9ef7-75bd81c00000 --allowlist-name list2 --exclude excludelist.txt -f allowlist.txt -c add' (Expected 0, got 0)
> ```
> 
> Also, keylime_tenant probably should not exit with 0 in such a situation but that is a different issue I guess.

Thanks for pointing the broken test out @kkaarreell!

I've got some interesting findings after digging into the tenant/verifier code - while it is true that my changes broke that test, I don't know if the test itself is valid at present given the existing codebase, because the allowlist CRD API (which is exposed through the tenant as the `addallowlist`, `showallowlist`, and `deleteallowlist` commands) isn't yet fully implemented on the verifier side. Specifically, this means that even though the endpoints (post, get, delete) exist and work from the tenant's perspective, they don't actually store the allowlist itself - only the `tpm_policy` that the tenant sends alongside the allowlist. (side note, this is why the other tests for `showallowlist` and `deleteallowlist` work - they are getting and deleting _something_ from the verifier's database, it's just not the full allowlist.)  

In short, adding an agent using named allowlists isn't actually supported at present to begin with (because named allowlists themselves aren't supported yet). So why did the test break - perhaps more importantly, why did it pass in the first place if it's not actually working?

It turns out that, at present, the tenant will just throw out the `allowlist_name` argument when executing the `add` command for initializing an agent and send an empty allowlist to the verifier. Prior to this PR, allowlists were processed entirely on tenant, so the verifier would happily take the empty allowlist and everything else would work - even though the allowlist it was using didn't actually have anything in it. That's changed with my PR - since allowlist processing now happens on the verifier, I made it so that it would fail if no allowlist was provided. This is where I could use the input of other Keylime contributors (@THS-on @mpeters) - **do we want the verifier to return an error if no allowlist data is provided, or should we assume an empty allowlist, as before?**

(Side note, my next task after this PR is completing the implementation of the allowlist CRD API - so hopefully this test isn't invalid for much longer 😄)

THS-on:
> MITM is still very much a concern, and this doesn't fix that.

Maybe I'm missing something, but I disagree with this. The connection is protected using a mTLS connection. You can only MITM a connection between the verifier if you already have a key and certificate trusted by the verifier. If you have that then you can do what ever you want at the verifier, so there is no reason for MITM a connection.

> I also want to clarify that I didn't intend the addition of "bundles" in this PR to be a part of the new IMA policy changes, more just a way to send additional artifacts alongside the allowlist in its current format (specifically, associated signing materials). My understanding was the IMA policy evolution would be a larger effort around the allowlist format itself, which would then render this obsolete - but in the meantime, this is a good stopgap measure.

Ok, I then propose the following steps:

* Get this merged with the temporary bundle format
* Implement the new IMA policy format with the temporary bundle format
* Move everything to a well specified bundle format

mbestavros:
> Maybe I'm missing something, but I disagree with MITM is a concern. The connection is protected using a mTLS connection. You can only MITM a connection between the verifier if you already have a key and certificate trusted by the verifier. If you have that then you can do what ever you want at the verifier, so there is no reason for MITM a connection.

@THS-on Right, apologies my wording was not precise. I meant that this change wouldn't _impact_ MITM security, for better or worse. It's a security improvement on the tenant side, but doesn't impact transit. My mistake 😅 

> Ok, I then propose the following steps:

> - Get this merged with the temporary bundle format
> - Implement the new IMA policy format with the temporary bundle format
> - Move everything to a well specified bundle format

Sounds like a plan to me!

mbestavros:
Some updates: I've made various fixes based on testing and adjusted `test_restful.py` with better test allowlist data, and tidied up lints and formatting. All tests pass now.

@mpeters @THS-on any additional feedback at this stage, or does this look good to go?

keylime/ima/ima.py
```diff
@@ -477,7 +498,54 @@ def read_allowlist(al_path=None, checksum="", gpg_sig_file=None, gpg_key_file=No
...
+        super().__init__(self.message)
+
+
+# Reads allowlist bundles sent to the verifier, validates them, and processes them.
+def unbundle_allowlist(allowlist_bundle, force_verification=False):
+    logger.info("allowlist bundle: %s", allowlist_bundle)
+    allowlist_raw = base64.b64decode(allowlist_bundle["ima_policy"])
+
+    # Verify allowlist signatures if set to enforce
+    if force_verification or config.getboolean("cloud_verifier", "enforce_allow_list_signatures", fallback=False):
```
THS-on, line 525: Move the config `enforce_allow_list_signatures` option to the caller. 
```suggestion
    if force_verification:
```

keylime/cloud_verifier_tornado.py
```diff
@@ -421,6 +422,21 @@ def post(self):
                     logger.warning("POST returning 400 response. Expected non zero content length.")
                 else:
                     json_body = json.loads(self.request.body)
+
+                    # Handle allowlists
+                    allowlist_bundle = json.loads(json_body.get("ima_policy"))
```
THS-on, line 427: This is an API change, but because it is only the backbend API incrementing the minor version is fine.

keylime/ima/ima.py
```diff
@@ -477,7 +498,54 @@ def read_allowlist(al_path=None, checksum="", gpg_sig_file=None, gpg_key_file=No
                 f"Checksum of allowlist does not match! Expected {checksum}, Calculated {calculated_checksum}"
             )
 
+    # Return artifacts as a JSON bundle, ready for sending
+    bundle = {
+        "ima_policy": base64.b64encode(alist_bytes).decode(),
+        "key": base64.b64encode(al_key).decode(),
+        "sig": base64.b64encode(al_sig).decode(),
+        "checksum": checksum,
```
THS-on, line 506: The checksum is never validated, right?

keylime/ima/ima.py
```diff
@@ -477,7 +498,54 @@ def read_allowlist(al_path=None, checksum="", gpg_sig_file=None, gpg_key_file=No
...
+    def __init__(self, message, code):
+        self.message = message
+        self.code = code
+        super().__init__(self.message)
+
+
+# Reads allowlist bundles sent to the verifier, validates them, and processes them.
+def unbundle_allowlist(allowlist_bundle, force_verification=False):
+    logger.info("allowlist bundle: %s", allowlist_bundle)
+    allowlist_raw = base64.b64decode(allowlist_bundle["ima_policy"])
```
THS-on, line 522: So the new API requires to always send a bundle with at least the `ima_policy` inside it. This needs then to be documented somewhere what is optional and what not.

THS-on (changes requested):
I still have to test it in more depth with my test setup, but besides some small changes this LGTM.

keylime/ima/ima.py
```diff
@@ -477,7 +498,54 @@ def read_allowlist(al_path=None, checksum="", gpg_sig_file=None, gpg_key_file=No
...
+
+        if not signing.verify_signature(allowlist_sig_key, allowlist_sig, allowlist_raw):
+            raise SignatureValidationError(message="Signature verification for allowlist failed!", code=401)
+        logger.info("Allowlist signature validation passed.")
+
+    return format_allowlist(allowlist_raw, allowlist_bundle["checksum"])
+
+
+# Manipulates allowlists into database-ready format.
+def format_allowlist(alist_bytes, checksum=""):
```
mpeters, line 546: Instead of `format_allowlist()` maybe something like `canonicalize_allowlist()`. The "format" is a bit too generic. It's about making sure they are all consistent when serializing into the DB so something like "canonicalize" is a better choice IMO.

keylime/tenant.py
```diff
@@ -639,7 +633,7 @@ def do_cv(self):
             "cloudagent_ip": self.cv_cloudagent_ip,
             "cloudagent_port": self.agent_port,
             "tpm_policy": json.dumps(self.tpm_policy),
-            "allowlist": json.dumps(self.allowlist),
```
mpeters, line 642: As Thore mentioned, this is an API change. Did you look at making it backwards compatible? If `allowlist` is provided treat it the same as before, if `ima_policy` is used do the new bundle processing? Not saying we have to have it that way, but if we're going to break backwards compat it should be for a good reason.

test/test_ima_verification.py
```diff
@@ -269,8 +270,12 @@ def test_read_allowlist(self):
         allowlist_bad_checksum = "4c143670836f96535d9e617359b4d87c59e89e633e2773b4d7feae97f561b3dc"
 
         # simple read, no fancy verification
-        al_data = ima.read_allowlist(allowlist_file)
-        self.assertIsNotNone(al_data, "AllowList data is present")
+        al_bundle = ima.read_allowlist(allowlist_file)
+        self.assertIsNotNone(al_bundle, "AllowList bundle data is present")
```
mpeters, line 274: Are you asserting the allowlist data is there or the ima_policy bundle is there? These assertion messages are actually neither of those.

test/test_restful.py
```diff
@@ -337,6 +338,9 @@ class TestRestful(unittest.TestCase):
     tpm_policy = {}
     metadata = {}
     allowlist = {}
+    excllist = {}
+    allowlist_bundle = {}
```
mpeters, line 342: `ima_bundle` ?

mpeters (changes requested):

keylime/ima/ima.py
mbestavros: @THS-on Not necessarily - this code is only called if a valid bundle was found in the first place: https://github.com/keylime/keylime/pull/994/files#diff-67033196f7dec303b3ea55793c6c13fb4b03a72a5515239475a871c3d638dfe4R428-R430

If not, the verifier will now populate the allowlist entry in the database with a default allowlist: https://github.com/keylime/keylime/pull/994/files#diff-67033196f7dec303b3ea55793c6c13fb4b03a72a5515239475a871c3d638dfe4R436-R438

I'm not sure if this is behavior we want, but I believe tests break if we don't put anything in the database for `allowlist`. Open to suggestions.

keylime/ima/ima.py
```diff
@@ -477,7 +498,54 @@ def read_allowlist(al_path=None, checksum="", gpg_sig_file=None, gpg_key_file=No
...
+class SignatureValidationError(Exception):
+    def __init__(self, message, code):
+        self.message = message
+        self.code = code
+        super().__init__(self.message)
+
+
+# Reads allowlist bundles sent to the verifier, validates them, and processes them.
+def unbundle_allowlist(allowlist_bundle, force_verification=False):
+    logger.info("allowlist bundle: %s", allowlist_bundle)
```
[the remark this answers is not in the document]
mbestavros: Oh, whoops. Did not mean to leave that in there.

keylime/ima/ima.py
mbestavros: The checksum value is also used during allowlist processing (which is now on the verifier side), hence why it's being sent along with the signing mateials. I suppose if we wanted we could also do a checksum check on the verifier side as well.

keylime/ima/ima.py
THS-on: What I meant is that I cannot post the old allowlist format to the verifier anymore I need to wrap it in a bundle when I want to use the IMA feature, which is fine.

> I'm not sure if this is behavior we want, but I believe tests break if we don't put anything in the database for allowlist. Open to suggestions.

This is fine for now.

keylime/ima/ima.py
```diff
@@ -425,49 +445,50 @@ def process_allowlists(allowlist, exclude):
...
-        calculated_checksum = sha256.hexdigest()
-        alist_raw = alist_bytes.decode("utf-8")
-        logger.debug("Loaded allowlist from %s with checksum %s", al_path, calculated_checksum)
+        # Purposefully die if path doesn't exist
+        with open(alist, "rb") as alist_f:
+            logger.debug("Loading allowlist from %s", alist)
+            alist_bytes = alist_f.read()
+
+    elif isinstance(alist, list):
+        alist_bytes = json.dumps(alist).encode()
```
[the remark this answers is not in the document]
mbestavros: So prior to this, signature checks only happened if an allowlist was provided as a file. if a user provided an allowlist in list format directly to the tenant and a signature/key, the tenant would just throw the signing materials out.

You raise a valid point, and I'm open to suggestions, but I don't expect legacy setups to have used it before anyway.

test/test_ima_verification.py
mbestavros: I'm asserting that the ima_policy bundle is there. `read_allowlist` now returns bundles directly, and a little further down, we also validate that the allowlist data is present within the bundle.

I can change the wording of these tests to `ima_policy` to be more consistent.

keylime/ima/ima.py
mbestavros: Agreed. We have a lot of these functions - `update_allowlist`, `process_allowlist`, `format_allowlist`...

keylime/tenant.py
mbestavros: Frankly, I didn't consider it, no. Though I think it would be good as a sanity check to lay out how the new bundle format works before introducing a backwards compatibility fix - I'd like to hear what you and Thore have to say about it, and make sure we think it's future-proof for the upcoming format shift for IMA policies.

I'll do that in a separate comment.

THS-on: I propose the following:
- Keep the `allowlist` option in the API. Treat all the data sent to it as before.
- Add the whole bundle feature to `ima_policy`, use it in the tenant only if a signature etc. is provided.
- Mark the `ima_policy` option as experimental until we have the new policy and bundle format in place.
- Once the new formats are implemented deprecate and remove `allowlist` from the API.

@mbestavros, @mpeters what do you think about that?

mbestavros:
@THS-on @mpeters You've both raised a good point about API levels and backwards compat. I'm happy to write some code to maintain backwards compat, but if we do I'd like a sanity check to make sure the bundle format I'm using makes sense for the foreseeable future (perhaps including the IMA policy format overhaul).

## Keylime at present

payload sent to the verifier:

```json
"allowlist": <processed allowlist>
```

stored in the verifier's database:

```json
"allowlist": <processed allowlist>
```

## After this PR

payload sent to the verifier: (note that the exclusion list needs to be sent separately because they aren't combined until allowlist processing, which now happens on the verifier after transit)

```json
"ima_policy": {
    "ima_policy": <raw allowlist>,
    "key": <pubkey>,
    "sig": <signature>,
    "checksum": <checksum>,
    "excllist": <exclusion list>   
}
```

stored in the verifier's database:

```json
"allowlist": <processed allowlist>
```

## The future

payload sent to the verifier:

```json
"ima_policy": {
    "ima_policy": <raw IMA policy in new format?>,
    "key": <pubkey>,
    "sig": <signature>,
    "checksum": <checksum>
}
```

stored in the verifier's database:

```json
"ima_policy": {
    "ima_policy": <raw IMA policy in new format?>,
    "key": <pubkey>,
    "sig": <signature>,
    "checksum": <checksum>
}
```

Does this make sense?

mpeters:
> @THS-on @mpeters You've both raised a good point about API levels and backwards compat. I'm happy to write some code to maintain backwards compat, but if we do I'd like a sanity check to make sure the bundle format I'm using makes sense for the foreseeable future (perhaps including the IMA policy format overhaul).

I'm not sure we can make guarantees that the format you're using for IMA policy will last for the "foreseeable future". These things evolve as people use them and that other enhancement still needs to be implemented. But that shouldn't really matter for this PR.

> ## After this PR
> payload sent to the verifier: (note that the exclusion list needs to be sent separately because they aren't combined until allowlist processing, which now happens on the verifier after transit)

Why is that? The old way combined the exclude into the allowlist and called them both the allowlist. Are you not bundling the exclude list in with the ima_policy bundle?

> ```json
> "ima_policy": {
>     "ima_policy": <raw allowlist>,
>     "key": <pubkey>,
>     "sig": <signature>,
>     "checksum": <checksum>,
>     "excllist": <exclusion list>   
> }
> ```
> 
> stored in the verifier's database:
> 
> ```json
> "allowlist": <processed allowlist>
> ```

> ## The future
> payload sent to the verifier:
> 
> ```json
> "ima_policy": {
>     "ima_policy": <raw IMA policy in new format?>,
>     "key": <pubkey>,
>     "sig": <signature>,
>     "checksum": <checksum>
> }
> ```
> stored in the verifier's database:
> 
> ```json
> "ima_policy": {
>     "ima_policy": <raw IMA policy in new format?>,
>     "key": <pubkey>,
>     "sig": <signature>,
>     "checksum": <checksum>
> }
> ```

This would include the exclude list in both of these correct?

mbestavros:
@THS-on Awesome, thanks for following up. I'll be pushing up another draft of the PR relatively soon with all the changes discussed here.

@mpeters to be clear, I'm not proposing this bundle structure as a first draft of the new IMA policy format - as you say, that's happening through the enhancements process.

I'm just proposing a format to carry the raw, user-provided artifact (currently the raw allowlist file plus exclusion list as separate entries, but in the future it would be the combined, new-format IMA policy as a single entry) from the tenant over to the verifier. I'm wondering if the format I've proposed makes sense for that specific purpose, and won't need to be changed once the IMA policy evolution happens.

keylime/cloud_verifier_tornado.py
```diff
@@ -421,6 +422,21 @@ def post(self):
...
+                    if allowlist_bundle:
+                        try:
+                            allowlist = ima.unbundle_allowlist(allowlist_bundle)
```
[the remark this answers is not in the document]
mbestavros: @mpeters I think I agree with `allowlist_bundle` being renamed to `ima_policy_bundle`, but the `allowlist` variable here really is just the allowlist, not a full IMA policy. On the next line, it's combined with the exclusion list found in `allowlist_bundle` (soon to be `ima_policy_bundle`) to create the final `ima_policy`.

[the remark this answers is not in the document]
mbestavros: @mpeters I've pushed an update with naming changes - giving everything another read, I think the `process_allowlists` function is the most confusing, since that takes in both an allowlist and exclusion list and spits out a combined "IMA policy". I've renamed it `process_ima_policy` accordingly.

Let me know how it reads now.

mbestavros:
@THS-on @mpeters I've pushed an update that should hopefully address most of the feedback we've discussed.

Most notably, I've tweaked a few things about the bundle transmission - the bundle is now sent under the `ima_policy_bundle` key in the payload, instead of just `ima_policy` before. This leaves the `ima_policy` key available in the future for when the IMA policy format evolution is complete and (potentially) the temporary bundle format would no longer be needed.

I've also tweaked how the CV handles allowlists. Instead of requiring the new bundle format, the verifier will maintain backwards-compatible behavior as before if an allowlist is sent using the legacy `allowlist` key in the POST payload.

I've also made some miscellaneous small fixes and naming updates based on your comments. Would appreciate another look whenever you both have the chance - thank you!

test/test_restful.py
mbestavros: I've made this and other naming changes in my latest push.

test/test_ima_verification.py
mbestavros: I've updated the assertion message in my latest push.

keylime/tenant.py
mbestavros: @THS-on @mpeters I've implemented backwards compatibility on the CV, so nothing should functionally change if a legacy API request is made to an updated verifier anymore.

Given how much refactoring was done to allowlist reading and processing in this PR (and to the associated tests) to defer allowlist processing until the verifier, I'd prefer it if we didn't have to maintain legacy tenant functionality. It would add some complexity managing tests and functionality between the different types of inputs. But I can still look into it if that's preferred.

keylime/ima/ima.py
mbestavros: I've changed this in my latest push.

keylime.conf
```diff
@@ -333,6 +333,11 @@ severity_policy = [{"event_id": ".*", "severity_label": "emergency"}]
 # Set to True to treat ToMToU entries as errors.
 tomtou_errors = False
 
+# If enabled, the Keylime verifier will require that a file signature and
+# associated key be sent alongside newly-created allowlists, and will perform a
+# signature check before storing them in the database.
+enforce_allow_list_signatures = False
```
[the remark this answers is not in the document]
mbestavros: Done in my latest push.

keylime/ima/ima.py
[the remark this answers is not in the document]
mbestavros: Fixed in my latest push.

keylime/ima/ima.py
mbestavros: @THS-on My latest push fixes this - it will now maintain backwards compatibility with the old allowlist format without a bundle.

keylime/ima/ima.py
THS-on: I would remove the `config.getboolean("cloud_verifier", "enforce_allow_list_signatures", fallback=False` and instead set the `force_verification` parameter when the `unbundle_allowlist(..)` is called e.g in `post(..)`.

keylime/cloud_verifier_tornado.py
```diff
@@ -421,6 +422,24 @@ def post(self):
                     logger.warning("POST returning 400 response. Expected non zero content length.")
                 else:
                     json_body = json.loads(self.request.body)
+
+                    # Handle allowlists/IMA policies
+                    ima_policy_bundle = json.loads(json_body.get("ima_policy_bundle"))
+                    if ima_policy_bundle:
+                        try:
+                            allowlist = ima.unbundle_allowlist(ima_policy_bundle)
```
mpeters, line 430: This is still confusingly named. `unbundle_allowlist` returns an allowlist... but not really. You extract the ima_policy from the allowlist but then used it as the key for agent_data's "allowlist". It seems to get things backward. An ima_policy would include an allowlist (along with an exclude list, signatures, etc, right?). So maybe `unbundle_allowlist` should be renamed `unbundle_ima_policy` and should return an ima_policy which you then pass to `process_ima_policy` to get an `allowlist` which you then pass as the `allowlist` key to the `agent_data`.

Or maybe I'm just more confused? This might seem nit-picky, but naming things is important to understand the flow. Something shouldn't be called an allowlist if it's more than an allowlist, etc. And it shouldn't be called an ima_policy if it's actually an allowlist.

keylime/cloud_verifier_tornado.py
```diff
@@ -421,6 +422,24 @@ def post(self):
...
+                            ima_policy = json.dumps(ima.process_ima_policy(allowlist, ima_policy_bundle["excllist"]))
+                        except ima.SignatureValidationError as e:
+                            web_util.echo_json_response(self, e.code, e.message)
+                            logger.warning(e.message)
+                            return
+                    else:
+                        # Read legacy allowlist if present in payload
+                        ima_policy = json_body.get("allowlist", None)
```
mpeters, line 438: Again, this variable should be called `allowlist` right?

keylime/ima/ima.py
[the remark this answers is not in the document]
mpeters: It's not about legacy compatibility, it's about signature matching. If someone provides a signature and a python structure, how do we know we can do the match correctly? Signatures should only really be about files that are bit-for-bit equivalent to what the user provided. Anything else is subject to some serialization errors/inconsistencies.

This is why file uploads should be b64 encoded. Then on the server side the b64 decoded bytes are treated as if they were a file read into memory. We can't do signature checking on python structures.

keylime/tenant.py
mpeters: I'm fine if we drop tenant verification of the signatures if it's now done at the verifier.

mpeters (changes requested):
It's getting close. I'm still not clear on the variable/function naming going on.

keylime/cloud_verifier_tornado.py
mbestavros: @mpeters I think the _most accurate_ naming would be if the `allowlist` key in `agent_data` changed to `ima_policy`. The "allowlist" sent over the wire prior to this PR and put into that key was actually closer to an IMA policy - it contained both `allowlist` and `excllist` subkeys. Hence why I named the variable I'm putting into `agent_data` as `ima_policy`. You're right about `unbundle_allowlist`, though. I'll change that.

My understanding is that if we wanted to change how things are named in `agent_data`, we'd need to change the database schema as well as update it where it's accessed. I'm not opposed to that, but IIRc it's not just a simple rename in that case.

mpeters: > @mpeters I think the _most accurate_ naming would be if the `allowlist` key in `agent_data` changed to `ima_policy`. 

You're right and that's probably where my confusion is coming from.

> My understanding is that if we wanted to change how things are named in `agent_data`, we'd need to change the database schema as well as update it where it's accessed. I'm not opposed to that, but IIRC it's not just a simple rename in that case (wouldn't it also involve a migration of some sort?).

We don't need to change that database field with this work. We can/should leave that to the larger IMA policy re-work.

mbestavros: @mpeters Sounds good. I should have made that detail clearer. I'll rename the `process_allowlist` function - otherwise, does the rest of the naming make sense given that context?

mpeters: yes

keylime/ima/ima.py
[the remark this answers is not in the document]
mbestavros: That makes sense to me. And we do b64 encode the allowlist for transmission, so in theory the verifier-side check should behave correctly regardless. I think we can just skip signature verification on the tenant side if input as a Python list.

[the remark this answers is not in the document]
mbestavros: Newest version implements this change.

mbestavros:
@mpeters @THS-on My latest push includes some minor fixes for naming as pointed out by Michael. Also, the tenant will no longer attempt to verify signing materials for allowlists provided as a list instead of a file (Michael pointed out this could be unreliable).

@THS-on have you had a chance to test this?

keylime/ima/ima.py
```diff
@@ -472,7 +492,53 @@ def read_allowlist(al_path=None, checksum="", gpg_sig_file=None, gpg_key_file=No
...
+        self.code = code
+        super().__init__(self.message)
+
+
+# Reads allowlist bundles sent to the verifier, validates them, and processes them. Returns an allowlist.
+def unbundle_ima_policy(allowlist_bundle, force_verification=False):
+    allowlist_raw = base64.b64decode(allowlist_bundle["ima_policy"])
+
+    # Verify allowlist signatures if set to enforce
+    if force_verification or config.getboolean("cloud_verifier", "require_allow_list_signatures", fallback=False):
```
THS-on, line 518: Can you remove `config.getboolean("cloud_verifier", "require_allow_list_signatures", fallback=False)` here and move to the `post()` function where this function is called in the verifier?

keylime/ima/ima.py
```diff
@@ -472,7 +492,53 @@ def read_allowlist(al_path=None, checksum="", gpg_sig_file=None, gpg_key_file=No
...
+
+class SignatureValidationError(Exception):
+    def __init__(self, message, code):
+        self.message = message
+        self.code = code
+        super().__init__(self.message)
+
+
+# Reads allowlist bundles sent to the verifier, validates them, and processes them. Returns an allowlist.
+def unbundle_ima_policy(allowlist_bundle, force_verification=False):
```
THS-on, line 514:
```suggestion
def unbundle_ima_policy(allowlist_bundle, force_verification=True):
```
I think this function should be secure by default

THS-on (commented):
Besides the two small code comments, I think this is ready. 

Thank you for implementing it.

keylime/ima/ima.py
mbestavros: @THS-on Ahhh, I understand what you were asking for now. I agree, that makes sense.

I've renamed the option to `verify` from `force_verification` - I think that is more consistent with how it's supposed to be used now - and moved the config check to the verifier code.

And thank you so much for your reviews! They have been very helpful in getting this ready 😄

THS-on (approved):
> Any additional feedback, or is this ready to merge?

From my side it should be ready. We will refine the code with the other proposals if necessary.  

Once this PR is merged can you also update the documentation (https://keylime-docs.readthedocs.io/en/latest/user_guide/runtime_ima.html), to better describe the signature feature?

mpeters (approved):

PR #999, opened 2022-05-20
ima_dm: enable support for dm_target_update events

THS-on:
This event is currently not supported in upstream kernels, but we already have
experimental support for it. Adding it to the list of valid names enables full
support for this event.


THS-on:
New attempt at adding upstream support for this can be found here: https://lore.kernel.org/linux-integrity/20220520142605.270625-1-public@thson.de/T/

PR #1005, opened 2022-05-25
remove deprecated distutuils and setDeamon() usage

THS-on:


aplanas:
Can you split the commit in two different ones?

The `setDaemon` seems OK to me, as the parameter is already present in 3.6

THS-on:
> Can you split the commit in two different ones?

Yes, done :)

keylime/ima/ima_dm.py
```diff
@@ -367,6 +366,22 @@ def invalid(self) -> Failure:
...
+def _strtobool(val):
+    """Convert a string representation of truth to true (1) or false (0).
+
+    True values are 'y', 'yes', 't', 'true', 'on', and '1'; false values
+    are 'n', 'no', 'f', 'false', 'off', and '0'.  Raises ValueError if
+    'val' is anything else.
+    """
+    val = val.lower()
+    if val in ("y", "yes", "t", "true", "on", "1"):
+        return 1
```
aplanas, line 378: Should return True?

keylime/ima/ima_dm.py
```diff
@@ -367,6 +366,22 @@ def invalid(self) -> Failure:
...
+    if val in ("n", "no", "f", "false", "off", "0"):
+        return 0
```
aplanas, line 380: `False`?

keylime/ima/ima_dm.py
THS-on: This is the same function as in the python standard lib, but I changed that because it does not make a difference is our case.

keylime/ima/ima_dm.py
```diff
@@ -367,6 +366,22 @@ def invalid(self) -> Failure:
         return failure
 
 
+def _strtobool(val: str) -> bool:
```
aplanas, line 369: maintain the original name now that it is not imported?

keylime/ima/ima_dm.py
```diff
@@ -385,7 +400,7 @@ def _check_attr(attr: Optional[Union[int, str, bool]], reference_value: Optional
 
         if isinstance(attr, str):
             try:
-                return reference_value == bool(strtobool(attr))
+                return reference_value == bool(_strtobool(attr))
```
aplanas, line 403: Now that we return a bool, we can skip the conversion

keylime/ima/ima_dm.py
THS-on: No because it is only used internally of this module.

keylime/ima/ima_dm.py
THS-on: Yes, I'll fix that.

keylime/ima/ima_dm.py
aplanas: agreed

THS-on: So `_strtobool` is fine right?

aplanas (approved):
The `daemon` parameter is pre-3.6 so we are OK

keylime/ima/ima_dm.py
aplanas: yes, makes sense

mpeters (approved):

maugustosilva (approved):

Issue #1007, opened 2022-05-26
Traceback when logging incorrect PCR bank value


kkaarreell:
Is your an issue a feature request? If so, please raise it as an [enhancement](https://github.com/keylime/enhancements)

# Environment
 * OS / version: Fedora 35
 * Processor architecture: x86_64
 * TPM Manufacturer: swtpm
 * Keylime version: current upstream 34a7165

# Description 
In measured boot scenario when PCR bank value does not match a traceback appears in verifier log due to an incorrect parameter order.

## Expected behavior vs. actual behavior 
No traceback in the verifier log

## Steps to reproduce problem
1. see above

# Relevant logs

```
keylime_verifier[16549]: --- Logging error ---
keylime_verifier[16549]: Traceback (most recent call last):
keylime_verifier[16549]:   File "/usr/lib64/python3.10/logging/__init__.py", line 1100, in emit
keylime_verifier[16549]:     msg = self.format(record)
keylime_verifier[16549]:   File "/usr/lib64/python3.10/logging/__init__.py", line 943, in format
keylime_verifier[16549]:     return fmt.format(record)
keylime_verifier[16549]:   File "/usr/lib64/python3.10/logging/__init__.py", line 678, in format
keylime_verifier[16549]:     record.message = record.getMessage()
keylime_verifier[16549]:   File "/usr/lib64/python3.10/logging/__init__.py", line 368, in getMessage
keylime_verifier[16549]:     msg = msg % self.args
keylime_verifier[16549]: TypeError: %d format: a real number is required, not str
keylime_verifier[16549]: Call stack:
keylime_verifier[16549]:   File "/usr/local/bin/keylime_verifier", line 33, in <module>
keylime_verifier[16549]:     sys.exit(load_entry_point('keylime==6.4.0', 'console_scripts', 'keylime_verifier')())
keylime_verifier[16549]:   File "/usr/local/lib/python3.10/site-packages/keylime-6.4.0-py3.10.egg/keylime/cmd/verifier.py", line 21, in main
keylime_verifier[16549]:     cloud_verifier_tornado.main()
keylime_verifier[16549]:   File "/usr/local/lib/python3.10/site-packages/keylime-6.4.0-py3.10.egg/keylime/cloud_verifier_tornado.py", line 1174, in main
keylime_verifier[16549]:     process.start()
keylime_verifier[16549]:   File "/usr/lib64/python3.10/multiprocessing/process.py", line 121, in start
[...]

PR #1008, opened 2022-05-26
Fix order of parameters in an error message

kkaarreell:
Fixes https://github.com/keylime/keylime/issues/1007


THS-on (approved):
Thanks for finding it. LGTM can be merged once the CI passes.

PR #1009, opened 2022-05-27
Add test /functional/measured-boot-swtpm-sanity to Packit CI plan

kkaarreell:


kkaarreell:
The newly added test is not supposed to run on CentOS Stream yet, it is failing there due to old tpm2-tools.
I have updated test metadata to exclude it.

aplanas (approved):

mpeters (approved):

PR #1014, opened 2022-06-03
RFC: Use python3-gpg instead of python3-gnupg

sergio-correia:
The former uses GPGME and is the recommended way of using GnuPG from
applications, by the GnuPG initiative, as it provides a better
documented API [1][2].

python-gpg is also already present in some distros, e.g. Fedora and
CentOS Stream, as it is a dependency of their package manager (dnf).

It is also available in other major distros such as Debian, Ubuntu,
OpenSUSE, so no disruptions are to be expected regarding packaging.

[1] https://gnupg.org/software/gpgme/index.html
[2] https://wiki.python.org/moin/GnuPrivacyGuard

### DIFF
--- a/keylime/signing.py
+++ b/keylime/signing.py
@@ -1,6 +1,6 @@
 import tempfile
 
-import gnupg
+import gpg
 from cryptography.exceptions import InvalidSignature
 from cryptography.hazmat.primitives import hashes
 from cryptography.hazmat.primitives.asymmetric import ec
@@ -47,19 +47,18 @@ def verify_signature(key, sig, file):
     try:
         # PGP
         if key_header == "-----BEGIN PGP PUBLIC KEY BLOCK-----":
-            gpg = gnupg.GPG()
-            logger.debug("Importing GPG key")
-            gpg_imported = gpg.import_keys(key.decode("utf-8"))
-            if gpg_imported.count == 1:  # pylint: disable=E1101
-                logger.debug("GPG key successfully imported")
-            else:
-                raise Exception("Unable to import GPG key")
+            verified = False
+            with tempfile.TemporaryDirectory() as gpg_homedir:
+                ctx = gpg.Context(home_dir=gpg_homedir)
+                try:
+                    logger.debug("Importing GPG key")
+                    result = ctx.key_import(key)
+                except Exception as e:
+                    raise Exception("Unable to import GPG key") from e
 
-            # The Python PGP library won't let you read a signature from memory, hence this hack.
-            with tempfile.NamedTemporaryFile() as temp_sig:
-                temp_sig.write(sig)
-                temp_sig.flush()
-                verified = gpg.verify_data(temp_sig.name, file)
+                if result is not None and hasattr(result, "considered") is True:
+                    _, result = ctx.verify(file, sig)
+                    verified = result.signatures[0].status == 0
 
         # OpenSSL
         elif key_header == "-----BEGIN PUBLIC KEY-----":
--- a/installer.sh
+++ b/installer.sh
@@ -58,7 +58,7 @@ case "$ID" in
         echo "${ID} selected."
         PACKAGE_MGR=$(command -v apt-get)
         PYTHON_PREIN="git patch"
-        PYTHON_DEPS="python3 python3-pip python3-dev python3-setuptools python3-zmq python3-tornado python3-cryptography python3-requests python3-psutil gcc g++ libssl-dev swig python3-yaml python3-gnupg python3-lark wget python3-alembic" 
+        PYTHON_DEPS="python3 python3-pip python3-dev python3-setuptools python3-zmq python3-tornado python3-cryptography python3-requests python3-psutil gcc g++ libssl-dev swig python3-yaml python3-gpg python3-lark wget python3-alembic"
         if [ "$(uname -m)" = "x86_64" ]; then
             PYTHON_DEPS+=" libefivar-dev"
         fi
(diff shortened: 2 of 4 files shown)


sergio-correia:
@aplanas: could you confirm this would be OK with OpenSUSE?

keylime/signing.py
```diff
@@ -49,19 +49,18 @@ def verify_signature(key, sig, file):
...
-            raise Exception("Unable to import GPG key")
-
-        # The Python PGP library won't let you read a signature from memory, hence this hack.
-        with tempfile.NamedTemporaryFile() as temp_sig:
-            temp_sig.write(sig)
-            temp_sig.flush()
-            verified = gpg.verify_data(temp_sig.name, file)
+        verified = False
+        with tempfile.TemporaryDirectory() as gpg_homedir:
+            ctx = gpg.Context(home_dir=gpg_homedir)
```
THS-on, line 54: Is there a reason we require the temp dir?
What happens if `home_dir=None` is set?

THS-on (changes requested):
The change makes sense. Can you also update the `requirements.txt`?

aplanas:
> @aplanas: could you confirm this would be OK with OpenSUSE?

Sergio, thanks a lot for the ping. I checked it and is present in openSUSE Tumbleweed, and in Leap 15.4 and SLE 15SP4, but not in olders.

This and the python-lark-parser dependency will represent a cutting point. But I guess that it is OK

aplanas:
Maybe this needs to be changed too: https://github.com/keylime/keylime/blob/master/requirements.txt#L13

keylime/signing.py
sergio-correia: If we don't set one, it will use the engine default, which would be e.g. `~/.gnupg`. Looking again, this seems to be what the current code is doing/using, so I guess we could remove this temp dir?

THS-on: No, the temp dir is then better in this case.

ueno:
This looks good, but I suspect it might be easier to just call out the `gpgv` program; that would avoid `gpg-agent` being launched in background:
```python
import subprocess
import tempfile
import base64
import io

def verify_signature(key, sig, file):
    with tempfile.NamedTemporaryFile() as keyring:
        # gpgv --keyring only accepts binary blob
        lines = key.read().decode('ascii').split("\n")
        base64.decode(io.BytesIO(''.join(lines[2:-2]).encode('ascii')), keyring)
        keyring.flush()
        status = subprocess.run([
            "gpgv", "--keyring", keyring.name, sig.name, file.name
        ])
        return status.returncode == 0

assert verify_signature(open("test-data/files/allowlist-pgp-key.pgp", "rb"),
                        open("test-data/files/allowlist-pgp-sig.sig", "rb"),
                        open("test-data/files/allowlist.json", "rb"))
```

sergio-correia:
> This looks good, but I suspect it might be easier to just call out the `gpgv` program; that would avoid extra package dependency but also `gpg-agent` being launched in background:

Good points. While I am not a big fan of launching processes in the background, in this particular case, this is what these python modules are already doing [1]. 

If we move to use `gpgv`, we would be using something that is only able to verify signatures, which is the exact use case here.

It would also look good on the dependency front -- one of the motivations to replace `python3-gnupg` was that it is not even packaged in distros like CentOS Stream 9. Then, the proposed `python3-gpg` module proved to be an issue with some older OpenSUSE versions as well as Ubuntu 18.04, which does not have the minimum required version. It looks like `gpgv` is available even in those distros, so this looks better, IMO.

What do you think, @THS-on, @aplanas, @mpeters?

[1] https://blogs.gnome.org/hughsie/2017/08/18/shipping-pkcs7-signed-metadata-and-firmware/

mpeters:
I agree, if the python module is launching another process in the background then we don't really have any downside of using `gpgv` ourselves and lot of upsides.

aplanas:
> What do you think, @THS-on, @aplanas, @mpeters?
> 
> [1] https://blogs.gnome.org/hughsie/2017/08/18/shipping-pkcs7-signed-metadata-and-firmware/

I read the link and it is not clear to me if there is a `gpg-agent` subprocess there or not. I have only some comments:

* python-gpgme is using ctypes and is not calling any app directly.
* the gpgv code seems simple, but there are missing bits: the full path needs to be set (to avoid some binary injection via PATH), better error handling (what to do if the binary is not there)
* as a nit, IMHO the opening of the sig and file seems not needed, and key seems better to be opened in the function itself.

In my naive PoV (that is not able to say if there is an agent launched there or not, and what is the impact for that), I prefer the Python library as a general better approach. But if the binary call code gets improved, that should be OK too.

sergio-correia:
Thanks, @aplanas. You are indeed right, `python-gpg` is using ctypes and not calling any app directly, which is also the approach I prefer; `python-gnupg` is the one issuing a call to `gpg` in background.

I would go with `python-gpg` as it has this additional improvement over the module we currently use, but apart from the issue you mentioned with older OpenSUSE, Ubuntu 18.04 packages version 1.10, and 1.11+ is required [1].

[1] https://gnupg-devel.gnupg.narkive.com/Hi4HECyN/hkp4py-hkp-hkps-bindings-for-python#post7

keylime/signing.py
```diff
@@ -30,7 +30,13 @@ def verify_signature_from_file(key_file, filename, sig_file, file_description):
     with open(filename, "rb") as file_f:
         file = file_f.read()
 
-    if verify_signature(key, sig, file):
+    verified = False
+    try:
+        verified = verify_signature(key, sig, file)
+    except Exception as e:
+        logger.debug("Unable to verify signature: %s", e)
```
aplanas, line 37: I think I commented in the other PR, but should be a warning here?

mpeters (approved):

PR #1017, opened 2022-06-07
Merge keylime-docs into main keylime repo

mpeters:
This commit merges the keylime-docs repo into the main keylime repo while preserving the history of both.


THS-on:
Do we want to fix the DCO before merging? Also do we want to keep the full history or do some cleanup before merging?

mpeters:
> Do we want to fix the DCO before merging? 

I started that but it led to conflicts (since a couple of filenames exist in both before the latest reorg in keylime-docs) and I thought rewriting history just to add sign-offs might be a little much. But I'm open to further exploring that.

> Also do we want to keep the full history or do some cleanup before merging?

What kind of cleanup did you have in mind? Just copying them in as-is is pretty simple, but I thought keeping the full history might be useful. But again, I'm willing to change my mind on that.

mpeters:
Abandoning this way of doing it. After dealing with the merge conflicts because there were several files that shared the same paths, I realized that the history is actually preserved in the old repo if we ever need it. So just making that one read-only and copying the files into the keylime proper repo is a better solution.

PR #1018, opened 2022-06-07
Adding files from keylime-docs into main repo

mpeters:
Doing this instead of a merge because it's cleaner and the old
history is still preserved in that old repo if we ever need it

Also moved the existing doc/ dir to docs/old since most of it
needs to be triaged


sergio-correia (approved):

THS-on (approved):
LGTM.

PR #1024, opened 2022-06-10
ci: Run lint with Python 3.6 as well

ueno:
As the minimal supported version of Python is 3.6, run the lint check
in the CI with that version to catch any usage of >= 3.7 features.

This utilizes the multiple Python interpreter support in Fedora:
https://developer.fedoraproject.org/tech/languages/python/multiple-pythons.html

Signed-off-by: Daiki Ueno <dueno@redhat.com>

### DIFF
--- a/keylime/revocation_notifier.py
+++ b/keylime/revocation_notifier.py
@@ -72,9 +72,9 @@ def stop_broker():
         logger.info("Stopping revocation notifier...")
         broker_proc.terminate()
         broker_proc.join(5)
-        if broker_proc.is_alive():
+        if broker_proc.is_alive() and sys.version_info >= (3, 7):
             logger.debug("Killing revocation notifier because it did not terminate after 5 seconds...")
-            broker_proc.kill()
+            broker_proc.kill()  # pylint: disable=E1101
 
 
 def notify(tosend):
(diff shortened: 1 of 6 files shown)


test/test_fs_util.py
```diff
@@ -8,7 +8,7 @@ class TestChDir(unittest.TestCase):
     @patch("keylime.fs_util.os.path.exists")
     @patch("keylime.fs_util.os.makedirs")
     @patch("keylime.fs_util.os.chdir")
-    def test_ch_dir_present(self, chdir_mock, makedirs_mock, exists_mock):
+    def test_ch_dir_present(self, chdir_mock, makedirs_mock, exists_mock):  # pylint: disable=R
```
aplanas, line 11: What refactor complain is producing this code? (just for curiosity)

ueno: It is [`old-no-self-use`](https://pylint.pycqa.org/en/latest/user_guide/messages/refactor/old-no-self-use.html); as the ID of this error has changed, I couldn't find a way to disable it in both Python 3.6 and 3.10.

keylime/revocation_notifier.py
```diff
@@ -72,7 +72,7 @@ def stop_broker():
         logger.info("Stopping revocation notifier...")
         broker_proc.terminate()
         broker_proc.join(5)
-        if broker_proc.is_alive():
+        if broker_proc.is_alive() and hasattr(broker_proc, "kill"):
```
aplanas, line 75: I would use

```python
if sys.version_info >= (3, 7):
    logger.debug("Killing revocation notifier because it did not terminate after 5 seconds...")
    broker_proc.kill()
```

aplanas (approved):

PR #1032, opened 2022-06-17
Full removal of the tenant WebApp

THS-on:
The WebApp was unmaintained and did not support more of the newer features
of Keylime. Therefore it gets removed. If there is enough interest a new
WebApp can be developed by directly using the REST APIs.


mpeters:
I'm in favor of this, but will give it a couple of days to percolate through the community just to make sure no one objects since it's a big removal. If someone does object, they will need to step up and take ownership to modernize it, otherwise it's just dead code hanging around again.

maugustosilva:
Neither myself nor my employer (IBM) has any interest or need in maintaining the Web Application code. I really do not see the case for a Web UI at this point, and I strongly believe that there are other, more scalable, more robust methods (e.g., "single pane of glass monitoring").

aplanas (approved):
Thanks a lot

THS-on:
@mpeters all major contributing parties (IBM, SUSE, Red Hat and Lernstick) are for the removal, so I can think we can merge this.

sergio-correia (approved):

PR #1037, opened 2022-06-20
Complete implementation of the Allowlists API

mbestavros:
This is a WIP, draft implementation of a completed Allowlists API. When complete, this will move allowlist storage to their own database, enable allowlist CRD operations through the Keylime tenant, and allow for agents to be provisioned using existing allowlists on the verifier.

This work is still WIP; some things may be rough around the edges. But it is functionally complete.

### DIFF
--- a/keylime/cloud_verifier_common.py
+++ b/keylime/cloud_verifier_common.py
@@ -28,7 +28,7 @@ def get_AgentAttestStates():
     return AgentAttestStates.get_instance()
 
 
-def process_quote_response(agent, json_response, agentAttestState) -> Failure:
+def process_quote_response(agent, ima_policy, json_response, agentAttestState) -> Failure:
     """Validates the response from the Cloud agent.
 
     This method invokes an Registrar Server call to register, and then check the quote.
@@ -156,7 +156,7 @@ def process_quote_response(agent, json_response, agentAttestState) -> Failure:
         agent["ak_tpm"],
         agent["tpm_policy"],
         ima_measurement_list,
-        agent["allowlist"],
+        ima_policy.ima_policy,
         algorithms.Hash(hash_alg),
         ima_keyrings,
         mb_measurement_list,
@@ -216,9 +216,9 @@ def prepare_get_quote(agent):
 
 
 def process_get_status(agent):
-    allowlist = json.loads(agent.allowlist)
-    if isinstance(allowlist, dict) and "allowlist" in allowlist:
-        al_len = len(allowlist["allowlist"])
+    allowlist_json = json.loads(agent.ima_policy.ima_policy)
+    if isinstance(allowlist_json, dict) and "allowlist" in allowlist_json:
+        al_len = len(allowlist_json["allowlist"])
     else:
         al_len = 0
 
@@ -290,12 +290,12 @@ def prepare_error(agent, msgtype="revocation", event=None):
     return tosend
 
 
-def validate_agent_data(agent_data):
+def validate_ima_policy_data(agent_data):
     if agent_data is None:
         return False, None
 
     # validate that the allowlist is proper JSON
-    lists = json.loads(agent_data["allowlist"])
+    lists = json.loads(agent_data)
 
     # Validate exlude list contains valid regular expressions
     is_valid, _, err_msg = validators.valid_exclude_list(lists.get("exclude"))
--- a/keylime/cloud_verifier_tornado.py
+++ b/keylime/cloud_verifier_tornado.py
@@ -10,6 +10,7 @@ from multiprocessing import Process
 import tornado.ioloop
 import tornado.web
 from sqlalchemy.exc import SQLAlchemyError
+from sqlalchemy.orm import joinedload
 from sqlalchemy.orm.exc import NoResultFound
 
 from keylime import api_version as keylime_api_version
@@ -81,7 +82,6 @@ def _from_db_obj(agent_db_obj):
         "tpm_policy",
         "meta_data",
         "mb_refstate",
-        "allowlist",
         "ima_sign_verification_keys",
         "revocation_key",
         "accept_tpm_hash_algs",
@@ -114,6 +114,7 @@ def _from_db_obj(agent_db_obj):
 def verifier_db_delete_agent(session, agent_id):
     get_AgentAttestStates().delete_by_agent_id(agent_id)
     session.query(VerfierMain).filter_by(agent_id=agent_id).delete()
+    session.query(VerifierAllowlist).filter_by(name=agent_id).delete()
     session.commit()
 
 
@@ -270,7 +271,12 @@ class AgentsHandler(BaseHandler):
                 return
 
             try:
-                agent = session.query(VerfierMain).filter_by(agent_id=agent_id).one_or_none()
+                agent = (
+                    session.query(VerfierMain)
+                    .options(joinedload(VerfierMain.ima_policy))
+                    .filter_by(agent_id=agent_id)
+                    .one_or_none()
+                )
             except SQLAlchemyError as e:
                 logger.error("SQLAlchemy Error: %s", e)
 
@@ -285,9 +291,14 @@ class AgentsHandler(BaseHandler):
                 agent_list = None
 
                 if ("verifier" in rest_params) and (rest_params["verifier"] != ""):
-                    agent_list = session.query(VerfierMain).filter_by(verifier_id=rest_params["verifier"]).all()
+                    agent_list = (
+                        session.query(VerfierMain)
+                        .options(joinedload(VerfierMain.ima_policy))
+                        .filter_by(verifier_id=rest_params["verifier"])
+                        .all()
+                    )
                 else:
-                    agent_list = session.query(VerfierMain).all()
+                    agent_list = session.query(VerfierMain).options(joinedload(VerfierMain.ima_policy)).all()
 
                 json_response = {}
                 for agent in agent_list:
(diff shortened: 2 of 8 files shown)


keylime/cloud_verifier_tornado.py
```diff
@@ -433,11 +443,58 @@ def post(self):
...
+                        # Don't allow owerwriting existing IMA policies
+                        if ima_policy and ima_policy_stored:
+                            web_util.echo_json_response(
+                                self, 409, f"Allowlist with name {ima_policy_name} already exists")
+                            logger.warning(
+                                "Allowlist with name %s already exists", ima_policy_name)
+                            return
+
+                    if not ima_policy_name and not ima_policy:
+                        logger.info("Allowlist data not provided with request! Using default allowlist.")
```
mpeters, line 467: Maybe we can take this opportunity to improve this message to let the user know the default allowlist is empty. Maybe something like "Allowlist data not provided with request! Using default empty allowlist"

keylime/migrations/versions/a72aec03d720_migrate_allowlists_to_separate_table.py
```diff
@@ -0,0 +1,77 @@
...
+def downgrade_registrar():
+    pass
+
+
+def upgrade_cloud_verifier():
+
+    # Migrate existing agent info to the allowlists table.
+    conn = op.get_bind()
+
+    res = conn.execute("SELECT agent_id, tpm_policy, allowlist FROM verifiermain")
```
mpeters, line 43: I thought we already have an `allowlist` table. Is data still populated in the `verifiermain` table for the allowlists? Is this a leftover from the partial allowlist API functionality from before?

keylime/cloud_verifier_tornado.py
```diff
@@ -1005,21 +1071,27 @@ async def process_agent(agent, new_operational_state, failure=Failure(Component.
         except SQLAlchemyError as e:
             logger.error("SQLAlchemy Error: %s", e)
 
+        # Load agent's IMA policy
+        try:
+            ima_policy = session.query(VerfierMain).options(joinedload(VerfierMain.ima_policy)).filter_by(agent_id=str(agent["agent_id"])).first().ima_policy
```
mpeters, line 1076: This isn't the most efficient way to query the `ima_policy` table. It's basically querying the `verifiermain` table and joining it to the `ima_policy` table and then you discard everything but the allowlist. It's more efficient to query the `ima_policy` table directly filtering on the `agent_id`.

mpeters: But if you already have the agent object loaded from the DB and you used `joinedload()` when you retrieved it, then it should already have the `ima_policy` available as an attribute on the `agent` object.

keylime/db/verifier_db.py
```diff
@@ -24,7 +25,8 @@ class VerfierMain(Base):
     public_key = Column(String(500))
     tpm_policy = Column(JSONPickleType(pickler=JSONPickler))
     meta_data = Column(String(200))
-    allowlist = Column(Text().with_variant(Text(429400000), "mysql"))
+    ima_policy = relationship("VerifierAllowlist", back_populates="agent", uselist=False)
```
mpeters, line 28: Let's take this opportunity to get the naming fixed/right. It's super weird to call something `ima_policy` in one place and `allowlist` in another. If the table is named `allowlist` then the foreign key should be as well.

So does it make more sense to have the table be `ima_policy` and the attribute be the `allowlist` or vice versa?

keylime/migrations/versions/4329e2d14944_associate_moved_allowlists_to_agents.py
```diff
@@ -0,0 +1,55 @@
+"""Associate moved allowlists to agents (2/2)
```
mpeters, line 1: Is there a reason you are using 2 migrations here? Typically a single change will use a single migration just so everything is done at once.

keylime/cloud_verifier_tornado.py
```diff
@@ -270,7 +270,7 @@ def get(self):
                 return
 
             try:
-                agent = session.query(VerfierMain).filter_by(agent_id=agent_id).one_or_none()
+                agent = session.query(VerfierMain).options(joinedload(VerfierMain.ima_policy)).filter_by(agent_id=agent_id).one_or_none()
```
mpeters, line 273: `joinedload()` should only be needed if we are going to reference the new `ima_policy` object that's attached to the `agent` object. If it's not used then it's extra overhead on the DB to do the join and marshall the data. Probably not a big deal when dealing with a single object, but it will definitely be noticeable for the bulk responses if there are a lot of entries.

mpeters (changes requested):

keylime/migrations/versions/a72aec03d720_migrate_allowlists_to_separate_table.py
mbestavros: To clarify: this is grabbing the `allowlist` **column** present within the `verifiermain` **table**. The migration then crafts it into an object that is then bulk inserted into the separate `allowlists` **table**.

keylime/cloud_verifier_tornado.py
mbestavros: I've removed this query in favor of piggybacking on the existing query for `stored_agent` a bit further up. Should be more efficient.

keylime/cloud_verifier_tornado.py
mbestavros: I've checked, and all the `joinedload` statements I added are justified - in that they eventually are used to retrieve the `ima_policy` from the agent.

In the bulk case, `process_get_status` (which does ask for `ima_policy`) is called on each agent within the bulk call, so it doesn't seem like we can avoid it for now (short of leaving out any allowlist/IMA policy metadata from the agent status call).

keylime/migrations/versions/4329e2d14944_associate_moved_allowlists_to_agents.py
mbestavros: I was running into issues attempting to populate columns (specifically the new foreign key column) that hadn't been committed yet. Adding a second migration to populate the new columns after they'd been committed by the first migration seems to work, so that's what I did. And vice versa for the downgrade step.

keylime/cloud_verifier_tornado.py
mbestavros: Done!

keylime/db/verifier_db.py
mbestavros: I think it might make the most sense to postpone any comprehensive work around making naming consistent for a different pull request. This one is pretty hefty as is.

mbestavros:
@mpeters @THS-on Tagging you both for review - it's finally ready! There's one test failure on the testing farm tests (seems like the quote is incorrect when adding an agent with a previously-uploaded allowlist) but given where the rest of the code is, I figure I should open it up for full review.

keylime/cloud_verifier_common.py
```diff
@@ -290,12 +290,12 @@ def prepare_error(agent, msgtype="revocation", event=None):
     return tosend
 
 
-def validate_agent_data(agent_data):
+def validate_allowlist_data(agent_data):
```
THS-on, line 293: Rename to `validate_ima_policy_data`?

We can add more sanity checks then in the future, because currently it only validates the regexes for the exclude list.

keylime/cloud_verifier_tornado.py
```diff
@@ -418,7 +429,17 @@ def post(self):
                     json_body = json.loads(self.request.body)
 
                     # Handle allowlists/IMA policies
+
+                    # How each pair of inputs should be handled:
+                    # - No name, no policy: use default empty policy
+                    # - Name, no policy: fetch existing policy from DB
+                    # - No name, policy: store policy using agent UUID as name
```
THS-on, line 436: Do we delete the policy after the agent is removed?

keylime/cloud_verifier_tornado.py
```diff
@@ -1005,21 +1085,24 @@ async def process_agent(agent, new_operational_state, failure=Failure(Component.
         except SQLAlchemyError as e:
             logger.error("SQLAlchemy Error: %s", e)
 
+        # Load agent's IMA policy
+        ima_policy = stored_agent.ima_policy
```
THS-on, line 1089: Policies can get quite big, maybe in the future we can cache the policies loaded in a process, but for now this is comparable to the old behavior.

THS-on (commented):
Only a few general comments. The overall design LGTM.


Can you squash your commits?

keylime/cloud_verifier_tornado.py
mbestavros: @THS-on I was wondering the best way to go about this... there are some traps we could fall into.

For example, the user could add a policy named `list1` then add an agent using that new list. But if we delete the agent, we probably don't want also delete the policy, right? That list was added separately from the agent, and the user will expect it to stick around after deleting an agent that happened to use it.

But if we add a policy alongside the agent directly, without a name, the implication is likely that the user expects the policy to go away along with the agent when deleted.

The strategy I settled on, for now, is deleting the policy if it's named with the UUID of the agent that will be deleted. This will allow named policies to stick around while also deleting policies that were created without a name.

keylime/cloud_verifier_common.py
mbestavros: Done!

mbestavros:
@mpeters @THS-on I've made some minor changes based on feedback, and squashed/signed my commits into one.

Also, the one failing Packit test was due to a misconfigured test setup, which should be fixed with this PR: https://github.com/RedHat-SP-Security/keylime-tests/pull/169

mpeters (approved):

Issue #1038, opened 2022-06-23
EACCESS on /dev/tpmrm0 due to missing supplementary group membership after dropping privs


berrange:
Is your an issue a feature request? If so, please raise it as an [enhancement](https://github.com/keylime/enhancements)

# Environment
 * OS / version: Fedora 36
 * Processor architecture: x86_64
 * TPM Manufacturer: swtpm 
 * Keylime version: 6.4.1

# Description 
Starting keylime_agent crashes immediately with
```
2022-06-23 08:24:11.879 - keylime.cloudagent - INFO - Dropped privileges to keylime:keylime
2022-06-23 08:24:11.895 - keylime.tpm - INFO - TPM2-TOOLS Version: 5.2
Traceback (most recent call last):
  File "/usr/bin/keylime_agent", line 33, in <module>
    sys.exit(load_entry_point('keylime==6.4.1', 'console_scripts', 'keylime_agent')())
  File "/usr/lib/python3.10/site-packages/keylime/cmd/agent.py", line 14, in main
    keylime_agent.main()
  File "/usr/lib/python3.10/site-packages/keylime/keylime_agent.py", line 685, in main
    (ekcert, ek_tpm, aik_tpm) = instance_tpm.tpm_init(
  File "/usr/lib/python3.10/site-packages/keylime/tpm/tpm_main.py", line 907, in tpm_init
    self.warn_emulator()
  File "/usr/lib/python3.10/site-packages/keylime/tpm/tpm_abstract.py", line 116, in warn_emulator
    if self.is_emulator():
  File "/usr/lib/python3.10/site-packages/keylime/tpm/tpm_main.py", line 903, in is_emulator
    return self.get_tpm_manufacturer() == "SW"
  File "/usr/lib/python3.10/site-packages/keylime/tpm/tpm_main.py", line 870, in get_tpm_manufacturer
    retDict = self.__run(["tpm2_getcap", "properties-fixed"])
  File "/usr/lib/python3.10/site-packages/keylime/tpm/tpm_main.py", line 233, in __run
    raise Exception(
[...]

PR #1039, opened 2022-06-23
keylime: initialize supplementary groups when dropping privileges

berrange:
On some distributions the keylime user's primary group may not have
permission to access /dev/tpmrm0 devices. For example, on Fedora,
/dev/tpmrm0 is owned root:tss, but keylime runs keylime:keylime,
with 'tss' as a supplementary group.

We need to load supplementary groups when dropping privileges,
otherwise we'll be unable to access the TPM device.

Resolves: #1038
Signed-off-by: Daniel P. Berrangé <berrange@redhat.com>


mpeters (approved):

mpeters:
Thanks for finding and fixing this

PR #1042, opened 2022-07-04
Fix broken URLs in README (Additional Reading)

mayaCostantini:
Fix URLs not found in project main README Additional Reading section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -411,18 +411,17 @@ Please, see [TESTING.md](TESTING.md) for details.
 
 ## Additional Reading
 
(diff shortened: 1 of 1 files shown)


aplanas (changes requested):
Thanks for the fix!

IMHO it should point to the `raw` version for the documents instead of the `blob` one

sergio-correia (approved):
Thanks for fixing these!

mpeters:
> Basically LGTM.
> 
> If anyone has a comment about keeping the `pptx` as raw (as it is already done in this PR), but pointing the `md`, or `pdf` to the blobs because github is able to render it, I would be OK too.

Yeah, that could be helpful to keep them formatted for new users. @mayaCostantini would you be willing to make that change to this PR?

aplanas (approved):

PR #1045, opened 2022-07-06
ima_ast: Allow whitespaces and multibyte characters in "n" and "n-ng"

ueno:
This series attempts to make the path parsing logic robuster to tolerate whitespace/multibyte characters in "n" and "n-ng" components, by doing:
- extract "sig" or "buf" first and then split the rest, so we do not need to reconstruct the "n" and "n-ng" components
- always split IMA entry with a limit (maxsplit=1)
- calculate the name length in bytes instead of characters


mpeters:
Is this to deal with paths that have whitespace or other special characters in them? Does IMA just puts those characters in the log without escaping?

test/test_ima_ast.py
```diff
@@ -22,7 +22,7 @@
     ),
     "ima-sig-missing": (
         ast.ImaSig,
-        "10 5426cf3031a43f5bfca183d79950698a95a728f6 ima-sig sha256:f1125b940480d20ad841d26d5ea253edc0704b5ec1548c891edf212cb1a9365e /lib/modules/5.4.48-openpower1/kernel/drivers/usb/common/usb-common.ko",
+        "10 5426cf3031a43f5bfca183d79950698a95a728f6 ima-sig sha256:f1125b940480d20ad841d26d5ea253edc0704b5ec1548c891edf212cb1a9365e /lib/modules/5.4.48-openpower1/kernel/drivers/usb/common/usb-common.ko ",
```
THS-on, line 25: What is the reason behind adding a space before the newline? 

I don't think the IMA log has one there.

ueno: IMA log does have it:
```console
$ tmt run -vvv prepare discover -h fmf -t 'configure_kernel_ima_module/ima_policy_signing' provision -h virtual -i Fedora-35 -c system execute login finish
...
[root@testcloud tree]# head -1 /sys/kernel/security/integrity/ima/ascii_runtime_measurements | od -x --endian=big
0000000 3130 2034 6633 3865 6638 6638 3262 6263
0000020 3261 3733 6632 3136 3963 3537 6666 3563
0000040 3736 6531 3463 6533 3533 6420 696d 612d
0000060 7369 6720 7368 6132 3536 3a30 3030 3030
0000100 3030 3030 3030 3030 3030 3030 3030 3030
*
0000160 3030 3030 3030 3030 3030 3020 626f 6f74
0000200 5f61 6767 7265 6761 7465 200a
0000214
```
as shown above, there is a space followed by a newline (`200a`) at the end.

The [code](https://elixir.bootlin.com/linux/v5.12/source/security/integrity/ima/ima_fs.c#L244) also backs up the behavior.

THS-on: Hmm on my Debian VM with a 5.18 kernel I see this:
```
root@debian-11:~# head -1 /sys/kernel/security/integrity/ima/ascii_runtime_measurements | od -x --endian=big
0000000 3130 2036 3462 6533 6334 3831 3830 3632
0000020 3766 3831 6133 3531 3363 6139 3336 6335
0000040 3566 3466 6337 3566 3563 6320 696d 612d
0000060 6e67 2073 6861 3235 363a 6462 3934 3931
0000100 3161 3239 3163 3265 3564 6563 3531 6630
0000120 3833 3766 3663 3936 3332 3538 3238 6361
0000140 6462 3137 3561 3963 6266 6665 3337 3933
0000160 3535 3339 3733 3634 6231 2062 6f6f 745f
0000200 6167 6772 6567 6174 650a
0000212
root@debian-11:~# cat /sys/kernel/security/ima/ascii_runtime_measurements
10 64be3c48180627f81a3513ca936c55f4fc75f5cc ima-ng sha256:db94911a291c2e5dec51f0837f6c96325828cadb175a9cbffe379355397364b1 boot_aggregate
```

ueno: That's because the template is "ima-ng", which consists of 2 fields, so there shouldn't be a second delimiter.
On the other hand, the above change is for "ima-sig", which consists of 3 fields, so there should be 2 delimiters (even if the third field (sig) is empty).

THS-on: You are right! The code adds a space even if the next field is empty. 

I missed that `ima-ng` and `ima-sig` behave differently. Thanks for pointing that out.

keylime/cmd/ima_emulator_adapter.py
```diff
@@ -24,7 +24,7 @@ def measure_list(file_path, position, ima_hash_alg, pcr_hash_alg, search_val=Non
             search_val = codecs.decode(search_val.encode("utf-8"), "hex")
 
         for line in lines:
-            line = line.strip()
+            line = line.strip("\n")
```
THS-on, line 27: Maybe add a comment here why we not strip all whitespace.

THS-on (approved):
The changes LGTM. Thank you for improving the parser.

If you have the time maybe add some `ima-sig` entries to the tests that are missing the space (which then should cause the parser to fail).

Issue #73, opened 2019-03-08
Update README with new meeting details


lukehinds:
It has been decided that we will meet every Wednesday at 15:00 UTC on the gitter keylime channel. 

The readme should be updated, so that new contributors can find the information.

PR #1060, opened 2022-07-26
Adds a per-agent counter for "successfull attestations" on Keylime.

maugustosilva:
This counter, "attestations" is now available via API calls (e.g.,
`keylime_tenant -c status`) and it is stored on the database to preserve
state in case the `verifier` is restarted. Given the availability of
this information, the old "first_verified" attribute is now removed.
This PR is the first one on a series of PRs to implement "Durable
Attestation", also known as "Offline Attestation" (enhancement #73).

### DIFF
--- a/keylime/cloud_verifier_common.py
+++ b/keylime/cloud_verifier_common.py
@@ -166,9 +166,7 @@ def process_quote_response(agent, json_response, agentAttestState) -> Failure:
     failure.merge(quote_validation_failure)
 
     if not failure:
-        # set a flag so that we know that the agent was verified once.
-        # we only issue notifications for agents that were at some point good
-        agent["first_verified"] = True
+        agent["attestation_count"] += 1
 
         # has public key changed? if so, clear out b64_encrypted_V, it is no longer valid
         if received_public_key != agent.get("public_key", ""):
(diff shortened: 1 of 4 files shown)


keylime/cloud_verifier_common.py
```diff
@@ -166,9 +166,9 @@ def process_quote_response(agent, json_response, agentAttestState) -> Failure:
     failure.merge(quote_validation_failure)
 
     if not failure:
-        # set a flag so that we know that the agent was verified once.
-        # we only issue notifications for agents that were at some point good
-        agent["first_verified"] = True
+        agent["attestations"] += 1
```
mpeters, line 169: Overall really like the change, but I wonder about the field name. I think `attestation_count` would be a better name since it won't be mistaken for being a list of attestation objects or something similar.

maugustosilva: fair enough, will make the change

aplanas (approved):
LGTM, but do we need to update the API version?

THS-on:
@aplanas it is not required because only the backend is affected, but we can bump the API to 2.2 to indicate the new field.

@maugustosilva can you squash your commits?

mpeters (approved):

PR #1061, opened 2022-07-27
Extracts (and uses) TPM clock information from the quote

maugustosilva:
With this chnage, the `verifier` will now use the `tpm2_print` command
to extract clock information from the quote. It will then uses this
information to make decisions about the attestation of the agent (i.e.,
the quote timestamp has to monotonically grow in a TPM which wasn't
restarted/reset). In order to make this comparison the clock information
from the previous quote is stored on the database and then both
timestamps are compared.

While this is not strictly needed for the current implementation of the
"online attestation", it will add yet another attestation check to it,
and is crucial to  "Durable Attestation", also known as "Offline
Attestation" (enhancement #73).

### DIFF
--- a/keylime/agentstates.py
+++ b/keylime/agentstates.py
@@ -1,8 +1,46 @@
+import sys
 import threading
 
 from keylime.common.algorithms import Hash
 from keylime.ima.ast import get_FF_HASH, get_START_HASH
 from keylime.ima.file_signatures import ImaKeyrings
+
+if sys.version_info >= (3, 7):
+    from dataclasses import dataclass
+else:
+    from keylime.backport_dataclasses import dataclass
+
+
+@dataclass
+class TPMClockInfo:
+    clock: int
+    resetcount: int
+    restartcount: int
+    safe: int
+
+    @classmethod
+    def from_dict(cls, data) -> "TPMClockInfo":
+        dclki = {}
+        if "clockInfo" in data:
+            dclki = data["clockInfo"]
+
+        if "clock" in data:
+            dclki = data
+
+        return cls(
+            clock=dclki.get("clock", 0),
+            resetcount=dclki.get("resetCount", 0),
+            restartcount=dclki.get("restartCount", 0),
+            safe=dclki.get("safe", 1),
+        )
+
+    def to_dict(self):
+        data = {}
+        data["clock"] = self.clock
+        data["resetCount"] = self.resetcount
+        data["restartCount"] = self.restartcount
+        data["safe"] = self.safe
+        return data
 
 
 class TPMState:
@@ -56,6 +94,7 @@ class AgentAttestState:
         self.agent_id = agent_id
         self.next_ima_ml_entry = 0
         self.set_boottime(0)
+        self.tpm_clockinfo = TPMClockInfo(clock=0, resetcount=0, restartcount=0, safe=1)
 
         self.tpm_state = TPMState()
         self.ima_pcrs = set()
@@ -122,6 +161,14 @@ class AgentAttestState:
     def set_boottime(self, boottime):
         """Set the boottime of the system"""
         self.boottime = boottime
+
+    def get_tpm_clockinfo(self):
+        """Return the clock info extracted from a TPM quote"""
+        return self.tpm_clockinfo
+
+    def set_tpm_clockinfo(self, tpm_clockinfo):
+        """Set the clock info with information extracted from a TPM quote"""
+        self.tpm_clockinfo = tpm_clockinfo
 
     def is_expected_boottime(self, boottime):
         """Check whether the given boottime is the expected boottime"""
--- a/keylime/cloud_verifier_tornado.py
+++ b/keylime/cloud_verifier_tornado.py
@@ -99,6 +99,7 @@ def _from_db_obj(agent_db_obj):
         "mtls_cert",
         "ak_tpm",
         "attestation_count",
+        "tpm_clockinfo",
     ]
     agent_dict = {}
     for field in fields:
--- a/keylime/tpm/tpm_main.py
+++ b/keylime/tpm/tpm_main.py
@@ -20,6 +20,7 @@ from cryptography.hazmat.primitives.asymmetric import padding
 from packaging.version import Version
 
 from keylime import cmd_exec, config, keylime_logging, secure_mount, tpm_ek_ca
+from keylime.agentstates import TPMClockInfo
 from keylime.common import algorithms, retry
 from keylime.failure import Component, Failure
 from keylime.tpm import tpm2_objects, tpm_abstract
@@ -1051,6 +1052,64 @@ class tpm(tpm_abstract.AbstractTPM):
         retDict = self.__run(command, lock=False)
         return retDict
 
+    def __tpm2_printquote(self, quoteFile):
+        command = ["tpm2_print", "-t", "TPMS_ATTEST", quoteFile]
+        retDict = self.__run(command, lock=False)
+        return retDict
+
+    def _tpm2_printquote(self, quote, compressed):
+        """Get TPM timestamp info from quote
+        :param quote: quote data in the format 'r<b64-compressed-quoteblob>:<b64-compressed-sigblob>:<b64-compressed-pcrblob>
+        :param compressed: if the quote data is compressed with zlib or not
+        :returns: Returns the 'retout' from running tpm2_print and True in case of success, None and False in case of error.
+        This function throws an Exception on bad input.
+        """
+
+        if quote[0] != "r":
+            raise Exception(f"Invalid quote type {quote[0]}")
+        quote = quote[1:]
+
+        quote_tokens = quote.split(":")
+        if len(quote_tokens) < 3:
+            raise Exception(f"Quote is not compound! {quote}")
+
+        quoteblob = base64.b64decode(quote_tokens[0])
+
+        if compressed:
+            logger.warning("Decompressing quote data which is unsafe!")
+            quoteblob = zlib.decompress(quoteblob)
+
+        qfd = -1
+        quoteFile = None
+
+        try:
+            # write out quote
+            qfd, qtemp = tempfile.mkstemp()
+            with open(qtemp, "wb") as quoteFile:
+                quoteFile.write(quoteblob)
+
+            retDict = self.__tpm2_printquote(quoteFile.name)
+            retout = retDict["retout"]
+            reterr = retDict["reterr"]
+            code = retDict["code"]
+        except Exception as e:
+            logger.error("Error printing quote: %s", str(e))
+            logger.exception(e)
+            return None, False
+        finally:
+            for fd in [qfd]:
+                if fd >= 0:
+                    os.close(fd)
+            for fi in [quoteFile]:
+                if fi is not None:
+                    os.remove(fi.name)
+
+        if len(retout) < 1 or code != tpm_abstract.AbstractTPM.EXIT_SUCESS:
+            logger.error("Failed to print quote info, output: %s", reterr)
+            return None, False
+
+        return retout, True
+
     def _tpm2_checkquote(self, aikTpmFromRegistrar, quote, nonce, hash_alg, compressed):
         """Write the files from data returned from tpm2_quote for running tpm2_checkquote
         :param aikTpmFromRegistrar: AIK used to generate the quote and is needed for verifying it now.
@@ -1157,13 +1216,26 @@ class tpm(tpm_abstract.AbstractTPM):
         if hash_alg is None:
             hash_alg = self.defaults["hash"]
 
+        # First and foremost, the quote needs to be validated
         retout, success = self._tpm2_checkquote(aikTpmFromRegistrar, quote, nonce, hash_alg, compressed)
         if not success:
             # If the quote validation fails we will skip all other steps therefore this failure is irrecoverable.
             failure.add_event(
-                "quote_validation", {"message": "Quote validation using tpm2-tools", "data": retout}, False
+                "quote_validation", {"message": "Quote data validation using tpm2-tools", "data": retout}, False
             )
             return failure
+
+        # Only after validating the quote, the TPM clock information can be extracted from it.
+        clock_failure, current_clock_info = self.check_quote_timing(agentAttestState.get_tpm_clockinfo(), quote)
+        if clock_failure:
+            failure.add_event(
+                "quote_validation",
+                {"message": "Validation of clockinfo from quote using tpm2-tools", "data": clock_failure},
+                False,
+            )
+            return failure
+        if current_clock_info:
+            agentAttestState.set_tpm_clockinfo(current_clock_info)
 
         pcrs = []
         jsonout = config.yaml_to_dict(retout, logger=logger)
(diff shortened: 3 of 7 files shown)


keylime/tpm/tpm_main.py
```diff
@@ -1051,6 +1051,64 @@ def __tpm2_checkquote(self, pubaik, nonce, quoteFile, sigFile, pcrFile, hash_alg
...
+        if compressed:
+            logger.warning("Decompressing quote data which is unsafe!")
+            quoteblob = zlib.decompress(quoteblob)
+
+        qfd = -1
+        quoteFile = None
+
+        try:
+            # write out quote
+            qfd, qtemp = tempfile.mkstemp()
```
aplanas, line 1086: You can avoid all the dancing (getting the fd, the name, the closing and the removal in the "finally" section) using `NamedTemporaryFile` (https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile):

```python
# Not tested, so 100% wrong
with tempfile.NamedTemporaryFile() as fd:
    fd.write(quoteblob)
    retDict = self.__tpm2_printquote(fd.name)
    retout = retDict["retout"]
    reterr = retDict["reterr"]
    code = retDict["code"]
except Exception as e:
    logger.error("Error printing quote: %s", str(e))
    logger.exception(e)
    return None, False
# No finally here
```

keylime/tpm/tpm_main.py
```diff
@@ -1051,6 +1051,64 @@ def __tpm2_checkquote(self, pubaik, nonce, quoteFile, sigFile, pcrFile, hash_alg
...
+                quoteFile.write(quoteblob)
+
+            retDict = self.__tpm2_printquote(quoteFile.name)
+            retout = retDict["retout"]
+            reterr = retDict["reterr"]
+            code = retDict["code"]
+        except Exception as e:
+            logger.error("Error printing quote: %s", str(e))
+            logger.exception(e)
+            return None, False
```
aplanas, line 1097: We should avoid Go error idioms here. Return `retout` if all is OK, and return None if there is something wrong.  If is too wrong maybe even re-raise the exception and manage it in upper levels.

keylime/tpm/tpm_main.py
```diff
@@ -1200,6 +1270,44 @@ def check_quote(
...
 
+    def check_quote_timing(self, previous_clockinfo, quote):
+        # Sanity check quote clock information
+
+        current_clockinfo = {}
+
+        if previous_clockinfo:
+            retout, success = self._tpm2_printquote(quote, False)
+            if not success:
+                return False, "tpm2_print failed with " + str(retout), current_clockinfo
```
aplanas, line 1281: As commented IMHO `success` should be dropped. Also in the current code, if `success` is `False`, `retout` is `None`, so the current message will not help.

keylime/tpm/tpm_main.py
```diff
@@ -1200,6 +1270,44 @@ def check_quote(
...
+            current_clockinfo = config.yaml_to_dict(retout, logger=logger)["clockInfo"]
+
+            if "clock" in previous_clockinfo and "clock" in current_clockinfo:
+                resetdiff = current_clockinfo["resetCount"] - previous_clockinfo["resetCount"]
+                restartdiff = current_clockinfo["restartCount"] - previous_clockinfo["restartCount"]
+
+                if resetdiff < 0:
+                    return "resetCount value decreased on TPM between two consecutive quotes", False, current_clockinfo
+
+                if restartdiff < 0:
```
aplanas, line 1292: This is very interesting. Can this happen?

keylime/tpm/tpm_main.py
```diff
@@ -1200,6 +1270,44 @@ def check_quote(
...
+            if not success:
+                return False, "tpm2_print failed with " + str(retout), current_clockinfo
+
+            current_clockinfo = config.yaml_to_dict(retout, logger=logger)["clockInfo"]
+
+            if "clock" in previous_clockinfo and "clock" in current_clockinfo:
+                resetdiff = current_clockinfo["resetCount"] - previous_clockinfo["resetCount"]
+                restartdiff = current_clockinfo["restartCount"] - previous_clockinfo["restartCount"]
+
+                if resetdiff < 0:
```
aplanas, line 1289: Should be OK this case if `resetCount` increased?

keylime/tpm/tpm_main.py
THS-on: You need to add `fd.seek(0)` after the write, because otherwise the content might not be not actually written into the file.

keylime/tpm/tpm_main.py
```diff
@@ -1157,11 +1216,22 @@ def check_quote(
         if hash_alg is None:
             hash_alg = self.defaults["hash"]
 
+        msg, success, clock_info = self.check_quote_timing(clock_info, quote)
```
THS-on, line 1219: The time checks needs to be after the quote validation because `tpm2_print` does not validate the data structure.
Can you add also a comment that states that, so the order does not get changed accidentally?

keylime/tpm/tpm_main.py
```diff
@@ -1200,6 +1270,44 @@ def check_quote(
...
+                    return (
+                        "restartCount value decreased on TPM between two consecutive quotes",
+                        False,
+                        current_clockinfo,
+                    )
+
+                if not (resetdiff and restartdiff):
+                    clockdiff = current_clockinfo["clock"] - previous_clockinfo["clock"]
+
+                    if clockdiff <= 0:
```
THS-on, line 1302: This only checks if the time is strictly increasing. Do we want to check against the time actually passed? Maybe with an configurable error range?

THS-on (changes requested):

keylime/tpm/tpm_main.py
maugustosilva: Again, same as above. I agree with you both, but for this PR, I have tried to follow the pattern already on the code for other commands. Not opposed to change it for **all** commands in a subsequent PR, dedicated to "code improvement/technical debt reduction"

keylime/tpm/tpm_main.py
maugustosilva: I imagine that this would happen only in case of a "malformed" or "corrupted" quote.

keylime/tpm/tpm_main.py
maugustosilva: Same case as above, I imagine that only a purposefully corrupted quote would result in a case like the above.

keylime/tpm/tpm_main.py
maugustosilva: Fair enough, will modify accordingly

keylime/tpm/tpm_main.py
```diff
@@ -1200,6 +1270,44 @@ def check_quote(
             hash_alg,
         )
 
+    def check_quote_timing(self, previous_clockinfo, quote):
+        # Sanity check quote clock information
```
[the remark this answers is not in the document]
maugustosilva: You mean, making sure that the `safe` attribute on the JSON object is set to `1`?

keylime/tpm/tpm_main.py
maugustosilva: Very good point, I will change accordingly

keylime/tpm/tpm_main.py
maugustosilva: Hmmm... good question. I can see a good case for an extra parameter on `keylime.conf` but I propose we leave it for a future PR. In the meantime it gives us a bit of time to gain operational experience for this code at scale and in production.

keylime/tpm/tpm_main.py
THS-on: If you have the time can you do it the over way around e.g. move the current code to NamedTemporaryFile and then add your PR?

keylime/tpm/tpm_main.py
THS-on: Maybe add a comment what this checks and also for the other sanity checks.

keylime/tpm/tpm_main.py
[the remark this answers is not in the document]
THS-on: Yes I exactly mean that.

THS-on:
@maugustosilva for the clock validation state maybe use a dataclass class with from and to json methods implemented.
This gives more type safety and I like to eliminate the usage of dicts for holding complex objects where possible.

keylime/tpm/tpm_main.py
maugustosilva: Fair, will do.

keylime/tpm/tpm_main.py
mpeters: I'm also ok if we open a ticket to track this tech debt reduction and letting this PR move forward.

keylime/tpm/tpm_main.py
mpeters: I'm also ok if we open a ticket to track this tech debt reduction and letting this PR move forward.

mpeters (approved):

aplanas (approved):
I have questions about how to manage resetCount and restartCount, but if I read the code correctly, this logic is more strict that the alternatives, so LGTM.

keylime/tpm/tpm_main.py
maugustosilva: I just created an issue to capture it https://github.com/keylime/keylime/issues/1067

keylime/agentstates.py
```diff
@@ -56,6 +94,7 @@ def __init__(self, agent_id):
         self.agent_id = agent_id
         self.next_ima_ml_entry = 0
         self.set_boottime(0)
+        self.set_tpm_clockinfo({"start": True})
```
THS-on, line 97: Is there a reason to not use a `TPMClockInfo` object here?

keylime/cloud_verifier_tornado.py
```diff
@@ -472,6 +472,7 @@ def post(self):
                     agent_data["verifier_ip"] = config.get("cloud_verifier", "cloudverifier_ip")
                     agent_data["verifier_port"] = config.get("cloud_verifier", "cloudverifier_port")
                     agent_data["attestation_count"] = 0
+                    agent_data["tpm_clockinfo"] = {"start": True}
```
THS-on, line 475: The agent_data needs probably to be JSON for the DB storage, but a nicer way is to either specify a Pickler for `TPMClockInfo` (just calling `to_json()`) or to explicitly construct a `TPMClockInfo` first and then calling `to_json()`. It is currently not clear when there is a dict or a `TPMClockInfo` object.

keylime/tpm/tpm_main.py
```diff
@@ -1157,13 +1217,26 @@ def check_quote(
...
         if not success:
             # If the quote validation fails we will skip all other steps therefore this failure is irrecoverable.
             failure.add_event(
-                "quote_validation", {"message": "Quote validation using tpm2-tools", "data": retout}, False
+                "quote_validation", {"message": "Quote data validation using tpm2-tools", "data": retout}, False
+            )
+            return failure
+
+        # Only after validating the quote, the TPM clock information can be extracted from it.
+        clock_failure, clock_info = self.check_quote_timing(clock_info, quote)
```
THS-on, line 1230: Maybe name the parameter `previous_clock_info` to avoid confusion.

keylime/tpm/tpm_main.py
```diff
@@ -1200,6 +1273,48 @@ def check_quote(
...
+
+        current_clockinfo = {}
+
+        if previous_clockinfo:
+
+            pclki = TPMClockInfo.from_json(previous_clockinfo)
+
+            retout, success = self._tpm2_printquote(quote, False)
+            if not success:
+                return "tpm2_print failed with " + str(retout), current_clockinfo
```
THS-on, line 1287: You can also construct a failure object and return that. This is more expressive than a string.

keylime/tpm/tpm_main.py
```diff
@@ -1200,6 +1273,48 @@ def check_quote(
...
+                return "clock safe flag is disabled", current_clockinfo
+
+            if not (resetdiff and restartdiff):
+                if cclki.clock - pclki.clock <= 0:
+                    return (
+                        "clock timestamp did issued by TPM did not increase between two consecutive quotes",
+                        current_clockinfo,
+                    )
+
+            current_clockinfo = cclki.to_json()
```
THS-on, line 1314: Why not return `TPMClockInfo`. I would to the json conversion as close as possible to the database storage.

keylime/agentstates.py
```diff
@@ -1,9 +1,47 @@
...
+            dclki = data
+
+        return cls(
+            clock=dclki.get("clock", 0),
+            resetcount=dclki.get("resetCount", 0),
+            restartcount=dclki.get("restartCount", 0),
+            safe=dclki.get("safe", 1),
+        )
+
+    def to_json(self):
```
THS-on, line 37: `to_json()` is probably a bit confusing because we are not actually converting to json. It produces a `dict`.

THS-on (changes requested):
Mostly Some changes regarding where to use `TPMClockInfo` and the json conversion.

The checks itself LGTM.

keylime/agentstates.py
maugustosilva: good catch, will change

keylime/cloud_verifier_tornado.py
maugustosilva: agree, will change

keylime/tpm/tpm_main.py
maugustosilva: +1

keylime/tpm/tpm_main.py
maugustosilva: Not sure, will keep it as is for now

keylime/tpm/tpm_main.py
maugustosilva: Cannot disagree, will change

maugustosilva:
@THS-on modified as requested

THS-on:
@maugustosilva I think you forgot to push your renaming `to_json` to `to_dict`.

mpeters (approved):

keylime/agentstates.py
```diff
@@ -1,9 +1,47 @@
...
+
+@dataclass
+class TPMClockInfo:
+    clock: int
+    resetcount: int
+    restartcount: int
+    safe: int
+
+    @classmethod
+    def from_json(cls, data) -> "TPMClockInfo":
```
THS-on, line 22: Maybe rename to `from_dict` just for consistency reasons. I missed that last round.

THS-on (commented):
Just one final small change. Otherwise LGTM.

Thank you for adding support for this!

THS-on:
@maugustosilva the linter now fails. Can you correct that?

keylime/agentstates.py
```diff
@@ -186,4 +233,4 @@ def add(self, agent_id, boottime, ima_pcrs_dict, next_ima_ml_entry, learned_ima_
         agentAttestState.set_boottime(boottime)
         agentAttestState.set_ima_pcrs(ima_pcrs_dict)
         agentAttestState.set_next_ima_ml_entry(next_ima_ml_entry)
-        agentAttestState.set_ima_keyrings(ImaKeyrings.from_json(learned_ima_keyrings))
+        agentAttestState.set_ima_keyrings(ImaKeyrings.from_dict(learned_ima_keyrings))
```
THS-on, line 236: Here you accidentally changed it for the ImaKeyrings class

kgoldman:
> I have questions about how to manage resetCount and restartCount, but if I read the code correctly, this logic is more strict that the alternatives, so LGTM.

BTW, the resetcount and restart count are obfuscated when the key is in the storage hierarchy, do don't depend on it being meaningful.

keylime/agentstates.py
maugustosilva: ooops, poor attempt at multi-tasking, fixed now

THS-on (approved):

THS-on:
> BTW, the resetcount and restart count are obfuscated when the key is in the storage hierarchy, do don't depend on it being meaningful.

Yes that we will keep in mind. The current agent should have the AK under the endorsement hierarchy.

In the future we will implement finer control on what is checked depending on the key and general setup.

PR #1064, opened 2022-07-29
Fixes #1063 in a minimalistic way, by making log output configurable

### LINKED ISSUE #1063: Keylime "server" components (`verifier` and `registrar`) always output to a log file (`/var/log/keylime/<component>.log`), even when run inside containers
# Environment
 * OS / version: Fedora Core 35
 * TPM Manufacturer: swtpm
 * Keylime version: post-6.4.2

# Description 
When executed inside containers, Keylime "server" components (`verifier` and `registrar`) send log entries to both `stdout` and to a log file (`/var/log/keylime/<component>.log`). This is both redundant (since the container's stdout/stderr can be already redirected to a remote logging server) and harmful (the disk space on the nodes running the container fills out with entries in `/var/log/keylime`).

An user should be able to selectively disable the creation of log files (when running inside a container)

## Steps to reproduce problem
1. Start any "server" component in a container 
2. Watch log output with `docker logs`
3. Get inside the docker (`docker exec -it bash`) and see that a log file has been created inside `/var/log/keylime`

maugustosilva:
This PR just removes the hard-coded log file destination, replacing it
with two attributes on `keylime.conf`. In the future, more work will be
needed in order to make the logging system more robust and configurable,
but this minimalistic set of changes fixes a clear and present
operational problem (in production deployments)

### DIFF
--- a/keylime.conf
+++ b/keylime.conf
@@ -249,7 +249,6 @@ database_pool_sz_ovfl=5,10
 
 auto_migrate_db = True
 
-
 # The number of worker processes to use for the cloud verifier.
 # Set to "0" to create one worker per processor.
 multiprocessing_pool_num_workers = 0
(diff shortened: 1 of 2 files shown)


mpeters:
So is the recommendation that containers leave it blank? If so, maybe we should put that in the conf documentation.

mpeters (approved):

Issue #1040, opened 2022-06-29
Better error handling and algo validation for `create_allowlist.sh`


lkatalin:
# Environment
 * OS / version: Fedora 36
 * Processor architecture: x86_64
 * TPM Manufacturer: N/A
 * Keylime version: latest

# Description 
If I accidentally run `./scripts/create_allowlist.sh` with incorrect parameters, the error handling is not very helpful in the following cases:

1. I use an incorrect or nonexistent hash algorithm
2. I pass the hash algorithm as the wrong parameter because I did not see that I must specify a filename due to the syntax of the usage output

Proposed solution:
I think it would help to have flags here for the file name and the hash alg, instead of relying on parameter order.
If we don't want flags, I think at the least we should validate input and the `usage` should be updated so that the two parameters look similar syntactically, ex:
```
Usage:  create_allowlist.sh list.txt sha256sum
```
OR
```
Usage:  create_allowlist.sh [filename] [hash-algo]
```
Ideally we would have both in the output, one as a specification and the other as an example.

## Expected behavior vs. actual behavior 
I would expect the usage output to be more clear, some validation on the two inputs, clearer error messages, and ideally flags.

```
[f36-keylime@localhost-live keylime]$ ./scripts/create_allowlist.sh -out list.txt -hash sha256
Error: hash alg must be either sha256sum or sha1sum

[f36-keylime@localhost-live keylime]$ ./scripts/create_allowlist.sh
Usage:  create_allowlist.sh [filename] [hash-algo]
Example:  create_allowlist.sh list.txt sha256sum
```

### Example 1: hash alg not validated, error handling not present
```
[f36-keylime@localhost-live keylime]$ ./scripts/create_allowlist.sh list.txt sha256
Writing allowlist to /home/f36-keylime/keylime/list.txt with sha256...
find: ‘sha256’: No such file or directory
find: ‘sha256’: No such file or directory
find: ‘sha256’: No such file or directory
```

### Example 2: Script interprets hash alg as filename, and uses default hash alg
```
[f36-keylime@localhost-live keylime]$ sudo ./scripts/create_allowlist.sh sha256
Writing allowlist to /home/f36-keylime/keylime/sha256 with sha1sum...
```

lkatalin:
This is actually not quite finished, the second part (validating the hash algo passed in) still needs to be implemented.

maugustosilva:
Oops, my bad @lkatalin.

PR #1065, opened 2022-07-31
Handle output file and algo validation errors

mayaCostantini:
Related to #1040 

Pass output file name and hash algorithm as flags:
`Usage:  create_allowlist.sh -o [filename] -h [hash-algo]`

### DIFF
--- a/scripts/create_allowlist.sh
+++ b/scripts/create_allowlist.sh
@@ -35,9 +35,32 @@ fi
 
 if [ $# -lt 1 ]
 then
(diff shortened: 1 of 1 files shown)


lkatalin (approved):
@mayaCostantini This is great! Thank you. I tried the script and it now works as I would expect (minus the validation of the `-h` parameter, which can be implemented separately). That is: without `-o` it will not run, the parameters can be passed in either order, and the hash alg defaults to `sha1sum` unless another valid algo is passed in with `-h`.

mpeters (approved):

PR #23, opened 2019-01-07
Implements pip requirements standard for test

lukehinds:
Start using pip standard for package versioning control as per [0]

The current versions in `test-requirements.yml` are the latest
available releases on pypi.

Closes #20

[0] https://pip.pypa.io/en/latest/user_guide/#requirements-files

PR #1068, opened 2022-08-03
Split and simplify configuration

ansasaki:
Add scripts to convert configuration from the old format to the new split format.

The script extracts the values from the given configuration file and
replaces the fields in the templates.  The templates for the
configuration files in the new split format are also added.

The templates to convert the old single configuration file to the new
split format are added with the convertion script, as well as the
'adjust.py' script to adjust values that changed from the old
configuration option to the new format.

The result of the converted default configuration is also addded in the
'config' directory.

The new configuration file is described in:
https://github.com/keylime/enhancements/blob/master/72_config_and_simplify_tls.md

Signed-off-by: Anderson Toshiyuki Sasaki <ansasaki@redhat.com>


aplanas:
Really neat! I still need to review the code at full but IIUC the plan from now on is that, for any change in the configuration file will require a new set of templates.

I have some general questions:

* Can we avoid using a triplet for the semver? Seems excessive to have a patch level.
* We have at least two versions (we have more, but untracked). I think we should start with `1.0` or `2.0`.
* A glimpse of the code seems to apply a map from the current configs into the last template available. I think that the general algorithm should apply from all the missing one, in a similar way that the SQL migrations
* In the same way, maybe we can think later on a plugin mechanism that can provide default values based on the current deployment, so the dictionary can be enriched migration by migration.
* We duplicate the examples in `config/*` from the ones in the templates, maybe we can generate them from the last template during the package build process so we know for sure that they are always in sync.

Of course the last 3 points are just thoughts for future work, not for this PR. But I love this revamping!

ansasaki:
> Really neat! I still need to review the code at full but IIUC the plan from now on is that, for any change in the configuration file will require a new set of templates.

Yes. The idea is to provide, for each new config version,  the `mapping.json` file to control the replacements (and defaults for missing options), the `adjust.py` for special adjustments needed, and the templates with the patterns to replace. My plan was to be able to use this same script for configuration upgrades in the future, but currently it is not checking the current config version number to decide the needed transformation.

> 
> I have some general questions:
> 
>     * Can we avoid using a triplet for the semver? Seems excessive to have a patch level.

I think we can avoid using the triplet, no problem.

> 
>     * We have at least two versions (we have more, but untracked). I think we should start with `1.0` or `2.0`.

OK, I think it makes sense, I'll start with `2.0` for the new configuration format (and consider the old config `1.0`). 

> 
>     * A glimpse of the code seems to apply a map from the current configs into the last template available. I think that the general algorithm should apply from all the missing one, in a similar way that the SQL migrations

Sorry, I'm not familiar with the SQL migrations. My idea was to generate the configuration files for the latest version if not specified as an initial policy. Ideally we would implement the configuration version check and apply all the transformations on the way. I mean: if the latest config version is 6.0 and we have the 3.0 version to convert, it should apply all transformations from 3.0 to 4.0, 4.0 to 5.0, and finally 5.0 to 6.0.

Maybe I'll squeeze this change in this PR to have the complete "config upgrade" script ready from the beginning.

>     * In the same way, maybe we can think later on a plugin mechanism that can provide default values based on the current deployment, so the dictionary can be enriched migration by migration.

Thinking on more "smart" adjustments needed, I added the `adjust.py` script which is dynamically loaded and executed to apply transformations to the configuration. So what the `convert_config.py` script does:

- Reads the current installed configuration (or the provided input file) to extract the old options values
- Run the `adjust.py` script which transforms values from the old configuration when needed (e.g. replace `CV` with `default` in some options if it was the value found in the config)
- Use the resulting dictionary to replace the values in the templates (for the version specified in the `mapping.json`)

> 
>     * We duplicate the examples in `config/*` from the ones in the templates, maybe we can generate them from the last template during the package build process so we know for sure that they are always in sync.

It makes sense to generate the configs instead of having a static copy based on the latest version. I'll change this.

> 
> 
> Of course the last 3 points are just thoughts for future work, not for this PR. But I love this revamping!

I think it makes sense to add these changes here to have the "config upgrade" solution ready from the beginning

ansasaki:
I modified the `convert_config.py` script to apply each transformation for versions found in the `templates` folder in crescent order. I also added the `--version` option to specify a target version to stop applying transformations. This makes the script usable for configuration upgrades (but requires mappings and templates for each version to be provided).

This changes the version used for configuration files to not use semver triplets for version numbers, but the `MAJOR.MINOR` format.

I also removed the pre-generated config files.

Note that this PR still does not make the changes to the code to use the configuration files in the new format, only adds the scripts and templates to generate the configuration files in the new format.

scripts/convert_config.py
```diff
@@ -0,0 +1,460 @@
...
+
+    if not flat:
+        # If no component was provided, use all available
+        components = COMPONENTS
+    else:
+        clean = list(set(flat))
+        for c in clean:
+            if not c in COMPONENTS:
+                print(f"Unknown component {c}, skipping")
+                clean.remove(c)
```
aplanas, line 162: Ahh Rust would complain here!

```python
>> a = [1, 2, 3, 4]
>>> for i in a:
...   print(i)
...   a.remove(i)
... 
1
3
```

scripts/convert_config.py
```diff
@@ -0,0 +1,460 @@
...
+            info = m[option]
+            try:
+                new[component][option] = old_config[info["section"]][info["option"]]
+            except Exception as e:
+                print(
+                    f"{e}: Using default value \"{info['default']}\" for \"{option}\"")
+                new[component][option] = info['default']
+
+    # If there is an adjust script, load and run it
+    adjust_script = os.path.abspath(os.path.join(version_dir, "adjust.py"))
```
aplanas, line 236: cool! :+1:

scripts/convert_config.py
```diff
@@ -0,0 +1,460 @@
...
+        raise Exception(f"Invalid version string {b}: expected 'MAJOR.MINOR'")
+
+    # After validating the strings format, split and compare
+    (a_major, a_minor) = map(lambda n: int(n), a.strip().split('.'))
+    (b_major, b_minor) = map(lambda n: int(n), b.strip().split('.'))
+
+    if a_major == b_major:
+        return a_minor - b_minor
+    else:
+        return a_major - b_major
```
aplanas, line 307: There is a nice trick.  You can compare tuples. With a code like this you can: validate the version string, extract the content and return a value that can be compared:

```python
>>> import re
>>> v1 = re.findall(r"^(\d+)\.(\d+)$", "1.2")[0]
>>> v2 = re.findall(r"^(\d+)\.(\d+)$", "1.3")[0]
>>> v3 = re.findall(r"^(\d+)\.(\d+)$", "2.0")[0]
>>> v1 < v2; v1 > v2; v2 < v3; v2 > v3
True
False
True
False
```

scripts/convert_config.py
```diff
@@ -0,0 +1,460 @@
...
+
+    if target_version and not target_version in versions:
+        raise Exception(f"Directory for target version {target_version} not "
+                        f"found in {templates}")
+
+    new = {}
+
+    # If the configuration file does not contain a version number, assume it is
+    # in the old format and use the minimum possible version
+    if "version" in old_config.keys():
```
aplanas, line 340: `"version" in old_config`

scripts/convert_config.py
```diff
@@ -0,0 +1,460 @@
...
+        old_version = old_config["version"]
+    else:
+        old_version = '0.0'
```
aplanas, line 343: All can be replaced with `old_version = old_config.get("version", "0.0")`

scripts/convert_config.py
```diff
@@ -0,0 +1,460 @@
...
+
+    for version in versions:
+        # Skip versions lower than the current version
+        if cmp_version(old_version, version) >= 0:
```
aplanas, line 347: Using the tuple trick you can do `old_version >= version`

scripts/convert_config.py
```diff
@@ -0,0 +1,460 @@
...
+        if not os.path.exists(args.templates):
+            raise Exception(
+                f"Templates directory {args.template} does not exist")
+
+        if not os.path.isdir(args.templates):
+            raise Exception(f"File {args.templates} is not a directory")
+
+        templates = args.templates
+    else:
+        templates = "./templates"
```
aplanas, line 435: If you add `default="./templates"` in the `add_argument`, instead of `None`, you can skip this line and also will run the validation (check the the patch exists) in all cases

scripts/convert_config.py
aplanas: Also, should be `"1.0"`?

aplanas (changes requested):
This looks amazing.

scripts/convert_config.py
ansasaki: Thank you for pointing this, I'll change the code.

scripts/convert_config.py
ansasaki: Apparently this doesn't work as expected. I tried this:
```
>>> import re
>>> v2 = re.findall(r"^(\d+)\.(\d+)$", "2.0")[0]
>>> v10 = re.findall(r"^(\d+)\.(\d+)$", "10.0")[0]
>>> v10 > v2
False
>>> v2 > v10
True
```
It could probably work if each string in the tuple was converted to `int`.

scripts/convert_config.py
ansasaki: Thanks, I'll replace it.

scripts/convert_config.py
aplanas:
```python
>>> (10, 0) > (2, 0)
True
```
Ah sorry, indeed should be `int`s.  This will do the job.

```python
v = tuple(int(x) for x in re.findall(r"^(\d+)\.(\d+)$", "2.0")[0])
```

THS-on: There is also the `Version` class from the packaging module that we already use in `api_version.py`.
More information here: https://packaging.pypa.io/en/latest/version.html

scripts/convert_config.py
ansasaki: Thanks, I'll replace everything with `old_version = old_config.get("version", "1.0")`

Since we changed the first version in templates to be `2.0`, we can use `1.0` here.

scripts/convert_config.py
ansasaki: The tuple trick could be used if we convert the version string to tuples of `int`. For now I'll keep the current implementation.

scripts/convert_config.py
ansasaki: Thanks, I'll change the code.

scripts/convert_config.py
ansasaki: I created an "intermediary" solution using the tuples idea to simplify sorting/comparisons. Please check.

scripts/convert_config.py
ansasaki: Using `old_version = old_config.get("version", "1.0")` was not straightforward as I initially thought since `old_config` is not always a normal python dictionary, but can also be a `RawConfigParser` which implements `get` differently. I worked around by catching the exceptions that can be raised. Please check if this is acceptable.

scripts/convert_config.py
ansasaki: As mentioned above, I used the tuple idea to implement an intermediary solution.

scripts/convert_config.py
```diff
@@ -0,0 +1,436 @@
...
+    is reached, otherwise all transformations to the latest version are applied
+    """
+
+    dirs = os.listdir(templates)
+
+    if not dirs:
+        raise Exception(f"No directories found in {templates} for versions")
+
+    # Get a sorted list of the available versions as tuples
+    versions = sorted(list(x for x in set(map(str_to_version, dirs)) if x is not None))
```
aplanas, line 291: Nit: there is no need to convert it to a `list`, `sorted` should use the generator directly.

aplanas (approved):
I did not review that the new fields for the configuration file match the RFC, not the conversion is correct. But this PR provides a mechanism to fix and improve the configuration files, so we should be good.

scripts/convert_config.py
ansasaki: Thanks, fixed!

ansasaki:
> @ansasaki this is looking good. Is it still a WIP?

Yes, but I believe it is almost done. I'm working on the changes in the entire code to use the new configuration format and I'm still debugging (but I'm close to get all services working)

mpeters:
Cool, just let me know when you're ready for a final review and merge.

ansasaki:
@mpeters I think it is now good enough for a final review. I managed to run a minimal attestation test (manually with all services running on the same machine) using the configuration files generated by the script from the default `keylime.conf`.

I want to debug a bit more and fix the tests before pushing the code changes, but I think the config conversion scripts and templates are good to go.

ansasaki:
@THS-on I'm thinking on reverting the change that drops support for encrypted private keys. Is there a particular reason why you added this change to the configuration enhancement proposal?

THS-on:
@ansasaki my idea was to do simplifications where possible. We had a mix of keys were only in some cases the encryption was supported. In my opinion there is no good reason to encrypt them because if you can read the keys than you also can probably read the keylime.conf to read the password.

If you have a good reason for not removing it we can revert it, but I think it adds more complexity than necessary.

ansasaki:
> @ansasaki my idea was to do simplifications where possible. We had a mix of keys were only in some cases the encryption was supported. In my opinion there is no good reason to encrypt them because if you can read the keys than you also can probably read the keylime.conf to read the password.
> 
> If you have a good reason for not removing it we can revert it, but I think it adds more complexity than necessary.

I understand your way of thinking, but keeping the support for password encrypted keys allows the user who wants to use an existing encrypted key. Some may see the removal of this possibility as a regression (although I agree that it does not add extra security).

I went on and added support for encrypted keys in all places. The change also allowed to make the keys/certificates usage more consistent.

THS-on:
> I understand your way of thinking, but keeping the support for password encrypted keys allows the user who wants to use an existing encrypted key. Some may see the removal of this possibility as a regression (although I agree that it does not add extra security).

The question is if someone actually used that feature in the past. I guess is that most users just use the default provided CA which has no encryption by default.


We agree that it does not add any security benefits, so I think the best thing is to remove it with this change because otherwise it adds unnecessarily complexity.  @mpeters @maugustosilva opinions?

mpeters:
If we think there are folks in the wild using this feature, then we would need a deprecation cycle. If we don't think anyone is actively using this, then I'm ok with removing it. It's only really useful if someone has existing certs and can't generate new ones, right?

THS-on:
> It's only really useful if someone has existing certs and can't generate new ones, right?

Yeah pretty much, we also agreed that for automigration we only support the basic setup (only the CV CA is used) of certificates.

THS-on:
@ansasaki if the password discussion is delaying some of your work then we just keep it in for now and the deprecate it later.

kkaarreell:
@ansasaki Hi, I tried to migrate existing tests to new config format but there are some tests failing. It is possible that I have overlooked something while changing options but I can't see what's wrong. Would you mind taking a look?
https://github.com/RedHat-SP-Security/keylime-tests/pull/182

kkaarreell:
I think the problem is that config files under `/etc/keylime/` won't appear on a test system with keylime installation. I can see you have a migration script but how is a user installing latest upstream keylime supposed to create those files? I think we should have them available in a repository. It would be weird if users would be expected to install old `/etc/keylime.conf` and then run the migration script every time. Would it make sense to include these new files in keylime repo and install them on a system with the installation script?

kkaarreell:
So I have split `/etc/keylime.conf` into multiple files under `/etc/keylime` using the provided script but.. where is the code that actually instructs keylime processes to use these new locations?
I tried running a tenant and got
```
keylime_tenant -v 127.0.0.1 -t 127.0.0.1 -u d432fbb3-d2f1-4a97-9ef7-75bd81c00000 --verify --allowlist allowlist.txt --exclude excludelist.txt --include payload-module --cert default -c add
Config file not found in ['/usr/etc/keylime.conf', '/etc/keylime.conf', '/root/.config/keylime.conf']. Please set environment variable KEYLIME_CONFIG or see /usr/local/lib/python3.10/site-packages/keylime-6.4.2-py3.10.egg/keylime/config.py for more details.
Falling back on package provided configuration
```
i.e. `/etc/keylime/tenant.conf` is not used.

ansasaki:
>I think the problem is that config files under /etc/keylime/ won't appear on a test system with keylime installation. I can see you have a migration script but how is a user installing latest upstream keylime supposed to create those files? I think we should have them available in a repository. It would be weird if users would be expected to install old /etc/keylime.conf and then run the migration script every time. Would it make sense to include these new files in keylime repo and install them on a system with the installation script?

In my first version of this PR I had the generated configuration files also added in a `config` directory, but @aplanas suggested that it is better to generate them as part of the installation script

> So I have split `/etc/keylime.conf` into multiple files under `/etc/keylime` using the provided script but.. where is the code that actually instructs keylime processes to use these new locations? I tried running a tenant and got

The code that uses the split files is not in this branch, and still WIP in a separate branch: https://github.com/ansasaki/keylime/tree/new_conf

I'm still fixing the tests there, probably there are still few bugs around the code too, and I have to fix the installation to generate the new configuration files. I created this separate PR for the scripts/templates because the code is independent from the rest of the changes and could be reviewed separately .

Do you think it is better to add the code to this PR or should I open a separate one?

kkaarreell:
At least from testing perspective it would be nice to have these two merged, even temporarily in a separate branch, just for the test execution.

aplanas:
> In my first version of this PR I had the generated configuration files also added in a config directory, but @aplanas suggested that it is better to generate them as part of the installation script

The idea is that we remove the problem of diverging config files. For packaging and testing we need to generate and edit the config file in any case.

@kkaarreell why for testing cannot be updated the scripts now to generate the files if they are not there? In that way it will be compatible with the old code base and ready for the new

kkaarreell:
> > In my first version of this PR I had the generated configuration files also added in a config directory, but @aplanas suggested that it is better to generate them as part of the installation script
> 
> The idea is that we remove the problem of diverging config files. For packaging and testing we need to generate and edit the config file in any case.

I see. I am fine with that, I was just confused that this is not happening in the installation script.

> @kkaarreell why for testing cannot be updated the scripts now to generate the files if they are not there? In that way it will be compatible with the old code base and ready for the new

I was going to do that but now I have just learned that the code that actually utilize these new options is on a different place. If installation script is going to do that why to duplicate this functionality in tests?

aplanas:
> I was going to do that but now I have just learned that the code that actually utilize these new options is on a different place. If installation script is going to do that why to duplicate this functionality in tests?

Yes, exactly. You are right.

kkaarreell:
@ansasaki I believe I have fixed that syntax error in /setup/install_upstream_keylime,  please reschedule tests

kkaarreell:
Hi @ansasaki , I can see that `/etc/keylime` population is happening in `installer.sh` which I am not using. I will therefore update tests to copy config files to its destination. 
While working on that, I have noticed
```
Reading configuration from ['/etc/keylime/agent.conf']
  File "/usr/local/bin/keylime_tenant", line 33, in <module>
    sys.exit(load_entry_point('keylime==6.4.3', 'console_scripts', 'keylime_tenant')())
  File "/usr/local/bin/keylime_tenant", line 25, in importlib_load_entry_point
    return next(matches).load()
  File "/usr/lib64/python3.10/importlib/metadata/__init__.py", line 162, in load
    module = import_module(match.group('module'))
  File "/usr/lib64/python3.10/importlib/__init__.py", line 126, in import_module
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 883, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/usr/local/lib/python3.10/site-packages/keylime-6.4.3-py3.10.egg/keylime/cmd/tenant.py", line 3, in <module>
  File "/usr/local/lib/python3.10/site-packages/keylime-6.4.3-py3.10.egg/keylime/tenant.py", line 36, in <module>
    class Tenant:
  File "/usr/local/lib/python3.10/site-packages/keylime-6.4.3-py3.10.egg/keylime/tenant.py", line 89, in Tenant
    tpm_instance = tpm()
  File "/usr/local/lib/python3.10/site-packages/keylime-6.4.3-py3.10.egg/keylime/tpm/tpm_main.py", line 94, in __init__
    self.defaults["hash"] = algorithms.Hash(",".join(defaultHash))
  File "/usr/lib64/python3.10/enum.py", line 384, in __call__
    return cls.__new__(cls, value)
  File "/usr/lib64/python3.10/enum.py", line 701, in __new__
    raise ve_exc
ValueError: 'sha512, sha384, sha256, sha1' is not a valid Hash
```
Clearly the code expects just one hash (not list of hashes) to be stored in `tpm_hash_alg`, however `tpm_hash_alg = ['sha256']` seems to be unnecessarily complicated. Should it be changed from list to a string?

kkaarreell:
Also, do you plan to use config defaults defined in a code, in case the option is missing in a config file?
Accidentally, I have also seen
```
                out: Traceback (most recent call last):
                out:   File "/usr/local/bin/keylime_tenant", line 33, in <module>
                out:     sys.exit(load_entry_point('keylime==6.4.3', 'console_scripts', 'keylime_tenant')())
                out:   File "/usr/local/bin/keylime_tenant", line 25, in importlib_load_entry_point
                out:     return next(matches).load()
                out:   File "/usr/lib64/python3.10/importlib/metadata/__init__.py", line 162, in load
                out:     module = import_module(match.group('module'))
                out:   File "/usr/lib64/python3.10/importlib/__init__.py", line 126, in import_module
                out:     return _bootstrap._gcd_import(name[level:], package, level)
                out:   File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
                out:   File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
                out:   File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
                out:   File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
                out:   File "<frozen importlib._bootstrap_external>", line 883, in exec_module
                out:   File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
                out:   File "/usr/local/lib/python3.10/site-packages/keylime-6.4.3-py3.10.egg/keylime/cmd/tenant.py", line 3, in <module>
                out:     from keylime import keylime_logging, tenant
                out:   File "/usr/local/lib/python3.10/site-packages/keylime-6.4.3-py3.10.egg/keylime/tenant.py", line 36, in <module>
                out:     class Tenant:
                out:   File "/usr/local/lib/python3.10/site-packages/keylime-6.4.3-py3.10.egg/keylime/tenant.py", line 89, in Tenant
                out:     tpm_instance = tpm()
                out:   File "/usr/local/lib/python3.10/site-packages/keylime-6.4.3-py3.10.egg/keylime/tpm/tpm_main.py", line 62, in __init__
                out:     defaultEncrypt = config.getlist("agent", "tpm_encryption_alg")
                out:   File "/usr/local/lib/python3.10/site-packages/keylime-6.4.3-py3.10.egg/keylime/config.py", line 247, in getlist
                out:     read = get_config(component).get(section, option)
                out:   File "/usr/lib64/python3.10/configparser.py", line 793, in get
                out:     raise NoOptionError(option, section)
                out: configparser.NoOptionError: No option 'tpm_encryption_alg' in section: 'agent'
```

kkaarreell:
> Hi @ansasaki , I can see that `/etc/keylime` population is happening in `installer.sh` which I am not using. I will therefore update tests to copy config files to its destination. 

This is now pushed. But basic attestation is failing with
```
                out: Aug 25 22:15:34 testcloud keylime_registrar[19174]:   File "/usr/local/lib/python3.10/site-packages/keylime-6.4.3-py3.10.egg/keylime/cmd/registrar.py", line 1, in <module>
                out: Aug 25 22:15:34 testcloud keylime_registrar[19174]:     import keylime.cmd.migrations_apply
                out: Aug 25 22:15:34 testcloud keylime_registrar[19174]:   File "/usr/local/lib/python3.10/site-packages/keylime-6.4.3-py3.10.egg/keylime/cmd/migrations_apply.py", line 5, in <module>
                out: Aug 25 22:15:34 testcloud keylime_registrar[19174]:     from keylime import keylime_logging
                out: Aug 25 22:15:34 testcloud keylime_registrar[19174]:   File "/usr/local/lib/python3.10/site-packages/keylime-6.4.3-py3.10.egg/keylime/keylime_logging.py", line 12, in <module>
                out: Aug 25 22:15:34 testcloud keylime_registrar[19174]:     logdest = config.get(svc, "log_destination", fallback="")
                out: Aug 25 22:15:34 testcloud keylime_registrar[19174]:   File "/usr/local/lib/python3.10/site-packages/keylime-6.4.3-py3.10.egg/keylime/config.py", line 289, in get
                out: Aug 25 22:15:34 testcloud keylime_registrar[19174]:     return get_config(component).get(section, option, fallback=fallback)
                out: Aug 25 22:15:34 testcloud keylime_registrar[19174]:   File "/usr/local/lib/python3.10/site-packages/keylime-6.4.3-py3.10.egg/keylime/config.py", line 212, in get_config
                out: Aug 25 22:15:34 testcloud keylime_registrar[19174]:     _config[component].read_string(file.read_text("utf-8"))
                out: Aug 25 22:15:34 testcloud keylime_registrar[19174]:   File "/usr/lib64/python3.10/pathlib.py", line 1132, in read_text
                out: Aug 25 22:15:34 testcloud keylime_registrar[19174]:     with self.open(mode='r', encoding=encoding, errors=errors) as f:
                out: Aug 25 22:15:34 testcloud keylime_registrar[19174]:   File "/usr/lib64/python3.10/pathlib.py", line 1117, in open
                out: Aug 25 22:15:34 testcloud keylime_registrar[19174]:     return self._accessor.open(self, mode, buffering, encoding, errors,
                out: Aug 25 22:15:34 testcloud keylime_registrar[19174]: FileNotFoundError: [Errno 2] No such file or directory: '/usr/local/lib/python3.10/site-packages/keylime-6.4.3-py3.10.egg/keylime/config/registrar.conf'
```

kkaarreell:
And also
```
               out: Traceback (most recent call last):
                out:   File "/usr/local/bin/keylime_ima_emulator", line 33, in <module>
                out:     sys.exit(load_entry_point('keylime==6.4.3', 'console_scripts', 'keylime_ima_emulator')())
                out:   File "/usr/local/bin/keylime_ima_emulator", line 25, in importlib_load_entry_point
                out:     return next(matches).load()
                out:   File "/usr/lib64/python3.10/importlib/metadata/__init__.py", line 162, in load
                out:     module = import_module(match.group('module'))
                out:   File "/usr/lib64/python3.10/importlib/__init__.py", line 126, in import_module
                out:     return _bootstrap._gcd_import(name[level:], package, level)
                out:   File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
                out:   File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
                out:   File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
                out:   File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
                out:   File "<frozen importlib._bootstrap_external>", line 883, in exec_module
                out:   File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
                out:   File "/usr/local/lib/python3.10/site-packages/keylime-6.4.3-py3.10.egg/keylime/cmd/ima_emulator_adapter.py", line 14, in <module>
                out:     tpm_instance = tpm(need_hw_tpm=True)
                out:   File "/usr/local/lib/python3.10/site-packages/keylime-6.4.3-py3.10.egg/keylime/tpm/tpm_main.py", line 82, in __init__
                out:     raise Exception(f"Unsupported encryption algorithm specified: {str(defaultEncrypt)}!")
                out: Exception: Unsupported encryption algorithm specified: ['ecc', ' rsa']!
```
There is an extra space in ' rsa'.

ansasaki:
> Clearly the code expects just one hash (not list of hashes) to be stored in `tpm_hash_alg`, however `tpm_hash_alg = ['sha256']` seems to be unnecessarily complicated. Should it be changed from list to a string?

The code expects lists in the form of python code, as agreed before, and as specified in the proposal.

`"['sha512', 'sha384', 'sha256', 'sha1']"` should be accepted, but not `"sha512, sha384, sha256, sha1"`

This was one of the main goals of this change, to use a single syntax for lists. So no, we should keep the list format instead of string.

ansasaki:
> Also, do you plan to use config defaults defined in a code, in case the option is missing in a config file?
> Accidentally, I have also seen

`tpm_encryption_alg` is a typo, should be `tpm_encryption_algs`. So it is not a missing value in the config file, but an invalid option.

ansasaki:
> There is an extra space in ' rsa'.

This is a real bug, it should use `strip()` to remove the leading and trailing white spaces.

ansasaki:
> Hi @ansasaki , I can see that `/etc/keylime` population is happening in `installer.sh` which I am not using. I will therefore update tests to copy config files to its destination.

I modified the setup code to generate the split configuration files on `build` and install them on `install`

kkaarreell:
> > Clearly the code expects just one hash (not list of hashes) to be stored in `tpm_hash_alg`, however `tpm_hash_alg = ['sha256']` seems to be unnecessarily complicated. Should it be changed from list to a string?
> 
> The code expects lists in the form of python code, as agreed before, and as specified in the proposal.
> 
> `"['sha512', 'sha384', 'sha256', 'sha1']"` should be accepted, but not `"sha512, sha384, sha256, sha1"`
> 
> This was one of the main goals of this change, to use a single syntax for lists. So no, we should keep the list format instead of string.

What I mean is: Is keylime code actually able to work with `['sha512', 'sha256']`? Because former keylime.conf expected only one value (hash) there, not multiple hashes. And `['sha512', 'sha256']` didn't work for me, even though it is definitely possible I had a bug somewhere else.

ansasaki:
> What I mean is: Is keylime code actually able to work with `['sha512', 'sha256']` specifically in `tpm_hash_alg`? Because former keylime.conf expected only one value (hash) there, not multiple hashes. And `['sha512', 'sha256']` didn't work for me, even though it is definitely possible I had a bug somewhere else.

I think you are right. I'm checking the code and these values are used as parameters to tpm2_tools commands, like `tpm2_createek` which expects single values and not lists.

I'll change back the code to make these to be single values (and I need to take note to update the specification on the proposal as well)

Thank you for catching this!

kkaarreell:
Hi @ansasaki , I am sorry for the confusion. The last commit regarding `require_ek_check` didn't help. I believe it is not even necessary as the problem is somewhere in tests, for some reason this option is not updated. I think you can remove that commit.

kkaarreell:
> the problem is somewhere in tests, for some reason this option is not updated.
I have fixed that issue and few more. Please remove that commit and rerun tests, we should make some progress.

ansasaki:
> the problem is somewhere in tests, for some reason this option is not updated.
> I have fixed that issue and few more. Please remove that commit and rerun tests, we should make some progress.

I think that commit makes sense anyway, the EK cert checking should not happen unless it is required. I'll keep it and rerun the tests.

kkaarreell:
The condition is in a different part of the code
https://github.com/keylime/keylime/blob/efc921e55de1163f1114a2358c2e0eca64d1600f/keylime/tenant.py#L516

kkaarreell:
I have noticed a log message mentioning cloud_agent section, it may be worth grepping the code for old section names.

kkaarreell:
I cannot verify it ATM but logs suggests tpm_policy option is not converted properly or the default has changed or empty policy {}  is not handled properly. 
There are messages like
keylime.tpm - ERROR - PCR #23: .* from quote does not match expected value. 
while this PCR should not be used

ansasaki:
> I cannot verify it ATM but logs suggests tpm_policy option is not converted properly or the default has changed or empty policy {} is not handled properly. There are messages like keylime.tpm - ERROR - PCR #23: .* from quote does not match expected value. while this PCR should not be used

I'm trying to understand what should be the default for tpm policy. I saw in the code it is basically treated as mandatory, but we removed from the configuration file. That is why I added hardcoded defaults for it. I'll set the default policy as the empty policy `'{}'` in the tenant and see if it works better.

kkaarreell:
Test /functional/basic-attestation-with-custom-certificates is not configuring trusted CA certs as list, I will fix that.

kkaarreell:
There is one discrepancy.. Unlike other services agent does not have tls_dir and therefore trusted_client_ca is absolute path, not cert name in tls_dir.

kkaarreell:
> > I cannot verify it ATM but logs suggests tpm_policy option is not converted properly or the default has changed or empty policy {} is not handled properly. There are messages like keylime.tpm - ERROR - PCR #23: .* from quote does not match expected value. while this PCR should not be used
> 
> I'm trying to understand what should be the default for tpm policy. I saw in the code it is basically treated as mandatory, but we removed from the configuration file. That is why I added hardcoded defaults for it. I'll set the default policy as the empty policy `'{}'` in the tenant and see if it works better.

The default is not a problem. The issue is that even when `tpm_policy` is configured in a config file it won't be used since the respective `config.get` is missing.
https://github.com/keylime/keylime/blob/master/keylime/tenant.py#L155

As a quick check I have changed the code from your branch to
```
    def process_allowlist(self, args):
        # Set up PCR values
        tpm_policy = config.get("tenant", "tpm_policy")
        if tpm_policy is None:
            # Use default empty TPM policy
            tpm_policy = '{}'
        if "tpm_policy" in args and args["tpm_policy"] is not None:
            tpm_policy = args["tpm_policy"]
        self.tpm_policy = TPM_Utilities.readPolicy(tpm_policy)
        logger.info("TPM PCR Mask from policy is %s", self.tpm_policy["mask"])
```
and it fixed failing of `/functional/tpm_policy-sanity-on-localhost`. I didn't test other tests though.

kkaarreell:
I have updates tests accordingly, please rerun tests, there should be less failing ones.
/packit retest-failed

kkaarreell:
`--excludelist` should also have some reasonable default, now when it is not present in a config file. In `/functional/tenant-allowlist-sanity
` not defining `--excludelist` I can see
```
Aug 27 21:08:36 ip-172-31-25-204.us-east-2.compute.internal keylime_verifier[75860]: 2022-08-27 21:08:36.814 - tornado.application - ERROR - Uncaught exception POST /v2.1/allowlists/list1 (127.0.0.1)
Aug 27 21:08:36 ip-172-31-25-204.us-east-2.compute.internal keylime_verifier[75860]: HTTPServerRequest(protocol='https', host='127.0.0.1:8881', method='POST', uri='/v2.1/allowlists/list1', version='HTTP/1.1', remote_ip='127.0.0.1')
Aug 27 21:08:36 ip-172-31-25-204.us-east-2.compute.internal keylime_verifier[75860]: Traceback (most recent call last):
Aug 27 21:08:36 ip-172-31-25-204.us-east-2.compute.internal keylime_verifier[75860]:   File "/usr/lib64/python3.10/site-packages/tornado/web.py", line 1702, in _execute
Aug 27 21:08:36 ip-172-31-25-204.us-east-2.compute.internal keylime_verifier[75860]:     result = method(*self.path_args, **self.path_kwargs)
Aug 27 21:08:36 ip-172-31-25-204.us-east-2.compute.internal keylime_verifier[75860]:   File "/usr/local/lib/python3.10/site-packages/keylime-6.4.3-py3.10.egg/keylime/cloud_verifier_tornado.py", line 801, in post
Aug 27 21:08:36 ip-172-31-25-204.us-east-2.compute.internal keylime_verifier[75860]:     ima_policy = json.dumps(ima.process_ima_policy(allowlist, ima_policy_bundle["excllist"]))
Aug 27 21:08:36 ip-172-31-25-204.us-east-2.compute.internal keylime_verifier[75860]:   File "/usr/local/lib/python3.10/site-packages/keylime-6.4.3-py3.10.egg/keylime/ima/ima.py", line 429, in process_ima_policy
Aug 27 21:08:36 ip-172-31-25-204.us-east-2.compute.internal keylime_verifier[75860]:     exclude = read_excllist()
Aug 27 21:08:36 ip-172-31-25-204.us-east-2.compute.internal keylime_verifier[75860]:   File "/usr/local/lib/python3.10/site-packages/keylime-6.4.3-py3.10.egg/keylime/ima/ima.py", line 599, in read_excllist
Aug 27 21:08:36 ip-172-31-25-204.us-east-2.compute.internal keylime_verifier[75860]:     if os.path.exists(exclude_path):
Aug 27 21:08:36 ip-172-31-25-204.us-east-2.compute.internal keylime_verifier[75860]:   File "/usr/lib64/python3.10/genericpath.py", line 19, in exists
Aug 27 21:08:36 ip-172-31-25-204.us-east-2.compute.internal keylime_verifier[75860]:     os.stat(path)
Aug 27 21:08:36 ip-172-31-25-204.us-east-2.compute.internal keylime_verifier[75860]: TypeError: stat: path should be string, bytes, os.PathLike or integer, not NoneType
```

kkaarreell:
For the "without-revocation" plan, you also need to update it with
`    ref: new_config`

kkaarreell:
Any ideas what could be causing this?
```

::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
::   Test showallowlist
::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::

:: [ 15:18:49 ] :: [  BEGIN   ] :: Running 'keylime_tenant -c showallowlist --allowlist-name list1'
Reading configuration from ['/etc/keylime/registrar.conf']
Reading configuration from ['/etc/keylime/verifier.conf']
2022-08-28 15:18:49.910 - keylime.tpm - INFO - TPM2-TOOLS Version: 5.2
Reading configuration from ['/etc/keylime/agent.conf']
Applied configuration snippets from /etc/keylime/agent.conf.d
2022-08-28 15:18:49.913 - keylime.tenant - INFO - Setting up client TLS...
2022-08-28 15:18:49.914 - keylime.tenant - INFO - Using default client_cert option for tenant
2022-08-28 15:18:49.914 - keylime.tenant - INFO - Using default client_key option for tenant
2022-08-28 15:18:49.941 - keylime.tenant - INFO - TLS is enabled.
2022-08-28 15:18:49.941 - keylime.tenant - WARNING - Using default UUID d432fbb3-d2f1-4a97-9ef7-75bd81c00000
Show allowlist command response: 405.
{'code': 405, 'status': 'Not Implemented: Use /agents/ interface instead', 'results': {}}
:: [ 15:18:50 ] :: [   PASS   ] :: Command 'keylime_tenant -c showallowlist --allowlist-name list1' (Expected 0, got 0)
:: [ 15:18:50 ] :: [   FAIL   ] :: File '/var/tmp/rlRun_LOG.s5CN4AcJ' should contain '{'code': 200, 'status': 'Success', 'results': {'name': 'list1'' 
::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
::   Duration: 1s
::   Assertions: 1 good, 1 bad
::   RESULT: FAIL (Test showallowlist)

```
This test is not failing with upstream keylime.

ansasaki:
> Any ideas what could be causing this?

Yes, just found and fixed it, it was a typo

ansasaki:
> There is one discrepancy.. Unlike other services agent does not have tls_dir and therefore trusted_client_ca is absolute path, not cert name in tls_dir.

tl;dr; Let's modify the meaning of `trusted_client_ca = default` for the client. It should use the `$WORK_DIR/cv_ca` like the other components. This shouldn't affect the tests as @kkaarreell already modified the agent configuration to use absolute paths for `trusted_client_ca` instead of `default`

Long version:

So, I was investigating the remaining failing test (tpm-tests) and I believe it is related with this comment. Actually, the agent in the new configuration has a `tls_dir` option, but the default value for it differs from the other components. For the agent, the default `tls_dir` is the secure mount directory (`/var/lib/keylime/secure/`) while for the others is the `${WORK_DIR}/cv_ca` (by default `/var/lib/keylime/cv_ca`) to reuse the verifier generated files.

The problem is that when we use `default` for `trusted_client_ca` in the agent, it will trust CA certificates in `/var/lib/keylime/secure/cacert.crt` instead of `/var/lib/keylime/cv_ca/cacert.crt`, meaning that by default it will NOT trust the verifier generated certificates. 

To workaround this, I made the change in the `adjust.py` to use the following heuristic when setting the configuration file: if the verifier has `tls_dir = generate`, and the client has `trusted_client_ca = default`, then set the client `trusted_client_ca = ['/var/lib/keylime/cv_ca']`:

https://github.com/ansasaki/keylime/blob/8d37f7cf11fdcb832994cd90683493574fbc5c01/scripts/templates/2.0/adjust.py#L114-L128

This worked well for most of the test cases, but not  for the case where the `WORK_DIR` is modified (e.g. via `KEYLIME_DIR` env var). What happens is that the client has `trusted_client_ca = ['/var/lib/keylime/cv_ca']` while the verifier generates the files in `$WORK_DIR/cv_ca`, meaning that the client will not find the CA certificate.

ansasaki:
@mpeters @THS-on @aplanas OK, made the changes and with @kkaarreell help (thanks!) also modified the tests (in a separate branch).

This is ready to review.

Please let me know if you have anything to be fixed. And we can't forget to drop 6b631f400aab995bf1c039818076f54e28cfc5b8 commit before pushing.

keylime/config.py
```diff
@@ -73,45 +74,257 @@ def environ_bool(env_name, default):
...
+    "tenant": ["/etc/keylime/tenant.conf", "/usr/etc/keylime/tenant.conf"],
+    "registrar": ["/etc/keylime/registrar.conf", "/usr/etc/keylime/registrar.conf"],
+    "ca": ["/etc/keylime/ca.conf", "/usr/etc/keylime/ca.conf"],
+    "logging": ["/etc/keylime/logging.conf", "/usr/etc/keylime/logging.conf"],
+}
+
+# Paths to directories in which options can be overriden using configuration
+# snippets
+CONFIG_SNIPPETS_DIRS = {
+    "agent": ["/etc/keylime/agent.conf.d", "/usr/etc/keylime/agent.conf.d"],
```
aplanas, line 90: Technically "/usr/etc/keylime/*.conf.d" can be skipped, as the user can only deploy to "/etc/keylime/*.conf.d". But this is creating a very nice place for the distributor to put config snippet that does not come from upstream. I like it.

keylime/config.py
```diff
@@ -73,45 +74,257 @@ def environ_bool(env_name, default):
...
+    if not component:
+        # If no component was specified, return an empty configuration
+        raise Exception("No component provided to get_config")
+
+    if component not in _config:  # pylint: disable=too-many-nested-blocks
         # Use RawConfigParser, so we can also use it as the logging config
-        config = configparser.RawConfigParser()
+        _config[component] = configparser.RawConfigParser()
+
+        if not CONFIG_ENV or not isinstance(CONFIG_ENV, dict):
```
aplanas, line 173: Why here, and more later, are we validating a variable that is expected to be "static const" and declared only for this file?

Defensive programming is OK but if we do not expect that this can be changed, maybe is too much?

keylime/config.py
```diff
@@ -73,45 +74,257 @@ def environ_bool(env_name, default):
...
+    if read:
+        try:
+            l = ast.literal_eval(read)
+            if isinstance(l, list):
+                stripped = []
+                for s in l:
+                    if isinstance(s, str):
+                        s = s.strip()
+                    stripped.append(s)
+                return stripped
```
aplanas, line 259:
```suggestion
                return [i.strip() if isinstance(i, str) else i for i in l]
```

keylime/config.py
```diff
@@ -73,45 +74,257 @@ def environ_bool(env_name, default):
...
+
+def getlist(component, option, section=None):
+    if not section:
+        section = component
+
+    read = get_config(component).get(section, option)
+
+    if read:
+        try:
+            l = ast.literal_eval(read)
```
aplanas, line 252: I was wondering if this is safe or not. Seems that it is, as is a very focused evaluator. Maybe we should add a note.

keylime/config.py
```diff
@@ -73,45 +74,257 @@ def environ_bool(env_name, default):
...
+            )
+        except Exception as e:
+            raise Exception(
+                f"Failed to get list from config for component '{component}', section '{section}', option '{option}'"
+            ) from e
+
+    raise Exception(f"Could not find option '{option}' in section '{section}' of component '{component}'")
+
+
+def getdict(component, option, section=None):
```
aplanas, line 271: This is not used? Do we require it?

keylime/crypto.py
```diff
@@ -90,12 +90,20 @@ def rsa_export_pubkey(private_key):
...
+            format=serialization.PrivateFormat.TraditionalOpenSSL,
+            encryption_algorithm=serialization.NoEncryption(),
+        )
+
     return private_key.private_bytes(
         encoding=serialization.Encoding.PEM,
-        format=serialization.PrivateFormat.TraditionalOpenSSL,
-        encryption_algorithm=serialization.NoEncryption(),
+        format=serialization.PrivateFormat.PKCS8,
+        encryption_algorithm=serialization.BestAvailableEncryption(password.encode("utf-8")),
```
aplanas, line 106:
```python
        return private_key.private_bytes(
            encoding=serialization.Encoding.PEM,
            format=serialization.PrivateFormat.PKCS8 if password else serialization.PrivateFormat.TraditionalOpenSSL,
            encryption_algorithm=serialization.BestAvailableEncryption(password.encode("utf-8")) if password else serialization.NoEncryption(),
        )
```

aplanas (commented):
(I will continue tomorrow)

auto-ipsec/libreswan/src/local_action_update_crl.py
```diff
@@ -26,7 +23,7 @@ async def execute(json_revocation):
 
     secdir = secure_mount.mount()
 
-    cert_path = config.get("cloud_agent", "revocation_cert")
+    cert_path = common.get("agent", "revocation_cert")
```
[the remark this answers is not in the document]
THS-on: Agree

keylime/web_util.py
```diff
@@ -11,117 +10,195 @@
...
+    elif tls_dir == "default":
+        # Use the keys/certificates generated for the verifier
         tls_dir = os.path.abspath(os.path.join(config.WORK_DIR, "cv_ca"))
         if not os.path.exists(os.path.join(tls_dir, "cacert.crt")):
             raise Exception(
                 "It appears that the verifier has not yet created a CA and certificates, please run the verifier first"
             )
+    else:
+        # if it is relative path, convert to absolute in WORK_DIR
+        if tls_dir[0] != "/":
```
THS-on, line 126:
```suggestion
        if not os.path.isabs(tls_dir):
```

keylime/registrar_client.py
```diff
@@ -1,81 +1,30 @@
...
-ca_cert = False
-tls_enabled = False
 api_version = keylime_api_version.current_version()
 
 
-def init_client_tls(section):
-    global tls_cert_info
-    global tls_enabled
-    global ca_cert
+def getData(registrar_ip, registrar_port, agent_id, tls_context):
```
aplanas, line 12: `get_data`

keylime/requests_client.py
```diff
@@ -75,7 +66,32 @@ class HostNameIgnoreAdapter(HTTPAdapter):
     It is required because in most cases we don't know the hostname during certificate generation.
     """
 
+    _tls_context = None
```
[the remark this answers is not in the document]
aplanas: Right, no. Both `_tls_context` and `_ignore_hostame` are initialized in the `__init__`, so should be removed from the class. Also, they are accessed by `self`.

keylime/requests_client.py
```diff
@@ -75,7 +66,32 @@ class HostNameIgnoreAdapter(HTTPAdapter):
...
-        self.poolmanager = PoolManager(
-            num_pools=connections, maxsize=maxsize, block=block, strict=True, assert_hostname=False, **pool_kwargs
-        )
+        if self._ignore_hostname and self._tls_context:
+            self.poolmanager = PoolManager(
+                num_pools=connections,
+                maxsize=maxsize,
+                block=block,
+                strict=True,
+                assert_hostname=False,
```
aplanas, line 85: you can remove the `if` and the other branch with `assert_hostname=not (self._ignore_hostname and self._tls_context)` or creating a variable (maybe more clear) like `assert_hostname = False if self._ignore_hostname and self._tls_context else True`.

keylime/tenant.py
```diff
@@ -79,82 +93,61 @@ def __init__(self):
...
-        return tls_context, agent_mtls_context, verifier_ca_cert
-
     def process_allowlist(self, args):
         # Set up PCR values
-        tpm_policy = config.get("tenant", "tpm_policy")
         if "tpm_policy" in args and args["tpm_policy"] is not None:
             tpm_policy = args["tpm_policy"]
+        else:
+            # Use default empty TPM policy
+            tpm_policy = "{}"
```
aplanas, line 150: All the block can be replaced with `tpm_policy = args.get("tpm_policy") or {}`. If "None" is not expected, `tpm_policy = args.get("tpm_policy", {})` can work. If args is a config, then `fallback={}`.

keylime/tenant.py
```diff
@@ -1249,27 +1238,21 @@ def do_add_allowlist(self, args):
         self.process_allowlist(args)
         data = {"tpm_policy": json.dumps(self.tpm_policy), "ima_policy_bundle": json.dumps(self.allowlist)}
         body = json.dumps(data)
-        cv_client = RequestsClient(self.verifier_base_url, self.tls_cv_enabled, ignore_hostname=True)
+        cv_client = RequestsClient(self.verifier_base_url, True, tls_context=self.tls_context, ignore_hostname=True)
```
aplanas, line 1241: I feel a bit annoyed by this "True".  I suggest to make RequestsClient constructor to set `tls_enabled=True` and drop the parameter during the call?

keylime/web_util.py
```diff
@@ -11,117 +10,195 @@
...
+    elif tls_dir == "default":
+        if component in ("verifier", "registrar", "tenant"):
+            # Use the keys/certificates generated for the verifier
+            tls_dir = os.path.abspath(os.path.join(config.WORK_DIR, "cv_ca"))
+        elif component == "agent":
+            # For the agent, use the secure mount dir as the default directory
+            tls_dir = secure_mount.get_secdir()
+    else:
+        # if it is relative path, convert to absolute in WORK_DIR
+        if tls_dir[0] != "/":
```
aplanas, line 41: Not needed?
```
>>> import os.path
>>> WORK_DIR="/tmp"
>>> p1="/mydir"
>>> p2="mydir"
>>> os.path.join(WORK_DIR, p1)
'/mydir'
>>> os.path.join(WORK_DIR, p2)
'/tmp/mydir'
```

keylime/web_util.py
aplanas: also not needed? (see previous instance)

aplanas (commented):
Uau .. that is long!!

ansasaki:
I added a commit to address an issue found by @kkaarreell. I'll address the suggestions from the reviews shortly.

auto-ipsec/libreswan/src/local_action_update_crl.py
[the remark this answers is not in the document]
ansasaki: I'll add a commit dropping the demos.

keylime/ca_util.py
```diff
@@ -82,11 +82,18 @@ def cmd_mkcert(workingdir, name):
...
+                encoding=serialization.Encoding.PEM,
+                format=serialization.PrivateFormat.PKCS8,
+                encryption_algorithm=serialization.BestAvailableEncryption(password.encode("utf-8")),
+            )
+        else:
+            priv[0][name] = pk.private_bytes(
+                encoding=serialization.Encoding.PEM,
+                format=serialization.PrivateFormat.PKCS8,
+                encryption_algorithm=serialization.NoEncryption(),
+            )
```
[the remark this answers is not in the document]
ansasaki: Thanks! I knew python would have some "pythonic" way of doing this.

keylime/cloud_verifier_tornado.py
```diff
@@ -1221,21 +1226,20 @@ def main():
...
     called as a function by an external program."""
 
-    cloudverifier_port = config.get("cloud_verifier", "cloudverifier_port")
-    cloudverifier_host = config.get("cloud_verifier", "cloudverifier_ip")
-    cloudverifier_id = config.get(
-        "cloud_verifier", "cloudverifier_id", fallback=cloud_verifier_common.DEFAULT_VERIFIER_ID
-    )
+    cloudverifier_port = config.get("verifier", "port")
+    cloudverifier_host = config.get("verifier", "ip")
+    cloudverifier_id = config.get("verifier", "uuid", fallback=cloud_verifier_common.DEFAULT_VERIFIER_ID)
```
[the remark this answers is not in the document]
ansasaki: I'll drop it

keylime/config.py
ansasaki: Great, It was one of the goals, to have a easy way for the packager to add snippets on top of the default configuration.

keylime/config.py
ansasaki: To be honest, I only added these checks because `test_config.py` had test cases for such changes.

keylime/config.py
ansasaki: Cool, thanks!

keylime/config.py
ansasaki: The documentation states it is safe. It never executes the code, only parses a restrict subset of literal values.

See: https://docs.python.org/3/library/ast.html#ast.literal_eval

keylime/config.py
ansasaki: The `tpm_policy` was a dictionary, indeed. But it was dropped from the configuration file , following the enhancement proposal:

https://github.com/keylime/enhancements/blob/master/72_config_and_simplify_tls.md#removal-of-unused-default-values-from-the-configuration-file

The default  `tpm_policy` used by the tenant is the empty policy (`'{}'`)

I implemented this as the enhancement proposal asked for it:

https://github.com/keylime/enhancements/blob/master/72_config_and_simplify_tls.md#unification-of-data-types

In the end it is not used and I'll drop it.

keylime/cloud_verifier_tornado.py
```diff
@@ -569,13 +567,13 @@ def post(self):
...
                     # Prepare SSLContext for mTLS connections
-                    agent_mtls_cert_enabled = config.getboolean(
-                        "cloud_verifier", "agent_mtls_cert_enabled", fallback=False
-                    )
+                    agent_mtls_cert_enabled = config.getboolean("verifier", "enable_agent_mtls", fallback=False)
                     mtls_cert = agent_data["mtls_cert"]
                     agent_data["ssl_context"] = None
-                    if agent_mtls_cert_enabled and mtls_cert:
-                        agent_data["ssl_context"] = web_util.generate_agent_mtls_context(mtls_cert, mtls_options)
+                    if agent_mtls_cert_enabled and mtls_cert and mtls_cert != "disabled":
```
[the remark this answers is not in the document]
ansasaki: I'll extend the check on line 464 which will simplify the check later as we don't need to check again if the certificate was provided.

keylime/elchecking/policies.py
```diff
@@ -103,7 +103,7 @@ def evaluate(policy_name: str, refstate: RefState, eventlog: tests.Data) -> str:
     return tester.why_not({}, eventlog)
 
 
-imports = config.MEASUREDBOOT_IMPORTS
+imports = config.get("verifier", "measured_boot_imports", fallback="")
```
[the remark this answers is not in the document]
ansasaki: I didn't notice this option, thanks for pointing it. It wasn't mentioned in the enhancement proposal ad wasn't included in the configuration files at all.

I'll fix the mappings (and take a note to add this to the specification).

keylime/ima/ast.py
```diff
@@ -392,7 +392,7 @@ def invalid(self) -> Failure:
         if self.ima_template_hash == get_FF_HASH(self._ima_hash_alg):
             logger.warning("Skipped template_hash validation entry with FF_HASH")
             # By default ToMToU errors are not treated as a failure
-            if config.getboolean("cloud_verifier", "tomtou_errors", fallback=False):
+            if config.getboolean("verifier", "ignore_tomtou_errors", fallback=True):
```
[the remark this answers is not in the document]
ansasaki: I thought this was an intentional change you proposed.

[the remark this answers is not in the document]
THS-on: Yes, sorry. I forgot that.

keylime/requests_client.py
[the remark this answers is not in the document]
ansasaki: I'll drop the class variables.

keylime/tpm/tpm_main.py
```diff
@@ -58,11 +58,11 @@ def __init__(self, need_hw_tpm=False):
         self.supported["sign"] = set()
 
         # Grab which default algs the config requested
-        defaultHash = config.get("cloud_agent", "tpm_hash_alg")
-        defaultEncrypt = config.get("cloud_agent", "tpm_encryption_alg")
-        defaultSign = config.get("cloud_agent", "tpm_signing_alg")
+        defaultHash = config.get("agent", "tpm_hash_alg")
```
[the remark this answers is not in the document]
ansasaki: Yes, this is one of the goals, to make the components to work with just their own configuration files. 

We need to map places like this where components use values from other component's configuration files.

keylime/tenant.py
ansasaki: Yes, when I was writing the code I was a bit annoyed by this, but I thought the goal was to make it explicit that the hostname would be ignored when verifying certificates.

I'll change the code.

keylime/web_util.py
ansasaki: Interesting, I didn't know this was the behavior.

Reference: https://docs.python.org/3/library/os.path.html#os.path.join

keylime/requests_client.py
```diff
@@ -1,40 +1,31 @@
...
-import tempfile
-
 import requests
 from requests.adapters import HTTPAdapter
 from requests.packages.urllib3.poolmanager import PoolManager  # pylint: disable=import-error
 
 
 class RequestsClient:
-    def __init__(self, base_url, tls_enabled, ignore_hostname=False, verify_custom: str = False, **kwargs):
+    def __init__(self, base_url, tls_enabled, tls_context=None, ignore_hostname=False, **kwargs):
```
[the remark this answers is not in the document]
ansasaki: Thanks!

keylime/requests_client.py
ansasaki: Thanks for the suggestion! The only nit is that `assert_hostname` does not accept `True`, but only `False`, `None`, or the actual hostname.

keylime/tenant.py
ansasaki: Thanks!

ansasaki:
Hopefully now the components are loading their own configuration file (and possibly the shared configuration files `ca.conf` and `logging.conf`)

I'm aware of the conflict introduced by the recent merge of https://github.com/keylime/keylime/pull/1094. I'll fix it later.

ansasaki:
@maugustosilva Hi, I'd like to ask if you can use the [`get_tls_options()`](https://github.com/ansasaki/keylime/blob/00d80dea1934fedd66b28e86899e11902f418460/keylime/web_util.py#L214-L225) I introduced in this PR to obtain keys and certificates from the configuration file instead of changing the `init_mtls()` behavior. If it is possible, I will revert the changes you made to the `init_mtls()` behavior on #1094.

maugustosilva:
> @maugustosilva Hi, I'd like to ask if you can use the [`get_tls_options()`](https://github.com/ansasaki/keylime/blob/00d80dea1934fedd66b28e86899e11902f418460/keylime/web_util.py#L214-L225) I introduced in this PR to obtain keys and certificates from the configuration file instead of changing the `init_mtls()` behavior. If it is possible, I will revert the changes you made to the `init_mtls()` behavior on #1094.

If you can return full paths with this new [`get_tls_options()`](https://github.com/ansasaki/keylime/blob/00d80dea1934fedd66b28e86899e11902f418460/keylime/web_util.py#L214-L225), I would certainly prefer to use your solution.

ansasaki:
> > @maugustosilva Hi, I'd like to ask if you can use the [`get_tls_options()`](https://github.com/ansasaki/keylime/blob/00d80dea1934fedd66b28e86899e11902f418460/keylime/web_util.py#L214-L225) I introduced in this PR to obtain keys and certificates from the configuration file instead of changing the `init_mtls()` behavior. If it is possible, I will revert the changes you made to the `init_mtls()` behavior on #1094.
> 
> If you can return full paths with this new [`get_tls_options()`](https://github.com/ansasaki/keylime/blob/00d80dea1934fedd66b28e86899e11902f418460/keylime/web_util.py#L214-L225), I would certainly prefer to use your solution.

All paths returned by `get_tls_options()` are absolute paths. You can switch from the server to the client files by providing `is_client=True`.

ansasaki:
Today I manually tested the current code in this PR using the `Multihost/basic-attestation` tests with the changes I proposed in [this PR](https://github.com/RedHat-SP-Security/keylime-tests/pull/196) and, at least in the basic attestation scenario, each component worked well having only their own configuration file (plus `ca.conf` and `logging.conf`) in `/etc/keylime/`.

I used 4 machines like this:
1. running the verifier with only `/etc/keylime/verifier.conf`, `/etc/keylime/ca.conf`, and `/etc/keylime/logging.conf`
2. running the registrar with only `/etc/keylime/registrar.conf`, `/etc/keylime/ca.conf`, and `/etc/keylime/logging.conf`
3. running the tenant and an agent with `/etc/keylime/tenant.conf`, `/etc/keylime/agent.conf`, `/etc/keylime/ca.conf`, and `/etc/keylime/logging.conf`
4. running a second agent with `/etc/keylime/agent.conf`, `/etc/keylime/ca.conf`, and `/etc/keylime/logging.conf`

All worked as expected which makes me believe the latest changes I introduced were sufficient to make each component use their own configuration (plus `ca.conf` and `logging.conf`)

ansasaki:
@maugustosilva Hello, the configuration conversion script I mentioned is [this one](https://github.com/ansasaki/keylime/blob/conf_refactor/scripts/convert_config.py).

Using it you can point to a `keylime.conf` and an output directory and it will generate the split configuration files keeping the values found on the `keylime.conf` (or use a default value if the option is missing).
```
$ mkdir output
$ cd scripts
$ python3 convert_config.py --input ../keylime.conf --out ../output
```
From here I'll provide some details (that probably should go to some documentation later):

If no input file is found, it will use the configuration currently installed in the system (from `/usr/etc/keylime.conf` or `/etc/keylime.conf`). You can check the possible parameters accepted by the script by passing `--help` option.

The algorithms is basically this:

* Read the old `keylime.conf`
* For each version found (in crescent order) in the provided templates directory (`./templates` if no path was provided):
  * Apply the transformation defined by the `mapping.json` file
  * Run the `adjust.py` script for special adjustments
* Create the new configuration files by replacing the values in the templates from the latest version

The mappings are declared in a [JSON file](https://github.com/ansasaki/keylime/blob/conf_refactor/scripts/templates/2.0/mapping.json). It basically defines the option names in the old configuration file, the new option name in the new configuration file, and the default value to use in case the option is not found.

The goal of this script is to be used in future to upgrade configuration files. The plan is to add new mappings, templates, and adjust script for each configuration file version.

ansasaki:
@mpeters The code scanning from CodeQL is failing due to false positives that I cannot dismiss (which is understandable as the author of PRs should not be able to hide eventual leaks).

I think only a maintainer can waive the failures.

mpeters:
@ansasaki I verified that they were false positives and cleared them.

ansasaki:
@aplanas @THS-on @maugustosilva Could you please review this once more?

I addressed all the suggestions since the last review (the only one that was left out was a change when getting the `tpm_policy` from args in the tenant).

The other changes I added were related with making each component to use only their own configuration file and to address a request from @maugustosilva to add a way to configure the requests timeout.

Let me know if any change is needed. When this gets approved, we can then drop the commit that changes the branch used for testing before merging.

aplanas (approved):
Thanks a lot for this work. Code-wise LGTM.

THS-on (approved):
LGTM from the code side and my manual test setup now only captures a subset of the e2e test, so I'll trust the e2e tests.

Note: we should drop the temporary test branch change commit now.

ansasaki:
Great, thank you!

Note: After dropping the commit the tests will most probably fail due to the configuration changes.

sergio-correia (approved):
Thanks, @ansasaki!

ansasaki:
> Still debugging, but we will have to hold off the merging for now

@maugustosilva Could you provide a configuration example that caused the issue?

maugustosilva:
> > Still debugging, but we will have to hold off the merging for now
> 
> @maugustosilva Could you provide a configuration example that caused the issue?
> 
> Edit: I've seen something like this before, but it was caused by permission issues (not on configuration files, but on the used key and certificate files) Edit2: Probably not your problem as you are running as root, which is even more confusing how you managed to get that using default configurations.

Ok, the fix was **trivial** (a symlink to the new default file - `server-*` - on `/var/lib/keylime/cv_ca`). I withdraw my objection my good man! :-)

maugustosilva (approved):

ansasaki:
> > > Still debugging, but we will have to hold off the merging for now
> > 
> > 
> > @maugustosilva Could you provide a configuration example that caused the issue?
> > Edit: I've seen something like this before, but it was caused by permission issues (not on configuration files, but on the used key and certificate files) Edit2: Probably not your problem as you are running as root, which is even more confusing how you managed to get that using default configurations.
> 
> Ok, the fix was **trivial** (a symlink to the new default file - `server-*` - on `/var/lib/keylime/cv_ca`). I withdraw my objection my good man! :-)

I know what happened:

- The old `keylime.conf` value for `private_key_pw` option is `default`
- The converter uses this value to set the `server_key_password` option
- In your previous run the key generated was not encrypted (as it was not supported)
- Then with the new config, the verifier did not overwrite the keys as they were already present. When trying to load the file, since the password was set as `default`, it tried to decrypt the key and failed (for some reason the returned error is `FileNotFound`)
- I reproduced locally by running the verifier with the old configuration, then changing branches, convert the config with the converter, and running the new version.

I think the best is to modify the `adjust.py` script to heuristically set the `server_key_password` as empty when the `private_key_pw` in the old configuration is `default`.  @maugustosilva That would be sufficient for you?

maugustosilva:
> > > > Still debugging, but we will have to hold off the merging for now
> > > 
> > > 
> > > @maugustosilva Could you provide a configuration example that caused the issue?
> > > Edit: I've seen something like this before, but it was caused by permission issues (not on configuration files, but on the used key and certificate files) Edit2: Probably not your problem as you are running as root, which is even more confusing how you managed to get that using default configurations.
> > 
> > 
> > Ok, the fix was **trivial** (a symlink to the new default file - `server-*` - on `/var/lib/keylime/cv_ca`). I withdraw my objection my good man! :-)
> 
> I know what happened:
> 
> * The old `keylime.conf` value for `private_key_pw` option is `default`
> * The converter uses this value to set the `server_key_password` option
> * In your previous run the key generated was not encrypted (as it was not supported)
> * Then with the new config, the verifier did not overwrite the keys as they were already present. When trying to load the file, since the password was set as `default`, it tried to decrypt the key and failed (for some reason the returned error is `FileNotFound`)
> * I reproduced locally by running the verifier with the old configuration, then changing branches, convert the config with the converter, and running the new version.
> 
> I think the best is to modify the `adjust.py` script to heuristically set the `server_key_password` as empty when the `private_key_pw` in the old configuration is `default`. @maugustosilva That would be sufficient for you?

Yes, this might save us from having to deal with open "issues" during the transition phase between the old and new regime.

ansasaki:
> > > Still debugging, but we will have to hold off the merging for now
> > 
> > 
> > @maugustosilva Could you provide a configuration example that caused the issue?
> > Edit: I've seen something like this before, but it was caused by permission issues (not on configuration files, but on the used key and certificate files) Edit2: Probably not your problem as you are running as root, which is even more confusing how you managed to get that using default configurations.
> 
> Ok, the fix was **trivial** (a symlink to the new default file - `server-*` - on `/var/lib/keylime/cv_ca`). I withdraw my objection my good man! :-)

OK, the actual issue is the change in the `default` meaning for the server key and certificate as you already figured out. Previously we used the FQDN to name the keys and certificates files for the verifier (probably I need some rest :)).

An option is to rename the files to `server-*` in `/var/lib/keylime/cv_ca`, but I don't know if it is worth the effort to make this part of the conversion script.

An alternative option is to make the verifier to check for the existence of all files before deciding not to generate new keys, which will make it to generate new keys (maybe this is the best approach).

ansasaki:
@maugustosilva The last commit I added should fix the corner case you found (thanks!) and will make testing with `tls_dir=generate` configuration option more resilient as it will re-generate the key and certificate if any of them are missing.

test/test_restful.py
```diff
@@ -108,13 +109,6 @@ def cmp(a, b):
 ek_tpm = None
 aik_tpm = None
 
-# Set up mTLS
-my_cert = config.get("tenant", "my_cert")
-my_priv_key = config.get("tenant", "private_key")
-cert = (my_cert, my_priv_key)
-tls_enabled = True
```
sergio-correia, line 115: Looks like this is still being used in further down in `test_035_test_delete_in_use_allowlist()`, which was added recently in #1088.

ansasaki: Yes, just fixed this, will push again.

ansasaki:
@THS-on @aplanas @maugustosilva Since your last review, I modified the verifier behavior when `tls_dir` is set as `generate` to check for the presence of all keys and certificates instead of checking only for the presence of the CA certificate to decide if it should generate new keys or not. If a key or certificate is missing they are re-generated.

Also, I took the liberty to separate the password that is used to protect the CA keystore from the password used to encrypt the server key, by adding a `password` option to the `ca.conf` (previously the password set in `server_key_password` was used for both). I'll update the enhancement proposal with this change. Let me know if this is acceptable.

If it is OK, then I'll drop again the temporary commit to change the branch used for the tests.

keylime/web_util.py
```diff
@@ -111,7 +111,12 @@ def init_tls_dir(component, logger=None):
             if logger:
                 logger.debug("Existing server certificate and key found in %s, not generating a new ones", tls_dir)
         else:
-            ca_util.cmd_mkcert(tls_dir, "server", password=password)
+            server_key_pw = config.get(component, "server_key_password")
+            if not server_key_pw:
+                server_key_pw = None
```
aplanas, line 116: This confuses me. If it is evaluated as False is because is `None`, `""` or `0` and you set as `None`, it is necessary or is the logic reverted and you want:

```python
if server_key_pw:
    server_key_pw = str(server_key_pw)
```

ansasaki: This is to avoid the empty string `""` as the underlying code treat `None` differently from `""`. The goal is to not encrypt the key with a password if the provided password is the empty string.

ansasaki: Ah, I think you are right and this is not needed anymore as the underlying code was also modified.
Let me check if it is true.

aplanas (approved):

keylime/web_util.py
ansasaki: Applied your suggestion, thanks!

ansasaki:
Dropped the temporary commit to change the branch used for tests.

Thank you all for you patience, I think this is ready to merge.

PR #1069, opened 2022-08-08
Throw an error if --exclude is used without --allowlist

mbestavros:

### DIFF
--- a/keylime/tenant.py
+++ b/keylime/tenant.py
@@ -1497,6 +1497,8 @@ def main(argv=sys.argv):  # pylint: disable=dangerous-default-value
     args = parser.parse_args(argv[1:])
 
     # Make sure argument dependencies are enforced
+    if args.ima_exclude and not args.allowlist:
+        parser.error("--exclude cannot be used without an --allowlist")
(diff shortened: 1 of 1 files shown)


keylime/tenant.py
```diff
@@ -205,6 +205,8 @@ def process_allowlist(self, args):
             # Process allowlists
             al_data["excllist"] = excl_data
             self.allowlist = al_data
+        elif excl_data is not None:
+            raise UserError("--allowlist must be used alongside --exclude")
```
mpeters, line 209: Maybe a slightly better wording is something like `--exclude cannot be used without an --allowlist`.

mpeters (changes requested):

keylime/tenant.py
mbestavros: Done!

aplanas (changes requested):
Is this validating command line parameters invariant? If so it should be done in https://github.com/keylime/keylime/blob/2a846b448d974ef7f8f5adda8a059ce2fafca63d/keylime/tenant.py#L1501

mpeters (approved):

Issue #1073, opened 2022-08-11
Since allowlist API implementation Keylime "local tests" show now SQL-related suspicious error.


maugustosilva:
# Environment
 * OS / version: Fedora Core 35
 * Processor architecture: x86_64
 * TPM Manufacturer: swtpm
 * Keylime version: commit `1a41a8ae91242d4179a81c86072ce89e0d225eb1`

# Description 
I see what seems to be a - silent, incorrectly ignored - error while running tests on `keylime`. I don't remember seeing the error before the changes in the database introduced by the `allowlist` API (@mbestavros please confirm)

## Expected behavior vs. actual behavior 
Local tests should complete without displaying errors.

## Steps to reproduce problem
1. `git clone https://github.com/keylime/keylime.git`
2. `cd ~/keylime/.ci; ./run_local.sh ${PWD}/..`

# Relevant logs
```
Captured stdout for test.test_restful.TestRestful.test_040_agent_quotes_integrity_get

2022-08-11 16:54:21.491 - keylime.cloudverifier - ERROR - (sqlite3.IntegrityError) UNIQUE constraint failed: allowlists.name
[SQL: INSERT INTO allowlists (name, tpm_policy, ima_policy) VALUES (?, ?, ?)]
[parameters: ('d432fbb3-d2f1-4a97-9ef7-75bd81c00000', None, '{"allowlist": {"meta": {"version": 5}, "release": 0, "hashes": {"boot_aggregate": ["0000000000000000000000000000000000000000", "000000000000000000000 ... (29 characters truncated) ... 00000000000000"]}, "keyrings": {}, "ima": {"ignored_keyrings": [], "log_hash_alg": "sha1", "dm_policy": null}, "ima-buf": {}}, "exclude": "WyIqIl0="}')]
(Background on this error at: http://sqlalche.me/e/13/gkpj)
Traceback (most recent call last):
  File "/usr/lib64/python3.9/site-packages/sqlalchemy/engine/base.py", line 1276, in _execute_context
    self.dialect.do_execute(
  File "/usr/lib64/python3.9/site-packages/sqlalchemy/engine/default.py", line 608, in do_execute
    cursor.execute(statement, parameters)
sqlite3.IntegrityError: UNIQUE constraint failed: allowlists.name

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/keylime/cloud_verifier_tornado.py", line 498, in post
    session.commit()
[...]

PR #1074, opened 2022-08-11
Refactor allowlist handling on verifier to prevent premature DB writes

mbestavros:
This PR refactors how allowlists are handled on the verifier, with the goal of catching a few edge cases where database writes could have happened before the verifier realized that an incoming request was incorrect.

Concrete example: user attempts to add an agent and allowlist, but provides an agent UUID that is already in use. Prior to this PR, adding the agent would fail with an error, while adding the allowlist would silently succeed. This would cause additional problems later on. This is fixed by making sure that the verifier won't attempt any database writes of any kind until all sanity checks have succeeded.

This PR also fixes a broken test that used incorrect dummy data (specifically, `test_040_agent_quotes_integrity_get` in `test_restful.py`).

Most of this PR is just moving existing code around; hopefully the end result also makes more logical sense to read than before. I've included a few comments where I thought they might be helpful.

Resolves https://github.com/keylime/keylime/issues/1073. Thanks to @maugustosilva for the report, which caused me to look further into this and write the fix.

### DIFF
--- a/keylime/cloud_verifier_tornado.py
+++ b/keylime/cloud_verifier_tornado.py
@@ -428,86 +428,6 @@ class AgentsHandler(BaseHandler):
                     logger.warning("POST returning 400 response. Expected non zero content length.")
                 else:
                     json_body = json.loads(self.request.body)
-
-                    # Handle allowlists/IMA policies
-
-                    # How each pair of inputs should be handled:
-                    # - No name, no policy: use default empty policy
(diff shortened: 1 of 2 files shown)


mbestavros:
CC @maugustosilva @THS-on

maugustosilva:
This PR seems to fix the problem, and all the end-to-end internal tests are properly passing.

THS-on (approved):

Issue #1080, opened 2022-08-19
tenant tpm_cert_store option is not processed properly


kkaarreell:
Is your an issue a feature request? If so, please raise it as an [enhancement](https://github.com/keylime/enhancements)

# Environment
 * OS / version: Fedora 35
 * Processor architecture: x86_64
 * TPM Manufacturer: swtpm
 * Keylime version: current upstream [d8968cb](https://github.com/keylime/keylime/commit/d8968cb6a011651d480f1dc96d7adfc0be0494e3)

# Description 
I have noticed the issue after configuring `tpm_cert_store` to a different directory.
`keylime_tenant` ended up with 
```
2022-08-19 21:33:52.155 - keylime.tpm - ERROR - No Root CA matched EK Certificate
2022-08-19 21:33:52.155 - keylime.tenant - WARNING - Invalid EK certificate
2022-08-19 21:33:52.156 - keylime.tenant - ERROR - TPM Quote from cloud agent is invalid for nonce: jv4lkOm4XhSIj1i8xpR0
```
I have noticed that in
https://github.com/keylime/keylime/blob/master/keylime/tpm_ek_ca.py#L25
the list is constructed in a weird way.
```
file_list = glob.glob(os.path.join(os.getcwd(), tpm_cert_store + "*.pem"))
```
I am not sure why `os.getcwd()` is there at all but given that `tpm_cert_store` is a directory it is my belief that the proper line should be
```
file_list = glob.glob(os.path.join(tpm_cert_store, "*.pem"))
```
since the user may not terminate the directory path with `/`. With this change `keylime_tenant` behaves as expected.

Edit: I guess that `os.getcwd()` is there in case `tpm_cert_store` represents relative path. I think this is rather misleading behavior and if that is really what's requested here then it should be mentioned in `keylime.conf` (resp. `keylime-agent.conf`).
@mpeters @THS-on Thoughts?
Edit2: Local path would make sense if we expect user to specify `tpm_cert_store` on cmdline but I don't see such an option for `keylime_tenant`\

PR #1082, opened 2022-08-22
Treat tpm_cert_store as absolute path

mpeters:
This fixes issue #1080

### DIFF
--- a/keylime/tpm_ek_ca.py
+++ b/keylime/tpm_ek_ca.py
@@ -22,7 +22,7 @@ def check_tpm_cert_store():
 
 
 def cert_loader():
-    file_list = glob.glob(os.path.join(os.getcwd(), tpm_cert_store + "*.pem"))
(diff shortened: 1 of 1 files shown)


keylime/tpm_ek_ca.py
```diff
@@ -22,7 +22,7 @@ def check_tpm_cert_store():
 
 
 def cert_loader():
-    file_list = glob.glob(os.path.join(os.getcwd(), tpm_cert_store + "*.pem"))
+    file_list = glob.glob(tpm_cert_store + "*.pem")
```
kkaarreell, line 25: @mpeters Please, use `tpm_cert_store, "*.pem"` instead. The former doesn't work when the directory path is not terminated with `/`.

mpeters: good catch.

kkaarreell:
LGTM, this is exactly the patch I was using.

kkaarreell (approved):

PR #1088, opened 2022-08-26
Prevent deletion of in-use allowlists + better error handling

mbestavros:
Previously, it was possible to delete an allowlist from the verifier while that allowlist was still in use by an agent. Doing so would break the agent on the verifier, and make it impossible to delete (and generally put the verifier in an unrecoverable state). This PR adds a safeguard against that scenario.

It also improves error messaging for `409` errors returned by the verifier; prior to the Allowlists API, that error was assumed to be an agent conflict, but now it can also be due to an allowlist conflict. Error messages should better reflect that now.

### DIFF
--- a/keylime/cloud_verifier_tornado.py
+++ b/keylime/cloud_verifier_tornado.py
@@ -509,7 +509,9 @@ class AgentsHandler(BaseHandler):
                         # Prevent overwriting existing IMA policies with name provided in request
                         if ima_policy and ima_policy_stored:
                             web_util.echo_json_response(
-                                self, 409, f"Allowlist with name {ima_policy_name} already exists"
+                                self,
+                                409,
+                                f"Allowlist with name {ima_policy_name} already exists. Please use a different name or delete the allowlist from the verifier.",
                             )
                             logger.warning("Allowlist with name %s already exists", ima_policy_name)
                             return
@@ -530,7 +532,11 @@ class AgentsHandler(BaseHandler):
                         raise e
 
                     if new_agent_count > 0:
-                        web_util.echo_json_response(self, 409, f"Agent of uuid {agent_id} already exists")
+                        web_util.echo_json_response(
+                            self,
+                            409,
+                            f"Agent of uuid {agent_id} already exists. Please use delete or update.",
+                        )
                         logger.warning("Agent of uuid %s already exists", agent_id)
                         return
 
@@ -739,7 +745,7 @@ class AllowlistHandler(BaseHandler):
 
         session = get_session()
         try:
-            session.query(VerifierAllowlist).filter_by(name=allowlist_name).one()
+            ima_policy = session.query(VerifierAllowlist).filter_by(name=allowlist_name).one()
         except NoResultFound:
             web_util.echo_json_response(self, 404, f"Allowlist {allowlist_name} not found")
             return
@@ -747,6 +753,19 @@ class AllowlistHandler(BaseHandler):
             logger.error("SQLAlchemy Error: %s", e)
             web_util.echo_json_response(self, 500, "Failed to get allowlist")
             raise
+
+        try:
+            agent = session.query(VerfierMain).filter_by(ima_policy_id=ima_policy.id).one_or_none()
+        except SQLAlchemyError as e:
+            logger.error("SQLAlchemy Error: %s", e)
+            raise
+        if agent is not None:
+            web_util.echo_json_response(
+                self,
+                409,
+                f"Can't delete allowlist as it's currently in use by agent {agent.agent_id}",
+            )
+            return
 
         try:
             session.query(VerifierAllowlist).filter_by(name=allowlist_name).delete()
--- a/keylime/cloud_verifier_tornado.py
+++ b/keylime/cloud_verifier_tornado.py
@@ -6,9 +6,12 @@ import sys
 import traceback
 from concurrent.futures import ThreadPoolExecutor
 from multiprocessing import Process
+from sqlite3 import Connection as SQLite3Connection
 
 import tornado.ioloop
 import tornado.web
+from sqlalchemy import event as saEvent
+from sqlalchemy.engine import Engine
 from sqlalchemy.exc import SQLAlchemyError
 from sqlalchemy.orm import joinedload
 from sqlalchemy.orm.exc import NoResultFound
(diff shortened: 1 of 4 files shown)


keylime/cloud_verifier_tornado.py
```diff
@@ -745,6 +745,15 @@ def delete(self):
             web_util.echo_json_response(self, 500, "Failed to get allowlist")
             raise
 
+        try:
+            agent = session.query(VerfierMain).filter_by(ima_policy_id=ima_policy.id).one_or_none()
+        except SQLAlchemyError as e:
+            logger.error("SQLAlchemy Error: %s", e)
+            raise
+        if agent is not None:
+            web_util.echo_json_response(self, 409, f"Allowlist in use by agent {agent.agent_id}")
```
mpeters, line 754: I would make this more descriptive. You've already provided the `409` code, so you can make the error message user-friendly. Something like "Can't delete allowlist as it's currently being used by agent ###".

keylime/tenant.py
```diff
@@ -662,7 +662,14 @@ def do_cv(self):
 
         if response.status_code == 409:
             # this is a conflict, need to update or delete it
-            logger.error("Agent %s already existed at CV. Please use delete or update.", self.agent_uuid)
+            message = response.json()["status"]
```
mpeters, line 665: If you make the text of the HTTP error message more user friendly then you can just re-use it here (for both of these cases) instead of trying to parse it and give some other take on it.

mpeters (changes requested):

mpeters:
This is a necessary but incomplete change. The API code should check if it's already in use and give a good error message, but there's still a race condition between when it's checked and when it's deleted and another request could try and use it in the meantime. So the enforcement needs to happen at the DB layer with foreign key constraints. Is that not happening for some reason? SQLite should support this.

mbestavros:
@mpeters I've pushed another commit that adds database-level referential integrity for SQLite. 

I've also tracked down a possible bug where database sessions aren't properly terminated when a SQLAlchemy error is thrown.

keylime/cloud_verifier_tornado.py
```diff
@@ -43,6 +46,13 @@
     logger.error("Error creating SQL engine or session: %s", err)
     sys.exit(1)
 
+@event.listens_for(Engine, "connect")
```
mpeters, line 49: Is this conditional on the engine being SQLite?

keylime/tenant.py
```diff
@@ -662,7 +662,7 @@ def do_cv(self):
 
         if response.status_code == 409:
             # this is a conflict, need to update or delete it
-            logger.error("Agent %s already existed at CV. Please use delete or update.", self.agent_uuid)
+            logger.error(json.dumps(response.json()))
```
mpeters, line 665: Is there a key in the json payload that has the error message? Iss this dumping a data structure?

mpeters (changes requested):

keylime/cloud_verifier_tornado.py
mbestavros: The event listener itself isn't, but the `PRAGMA foreign_keys=on` execution doesn't happen unless the database is SQLite. See here: https://github.com/keylime/keylime/pull/1088/files#diff-67033196f7dec303b3ea55793c6c13fb4b03a72a5515239475a871c3d638dfe4R51

keylime/tenant.py
mbestavros: @mpeters Yes - it's just printing out the entire JSON response from the verifier, wich includes the response code and error message. This is similar to how the [allowlist API commands](https://github.com/keylime/keylime/blob/master/keylime/tenant.py#L1277) handle responses without doing any parsing.

In general, the tenant's response handling is pretty inconsistent. Some commands (including the one I'm replacing) parse the response or response code in some form to deliver tailored output to the user, while others (notably the allowlist API) just forward whatever output came from the verifier directly.

I can tidy that up in a follow-up PR if we want.

mbestavros: Update - in hindsight, realized it was silly to not use the existing `_print_json_response` function if that's more or less all I'm doing, so that's been changed.

mpeters:
@mbestavros there's some lint failures that need to be fixed.

keylime/cloud_verifier_tornado.py
```diff
@@ -137,10 +146,10 @@ def store_attestation_state(agentAttestState):
             try:
                 session.add(update_agent)
             except SQLAlchemyError as e:
-                logger.error("SQLAlchemy Error on storing attestation state for agent %s: %s", agent_id, e)
+                logger.error("SQLAlchemy Error on storing attestation state: %s", e)
```
sergio-correia, line 149: This extra info (agent ID) was recently added by #1090. Is it intentional that you are removing it here? This also happens in other instances further on.

mbestavros: Very strange, this doesn't seem intentional... I might have messed up my rebase.

mpeters:
@mbestavros besides the lint failures, can you also add a test that covers this scenario so we make sure it doesn't break in the future?

mbestavros:
@mpeters I believe lints should pass now, and I've added an additional test in `test_restful` to cover the in-use deletion case. Anything else needed?

Thanks for the review!

test/test_restful.py
```diff
@@ -828,6 +828,17 @@ def test_034_cv_agent_post_invalid_exclude_list(self):
         json_response = response.json()
         self.assertIn("results", json_response, "Malformed response body!")
 
+    def test_035_test_delete_in_use_allowlist(self):
```
mpeters, line 831: This is a good test of this issue, but looking at the other tests, I don't see a test that successfully deletes an allowlist. `test_027_cv_allowlist_delete` and `test_035_test_delete_in_use_allowlist` both fail in different ways. Do we need another one to make sure deletes still work?

mbestavros: @mpeters Doesn't `test_027_cv_allowlist_delete` pass if successful? The error message is only returned if the status _doesn't_ match `204`, which is success.

mpeters: You're right, sorry I was reading the assert message and got my logic confused.

mpeters (approved):

PR #1090, opened 2022-08-29
Verifier log message improvements for large-scale testing.

maugustosilva:
While testing KeyLime at scale (hundreds to low thousands of rust
agents), the fact that several error/warn/debug log messages did NOT
include the agent UUID turned post-failure analysis significantly more
complex. This PR makes sure that all relevant messages are kept in a
single line, and all contain the agent's UUID as part of it.

### DIFF
--- a/keylime/cloud_verifier_common.py
+++ b/keylime/cloud_verifier_common.py
@@ -36,8 +36,10 @@ def process_quote_response(agent, ima_policy, json_response, agentAttestState) -
     failure = Failure(Component.QUOTE_VALIDATION)
     received_public_key = None
     quote = None
+    agent_id = None
     # in case of failure in response content do not continue
     try:
+        agent_id = agent["agent_id"]
         received_public_key = json_response.get("pubkey", None)
         quote = json_response["quote"]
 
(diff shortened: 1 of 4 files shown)


aplanas (commented):
Generally LGTM, but maybe would make more sense to adjust the logger config to show always the agent_id. The change will be much smaller and future proof: new log entries will have the agent_id included.

keylime/cloud_verifier_common.py
```diff
@@ -36,8 +36,10 @@ def process_quote_response(agent, ima_policy, json_response, agentAttestState) -
     failure = Failure(Component.QUOTE_VALIDATION)
     received_public_key = None
     quote = None
+    agent_id = "NA"
```
THS-on, line 39: Why introduce a magic constant here instead of using None?

keylime/tenant.py
```diff
@@ -1542,14 +1542,14 @@ def main(argv=sys.argv):  # pylint: disable=dangerous-default-value
         mytenant.registrar_ip = args.registrar_ip
     if args.registrar_port is not None:
         mytenant.registrar_port = args.registrar_port
-
+    interval = config.getfloat("tenant", "retry_interval")
```
THS-on, line 1545: Using the retry interval also as a timeout value seems hacky. 
#1068 already adds some timeout to these requests, so you can drop it from this PR.

THS-on (changes requested):
I added some comments. The idea from @aplanas to always add the agent ID to the log is a good idea.

keylime/cloud_verifier_common.py
maugustosilva: Could be None, sure

keylime/tenant.py
maugustosilva: If I drop from this PR, pylint will fail

THS-on: It is probably better to have a hard coded timeout (like 60s) instead of using the retry interval setting for the timeout.

maugustosilva:
> Generally LGTM, but maybe would make more sense to adjust the logger config to show always the agent_id. The change will be much smaller and future proof: new log entries will have the agent_id included.

This seems like a good idea. The whole logging system is due to an overhaul anyway, and I am willing to open a new PR who carefully implements it in the future. For now, the lack of agent ID in several of the log entries is a clear and present problem for anyone attempting to debug keylime communications/timing problems at scale.

keylime/tenant.py
maugustosilva: Fair enough

mpeters (approved):

mpeters:
@THS-on are the latest changes from @maugustosilva good for this to be merged?

THS-on (approved):

THS-on:
@mpeters yes this can be merged.

Issue #1091, opened 2022-08-30
Potential for unnoticed verifier failures in multi-verifier operation


galmasi:
# Environment
 * OS / version: any (not relevant)
 * Processor architecture: any (not relevant)
 * TPM Manufacturer: any (not relevant)
 * Keylime version: 6.4.3 (current as of 8/30/22)

# Description 

We have seen strange attestation "non-failures" (false negatives) in multi-verifier keylime systems during ongoing large scale testing. At least one theory of such non-failures, made possible by the current code path, is that in a multi-verifier system certain verifiers can die, stopping attestation on all their assigned agents, *without anyone noticing*.

The specific setup in our scaling test involves multiple verifiers and registrars. These run on virtual machines. In steady state operation (that is, all agents being attested) there is no user (tenant) interaction with either the registrar or the verifier on such virtual machines ... the tenant queries the database by talking to exactly one verifier. 

The database does not contain information about *when* the last attestation attempt was made for any agent. Thus any verifier could potentially crash, not be restarted, and its absence will never be noticed. Agents managed by any particular verifier will not be reassigned until the tenant attempts to delete and re-add them. Agents may fail or break and the database will continue to state that the last attestation attempt passed.

We consider this an extremely serious situation -- a "false negative" in which keylime continues to claim "all is well" in a situation where large portions of a cluster are no longer being attested.

## Expected behavior vs. actual behavior 

Expected behavior: crashed verifiers -- better yet, verifiers inactive for some time X -- should cause all agents managed by the verifier to fail.

Actual behavior: crashed verifiers do not cause any change in system behavior -- all agents continue to show attested.

## Steps to reproduce problem

1. Create a multiple-verifier keylime system (single-database-multiple-verifiers) with at least 2 verifiers.
2. Set up "steady state" attestation in which every verifier has at least one agent to attest.
3. Pick one verifier that is not being talked to directly by the keylime tenant. Destroy the verifier's process without giving it time to mark the agent states (e.g. use SIGKILL to destroy the process; we experimentally murdered the entire docker process running that particular verifier).
4. Observe that the database no longer receives updates for the affected agents
5. Observe that keylime tenant queries will continue to mark the entire system as attested
6. Observe that the keylime agents previously being queried by the now-dead verifier are now idle.

## A complicated non-fix that should _not_ be considered

[...]

PR #1093, opened 2022-08-31
This commit solves #1091 by adding a per-agent verifier-issued epoch timestamp

maugustosilva:
A new DB column, containing an epoch timestamp (from the verifier)
records when a quote was received. This new column is then exposed as an
additional parameter on the API (whenever agent status is requested).

In case a verifier thread dealing with an  agent dies (while the
verifier process remains responsive) or in case of a multi-verifier
configuration (where a whole verifier could die, while one keeps
requesting status from a different verifier), there is a way for an user
to make a decision about the "age" of the last successful attestation.

### DIFF
--- a/keylime/db/verifier_db.py
+++ b/keylime/db/verifier_db.py
@@ -1,4 +1,4 @@
-from sqlalchemy import Column, ForeignKey, Integer, LargeBinary, PickleType, String, Text, schema
+from sqlalchemy import Column, Float, ForeignKey, Integer, LargeBinary, PickleType, String, Text, schema
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.orm import relationship
 
@@ -47,6 +47,7 @@ class VerfierMain(Base):
     ak_tpm = Column(String(500))
     mtls_cert = Column(String(2048), nullable=True)
     attestation_count = Column(Integer)
+    last_received_quote = Column(Float)
     tpm_clockinfo = Column(JSONPickleType(pickler=JSONPickler))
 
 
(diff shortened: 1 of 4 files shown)


keylime/db/verifier_db.py
```diff
@@ -47,6 +47,7 @@ class VerfierMain(Base):
     ak_tpm = Column(String(500))
     mtls_cert = Column(String(2048), nullable=True)
     attestation_count = Column(Integer)
+    last_received_quote = Column(Float)
```
mpeters, line 50: Why is this a float? I thought epoch timestamps were plain integers.

keylime/migrations/versions/a09cc94177f0_add_last_received_quote.py
```diff
@@ -0,0 +1,39 @@
...
+def upgrade_registrar():
+    pass
+
+
+def downgrade_registrar():
+    pass
+
+
+def upgrade_cloud_verifier():
+    op.add_column("verifiermain", sa.Column("last_received_quote", sa.Float))
```
mpeters, line 35: When the column is added, what is the default? I think for numbers that weren't set as nullable it's likely 0. Or do we want to be explicit and set 0 as the default since that's what we do when we create the object anyway.

keylime/db/verifier_db.py
maugustosilva: In python3, `time.time()` is a `float`

mpeters: I don't have strong feelings either way, but do we ever need subsecond precision for this? Ints are a lot more efficient than floats, but it's not like we're doing any sort of high speed calculation on this field so it's not really going to matter. So I guess whatever is easiest to use is best.

keylime/migrations/versions/a09cc94177f0_add_last_received_quote.py
maugustosilva: Good point, I have added `nullable=True`

mpeters (approved):

PR #1100, opened 2022-09-12
A fix for "per-agent verifier-issued epoch timestamp"

maugustosilva:
In some database backends, such as MySQL, the use of a float as column lead to representation and rounding errors. In order to preserve the information in the most backend-agnostic way possible, we will simply use string (varchar)

### DIFF
--- a/keylime/cloud_verifier_common.py
+++ b/keylime/cloud_verifier_common.py
@@ -181,7 +181,7 @@ def process_quote_response(agent, ima_policy, json_response, agentAttestState) -
     )  # TODO: change this to always False after initial update
     failure.merge(quote_validation_failure)
 
-    agent["last_received_quote"] = time.time()
+    agent["last_received_quote"] = int(time.time())
 
     if not failure:
         agent["attestation_count"] += 1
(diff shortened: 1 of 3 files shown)


mpeters:
I'm hesitant to make this a non-number type since it makes it difficult to do any math on the database and it can be useful for things like a last time something happened. But I get the problem with floats...

So I would prefer one of 2 approaches: Using the database's timestamp types and then convert in the db layer. Or use an integer (just truncate off the partial seconds off the timestamp).

maugustosilva:
@mpeters I fully understand your hesitancy, and I make it mine. 

Here is my problem: I really, really would like to capture at least fractions of seconds, and I am having a devil of time getting my version of  `MySQL` (8.0.x) to preserve this information (it keeps truncating it, while works fine on `sqlite`). I could go on the `sqlalchemy.dialects.mysql` route, but then I don't know how `PostgresSQL` will behave. Would you be willing to reconsider?

mpeters:
What about making it an integer of milliseconds instead of seconds? So multiply by a thousand on serialize and divide on deserialization?

maugustosilva:
@mpeters as discussed, I have switched to `integer`.

mpeters (approved):

mpeters:
@maugustosilva do you need to rebase to pick up the new config file changes?

maugustosilva:
@mpeters can we go ahead and merge it?

mpeters (approved):

PR #1101, opened 2022-09-14
Move SQLite ref integrity pragma to keylime_db

mpeters:
This will make sure it's consistent across verifier, registrar and tenant


mbestavros (approved):
Makes sense - and it's good we have another test for the referential integrity. LGTM!

PR #1103, opened 2022-09-15
Fix testing with the rust agent

ansasaki:
Fix testing with the rust agent (triggered by setting the `RUST_TEST` environment variable)

This depends on:

- https://github.com/keylime/rust-keylime/pull/448
  - Uses the `tpm_ownerpassword` when generating the EK
- https://github.com/keylime/rust-keylime/pull/432
  - Add the `tpm_ownerpassword` to the `keylime-agent.conf`

One test is still skipped due to https://github.com/keylime/rust-keylime/issues/447

Add a test case for the `/keys/verify` endpoint for the rust agent

### DIFF
--- a/test/test_restful.py
+++ b/test/test_restful.py
@@ -1016,6 +1016,10 @@ class TestRestful(unittest.TestCase):
     def test_076_agent_quotes_integrity_get(self):
         self.test_040_agent_quotes_integrity_get()
 
+    @unittest.skipIf(SKIP_RUST_TEST, "Testing against rust-keylime is disabled!")
+    async def test_077_agent_keys_verify_get(self):
+        await self.test_041_agent_keys_verify_get()
+
     def tearDown(self):
         """Nothing to bring down after each test"""
         return
--- a/test/test_restful.py
+++ b/test/test_restful.py
@@ -191,6 +191,9 @@ def tearDownModule():
     kill_cloudverifier()
     kill_registrar()
 
+    # Run tpm2_clear to allow other processes to use the TPM
+    subprocess.run("tpm2_clear", stdout=subprocess.PIPE, check=False)
+
 
 def launch_cloudverifier():
     """Start up the cloud verifier"""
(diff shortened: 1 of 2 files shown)


ansasaki:
Note that when `RUST_TEST=1` is set, the tests are executed first with the python agent, then with the rust agent.

The changes related with the owner password on rust agent are to allow it to access the TPM after the python agent took its onwership and set a password.

test/test_restful.py
```diff
@@ -367,7 +368,7 @@ def setUpClass(cls):
         cls.auth_tag = crypto.do_hmac(cls.K, tenant_templ.agent_uuid)
 
         # Prepare policies for agent
-        cls.tpm_policy = '{"22":["0000000000000000000000000000000000000001","0000000000000000000000000000000000000000000000000000000000000001","000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001","ffffffffffffffffffffffffffffffffffffffff","ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff","ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff"],"15":["0000000000000000000000000000000000000000","0000000000000000000000000000000000000000000000000000000000000000","000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"]}'
+        cls.tpm_policy = "{}"
```
lkatalin, line 371: Why the policy change here?

ansasaki: To simplify the test using the empty policy. I can revert this if you prefer.

lkatalin: No preference, just curiosity :)

lkatalin (approved):

mpeters (approved):

test/test_restful.py
kgoldman: Good security says that a policy should be empty when there is none.  While a policy of all zeros or all ones likely cannot be satisfied, empty is easier to test and validate.

I.e., it's the difference between
if (empty)
and
if (empty or all zeros or all ones)

PR #1104, opened 2022-09-15
New readthedocs location for keylime

mpeters:
After the move from the keylime-docs repo into the docs/ dir of the main repo, we need to update readthedocs to point to the new docs which will also follow the version tags.


keylime.conf
```diff
@@ -34,12 +34,12 @@ rsa_keyname = tci_rsa_key
 # Enable mTLS communication between agent, verifier and tenant.
 # Details on why setting it to "False" is generally considered insecure can be found
 # on https://github.com/keylime/keylime/security/advisories/GHSA-2m39-75g9-ff5r
-mtls_cert_enabled = True
```
ansasaki, line 37: I think the changes to the configuration file other than the URL for the documentation was not intentional, right?

ansasaki (changes requested):

keylime.conf
maugustosilva: @mpeters I have the same question. Shouldn't it be on a different PR?

mpeters: yes, sorry accidentally included my local dev changes without looking

THS-on (approved):

PR #1113, opened 2022-09-26
Implement IMA policy conversion script

mbestavros:
This script should serve as an upgrade path for the eventual IMA policy format overhaul, as laid out in Keylime enhancement https://github.com/keylime/enhancements/pull/71.

Note that this is NOT a complete implementation of the IMA policy format overhaul; it merely produces valid-formatted IMA policies according to the enhancement.

### DIFF
--- a/keylime/cmd/convert_ima_policy.py
+++ b/keylime/cmd/convert_ima_policy.py
@@ -1,0 +1,196 @@
+import argparse
+import copy
+import datetime
+import json
+import sys
+from os.path import basename
+
+from keylime.ima import file_signatures, ima
+
+# pylint: disable=pointless-string-statement
+"""
+This script converts legacy Keylime policies into the unified, next-generation
+IMA policy format. Both legacy allow/exclude lists as well as out-of-date IMA
+policy formats are accepted.
+
+Example usage:
+
+To convert an allowlist and exclude list to an IMA policy at `keylime-policy.json`:
+
+```
+python convert_ima_policy.py -a <allowlist path> -e <exclude list path> -o keylime-policy.json
+```
+
+To convert an existing IMA policy to the latest version:
+
+```
+python convert_ima_policy.py -i <ima policy path> -o keylime-policy.json
+```
+
+To view help and see all available options:
+
+```
+python convert_ima_policy.py -h
+```
+"""
+
+# Creates an IMA policy from provided legacy allowlist.
+def convert_legacy_allowlist(allowlist_path):
+    with open(allowlist_path, "r", encoding="utf8") as f:
+        alist_raw = f.read()
+
+    # Attempt to load JSON, and convert the appropriate format
+    try:
+        alist_json = json.loads(alist_raw)
+        print(f"{basename(allowlist_path)} appears to be JSON-formatted; converting to IMA policy")
+        ima_policy = _convert_json_allowlist(alist_json)
+    except Exception as _:
+        print(
+            f"{basename(allowlist_path)} is not JSON-formatted; attempting to convert to IMA policy from flat file format"
+        )
+        ima_policy = _convert_flat_format_allowlist(alist_raw)
+    return ima_policy
+
+
+# Converts JSON-format allowlist to JSON-format IMA policy
+def _convert_json_allowlist(alist_json):
+    ima_policy = copy.deepcopy(ima.EMPTY_IMA_POLICY)
+    ima_policy["meta"]["timestamp"] = str(datetime.datetime.now())
+    ima_policy["meta"]["generator"] = "convert_ima_policy.py"
+    for key in ima_policy.keys():
+        if key == "digests":
+            digests = alist_json.get("hashes")
+            if not digests:
+                print("Allowlist does not have a valid hash list!")
+            else:
+                ima_policy[key] = alist_json["hashes"]
+        elif key == "meta":
+            # Skip old metadata
+            continue
+        else:
+            to_migrate = alist_json.get(key)
+            if not to_migrate:
+                print(f"IMA policy field '{key}' not found in allowlist; using default value")
+            else:
+                ima_policy[key] = alist_json[key]
+    return ima_policy
+
+
+# Converts flat-format allowlist to JSON-format IMA policy
+def _convert_flat_format_allowlist(alist_raw):
+    ima_policy = copy.deepcopy(ima.EMPTY_IMA_POLICY)
+    ima_policy["meta"]["timestamp"] = str(datetime.datetime.now())
+    ima_policy["meta"]["generator"] = "convert-ima-policy"
+
+    lines = alist_raw.splitlines()
+    for line_num, line in enumerate(lines):
+        line = line.strip()
+        if len(line) == 0:
+            continue
+
+        pieces = line.split(None, 1)
+        if not len(pieces) == 2:
+            print(f"Line #{line_num} in Allowlist does not consist of hash and file path: {line}")
+            continue
+
+        (checksum_hash, path) = pieces
+
+        if path.startswith("%keyring:"):
+            entrytype = "keyrings"
+            path = path[len("%keyring:") :]  # remove leading '%keyring:' from path to get keyring name
+        else:
+            entrytype = "digests"
+
+        if path in ima_policy[entrytype]:
+            ima_policy[entrytype][path].append(checksum_hash)
+        else:
+            ima_policy[entrytype][path] = [checksum_hash]
+    return ima_policy
+
+
+# Updates an existing IMA policy to the latest version, and adds any provided input
+def update_ima_policy(policy, excludelist_path=None, verification_keys=None):
+    if policy["meta"]["version"] < ima.IMA_POLICY_CURRENT_VERSION:
+        print(
+            f"Provided policy has version {policy['meta']['version']}; latest policy has version {ima.IMA_POLICY_CURRENT_VERSION}. Updating to latest version."
+        )
+        updated_policy = copy.deepcopy(ima.IMA_POLICY_CURRENT_VERSION)
+        updated_policy["meta"]["timestamp"] = str(datetime.datetime.now())
+        updated_policy["meta"]["generator"] = "convert_ima_policy.py"
+        for key in updated_policy.keys():
+            if key == "meta":
+                continue
+            to_migrate = policy.get(key)
+            if not to_migrate:
+                print(f"IMA policy field '{key}' not found in existing IMA policy; using default value")
+            else:
+                updated_policy[key] = policy[key]
+        policy = updated_policy
+
+    excl_list = []
+    if excludelist_path:
+        with open(excludelist_path, "r", encoding="utf8") as f:
+            for line in f:
+                line = line.strip()
+                if line.startswith("#") or len(line) == 0:
+                    continue
+                excl_list.append(line)
+
+    verification_key_list = None
+    if verification_keys:
+        keyring = file_signatures.ImaKeyring().from_string(policy["ima_sign_verification_keys"])
+        for key in verification_keys:
+            pubkey, keyidv2 = file_signatures.get_pubkey_from_file(key)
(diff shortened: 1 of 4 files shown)


mbestavros:
An update for those following: I've implemented support for signing Keylime policies in the DSSE format (spec, for reference: https://github.com/secure-systems-lab/dsse).

I'm also marking this as a draft, since it currently relies on a Git dependency for in-toto's [downstream `securesystemslib` implementation](https://github.com/in-toto/securesystemslib). When DSSE support is upstreamed, this dependency can be made a normal `pip` dependency and will then be eligible for merge.

mbestavros:
@mpeters @THS-on I've dropped DSSE signing for now, since the status of required dependencies for that feature is in flux. The conversion script should be ready for review - please let me know your feedback!

scripts/convert_ima_policy.py
```diff
@@ -0,0 +1,173 @@
...
+"""
+
+# Creates an IMA policy from provided legacy allowlist.
+def create_ima_policy(allowlist_path):
+    with open(allowlist_path, "r") as f:
+        alist_raw = f.read()
+
+    p = re.compile(r"^\s*{")
+    if p.match(alist_raw):
+        print(f"{basename(allowlist_path)} is JSON-formatted; converting to IMA policy")
```
mpeters, line 44: Since you're just using a regex to determine if it's likely JSON, I would change the wording to "appears to be JSON-formatted;..." just in case there are errors it would then be easier to see that we were making assumptions.

scripts/convert_ima_policy.py
```diff
@@ -0,0 +1,173 @@
...
+        alist_json = json.loads(alist_raw)
```
mpeters, line 45: Should do this in a try/catch block because it's user input and will fail in all kinds of weird ways. Best to catch it and give a friendlier error message.

scripts/convert_ima_policy.py
```diff
@@ -0,0 +1,173 @@
...
+        ima_policy = copy.deepcopy(ima.EMPTY_IMA_POLICY)
+        ima_policy["meta"]["timestamp"] = str(datetime.datetime.now())
+        ima_policy["meta"]["generator"] = "convert-ima-policy"
+        for key in alist_json.keys():
```
mpeters, line 49: I would do this the other way around: loop through the keys from the empty policy we know is good, rather than whatever keys might be present in the user data. That way if there are missing keys or bad keys they aren't just ignored or copied over, but we can give actual error messages. You'll still need your special handling for hashes vs digests.

scripts/convert_ima_policy.py
```diff
@@ -0,0 +1,173 @@
...
+
+# Updates an existing IMA policy to the latest version, and adds any provided input
+def update_ima_policy(policy, excludelist_path, verification_keys):
+    if policy["meta"]["version"] < ima.IMA_POLICY_CURRENT_VERSION:
+        print(
+            f"Provided policy has version {policy['meta']['version']}; latest policy has version {ima.IMA_POLICY_CURRENT_VERSION}. Updating to latest version."
+        )
+        updated_policy = copy.deepcopy(ima.IMA_POLICY_CURRENT_VERSION)
+        updated_policy["meta"]["timestamp"] = str(datetime.datetime.now())
+        updated_policy["meta"]["generator"] = "convert-ima-policy"
```
mpeters, line 70: This should probably just be the script name instead of a something that's kinda like it.

scripts/convert_ima_policy.py
```diff
@@ -0,0 +1,173 @@
...
+        policy["verification-keys"].append(verification_key_list)
+
+    return policy
+
+
+# Converts flat-format allowlist to JSON-format IMA policy
+def convert_legacy_allowlist(alist_raw):
+    ima_policy = copy.deepcopy(ima.EMPTY_IMA_POLICY)
+    ima_policy["meta"]["timestamp"] = str(datetime.datetime.now())
+    ima_policy["meta"]["generator"] = "convert-ima-policy"
```
mpeters, line 107: Same as above.

scripts/convert_ima_policy.py
```diff
@@ -0,0 +1,173 @@
...
+
+    for line in alist_raw.splitlines():
+        line = line.strip()
+        if len(line) == 0:
+            continue
+
+        pieces = line.split(None, 1)
+        if not len(pieces) == 2:
+            print("Line in Allowlist does not consist of hash and file path: %s", line)
```
mpeters, line 116: I would phrase this as "Line #x in Allowlist..."

mpeters (changes requested):

scripts/convert_ima_policy.py
mbestavros: @mpeters Done in both loops of this type in the script, and I also added more debug info in case keys aren't found.

mbestavros:
@mpeters Addressed your feedback; ready for another review.

THS-on:
I think it is similar to the config script also a good idea to put this script under `keylime/cmd` because it is an official part of Keylime. 

In general it LGTM, but I still need to test it. Can you add some unit tests to that script?

mbestavros:
@THS-on I implemented a basic test suite. Let me know how it looks.

scripts/convert_ima_policy.py
```diff
@@ -0,0 +1,250 @@
...
+        sys.stderr.write("error: %s\n" % message)
+        self.print_help(sys.stderr)
+        sys.exit(2)
+
+
+if __name__ == "__main__":
+    main()
+
+
+class TestIMAVerification(unittest.TestCase):
```
aplanas, line 194: Can we move this out?

scripts/convert_ima_policy.py
```diff
@@ -0,0 +1,250 @@
...
+                    ima_policy[key] = alist_json[key]
+    else:
+        print(f"{basename(allowlist_path)} is not JSON-formatted; converting to IMA policy from flat file format")
+        ima_policy = convert_legacy_allowlist(alist_raw)
+    return ima_policy
+
+
+# Updates an existing IMA policy to the latest version, and adds any provided input
+def update_ima_policy(policy, excludelist_path=None, verification_keys=None, excl_list=[]):
+    if policy["meta"]["version"] < ima.IMA_POLICY_CURRENT_VERSION:
```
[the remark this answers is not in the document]
mbestavros: @aplanas No, version 1 is the only version as of now. I tried designing this function with an eye towards futureproofing - perhaps if we want to add new fields in a version 2, we can run it through this same function.

You are right that, as it stands, this code won't actually be used at present. Only when we create a version 2.

scripts/convert_ima_policy.py
```diff
@@ -0,0 +1,250 @@
...
+    p = re.compile(r"^\s*{")
+    if p.match(alist_raw):
+        print(f"{basename(allowlist_path)} appears to be JSON-formatted; converting to IMA policy")
+        try:
+            alist_json = json.loads(alist_raw)
+        except Exception as e:
+            print(f"Error loading JSON-formatted Keylime policy: {repr(e)}")
+        ima_policy = copy.deepcopy(ima.EMPTY_IMA_POLICY)
+        ima_policy["meta"]["timestamp"] = str(datetime.datetime.now())
+        ima_policy["meta"]["generator"] = "convert_ima_policy.py"
```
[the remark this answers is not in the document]
mbestavros: It's been present in older versions of allowlists internally, and some tests check for these fields, but we can drop if we feel they aren't useful.

scripts/convert_ima_policy.py
mbestavros: @aplanas (and @THS-on) Where do you propose would be the best spot to put these tests?

scripts/convert_ima_policy.py
```diff
@@ -0,0 +1,250 @@
...
+import sys
+import tempfile
+import unittest
+from os.path import basename
+
+from keylime.ima import file_signatures, ima
+
+"""
+This script converts legacy Keylime policies into the unified, next-generation
+IMA policy format.Both legacy allow/exclude lists as well as out-of-date IMA
```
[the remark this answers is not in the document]
mbestavros: Fixed!

scripts/convert_ima_policy.py
```diff
@@ -0,0 +1,250 @@
...
+```
+"""
+
+# Creates an IMA policy from provided legacy allowlist.
+def create_ima_policy(allowlist_path):
+    with open(allowlist_path, "r") as f:
+        alist_raw = f.read()
+
+    p = re.compile(r"^\s*{")
+    if p.match(alist_raw):
```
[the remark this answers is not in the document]
mbestavros: I no longer use regex in my latest version, so this is resolved.

scripts/convert_ima_policy.py
```diff
@@ -0,0 +1,250 @@
...
+        print(f"{basename(allowlist_path)} appears to be JSON-formatted; converting to IMA policy")
+        try:
+            alist_json = json.loads(alist_raw)
```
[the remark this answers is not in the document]
mbestavros: Done in my latest push.

scripts/convert_ima_policy.py
```diff
@@ -0,0 +1,250 @@
...
+                continue
+            else:
+                to_migrate = alist_json.get(key)
+                if not to_migrate:
+                    print(f"IMA policy field '{key}' not found in allowlist; using default value")
+                else:
+                    ima_policy[key] = alist_json[key]
+    else:
+        print(f"{basename(allowlist_path)} is not JSON-formatted; converting to IMA policy from flat file format")
+        ima_policy = convert_legacy_allowlist(alist_raw)
```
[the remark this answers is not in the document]
mbestavros: @aplanas I've spun out the JSON-format allowlist conversion into its own function and hopefully tidied up the naming. Let me know how it looks.

THS-on:
I would say this is nearly ready to go. Following changes I think are still necessary:
* drop changes to the create_allowlist script (currently not really useful)
* Add the script entry point to https://github.com/keylime/keylime/blob/master/setup.cfg
* squash commits together: 1x for script, 1x for the tests, 1x for the changes to ima.py

mbestavros:
@THS-on I've made your requested changes and squashed/tidied up my commits - there are now just three commits, one for the reference IMA policy in ima.py, one for the conversion script, and one for the test suite.

@mpeters Would love for you to have another look - anything else you think is needed before merge?

keylime/cmd/convert_ima_policy.py
```diff
@@ -0,0 +1,195 @@
...
+    if len(sys.argv) == 1:
+        parser.print_help()
+        parser.exit()
+
+    args = parser.parse_args()
+    if (bool(args.allowlist) and not bool(args.ima_policy)) and (not bool(args.allowlist) and bool(args.ima_policy)):
+        print("Either --allowlist or --ima_policy is required!")
+        sys.exit(1)
+    elif not args.output_file:
+        print("An output file path (-o, --output_file) is required to write new policy!")
```
THS-on, line 175: Here a `sys.exit(1)` missing:

```
An output file path (-o, --output_file) is required to write new policy!
allowlist is not JSON-formatted; attempting to convert to IMA policy from flat file format
  File "/usr/local/bin/keylime_convert_ima_policy", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.10/site-packages/keylime/cmd/convert_ima_policy.py", line 183, in main
    with open(args.output_file, "wb") as f:
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

THS-on (changes requested):
I now tested it a little bit more, I only found one small thing other wise LGTM!

THS-on (approved):

aplanas (approved):

mbestavros:
I'm assuming the codecov failure is spurious - @mpeters @THS-on how would I re-run that?

mpeters (approved):

PR #1115, opened 2022-09-29
Small changes required by enhancement #73 "Durable (Offline) Attestation"

maugustosilva:
First of all, several IMA and MB options on `tenant` are now processed on a separate "common" module. Second, a small change is introduced on the `verifier` method `process_quote_response`, where we pass an actual IMA policy (defined as a JSON) instead of a sqlalchemy object.

### DIFF
--- a/keylime/common/cli_opts.py
+++ b/keylime/common/cli_opts.py
@@ -1,0 +1,143 @@
+import sys
+
+from keylime import config, keylime_logging, measured_boot
+from keylime.ima import file_signatures, ima
+from keylime.tpm.tpm_abstract import TPM_Utilities
+
+logger = keylime_logging.init_logging("cli_opts")
+
+
+class UserError(Exception):
+    pass
+
+
+def enforce_pcrs(tpm_policy, policy_pcrs, protected_pcrs, pcr_use):
+    policy_pcrs = list(tpm_policy.keys())
+    policy_pcrs.remove("mask")
+
+    for _pcr in policy_pcrs:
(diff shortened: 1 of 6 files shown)


keylime/common/cli_opts.py
```diff
@@ -0,0 +1,143 @@
...
+    if args.allowlist_sig and not args.allowlist_sig_key:
+        return True, "--allowlist-sig must also have --allowlist-sig-key"
+    if args.allowlist_sig_url and not args.allowlist_sig_key:
+        return True, "--allowlist-sig-url must also have --allowlist-sig-key"
+    if args.allowlist_sig_key and not (args.allowlist_sig or args.allowlist_sig_url):
+        return True, "--allowlist-sig-key must have either --allowlist-sig or --allowlist-sig-url"
+    return False, None
+
+
+def process_allowlist(args):
```
mpeters, line 53: The rest of this module does seem like cli option handling, but `process_allowlist()` seems very different. I agree with pulling it out of the tenant, but maybe it should be in a separate module designed for policy related things? keylime/policy.pm ?

mpeters (changes requested):

keylime/common/cli_opts.py
maugustosilva: Well, it was sitting within the `tenant` module. But sure, I can certainly create a second module to process inputs to policy (**from the CLI**).

mpeters (approved):

PR #1121, opened 2022-10-06
Make Rust agent offical, improve docs

### LINKED ISSUE #1111: Mark Python agent with deprecation warning
After the `0.1.0` release of the Rust agent, it will be the official agent. The Python agent should be marked with a deprecation warning at this time to enable people to transition off of it over Q4 2022.

THS-on:
This adds the deprecation warnings for the Python agent and makes the Rust agent the official one.
I also updated the installation instructions to be more up to date.

Closes #1111

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,18 +38,19 @@ with secrets stored within an encrypted payload released once trust is establish
 ### Rust based Keylime Agent
 
 The verifier, registrar, and agent are all developed in Python and situated
-in this repository `keylime`. The agent is currently undergoing a port to the
-[Rust programming language](https://www.rust-lang.org), with this work taking
-place in the [rust-keylime repository](https://github.com/keylime/rust-keylime).
+in this repository `keylime`. The agent was ported to the
+[Rust programming language](https://www.rust-lang.org). The code can be found 
+in the [rust-keylime repository](https://github.com/keylime/rust-keylime).
 
 The decision was made to port the agent to Rust, as rust is a low-level
 performant systems language designed with security as a central tenet, by means
 of the rust compiler's ownership model.
 
-When the rust agent work is complete, the rust-keylime agent will become the
-recommended ongoing agent within Keylime. Until then the Python agent is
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -38,18 +38,19 @@ with secrets stored within an encrypted payload released once trust is establish
...
 
 The decision was made to port the agent to Rust, as rust is a low-level
 performant systems language designed with security as a central tenet, by means
 of the rust compiler's ownership model.
 
-When the rust agent work is complete, the rust-keylime agent will become the
-recommended ongoing agent within Keylime. Until then the Python agent is
-fully functioning and available to use as a remote monitoring system to interact
-with the keylime verifier and registrar.
+Starting with the 0.1.0 release of the Rust based Keylime agent, this agent this the official agent.
```
mpeters, line 49: Something about the wording here is off. Maybe it should be something like "this agent is now the official agent"

README.md
```diff
@@ -38,18 +38,19 @@ with secrets stored within an encrypted payload released once trust is establish
...
+
+| IMPORTANT: The Python version is deprecated and will be removed with the next major version (6.6.0)! |
```
mpeters, line 51: Are we sure we'll remove it in 6.6.0? I think it should be marked as deprecated longer than that, but I'm persuadable.

mpeters (changes requested):

README.md
THS-on: In the [roadmap](https://github.com/keylime/keylime/blob/master/ROADMAP.md#removal-of-the-python-agent) we planned the removal of Python agent by the end of Q4, so I think 3-4 months are fine.

I'll change it to 7.0.0 because this allows us to make one last 6.6.0 release before removing the agent in 7.0.0.

README.md
THS-on: Thanks, fixed.

README.md
maugustosilva: 7.0.0 it is, thanks.

mpeters (approved):

PR #1125, opened 2022-10-09
tpm_main: unescape UEFI eventlog strings

THS-on:
Newer versions of tpm2-tools escape the strings in the UEFI eventlog [1]. We
remove the qoutes and convert the escaped special characters back to normal
ones.

[1]: https://github.com/tpm2-software/tpm2-tools/commit/c78d258b2588aee535fd17594ad2f5e808056373

### DIFF
--- a/keylime/tpm/tpm_main.py
+++ b/keylime/tpm/tpm_main.py
@@ -68,7 +68,16 @@ class tpm(tpm_abstract.AbstractTPM):
         # Extract the full semver release number.
         self.tools_version = version_str.split("-")
 
-        if Version(self.tools_version[0]) >= Version("4.2"):
+        if Version(self.tools_version[0]) >= Version("5.4") or (
+            # Also mark first git version that introduces the change to the tpm2_eventlog format as 5.4
+            # See: https://github.com/tpm2-software/tpm2-tools/commit/c78d258b2588aee535fd17594ad2f5e808056373
+            Version(self.tools_version[0]) == Version("5.3")
+            and len(self.tools_version) > 1
+            and int(self.tools_version[1]) >= 24
+        ):
+            logger.info("TPM2-TOOLS Version: %s", self.tools_version[0])
+            self.tools_version = "5.4"
+        elif Version(self.tools_version[0]) >= Version("4.2"):
             logger.info("TPM2-TOOLS Version: %s", self.tools_version[0])
             self.tools_version = "4.2"
         elif Version(self.tools_version[0]) >= Version("4.0.0"):
@@ -84,7 +93,7 @@ class tpm(tpm_abstract.AbstractTPM):
     def __get_tpm_algorithms(self):
         if self.tools_version == "3.2":
             retDict = self.__run(["tpm2_getcap", "-c", "algorithms"])
-        elif self.tools_version in ["4.0", "4.2"]:
+        elif self.tools_version in ["4.0", "4.2", "5.4"]:
             retDict = self.__run(["tpm2_getcap", "algorithms"])
 
         output = config.convert(retDict["retout"])
@@ -118,7 +127,7 @@ class tpm(tpm_abstract.AbstractTPM):
         """Gets which PCRs are enabled with which hash algorithm"""
         if self.tools_version == "3.2":
             retDict = self.__run(["tpm2_getcap", "-c", "pcrs"])
-        elif self.tools_version in ["4.0", "4.2"]:
+        elif self.tools_version in ["4.0", "4.2", "5.4"]:
             retDict = self.__run(["tpm2_getcap", "pcrs"])
 
         output = config.convert(retDict["retout"])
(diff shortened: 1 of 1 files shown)


keylime/tpm/tpm_main.py
```diff
@@ -1576,6 +1583,32 @@ def __add_boot_aggregate(log: dict) -> None:
...
+        This converts it back to an unescaped string.
+        Example:
+            '"MokList\\0"' -> 'MokList'
+        """
+        if tpm.tools_version in ["3.2", "4.0", "4.2"]:
+            return
+
+        def recursive_unescape(data):
+            if isinstance(data, str):
+                if data.startswith('"') and data.endswith('\\0"'):
```
aplanas, line 1599: I am wondering 2 stuff:

1. maybe the datatype is wrong, and should be byte instead of str
2. is there ant guarantee that the null byte will appear only in the last position? I am thinking on other assets that are measured and part of the event log that will contain the null byte in the middle.  In this case the parser should be more genering.

keylime/tpm/tpm_main.py
```diff
@@ -68,7 +68,14 @@ def __get_tpm2_tools(self):
         # Extract the full semver release number.
         self.tools_version = version_str.split("-")
 
-        if Version(self.tools_version[0]) >= Version("4.2"):
+        if Version(self.tools_version[0]) >= Version("5.4") or (
+            Version(self.tools_version[0]) == Version("5.3")
+            and len(self.tools_version) > 1
+            and int(self.tools_version[1]) >= 24
```
aplanas, line 74: why so specific? is there a patched version of 5.3 that backport the feature in an specific distro? If so a comment would help.

keylime/tpm/tpm_main.py
THS-on: 1. It is actually a str, because it is part of the YAML output of tpm2-tools which is converted to a dict.
2. Kinda, there should be always a final trailing one. If not someone writes malformed strings. See this commit for more info: https://github.com/tpm2-software/tpm2-tools/commit/c78d258b2588aee535fd17594ad2f5e808056373

keylime/tpm/tpm_main.py
THS-on: When you build from git this is the first version that introduces that change. Because the change is only 5 days old, I don't think any distro has backported that yet.

I'll add a comment.

keylime/tpm/tpm_main.py
aplanas: Thanks for the pointer!

```
- All NUL bytes, including the final trailing NUL
    are displayed in escaped format.
```

Glazing the code seems to be more escapes ("\a", "\n", "\t", ...) at any point in the byte array.

IMHO would make sense to reflect this into the parser too.  Something trivial, like a search-replace.  The cost will be `10 * n` (10 escape character for a string of length n), as we will repeat the search-replace 10 times. So is still `O(n)`, so kind of OK.

keylime/tpm/tpm_main.py
aplanas: Would make sense then to leave it to 5.4 only.  The other option is to unconditionally do the scan and pay the price.

THS-on: Because of https://github.com/tpm2-software/tpm2-tools/pull/3144 you currently need to run a git snapshot if you want to attest a modern system that uses the latest Shim.

I still want to run the scan only on versions that require it, so my idea would be to keep the current check and then remove it once 5.4 is released with the necessary fixes.

keylime/tpm/tpm_main.py
THS-on: Yeah we should do that. My idea is the following

1. Check if string is in quotes
2. Remove all the escaped characters using search-replace. (I still need check if simple search-replace is enough)
3. Remove trailing null if present

THS-on: I now implemented the idea.

keylime/tpm/tpm_main.py
```diff
@@ -1576,6 +1585,49 @@ def __add_boot_aggregate(log: dict) -> None:
...
+        ]
+
+        def recursive_unescape(data):
+            if isinstance(data, str):
+                if data.startswith('"') and data.endswith('"'):
+                    data = data[1:-1]
+                    for orig, escaped in escaped_chars:
+                        data = data.replace(escaped, orig)
+                    return data.rstrip("\0")
+                return data
```
aplanas, line 1620: can be removed?

keylime/tpm/tpm_main.py
```diff
@@ -1576,6 +1585,49 @@ def __add_boot_aggregate(log: dict) -> None:
...
+            ("\\", "\\\\"),
+        ]
+
+        def recursive_unescape(data):
+            if isinstance(data, str):
+                if data.startswith('"') and data.endswith('"'):
+                    data = data[1:-1]
+                    for orig, escaped in escaped_chars:
+                        data = data.replace(escaped, orig)
+                    return data.rstrip("\0")
```
aplanas, line 1619: the return can be removed, and you can `data = data.rstrip("\0")`

aplanas (approved):
Some nits, but LGTM

keylime/tpm/tpm_main.py
THS-on: yes fixed.

aplanas (approved):

keylime/tpm/tpm_main.py
THS-on: Yes is now also fixed. Thank you for the fast review!

PR #1132, opened 2022-10-14
Draft: Do not install keylime nor configuration files during tests

ansasaki:
Use a virtual environment for the python components and temporary
directories for the generated files.  The dependencies are still
installed in the system.

The used configuration files are generated in a temporary directory with
the changes necessary to run the tests. The installed configuration
files are not modified.

The files generated when running run_tests.sh will be removed at the end
of the execution unless the environment variable KEEP_TMP_FILES=1.

If the tests are executed using the run_tests.sh script, all the
necessary dependencies are installed and environment variables set.

When running test_restful.py outside run_tests.sh with RUST_TEST=1, it
expects the KEYLIME_RUST_AGENT and KEYLIME_RUST_CONF environment
variables set, respectively, with the path to the rust agent binary and
the rust agent configuration file.

If the environment variables are not set, it tries to find the
rust agent binary and configuration in an adjacent directory (i.e. in
the same location run_tests.sh downloads it).


test/scripts
```diff
@@ -0,0 +1 @@
+../scripts
```
mpeters, line 1: What is the purpose of this file?

ansasaki: I don't like much the solution I came up with, but this is to allow importing the `convert_config.py` as a module and use inside the test to generate the configuration files.

ansasaki:
I see issues with missing `keylime` user as the default `run_as` option is set as `keylime:tss`. I'll modify the scripts later to create the user. Meanwhile I'll leave this as a draft.

test/scripts
ansasaki: I think the best here would be to follow the [suggestion](https://github.com/keylime/keylime/pull/1133#issuecomment-1282730523) from @THS-on and install the `convert_config.py` script along the other utilities which will also make it a keylime submodule that can be imported without this _hackish_ solution.

ansasaki:
Superseded by #1260

Issue #1112, opened 2022-09-26
Certificate files created in /var/lib/keylime/cv_ca have insufficient permissions


zkolpet:
Is your an issue a feature request? If so, please raise it as an [enhancement](https://github.com/keylime/enhancements)

# Environment
 * OS / version: Ubuntu 22.04
 * Processor architecture: N/A
 * TPM Manufacturer: N/A
 * Keylime version: v6.5.0

# Description 
Certificate files created in /var/lib/keylime/cv_ca have insufficient permissions. They are created with 600 permissions and we would need 640 permissions on those files. 

## Expected behavior vs. actual behavior 
We are running keylime_verifier service within one user and keylime_registrar service within different user but both users are from same group. Due to this, we need the certificate files in /var/lib/keylime/cv_ca to be created with 640 permissions (to have group read permissions also) so our custom registrar user is able to read those certificate files (via group read permissions) otherwise we face keylime_registrar service to fail starting.

## Steps to reproduce problem
1. start keylime_verifier and then keylime_registrar service as non-root user
2. keylime_registrar service fails to start due to being unable to read certificate files in /var/lib/keylime/cv_ca

## Code change suggestion:
in keylime/keylime/ca_util.py file
method: cmd_mkcert
part: finally

```
finally:
    # files inside /var/lib/keylime/cv_ca dir need to have at least read permissions for group
    chmodfiles("./*", 0o640)
    os.chdir(cwd)
```

plus create "chmodfiles" function somewhere in file:

```
def chmodfiles(path, permissions):
    files = glob.glob(path)
    for f in files:
        os.chmod(f, mode=permissions)
```

# Relevant logs

Attach any relevant log files that can help to debug your issue.

PR #1138, opened 2022-10-24
ca_util: allow users in the same group to read the created certificates and keys

THS-on:
This is required when for example the verifier and registrar share the certificates, but run with different users.

Fixes https://github.com/keylime/keylime/issues/1112

### DIFF
--- a/keylime/ca_util.py
+++ b/keylime/ca_util.py
@@ -75,6 +75,7 @@ def setpassword(pw):
 
 def cmd_mkcert(workingdir, name, password=None):
     cwd = os.getcwd()
+    mask = os.umask(0o037)
     try:
         fs_util.ch_dir(workingdir)
         priv = read_private()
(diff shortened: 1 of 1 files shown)


mpeters (approved):

keylime/ca_util.py
```diff
@@ -75,6 +75,7 @@ def setpassword(pw):
 
 def cmd_mkcert(workingdir, name, password=None):
     cwd = os.getcwd()
+    mask = os.umask(0o037)
```
aplanas, line 78: Is it possible to avoid the use of umask? I see that is later reset in the `finally`, this still creates a state in the process that affect any created file, and sometimes it is difficult to track them.

IIUC you want that the files are read by the group, so why `0o640` as you did in the `fdopen` is not enough?

THS-on: The Keylime services are starting with the mask `0o077` which turns the `0o640` to `0o600` again. The alternative would be to change the mask to `0o037` in general. I don't know if that is the better option.

maugustosilva (approved):

PR #1139, opened 2022-10-24
Chore fix type issues

mdrocco:
This PR provides a first batch of type fixes (see #929, "pyright" section). I will follow up with a further issue about the outstanding issues that I could not figure put how to fix without a better understanding of the current behavior.

### DIFF
--- a/keylime/cloud_verifier_common.py
+++ b/keylime/cloud_verifier_common.py
@@ -61,7 +61,7 @@ def process_quote_response(agent, ima_policy, json_response, agentAttestState) -
         )
 
     except Exception as e:
-        failure.add_event("invalid_data", {"message": "parsing agent get quote respone failed", "data": e}, False)
+        failure.add_event("invalid_data", {"message": "parsing agent get quote respone failed", "data": str(e)}, False)
         return failure
 
     # TODO: Are those separate failures?
@@ -123,7 +123,7 @@ def process_quote_response(agent, ima_policy, json_response, agentAttestState) -
         logger.error("TPM Quote for agent %s is using an unaccepted signing algorithm: %s", agent_id, sign_alg)
         failure.add_event(
             "invalid_sign_alg",
-            {"message": f"TPM Quote is using an unaccepted signing algorithm: {sign_alg}", "data": {sign_alg}},
+            {"message": f"TPM Quote is using an unaccepted signing algorithm: {sign_alg}", "data": sign_alg},
             False,
         )
         return failure
@@ -301,7 +301,7 @@ def prepare_error(agent, msgtype="revocation", event=None):
         tosend["signature"] = crypto.rsa_sign(signing_key, tosend["msg"])
 
     else:
-        tosend["signature"] = "none"
+        tosend["signature"] = b""
     return tosend
 
 
@@ -313,8 +313,14 @@ def validate_ima_policy_data(agent_data):
     lists = json.loads(agent_data)
 
     # Validate exlude list contains valid regular expressions
-    is_valid, _, err_msg = validators.valid_exclude_list(lists.get("exclude"))
+    is_valid, _, err_msg_from_validator = validators.valid_exclude_list(lists.get("exclude"))
     if not is_valid:
-        err_msg += " Exclude list regex is misformatted. Please correct the issue and try again."
+        if err_msg_from_validator is not None:
+            err_msg_base = err_msg_from_validator + " "
+        else:
+            err_msg_base = ""
+        err_msg = err_msg_base + "Exclude list regex is misformatted. Please correct the issue and try again."
+    else:
+        err_msg = err_msg_from_validator
 
     return is_valid, err_msg
--- a/keylime/cloud_verifier_tornado.py
+++ b/keylime/cloud_verifier_tornado.py
@@ -12,6 +12,7 @@ import tornado.ioloop
 import tornado.netutil
 import tornado.process
 import tornado.web
+from sqlalchemy.engine import Engine
 from sqlalchemy.exc import SQLAlchemyError
 from sqlalchemy.orm import joinedload, noload
 from sqlalchemy.orm.exc import NoResultFound
@@ -563,6 +564,7 @@ class AgentsHandler(BaseHandler):
                         logger.info("Allowlist data not provided with request! Using default empty allowlist.")
                         ima_policy = json.dumps(ima.process_ima_policy(ima.EMPTY_ALLOWLIST, []))
 
+                    ima_policy_stored = None
                     if ima_policy:
                         is_valid, err_msg = cloud_verifier_common.validate_ima_policy_data(ima_policy)
                         if not is_valid:
@@ -1123,8 +1125,10 @@ async def process_agent(agent, new_operational_state, failure=Failure(Component.
         # The administration application (tenant) can GET the status and act accordingly (delete/retry/etc).
         if new_operational_state in (states.FAILED, states.INVALID_QUOTE):
             assert failure, "States FAILED and INVALID QUOTE should only be reached with a failure message"
+            assert failure.highest_severity
 
             if agent.get("severity_level") is None or agent["severity_level"] < failure.highest_severity.severity:
+                assert failure.highest_severity_event
                 agent["severity_level"] = failure.highest_severity.severity
                 agent["last_event_id"] = failure.highest_severity_event.event_id
                 agent["operational_state"] = new_operational_state
@@ -1194,8 +1198,11 @@ async def process_agent(agent, new_operational_state, failure=Failure(Component.
                 logger.debug(
                     "Setting up callback to check agent ID %s again in %f seconds", agent["agent_id"], interval
                 )
+
                 # set up a call back to check again
-                cb = functools.partial(invoke_get_quote, agent, ima_policy, False, timeout=timeout)
+                async def cb():
+                    await invoke_get_quote(agent, ima_policy, False, timeout=timeout)
+
                 pending = tornado.ioloop.IOLoop.current().call_later(interval, cb)
                 agent["pending_event"] = pending
             return
@@ -1219,7 +1226,10 @@ async def process_agent(agent, new_operational_state, failure=Failure(Component.
                 await process_agent(agent, states.FAILED, failure)
             else:
                 agent["operational_state"] = states.GET_QUOTE
-                cb = functools.partial(invoke_get_quote, agent, ima_policy, True, timeout=timeout)
+
+                async def cb_quote_retry():
+                    await invoke_get_quote(agent, ima_policy, True, timeout=timeout)
+
                 agent["num_retries"] += 1
                 next_retry = retry.retry_time(exponential_backoff, interval, agent["num_retries"], logger)
                 logger.info(
@@ -1229,7 +1239,7 @@ async def process_agent(agent, new_operational_state, failure=Failure(Component.
                     maxr,
                     next_retry,
                 )
-                tornado.ioloop.IOLoop.current().call_later(next_retry, cb)
+                tornado.ioloop.IOLoop.current().call_later(next_retry, cb_quote_retry)
             return
 
         if main_agent_operational_state == states.PROVIDE_V and new_operational_state == states.PROVIDE_V_RETRY:
@@ -1244,7 +1254,10 @@ async def process_agent(agent, new_operational_state, failure=Failure(Component.
                 await process_agent(agent, states.FAILED, failure)
             else:
                 agent["operational_state"] = states.PROVIDE_V
-                cb = functools.partial(invoke_provide_v, agent)
+
+                async def cb_vprov_retry():
+                    await invoke_provide_v(agent)
+
                 agent["num_retries"] += 1
                 next_retry = retry.retry_time(exponential_backoff, interval, agent["num_retries"], logger)
                 logger.info(
@@ -1254,7 +1267,7 @@ async def process_agent(agent, new_operational_state, failure=Failure(Component.
                     maxr,
                     next_retry,
                 )
-                tornado.ioloop.IOLoop.current().call_later(next_retry, cb)
+                tornado.ioloop.IOLoop.current().call_later(next_retry, cb_vprov_retry)
             return
         raise Exception("nothing should ever fall out of this!")
 
@@ -1354,6 +1367,7 @@ def main():
 
     def server_process(task_id):
         logger.info("Starting server of process %s", task_id)
+        assert isinstance(engine, Engine)
         engine.dispose()
         server = tornado.httpserver.HTTPServer(app, ssl_options=ssl_ctx, max_buffer_size=max_upload_size)
         server.add_sockets(sockets)
--- a/keylime/crypto.py
+++ b/keylime/crypto.py
@@ -7,15 +7,14 @@ import secrets
 from ipaddress import IPv4Address, IPv6Address, ip_address
 from typing import List, Optional
 
-import cryptography.hazmat.primitives.asymmetric
-
 # Crypto implementation using python cryptography package
 from cryptography import exceptions, x509
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import hashes, serialization
 from cryptography.hazmat.primitives.asymmetric import padding
+from cryptography.hazmat.primitives.asymmetric.padding import MGF1, OAEP
 from cryptography.hazmat.primitives.asymmetric.rsa import generate_private_key
-from cryptography.hazmat.primitives.ciphers import Cipher, algorithms, modes
+from cryptography.hazmat.primitives.ciphers import AEADEncryptionContext, Cipher, algorithms, modes
 from cryptography.hazmat.primitives.kdf.pbkdf2 import PBKDF2HMAC
 from cryptography.x509.oid import NameOID
 
@@ -64,8 +63,10 @@ def rsa_sign(key, message, paddingt="internal"):
     if paddingt == "internal":
         _padding = padding.PSS(mgf=padding.MGF1(hashes.SHA256()), salt_length=padding.PSS.MAX_LENGTH)
 
-    if paddingt == "default":
+    elif paddingt == "default":
         _padding = padding.PKCS1v15()
+    else:
+        raise ValueError
 
     signature = key.sign(message, _padding, hashes.SHA256())
     return base64.b64encode(signature)
@@ -107,11 +108,12 @@ def rsa_export_privkey(private_key, password=None):
 
 
 def rsa_encrypt(key, message):
+
     """RSA encrypt message"""
     return key.encrypt(
         bytes(message),
-        cryptography.hazmat.primitives.asymmetric.padding.OAEP(
-            mgf=cryptography.hazmat.primitives.asymmetric.padding.MGF1(algorithm=hashes.SHA1()),
+        OAEP(
+            mgf=MGF1(algorithm=hashes.SHA1()),
             algorithm=hashes.SHA1(),
             label=None,
         ),
@@ -122,8 +124,8 @@ def rsa_decrypt(key, ciphertext):
     """RSA decrypt message"""
     return key.decrypt(
         ciphertext,
-        cryptography.hazmat.primitives.asymmetric.padding.OAEP(
-            mgf=cryptography.hazmat.primitives.asymmetric.padding.MGF1(algorithm=hashes.SHA1()),
+        OAEP(
+            mgf=MGF1(algorithm=hashes.SHA1()),
             algorithm=hashes.SHA1(),
             label=None,
         ),
@@ -190,6 +192,7 @@ def encrypt(plaintext, key):
     iv = generate_random_key(aes_block_size)
     encryptor = Cipher(algorithms.AES(key), modes.GCM(iv), backend=default_backend()).encryptor()
     cipher_text = encryptor.update(plaintext) + encryptor.finalize()
+    assert isinstance(encryptor, AEADEncryptionContext)
     return base64.b64encode(iv + cipher_text + encryptor.tag)
 
 
--- a/keylime/ima/ast.py
+++ b/keylime/ima/ast.py
@@ -29,11 +29,15 @@ COLON_BYTE = ord(":")
 
 
 def get_START_HASH(hash_alg: Hash) -> bytes:
-    return codecs.decode(b"0" * (hash_alg.get_size() // 4), "hex")
+    hsize = hash_alg.get_size()
+    assert hsize is not None
+    return codecs.decode(b"0" * (hsize // 4), "hex")
 
 
 def get_FF_HASH(hash_alg: Hash) -> bytes:
-    return codecs.decode(b"f" * (hash_alg.get_size() // 4), "hex")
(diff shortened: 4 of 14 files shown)


keylime/ima/ima.py
```diff
@@ -219,6 +219,7 @@ def _validate_ima_buf(
             failure = _validate_ima_ng(exclude_regex, allowlist, digest, path, hash_types="keyrings")
             if not failure:
                 # Add the key only now that it's validated (no failure)
+                assert ima_keyrings is not None
```
THS-on, line 222: Under normal conditions `ima_keyrings` is never None when an IMA validator is called, but for now I would rather check it explicitly instead of an assertion.

We definitely need to refactor the agent state at some point into a single object.

keylime/ima/ima_dm.py
```diff
@@ -12,14 +12,19 @@
...
 import lark
-from lark import Lark, Transformer, v_args
 from packaging import version
 
 from keylime.common.algorithms import Hash
 from keylime.failure import Component, Failure
 from keylime.ima import ast
 from keylime.ima.dm_grammar import DM_GRAMMAR
 
+if sys.version_info >= (3, 10):
```
THS-on, line 22: Maybe add a comment why this is needed for Python 3.10?

keylime/ima/ima_dm.py
```diff
@@ -293,7 +299,7 @@ def validate_table_load(self, event: "LoadEvent", match_key, digest) -> Failure:
         # Check "num_targets"
         # Note that we get actually could get multiple lines, but this does not happen for our use cases so we
         # treat it as a failure.
-        if event.device_metadata.num_targets != len(event.targets) or used_policy["table_load"][entry] != len(
+        if event.device_metadata.num_targets != len(event.targets) or used_policy["table_load"]["num_targets"] != len(
```
THS-on, line 302: That was definitely a bug. Good catch!

keylime/ima/ima_dm.py
```diff
@@ -395,6 +401,8 @@ def _check_attr(attr: Optional[Union[int, str, bool]], reference_value: Optional
     if attr is None and reference_value is not None:
         return False
 
+    assert attr is not None
```
THS-on, line 404: We can just drop the `and reference_value is not None` from line 401 right? Because that is we already returned if the value was None.

keylime/registrar_common.py
```diff
@@ -281,11 +281,16 @@ def do_POST(self):
...
-            (blob, key) = initialize_tpm.encryptAIK(
+            aik_enc = initialize_tpm.encryptAIK(
                 agent_id,
                 base64.b64decode(ek_tpm),
                 base64.b64decode(aik_tpm),
             )
+            if aik_enc is not None:
+                (blob, key) = aik_enc
+            else:
+                logger.warning("Agent %s failed encrypting AIK", agent_id)
```
THS-on, line 292: Before returning you should still send a response. E.g.:
```
web_util.echo_json_response(self, 400, f"Error: failed encrypting AK")
```

keylime/tenant.py
```diff
@@ -953,7 +959,7 @@ def do_quote(self):
                 response_body = response.json()
 
             except Exception as e:
-                if response.status_code in (503, 504):
+                if response is not None and response.status_code in (503, 504):
```
THS-on, line 962:
```suggestion
                if response is None or response.status_code in (503, 504):
```
If we don't get anything back than also retry

keylime/tenant.py
```diff
@@ -983,7 +989,7 @@ def do_quote(self):
             raise UserError(f"Status command response: {response.status_code} Unexpected response from Cloud Agent.")
 
         if "results" not in response_body:
-            raise UserError(f"Error: unexpected http response body from Cloud Agent: {str(response.status)}")
+            raise UserError(f"Error: unexpected http response body from Cloud Agent: {str(response_body)}")
```
THS-on, line 992: I don't know if we actually want to log the entire body. I think the status code is fine.

keylime/tpm/tpm_main.py
```diff
@@ -1059,6 +1111,10 @@ def __tpm2_checkquote(self, pubaik, nonce, quoteFile, sigFile, pcrFile, hash_alg
                 "-q",
                 nonce,
             ]
+        else:
```
THS-on, line 1114: This case should never happen. Calling sys.exit() is fine for the functions only used by the agent, but for the verifier i think it is not a good idea. Maybe instead we can only add the check for the old version and remove the check for the newer ones.

keylime/tpm/tpm_main.py
```diff
@@ -1331,12 +1387,14 @@ def sim_extend(self, hashval_1, hashval_0=None, hash_alg=None):
             hashval_0 = self.START_HASH(hash_alg)
 
         # compute expected value  H(0|H(data))
-        extendedval = self.hashdigest(
-            codecs.decode(hashval_0, "hex_codec")
-            + codecs.decode(self.hashdigest(hashval_1.encode("utf-8"), hash_alg), "hex_codec"),
+        hdata = self.hashdigest(hashval_1.encode("utf-8"), hash_alg)
+        assert hdata is not None
```
THS-on, line 1391: What can go wrong here that the value might be None?

THS-on (changes requested):
Thank you for this PR! I really going to improve the code quality.

keylime/backport_dataclasses.py
```diff
@@ -236,7 +236,7 @@ class Field:
 
     def __init__(self, default, default_factory, init, repr, hash, compare,
                  metadata):
-        self.name = None
+        self.name = ""
```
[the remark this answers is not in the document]
mdrocco: done

keylime/agentstates.py
```diff
@@ -18,6 +18,12 @@ class TPMClockInfo:
     restartcount: int
     safe: int
 
+    def __init__(self, clock, resetcount, restartcount, safe):
```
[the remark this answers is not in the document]
mdrocco: I reverted this change, I guess this "fix" is actually overridden by a better one somewhere else, as the type checker is not complaining if I remove it.

keylime/crypto.py
```diff
@@ -64,7 +64,8 @@ def rsa_sign(key, message, paddingt="internal"):
     if paddingt == "internal":
         _padding = padding.PSS(mgf=padding.MGF1(hashes.SHA256()), salt_length=padding.PSS.MAX_LENGTH)
 
-    if paddingt == "default":
+    else:
+        assert paddingt == "default"
```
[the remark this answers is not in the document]
aplanas: I guess is to make sure that paddingt is "internal" | "default". IMHO and `if`, `elif` and a `else` that thrown an exception communicates better.

keylime/ima/ima_dm.py
aplanas: Yes, that is funny. Does lark namespace for a fixed version changes based on the Python version? I never saw that : )

keylime/cli/options.py
```diff
@@ -29,4 +29,4 @@ def get_opts_error(args):
         return True, "--allowlist-sig-url must also have --allowlist-sig-key"
     if args.allowlist_sig_key and not (args.allowlist_sig or args.allowlist_sig_url):
         return True, "--allowlist-sig-key must have either --allowlist-sig or --allowlist-sig-url"
-    return False, None
+    return False, ""
```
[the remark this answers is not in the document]
mdrocco: done

keylime/cloud_verifier_tornado.py
```diff
@@ -1103,9 +1104,19 @@ async def process_agent(agent, new_operational_state, failure=Failure(Component.
         if new_operational_state in (states.FAILED, states.INVALID_QUOTE):
             assert failure, "States FAILED and INVALID QUOTE should only be reached with a failure message"
 
-            if agent.get("severity_level") is None or agent["severity_level"] < failure.highest_severity.severity:
-                agent["severity_level"] = failure.highest_severity.severity
-                agent["last_event_id"] = failure.highest_severity_event.event_id
+            if (
+                agent.get("severity_level") is None
+                or failure.highest_severity is not None
```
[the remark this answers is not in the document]
mdrocco: done

keylime/cloud_verifier_tornado.py
```diff
@@ -1173,8 +1184,11 @@ async def process_agent(agent, new_operational_state, failure=Failure(Component.
                 logger.debug(
                     "Setting up callback to check agent ID %s again in %f seconds", agent["agent_id"], interval
                 )
+
                 # set up a call back to check again
-                cb = functools.partial(invoke_get_quote, agent, ima_policy, False, timeout=timeout)
+                def cb():
+                    asyncio.run(invoke_get_quote(agent, ima_policy, False, timeout=timeout))
```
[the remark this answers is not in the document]
mdrocco: it is matter of typing rather than style: `partial()` returns an object that is not a function.

keylime/cloud_verifier_tornado.py
```diff
@@ -1333,6 +1353,7 @@ def main():
 
     def server_process(task_id):
         logger.info("Starting server of process %s", task_id)
+        assert isinstance(engine, Engine)
```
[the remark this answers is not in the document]
mdrocco: I am not sure how to do that, as `engine` is a global variable that gets assigned the return value of a function call. Feel free to suggest anything, I am not very familiar with type hints.

keylime/crypto.py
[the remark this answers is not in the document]
mdrocco: please take a look: https://github.com/keylime/keylime/pull/1139/commits/e6b2344b0d1ab9b4c0d6ccfbb8399e2fe5dccf5c

keylime/ima/ima.py
mdrocco: https://github.com/keylime/keylime/pull/1139/commits/e6b2344b0d1ab9b4c0d6ccfbb8399e2fe5dccf5c

keylime/ima/ima_dm.py
mdrocco: yes, at least that is what I inferred looking at the issues reported by the type checker and the code of the Lark module itself at different versions. Of course I might be wrong, feel free to double-check. I added a minimal comment for now: d84fa2c

keylime/ima/ima_dm.py
mdrocco: good point: 07ef85c

keylime/registrar_common.py
mdrocco: 668a1bd

keylime/tenant.py
mdrocco: thanks!

keylime/tenant.py
mdrocco: da2288aded96fcf8ec3a9b3ece9dce994e173e55

keylime/tpm/tpm_main.py
mdrocco: do you mean something like that? 71b33a9
If it looks good to you, I will apply the same change to all the spots where I added that "internal error" thing.

keylime/tpm/tpm_main.py
mdrocco: https://github.com/mdrocco/keylime/blob/71b33a96d084321be7dcb0ef78ba4c3a1edf8a66/keylime/tpm/tpm_main.py#L1395

`codecs.decode` does not accept `None` for the `obj` parameter.

keylime/api_version.py
```diff
@@ -42,19 +42,25 @@ def normalize_version(v):
...
         latest_minor = latest_minor_version(base_version)
         if latest_minor != "0":
             return latest_minor
     return base_version
 
 
 def major(v):
-    return version.parse(str(v)).major
+    version_ = version.parse(str(v))
+    if isinstance(version_, version.Version):
```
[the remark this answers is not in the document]
mdrocco: regardless what `parse` does with respect to exceptions, the assertion looks simpler and safer to me: 53f6d15

keylime/api_version.py
```diff
@@ -42,19 +42,25 @@ def normalize_version(v):
...
+        return version_.major
+    return 0
 
 
 def minor(v):
-    return version.parse(str(v)).minor
+    version_ = version.parse(str(v))
```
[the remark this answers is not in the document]
mdrocco: also targeted by 53f6d15

keylime/cloud_verifier_common.py
```diff
@@ -322,7 +322,7 @@ def validate_ima_policy_data(agent_data):
 
     # Validate exlude list contains valid regular expressions
     is_valid, _, err_msg = validators.valid_exclude_list(lists.get("exclude"))
-    if not is_valid:
+    if not is_valid and err_msg is not None:
```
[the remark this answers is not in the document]
mdrocco: 6ff10100c3938bc323bbbb06e2fc2b597e74442d

keylime/cloud_verifier_tornado.py
```diff
@@ -1198,7 +1212,10 @@ async def process_agent(agent, new_operational_state, failure=Failure(Component.
                 await process_agent(agent, states.FAILED, failure)
             else:
                 agent["operational_state"] = states.GET_QUOTE
-                cb = functools.partial(invoke_get_quote, agent, ima_policy, True, timeout=timeout)
+
+                def cb_quote_retry():
+                    asyncio.run(invoke_get_quote(agent, ima_policy, True, timeout=timeout))
+
```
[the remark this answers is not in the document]
mdrocco: Former point: since `invoke_get_quote` takes 4 arguments, partial application with 4 parameters is equivalent to the complete `run` call, to the best of my understanding.

Latter point: I am not getting your point, can you please explain what makes the change not equivalent?

keylime/config.py
```diff
@@ -27,7 +27,7 @@ def convert(data):
     if isinstance(data, bytes):
         return data.decode()
     if isinstance(data, dict):
-        return dict(map(convert, data.items()))
+        return dict(iter(map(convert, data.items())))
```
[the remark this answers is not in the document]
mdrocco: Without that "fix", the type checker says:

```console
error: Argument of type "map[str | dict[Any, Any] | tuple[Unknown, ...] | list[Unknown] | Unknown]" cannot be assigned to parameter "__iterable" of type "Iterable[tuple[_KT@dict, _VT@dict]]" in function "__init__"
    "map[str | dict[Any, Any] | tuple[Unknown, ...] | list[Unknown] | Unknown]" is incompatible with "Iterable[tuple[Any, Any]]"                                                         TypeVar "_T_co@Iterable" is covariant
        Type "str | dict[Any, Any] | tuple[Unknown, ...] | list[Unknown] | Unknown" cannot be assigned to type "tuple[Any, Any]"
          "str" is incompatible with "tuple[Any, Any]" (reportGeneralTypeIssues)
```

I am not sure the "fix" is the best thing to do, but at least makes the checker happy and does not modify the behavior, to the best of my understanding.

THS-on:
@mdrocco can you do a clean rebase on master, this makes it easier to review?

mdrocco:
I merged master into the branch, let me know it that works.

keylime/cloud_verifier_tornado.py
```diff
@@ -572,13 +573,13 @@ def post(self):
...
                             raise
 
-                    # Write the agent to the database, attaching associated stored policy
-                    try:
-                        session.add(VerfierMain(**agent_data, ima_policy=ima_policy_stored))
-                        session.commit()
-                    except SQLAlchemyError as e:
-                        logger.error("SQLAlchemy Error for agent ID %s: %s", agent_id, e)
-                        raise e
+                        # Write the agent to the database, attaching associated stored policy
```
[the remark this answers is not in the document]
THS-on: Yes make it None by default.

keylime/cloud_verifier_tornado.py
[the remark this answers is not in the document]
THS-on: In line 42 you can write:

```
    engine: Engine = DBEngineManager().make_engine("cloud_verifier")
```

keylime/cmd/ima_emulator_adapter.py
```diff
@@ -32,6 +32,7 @@ def measure_list(file_path, position, ima_hash_alg, pcr_hash_alg, search_val=Non
 
             entry = ast.Entry(line, None, ima_hash_alg=ima_hash_alg, pcr_hash_alg=pcr_hash_alg)
 
+            assert entry.pcr_template_hash is not None
```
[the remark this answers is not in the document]
THS-on: No we should just error in the constructor if one of the hashes returns None.

keylime/ima/ima_dm.py
THS-on: Can you investigate with which lark versions this happens?

keylime/tpm/tpm_main.py
THS-on: Yeah, but can `self.hashdigest(...)` return None?

pyproject.toml
```diff
@@ -14,3 +14,14 @@ include = '\.pyi?$'
 [tool.isort]
 profile = "black"
 line_length = 120
+
+[tool.pyright]
+include = ["keylime"]
+ignore = [
+    "keylime/backport_dataclasses.py",
+    "keylime/keylime_agent.py",
+    "keylime/elchecking/",
```
THS-on, line 23: The elchecking module should be verified in the long run, but it is fine to exclude it for this PR.

keylime/cloud_verifier_tornado.py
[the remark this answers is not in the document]
mdrocco: dd7b1c8

keylime/cloud_verifier_tornado.py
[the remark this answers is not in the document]
mdrocco: that raise a type error:

```console
Expression of type "MockConnection" cannot be assigned to declared type "Engine"
  "MockConnection" is incompatible with "Engine"
```

keylime/cmd/ima_emulator_adapter.py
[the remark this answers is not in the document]
mdrocco: bff9d1f

keylime/ima/ima_dm.py
mdrocco: My setup is Python 3.10.6 and Lark 1.1.3. If i Try the "old-style" import I get this for `Lark`:

```console
"Lark" is not exported from module "lark"
  Import from "lark.lark" instead
```

(same thing for `Transformer` and `v_args`)

keylime/tpm/tpm_main.py
mdrocco: I don't understand your question, the type checker is making the point about `codecs.decode`.

keylime/tpm/tpm_main.py
THS-on: Yeah, seems the better option.

keylime/tpm/tpm_main.py
THS-on: Ah ok, then keep the asserts. On another note I think it is fine to just assign it to same variable instead of using a new one or does the type checker than complain?

mdrocco: I am not sure what you mean by "assign it to the same variable", but as long as we are not re-typing the variable (which I second the checker if it complains about, as it makes the code unnecessarily obscure) I am happy to hear any simplification you have to propose.

keylime/tpm/tpm_main.py
mdrocco: great, I will apply the change everywhere.

mdrocco: done

keylime/cloud_verifier_tornado.py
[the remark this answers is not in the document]
mdrocco: @aplanas are you referring to the behavior with respect to async calls? I am struggling with it since I noticed my change breaks the code badly (e.g. integration tests are failing), maybe you have some suggestion based on your better understanding of Python's async and similar?

mdrocco:
@THS-on I rebased on master and squashed, as I screwed up with commit signatures, let me know if you have concerns with that. I will now move to debugging what I broke with those type "fixes".

mdrocco:
Thanks for checking, I changed the author.

keylime/ima/ima_dm.py
THS-on: Can you check if Lark 0.X versions and other Python versions, because making it conditionally on the Python version still does not seem to be right.

mdrocco: Apparently it is a Lark issue and they are working on it: https://github.com/lark-parser/lark/issues/1199. To my understanding, the "old-style" import is intended to be supported, so I would simply roll back to it.

mdrocco: done, hopefully the type checker will eventually be happy once they fix.

keylime/ima/ast.py
```diff
@@ -29,11 +29,15 @@
...
-    return codecs.decode(b"0" * (hash_alg.get_size() // 4), "hex")
+    hsize = hash_alg.get_size()
+    assert hsize is not None
+    return codecs.decode(b"0" * (hsize // 4), "hex")
 
 
 def get_FF_HASH(hash_alg: Hash) -> bytes:
-    return codecs.decode(b"f" * (hash_alg.get_size() // 4), "hex")
+    hsize = hash_alg.get_size()
+    assert hsize is not None
```
stefanberger, line 39: What is driving this that you add an assert here? Why would the hsize possibly be None?

stefanberger:
My general suggestion would be to split this PR up into many small easy-to-digest patches and PRs. There are some things that could go in very easily (changes in crypto.py for example) and others may not be so easy to digest...

mdrocco:
Sure, let us split the PR. Can you please mark the not-so-easy-to-digest changes? This would also help isolating what is making CI tests fail. Thanks!

keylime/ima/ast.py
```diff
@@ -370,7 +374,9 @@ def __init__(
             raise ParserError(f"No parser for mode {tokens[2]} implemented.")
         self.mode = mode(tokens[3])
         self._bytes = self.mode.bytes()
-        self.pcr_template_hash = self._pcr_hash_alg.hash(self._bytes)
+        pcr_hash_alg_ = self._pcr_hash_alg.hash(self._bytes)
+        assert pcr_hash_alg_ is not None
```
stefanberger, line 378: Can the hash() of something be None?

stefanberger:
It looks like there are many different types of changes bunched together in this PR. I would split it by individual types of changes.

stefanberger:
> This would also help isolating what is making CI tests fail

For that I would rebase on master.

keylime/ima/ast.py
mdrocco: https://github.com/mdrocco/keylime/blob/314dd2c8de2eeae5209b688903678f959f41ec3c/keylime/common/algorithms.py#L48

mdrocco:
Thanks @stefanberger for the thorough review, I will rebase on master.

mdrocco:
closing as split into #1164 #1173 #1180

PR #1140, opened 2022-10-24
Update sample ima-policy to exclude overlayfs

mpeters:
Since this is almost always the right default if there are containers


THS-on:
Maybe add a comment why we exclude it by default. Otherwise LGTM.

maugustosilva (approved):

PR #1142, opened 2022-10-25
This PR fixes a bug that prevented 6.5.x verifiers from interacting with 6.2. agents

maugustosilva:

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -26,7 +26,7 @@ author = "Keylime Developers"
 # The short X.Y version
 version = ""
 # The full version, including alpha/beta/rc tags
-release = "6.5.1"
+release = "6.5.3"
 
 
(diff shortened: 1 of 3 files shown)


docs/conf.py
```diff
@@ -26,7 +26,7 @@
 # The short X.Y version
 version = ""
 # The full version, including alpha/beta/rc tags
-release = "6.5.1"
+release = "6.5.3"
```
mpeters, line 29: I'm not sure what's happening with the keylime bot creating new releases. We should still be on 6.5.1 and I'm going to delete those bad releases.

mpeters (changes requested):

docs/conf.py
maugustosilva: Fixed

mpeters (approved):

PR #1145, opened 2022-10-27
Allow encrypting arbitraty binary input

ansasaki:
Do not require the input data to be utf-8 encoded and read as raw bytes instead. 
This allows binary files to be sent using the tenant `-f` sub-option of the `add` command, including zipped payloads.

This reverts #1141


THS-on (approved):
LGTM. It passes now all the tests. Any reason why this is marked as draft?

ansasaki:
> LGTM. It passes now all the tests. Any reason why this is marked as draft?

Friday was a public holiday in Czechia, and I left as a draft on Thursday before leaving :). Removing the draft state now.

PR #1147, opened 2022-10-28
Implement major Keylime policy overhaul

mbestavros:
This pull request implements support for the new IMA policy format detailed by @THS-on in Keylime enhancement proposal [#70](https://github.com/keylime/enhancements/blob/master/70_ima_policy.md).

This redesign carries several benefits:

- Allow and exclude lists are now merged together into one file, reducing the required number of tenant CLI options and generally simplifying policy configuration.
- Internally, the Keylime verifier no longer modifies IMA policies before storing them in its database. This not only greatly simplifies allowlist processing on both tenant and verifier, but also enables development of new features like continuous signature verification after policy creation.

Some additional notes:

- This work is intended to allow Keylime to cleanly break from older allow/exclude list formats, and as such is meant to be a breaking change. While a database migration is provided to ensure that existing allow/exclude lists are preserved, neither the Keylime tenant nor verifier will allow them as input.
- While self-signed IMA policies were also intended to be part of this work (see enhancement proposal [88](https://github.com/keylime/enhancements/pull/89)), the required dependencies for that feature are in flux, and as such this first implementation retains the legacy detached signature functionality. In the future, it's possible this functionality will be dropped in favor of DSSE-signed IMA policies.
- The pull request as-is encodes new-format IMA policies as `base64`. If desired, I can adjust the payload to be pure JSON instead.

Please also have a look at my [second pull request](https://github.com/keylime/keylime/pull/1113), which includes a separate conversion script for converting existing allow and exclude lists to a correctly-formatted IMA policy.

Would appreciate review and testing! Tagging @mpeters @THS-on @lukehinds @lkatalin @aplanas and whoever else would like to provide feedback.


THS-on:
> The pull request as-is encodes new-format IMA policies as base64. If desired, I can adjust the payload to be pure JSON instead.

From a signature verification standpoint base64 is better, because encoding and decoding JSON is not idempotent.

keylime/cloud_verifier_common.py
```diff
@@ -158,7 +158,14 @@ def process_quote_response(agent, ima_policy, json_response, agentAttestState) -
     agentAttestState.set_boottime(boottime)
 
     ima_keyrings = agentAttestState.get_ima_keyrings()
-    tenant_keyring = file_signatures.ImaKeyring.from_string(agent["ima_sign_verification_keys"])
+
+    # If ima_sign_verification_keys was provided to agent by tenant directly, use that. Otherwise, find keyring in IMA policy.
+    if agent["ima_sign_verification_keys"]:
+        verification_key_string = agent["ima_sign_verification_keys"]
```
THS-on, line 164: We should add a deprecation warning to this.

keylime/ima/ima.py
```diff
@@ -18,26 +18,20 @@
...
 
 
-# The version of the allowlist format that is supported by this keylime release
-ALLOWLIST_CURRENT_VERSION = 5
-
-# A correctly-formatted empty allowlist.
-#
-# IMA allowlists of versions older than 5 will not have the "log_hash_alg"
-# parameter. Hard-coding it to "sha1" is perfectly fine, and the fact one
-# specifies a different algorithm on the kernel command line (e.g., ima_hash=sha256)
```
THS-on, line 28: I think this comment should be still here, because otherwise the hardcoded value causes again some confusion.

keylime/ima/ima.py
```diff
@@ -466,160 +406,92 @@ def read_allowlist(alist=None, checksum="", al_sig_file=None, al_key_file=None):
...
-            for line in f:
-                line = line.strip()
-                if line.startswith("#") or len(line) == 0:
-                    continue
-                excl_list.append(line)
-
-        logger.debug("Loaded exclusion list from %s: %s", exclude_path, excl_list)
+    # validate that the allowlist is proper JSON
+    try:
+        lists = json.loads(ima_policy)
```
THS-on, line 470: Do we want to check against schema also?

keylime/ima/ima.py
```diff
@@ -245,8 +239,8 @@ def _process_measurement_list(
     if lists is not None:
```
THS-on, line 239: I think we can rename `lists` to `ima_policy` here. 

Also can we eliminate the `json.loads(lists)`? This function should be only called with an already parsed policy and if not we should fix that.

mbestavros: @THS-on Agreed on both counts. I've included a stub function called `deserialize_ima_policy` that is intended to translate the IMA policy DB object into something usable by the verifier; it doesn't do anything right now (the intention was to use it to load DSSE-formatted IMA policies, but that didn't make it in), but I could adapt it to return JSON-formatted policies. That would then be passed down the chain and we could eventually drop that `json.loads` call.

mbestavros: @THS-on From digging into the code, it seems like there are some tests that pass JSON-formatted strings to `process_measurement_list()`, but that's the only scenario I'm aware of where that function isn't processing a JSON dictionary. I think we can remove those tests.

keylime/ima/ima.py
mbestavros: I've added the comment back in my latest revision.

keylime/cloud_verifier_common.py
mbestavros: Done - I added a comment in the verifier, but the tenant will also print a message when the `ima_sign_verification_keys` argument is provided.

keylime/ima/ima.py
mbestavros: I think that would be a good idea, though, like you mentioned, we'd need to investigate the state of dependencies for that functionality before implementing.

In the meantime, I've extended this policy correctness check to also apply to the Keylime tenant (so whatever checks are present fail early).

mbestavros:
I've addressed @THS-on's initial review and made some small QoL improvements, including satisfying lints.

@kkaarreell From a quick glance at the Red Hat tests, it seems like most of the failures are due to tenant changes (mostly the dropping of the `--exclude` flag, and knock-on failures, but possibly more). It's likely that many tests will need to be rewritten with new test data before everything passes.

kkaarreell:
> @kkaarreell From a quick glance at the Red Hat tests, it seems like most of the failures are due to tenant changes (mostly the dropping of the `--exclude` flag, and knock-on failures, but possibly more). It's likely that many tests will need to be rewritten with new test data before everything passes.

Ideally we fix tests in a separate branch/PR and at the same time (temporarily) point tmt test plans in both PRs to the respective repo&branch counterpart so that these two PRs gets mutually tested.

keylime/ima/ima.py
THS-on: Yes and make sure that when the data is loaded from the database is a dictionary and not a string. This normally happens on startup of the verifier.

mbestavros: @THS-on Yes - you can see that any time we ask for a policy from the database, it has to go through `ima.deserialize_ima_policy()`, which at current just calls `json.loads()` on the database object - so it will always be JSON after. In the future we can add to this function to support things like DSSE in the DB.

keylime/cli/policies.py
```diff
@@ -43,6 +46,13 @@ def process_allowlist(args):
         # Auto-enable IMA (or-bit mask)
         tpm_policy["mask"] = hex(int(tpm_policy["mask"], 0) | (1 << config.IMA_PCR))
 
+        logger.info(
```
aplanas, line 49: Based on the text, would this be a `logger.warning?

keylime/cli/options.py
```diff
@@ -9,9 +9,7 @@ class UserError(Exception):
     pass
 
 
-def get_opts_error(args) -> Tuple[bool, str]:
-    if args.ima_exclude and not args.allowlist:
-        return True, "--exclude cannot be used without an --allowlist"
+def get_opts_error(args):
```
aplanas, line 12: Is OK to remove the type annotation? I think there is an ongoing PR that is enforcing this.

keylime/common/cli_opts.py
```diff
@@ -68,6 +69,13 @@ def process_allowlist(args):
         # Auto-enable IMA (or-bit mask)
         tpm_policy["mask"] = hex(int(tpm_policy["mask"], 0) | (1 << config.IMA_PCR))
 
+        logger.info(
```
aplanas, line 72: same

aplanas (commented):
Only minimal nits. I should deep a bit better in the code : )))

keylime/cli/options.py
mbestavros: Oops, I think that may have been something gone wrong with my rebase...

keylime/common/cli_opts.py
mbestavros: fixed!

keylime/cli/policies.py
mbestavros: Good catch, updated!

keylime/ima/ima.py
THS-on: Makes sense. Schema validation can be easily implemented after this PR is merged.

mbestavros:
@THS-on @mpeters @stefanberger @aplanas Though this needs a rebase (and likely a squash eventually), I've got an e2e test branch that passes against this PR! That means all CI tests now pass and it should be ready for final review. Please let me know any feedback you have - I'd like to get this on the road to merge sooner rather than later.

keylime/cli/policies.py
```diff
@@ -43,6 +46,13 @@ def process_allowlist(args):
         # Auto-enable IMA (or-bit mask)
         tpm_policy["mask"] = hex(int(tpm_policy["mask"], 0) | (1 << config.IMA_PCR))
 
+        logger.warning(
+            """
+            WARNING: Verification key support in the Keylime tenant is deprecated. Provide
+            verification keys as part of an IMA policy instead.
+            """
+        )
```
aplanas, line 54:
```suggestion
        logger.warning(
            "WARNING: Verification key support in the Keylime tenant is deprecated. Provide "
            "verification keys as part of an IMA policy instead."
        )
```

The `"""` will include the spaces.

keylime/ima/ima.py
```diff
@@ -625,46 +462,69 @@ def ima_policy_db_contents(ima_policy_name: str, ima_policy: str, tpm_policy: st
...
-                if line.startswith("#") or len(line) == 0:
-                    continue
-                excl_list.append(line)
+def verify_ima_policy(ima_policy, ima_policy_key=None, ima_policy_sig=None, verify_sig=True):
+    """
+    Verify that an IMA policy is valid. If provided IMA policy has a detached signature, verify the signature.
+    """
+
+    if ima_policy is None:
+        raise ImaValidationError("No IMA policy provided!", 400)
```
aplanas, line 484: For uniformity, `message="..", code=400`?

keylime/ima/ima.py
```diff
@@ -625,46 +462,69 @@ def ima_policy_db_contents(ima_policy_name: str, ima_policy: str, tpm_policy: st
...
+    try:
+        lists = json.loads(ima_policy)
+    except Exception as error:
+        raise ImaValidationError(message="IMA policy is not valid JSON!", code=400) from error
+
+    # Validate exlude list contains valid regular expressions
+    _, excl_err_msg = validators.valid_exclude_list(lists.get("exclude"))
+    if excl_err_msg:
+        raise ImaValidationError(
+            message=excl_err_msg + " Exclude list regex is misformatted. Please correct the issue and try again.",
```
aplanas, line 510: f-string

keylime/ima/ima.py
```diff
@@ -625,46 +462,69 @@ def ima_policy_db_contents(ima_policy_name: str, ima_policy: str, tpm_policy: st
...
+            code=400,
+        )
 
-    return excl_list
+
+def deserialize_ima_policy(ima_policy):
+    """
+    TODO: Extract IMA policy JSON from DSSE envelope, if applicable.
```
aplanas, line 517: Is this TODO real? If so add it as a comment in the code, not as a doc-string

packit-ci.fmf
```diff
@@ -99,8 +99,8 @@
 
   discover:
     how: fmf
-    url: https://github.com/RedHat-SP-Security/keylime-tests
-    ref: main
+    url: https://github.com/mbestavros/keylime-tests
```
aplanas, line 102: left over?

aplanas (commented):
I did not tests the changes ...

keylime/cloud_verifier_tornado.py
```diff
@@ -570,14 +556,29 @@ def post(self):
...
+                        logger.info("IMA policy data not provided with request! Using default empty IMA policy.")
+                        ima_policy = json.dumps(ima.EMPTY_IMA_POLICY)
 
                     if ima_policy:
-                        err_msg = cloud_verifier_common.validate_ima_policy_data(ima_policy)
-                        if err_msg:
-                            web_util.echo_json_response(self, 400, err_msg)
-                            logger.warning(err_msg)
+                        ima_policy_sig = json_body.get("ima_policy_sig")
+                        ima_policy_key = json_body.get("ima_policy_key")
```
THS-on, line 564: We need to document those options in the API docs somewhere. Once we move to DSSE the plan was to remove them again (at least the signature option), right?

THS-on (approved):
I still need to test it again, but in general it LGTM.

packit-ci.fmf
mbestavros: @aplanas This is necessary to run updated e2e tests, per @kkaarreell. Once this PR is ready for merge, we'll merge the updated tests and change this pointer back to the mainline test branch.

keylime/cloud_verifier_tornado.py
mbestavros: @THS-on For now, these options are the same as before, at least from the tenant API's perspective. Do we have documentation for the verifier API?

THS-on: > Do we have documentation for the verifier API?

Yes, here: https://github.com/keylime/keylime/blob/master/docs/rest_apis.rst

keylime/cloud_verifier_tornado.py
```diff
@@ -852,18 +853,26 @@ def post(self):
...
+            ima.verify_ima_policy(
+                ima_policy.encode(),
+                ima_policy_key,
+                ima_policy_sig,
+                verify_sig=config.getboolean("verifier", "require_allow_list_signatures", fallback=False),
+            )
+        except ima.ImaValidationError as e:
+            web_util.echo_json_response(self, e.code, e.message)
+            logger.warning(e.message)
+            return
```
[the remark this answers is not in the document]
mbestavros: @aplanas How do you mean, exactly?

keylime/ima/ima.py
mbestavros: @aplanas Thanks, fixed - I've also added a proper docstring to that function too.

keylime/ima/ima.py
mbestavros: Fixed, thanks!

keylime/ima/ima.py
mbestavros: Fixed, thanks!

keylime/cli/policies.py
mbestavros: Fixed, thanks!

keylime/cloud_verifier_tornado.py
mbestavros: @THS-on My latest commit adds documentation on the new payload fields. It also adds missing documentation on the Allowlist API from earlier, for completeness!

packit-ci.fmf
mbestavros: @aplanas See here: https://github.com/RedHat-SP-Security/keylime-tests/pull/256

mbestavros:
Okay - after a string of rebases on top of @stefanberger and @maugustosilva's work, I think this should be ready.

The new `generator` field in the allowlist/IMA policy metadata caused some headaches that needed to be fixed, but everything works again. I've also manually tested the database upgrade to be sure that it produces a correctly-formatted policy, including the `generator` field.

mpeters:
@mbestavros Is there any way to make this backwards compatible? Can we have it detect that an old format is being used, issue a deprecation warning and then convert? And then we drop the deprecation and back-compat in a future version.

We really need to avoid breaking changes without a period of deprecation.

mbestavros:
@mpeters @maugustosilva I can probably call the format conversion function at some point down the chain. If we want to maintain full compatibility (ex. with detached signature verification support) that would have to happen on the verifier, which might be complicated. It would be easier to just detect and convert at the tenant level.

On the subject of compatibility, I'll also note that at present, I do have an upgrade migration written, but no downgrade migration. Should I try to write one of those too?

mbestavros:
@mpeters I wonder if this might be our opportunity to do a clean switch of the tenant arguments... Instead of trying to detect the kind of policy provided to `--allowlist`, we instead introduce a _new_ tenant option called `--ima_policy` or `--policy` or similar. That would make the opt-in to the new IMA policy format explicit, and would allow us to completely remove the old tenant arguments when we deprecate allowlists.

mpeters:
> @mpeters @maugustosilva I can probably call the format conversion function at some point down the chain. Maintaining full compatibility (with detached signature verification support on the verifier) would probably be pretty complicated with the amount of re-architecting done here. It would be much easier to just detect and convert at the tenant level. Is there a preference around signatures?

I'm fine with the backcompat being at the tenant layer. I think that's just fine from a user's perspective.

> On the subject of compatibility, I'll also note that at present, I do have an upgrade migration written, but no downgrade migration. Should I try to write one of those too?

Yes, we should always have a downgrade in case people rollback their version if they don't like the upgrade.

mpeters:
> @mpeters I wonder if this might be our opportunity to do a clean switch of the tenant arguments... Instead of trying to detect the kind of policy provided to `--allowlist`, we instead introduce a _new_ tenant option called `--ima_policy` or `--policy` or similar. That would make the opt-in to the new IMA policy format explicit, and would allow us to completely remove the old tenant arguments when we deprecate allowlists.

I actually like this idea the best. It's really not an allowlist anymore, but a policy of which the allowlist is now a part. This will help with a lot of internal variable naming as well. And the old option can issue a deprecation warning and then we'll be free to remove it as needed.

mbestavros:
> > @mpeters I wonder if this might be our opportunity to do a clean switch of the tenant arguments... Instead of trying to detect the kind of policy provided to `--allowlist`, we instead introduce a _new_ tenant option called `--ima_policy` or `--policy` or similar. That would make the opt-in to the new IMA policy format explicit, and would allow us to completely remove the old tenant arguments when we deprecate allowlists.
> 
> I actually like this idea the best. It's really not an allowlist anymore, but a policy of which the allowlist is now a part. This will help with a lot of internal variable naming as well. And the old option can issue a deprecation warning and then we'll be free to remove it as needed.

Cool, I'll write that up. Does the community have any preference as to the argument name? `--policy` would give us the most latitude as policies evolve in the future, but might be too general. `--ima_policy` would fit best with the original proposal.

mbestavros:
@mpeters @maugustosilva My latest commit re-implements support for `--exclude` and introduces a set of new, separate `--policy` tenant arguments.

Now, when an allowlist/exclude list are provided to the appropriate arguments, the tenant will convert them to a compliant IMA policy automatically. This should have the benefit of maintaining much of the compatibility needed to keep existing deployments running while also allowing opt-in to the new policies.

I still need to update the end-to-end tests to conform to the new arguments, and add a downgrade migration for full compatibility. Let me know what you think!

maugustosilva:
> @mpeters @maugustosilva My latest commit re-implements support for `--exclude` and introduces a set of new, separate `--policy` tenant arguments.
> 
> Now, when an allowlist/exclude list are provided to the appropriate arguments, the tenant will convert them to a compliant IMA policy automatically. This should have the benefit of maintaining much of the compatibility needed to keep existing deployments running while also allowing opt-in to the new policies.
> 
> I still need to update the end-to-end tests to conform to the new arguments, and add a downgrade migration for full compatibility. Let me know what you think!

Sorry for being late in answering (just finished watching Brazil's game at the World Cup :-).

What I envision in the future is to have **both** IMA and MB (Measured Boot) "policies" as first-class objects in `keylime` (eventually on even on their separate tables). That being said, I propose the option to be named `--ima_policy` (instead of simply "policy"... in anticipation of a future `--mb_policy`, to be there together with already existing `--tpm_policy`). The current `--allowlist` and `--exclude` will be marked for deprecation in the future.

I believe a "downgrade" might not even be necessary, as long as we give time - a couple of releases - before effectively deprecating the other two older IMA-related options.

mpeters:
After some more conversations with folks, maybe we should use `file-policy` or `runtime-policy` instead of `ima-policy` so that it's not confused with the kernel IMA policy in `/etc/ima/ima-policy`.

mbestavros:
I think `runtime_policy` makes the most sense, especially since it may encapsulate things beyond plain files.

We're at a bit of a crossroads here; given this PR will now introduce new arguments, I'd like to get the naming right, even if we don't fully convert everything to the new naming scheme immediately. Especially around API endpoints on both the tenant and verifier.

maugustosilva:
> @mpeters I think `runtime_policy` makes the most sense, especially since it may encapsulate things beyond plain files.
> 
> We're at a bit of a crossroads here; given this PR will now introduce new arguments, I'd like to get the naming right, even if we don't fully convert everything to the new naming scheme immediately. Especially around API endpoints on both the tenant and verifier.



> @mpeters I think `runtime_policy` makes the most sense, especially since it may encapsulate things beyond plain files.
> 
> We're at a bit of a crossroads here; given this PR will now introduce new arguments, I'd like to get the naming right, even if we don't fully convert everything to the new naming scheme immediately. Especially around API endpoints on both the tenant and verifier.

I agree with you there, getting the name right is important. Since we already have `tpm_policy`, we could go with `ima_runtime_policy` and **for the future**, `mb_policy`.

mbestavros:
@maugustosilva Are you particularly attached to the extra `ima_` at the beginning of the arguments? I've already done the (fairly tedious) refactor to `runtime_policy`, and I'd like to avoid redoing it if I can.

maugustosilva:
> @maugustosilva Are you particularly attached to the extra `ima_` at the beginning of the arguments? I've already done the (fairly tedious) refactor to `runtime_policy`, and I'd like to avoid redoing it if I can.



> @maugustosilva Are you particularly attached to the extra `ima_` at the beginning of the arguments? I've already done the (fairly tedious) refactor to `runtime_policy`, and I'd like to avoid redoing it if I can.

Not really. We can proceed without this additional refactor.

maugustosilva:
@mbestavros I just tried this PR, and it looks like everything, even simple "identity" attestation in kind of broken. Should these parameters change?

```
[root@bmklqavm1 /]# /usr/local/bin/keylime_tenant -c update -t 100.127.0.15 -v 100.127.0.13 -u be32d72b-04af-5c07-b93a-86730100832a --no-verifier-check --cert /var/lib/keylime/ca/be32d72b-04af-5c07-b93a-86730100832a --tpm_policy '{"23":"0000000000000000000000000000000000000000000000000000000000000000"}'
2022-12-08 19:22:25.460 - keylime.tpm - INFO - TPM2-TOOLS Version: 5.1.1
2022-12-08 19:22:25.464 - keylime.tenant - INFO - Setting up client TLS...
2022-12-08 19:22:25.464 - keylime.tenant - INFO - Using default client_cert option for tenant
2022-12-08 19:22:25.465 - keylime.tenant - INFO - Using default client_key option for tenant
2022-12-08 19:22:25.465 - keylime.tenant - WARNING - Warning: agent mTLS is currently disabled, keys will be sent in the clear! This should only be used for testing.
2022-12-08 19:22:25.470 - keylime.tenant - INFO - TLS is enabled.
2022-12-08 19:22:25.472 - urllib3.connectionpool - DEBUG - Starting new HTTPS connection (1): 100.127.0.12:8891
2022-12-08 19:22:25.490 - urllib3.connectionpool - DEBUG - https://100.127.0.12:8891 "GET /v2.1/agents/be32d72b-04af-5c07-b93a-86730100832a HTTP/1.1" 200 None
2022-12-08 19:22:25.491 - keylime.tenant - WARNING - Warning: mTLS for agents is disabled: the identity of each node will be based on the properties of the TPM only. Unless you have strict control of your network, it is strongly advised that remote code execution should be disabled, by setting "payload_script=" and "extract_payload_zip=False" under "[agent]" in agent configuration file.
2022-12-08 19:22:25.492 - urllib3.connectionpool - DEBUG - Starting new HTTP connection (1): 100.127.0.15:9002
2022-12-08 19:22:25.497 - urllib3.connectionpool - DEBUG - http://100.127.0.15:9002 "GET /version HTTP/1.1" 200 69
2022-12-08 19:22:25.498 - keylime.cli.policies - INFO - TPM PCR Mask from policy is 0x800000
Please enter the password to decrypt your keystore:
2022-12-08 19:22:28.807 - keylime.ca-util - INFO - Creating cert package for be32d72b-04af-5c07-b93a-86730100832a in be32d72b-04af-5c07-b93a-86730100832a-pkg.zip
2022-12-08 19:22:28.835 - keylime.ca-util - INFO - Creating cert package for RevocationNotifier in RevocationNotifier-pkg.zip
2022-12-08 19:22:28.838 - urllib3.connectionpool - DEBUG - Starting new HTTPS connection (1): 100.127.0.13:8881
2022-12-08 19:22:28.856 - urllib3.connectionpool - DEBUG - https://100.127.0.13:8881 "DELETE /v2.1/agents/be32d72b-04af-5c07-b93a-86730100832a HTTP/1.1" 404 60
2022-12-08 19:22:28.856 - keylime.tenant - ERROR - Response code 404: agent id not found
2022-12-08 19:22:28.857 - keylime.tenant - WARNING - Connecting to agent without using mTLS!
2022-12-08 19:22:28.858 - urllib3.connectionpool - DEBUG - Starting new HTTP connection (1): 100.127.0.15:9002
2022-12-08 19:22:29.099 - urllib3.connectionpool - DEBUG - http://100.127.0.15:9002 "GET /v2.0/quotes/identity?nonce=QqwhZD0IHYAj2oPkRcHq HTTP/1.1" 200 2009
{'code': 200, 'status': 'Success', 'results': {'quote': 'r/1RDR4AYACIACzymoTTLKcKVbxhDZ9pkEbjN9fDjPjCL4fSa0wcXk6aHABRRcXdoWkQwSUhZQWoyb1BrUmNIcQAAAADdo8fbAAAAAwAAAAABIBkQIwAWNjYAAAABAAsDAAABACB4l2Bo6MOwA+ABKTA9hsc3Z6sRX5b77xB/HyAEUGxJcw==:ABQACwEAdveJCtUHwtJSAU+4qle6rkohUK7Ubp0RmMO4AROG6PyNfboIYtSiFQrP7IYqvtv340Jb+xBzi5hhTUAplNMgSYCr97mkCcsNx2O8scecP8Z7QO9UKYFmguNyf3LVXVLMG0EvlCdas2adcS7Wj8MgTMt2pbIs6/RL6ZW0ids6OIeqNtMHXHvXOUPwIlTBNHMzfdI+xkALz/XvWPcyiU8bgJnj1bGlCqQrM7k4fkQDi1VZIDSMklZ+IYUSypkGmr91UB+e1e0Gdv5qUECDRAj4i6+KhMRBr/YP76SNDaZsJa8JRjjq2YHoNVvyWLzGCVJJfX7jmGp4pVhaTlNcltCQVA==:AQAAAAsAAwAAAQAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAQAAAAEAAAAgAFEiUGw4CjUgWMO4E7APJS3PLPVNt/sWzwMZLfbcuaZfAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=', 'hash_alg': 'sha256', 'enc_alg': 'rsa', 'sign_alg': 'rsassa', 'pubkey': '-----BEGIN PUBLIC KEY-----\nMIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAxBNA0GgnaoN+gpaut0hb\n5qGDdp7NCmYIJ7dgPACbW5YbRSXXOzGPs4aQiZIEPp+SFvfFYvr5R2+/WLByd4YJ\nQjwxEdCpEkp5XMLTnPRFmiIiLhofhozaxe3yKM3t1zdHUrmfYRDiMI1rp3y75n7K\nMhJTDnhzSa7VVSldFXxqCvHT81gYk+R4EcvCfUCC6bNTXEfW1h6HdKcCZdQluy/g\nTgbwyUJMAn+w+YRW7hxq/UYS+wSAI9c8bDD4A+q+EG8TGkD3IWwe0m/SRiFv/4fK\nIa8zxavaHqME5574nnOMWsf6q2cH9CEEtxDJdEhC5QsvEGfPhmdsA9Y/PZT1KcAS\nWQIDAQAB\n-----END PUBLIC KEY-----\n'}}
2022-12-08 19:22:29.100 - keylime.tenant - DEBUG - Agent_quote received quote: r/1RDR4AYACIACzymoTTLKcKVbxhDZ9pkEbjN9fDjPjCL4fSa0wcXk6aHABRRcXdoWkQwSUhZQWoyb1BrUmNIcQAAAADdo8fbAAAAAwAAAAABIBkQIwAWNjYAAAABAAsDAAABACB4l2Bo6MOwA+ABKTA9hsc3Z6sRX5b77xB/HyAEUGxJcw==:ABQACwEAdveJCtUHwtJSAU+4qle6rkohUK7Ubp0RmMO4AROG6PyNfboIYtSiFQrP7IYqvtv340Jb+xBzi5hhTUAplNMgSYCr97mkCcsNx2O8scecP8Z7QO9UKYFmguNyf3LVXVLMG0EvlCdas2adcS7Wj8MgTMt2pbIs6/RL6ZW0ids6OIeqNtMHXHvXOUPwIlTBNHMzfdI+xkALz/XvWPcyiU8bgJnj1bGlCqQrM7k4fkQDi1VZIDSMklZ+IYUSypkGmr91UB+e1e0Gdv5qUECDRAj4i6+KhMRBr/YP76SNDaZsJa8JRjjq2YHoNVvyWLzGCVJJfX7jmGp4pVhaTlNcltCQVA==:AQAAAAsAAwAAAQAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAQAAAAEAAAAgAFEiUGw4CjUgWMO4E7APJS3PLPVNt/sWzwMZLfbcuaZfAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=
2022-12-08 19:22:29.101 - keylime.tenant - DEBUG - Agent_quote received public key: -----BEGIN PUBLIC KEY-----
MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAxBNA0GgnaoN+gpaut0hb
5qGDdp7NCmYIJ7dgPACbW5YbRSXXOzGPs4aQiZIEPp+SFvfFYvr5R2+/WLByd4YJ
QjwxEdCpEkp5XMLTnPRFmiIiLhofhozaxe3yKM3t1zdHUrmfYRDiMI1rp3y75n7K
MhJTDnhzSa7VVSldFXxqCvHT81gYk+R4EcvCfUCC6bNTXEfW1h6HdKcCZdQluy/g
TgbwyUJMAn+w+YRW7hxq/UYS+wSAI9c8bDD4A+q+EG8TGkD3IWwe0m/SRiFv/4fK
Ia8zxavaHqME5574nnOMWsf6q2cH9CEEtxDJdEhC5QsvEGfPhmdsA9Y/PZT1KcAS
WQIDAQAB
-----END PUBLIC KEY-----

2022-12-08 19:22:29.101 - keylime.tenant - DEBUG - Agent_quote received hash algorithm: sha256
2022-12-08 19:22:29.101 - keylime.tenant - DEBUG - Agent_quote received encryption algorithm: rsa
2022-12-08 19:22:29.101 - keylime.tenant - DEBUG - Agent_quote received signing algorithm: rsassa
2022-12-08 19:22:29.141 - keylime.cert_utils - DEBUG - EK cert matched cert: /var/lib/keylime/tpm_cert_store/VE_cert.pem
2022-12-08 19:22:29.142 - keylime.tenant - INFO - Checking EK with script /usr/local/bin/ek_check
2022-12-08 19:22:29.189 - keylime.cert_utils - DEBUG - External check script successfully to validated EK
2022-12-08 19:22:29.189 - keylime.cert_utils - INFO - ek_check output:
2022-12-08 19:22:29.189 - keylime.tenant - INFO - Quote from 100.127.0.15 validated
2022-12-08 19:22:29.190 - keylime.tenant - DEBUG - b64_encrypted_u: W1scJNfYkFb+q6ZC8alFnTEsBCpt3ZyeurxNhncBd2AnEPfNKPqqboJDzCzPR0Muz53wSiUl/FimtSDY0uYQzvqmMGgjQokun7cdHkno+uInyYA/yZ/Bo9wR1rSrqPwBmfkDen6mOGs7PbMCRXleAl2QCkz1wym3CzGmNFjD5D9vC0lhPKh8iTyTdBtr4Fhvk5Yi1dUz9ro66GAr4SwDlQ9uyl5tvnv95Nis0sy9ixS/6KUQ6R8ZS9ioQh4YXxcEZSTgFqBL8isKlbr/paWMNy+do1Y89mQC2NPKywM+fegHazjVn45NaFaL1HtjO0U0C/rrR+69e4K+oCxAuI7bcQ==
2022-12-08 19:22:29.190 - keylime.tenant - WARNING - Connecting to agent without using mTLS!
2022-12-08 19:22:29.193 - urllib3.connectionpool - DEBUG - Starting new HTTP connection (1): 100.127.0.15:9002
2022-12-08 19:22:29.196 - urllib3.connectionpool - DEBUG - http://100.127.0.15:9002 "POST /v2.0/keys/ukey HTTP/1.1" 200 0
2022-12-08 19:22:29.197 - keylime.tenant - DEBUG - b64_v: vc2dlPGAuGeMUErPVXwGtskMR40l/4bRpRIWt2wy0wU=
2022-12-08 19:22:29.198 - keylime.tenant - ERROR - Object of type bytes is not JSON serializable
  File "/usr/local/lib/python3.9/site-packages/keylime-6.5.2-py3.9.egg/keylime/cmd/tenant.py", line 10, in main
    tenant.main()
  File "/usr/local/lib/python3.9/site-packages/keylime-6.5.2-py3.9.egg/keylime/tenant.py", line 1585, in main
    mytenant.do_cvadd()
  File "/usr/local/lib/python3.9/site-packages/keylime-6.5.2-py3.9.egg/keylime/tenant.py", line 561, in do_cvadd
    json_message = json.dumps(data)
  File "/usr/lib64/python3.9/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
  File "/usr/lib64/python3.9/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib64/python3.9/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/usr/lib64/python3.9/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type bytes is not JSON serializable
```

mbestavros:
@maugustosilva Try now - I think I know what caused the issue. I'm also going to test.

maugustosilva:
Works now, without any need for modifications on our deployment framework.

LGTM

maugustosilva (approved):

kkaarreell:
Hi Mark, the failure on rawhide is expected. You would have to rebase to get rid of it.

mbestavros:
@kkaarreell Thanks for the heads up! Just rebased.

mpeters:
@mbestavros Can you squash these all down to a single commit? I'll continue reviewing, but it's better for the merge to be a single commit

keylime/migrations/versions/8c0f8ded1f90_convert_allowlists_to_ima_policies.py
```diff
@@ -0,0 +1,140 @@
...
+
+def upgrade_registrar():
+    pass
+
+
+def downgrade_registrar():
+    pass
+
+
+def upgrade_cloud_verifier():
```
mpeters, line 44: So the upgrade is changing the values in `ima_policy` but still leaving the old names for the tables and fields? Could we take this opportunity to upgrade those as well? If we're going to pull off the band-aid, let's get it all off.

test/test_ima_verification.py
```diff
@@ -3,8 +3,8 @@
 Copyright 2020 IBM Corporation
```
mpeters, line 3: When we have a big disruptive change like this where we have to continue to support some deprecated features, it's best to leave existing tests as-is as much as possible so we know we haven't broken something important before it's removed. Deprecated still means supported. So maybe the best way is to create a new `test_runtime_verification` script that mirrors this one but uses all of the new options/formats. This will allow use to test both old and new easily and when we fully remove the old path it's as easy as removing the test file. You can even put a comment saying that in the old file.

test/test_restful.py
```diff
@@ -373,12 +373,14 @@ def setUpClass(cls):
         # Allow targeting a specific API version (default latest)
         cls.api_version = "2.0"
 
-        # Set up allowlist bundles. Use invalid exclusion list regex for bad bundle.
```
mpeters, line 376: Similar comment here about deprecated testing. We will need to test both API usages for a while. So makes sense to have separate tests for each, in this case test methods rather just a new file.

mpeters (changes requested):

keylime/ima/ima.py
```diff
@@ -441,262 +422,146 @@ def process_measurement_list(
...
-                raise Exception("Allowlist has unsupported version {version}")
-        else:
-            logger.debug("Allowlist does not specify a version. Assuming current version %s", ALLOWLIST_CURRENT_VERSION)
 
-        alist["meta"]["generator"] = IMA_POLICY_GENERATOR.CompatibleAllowList
+def runtime_policy_db_contents(runtime_policy_name: str, runtime_policy: str, tpm_policy: str = "") -> Dict[str, Any]:
+    """Assembles a runtime policy dictionary to be written on the database"""
+    runtime_policy_db_format: Dict[str, Any] = {}
+    runtime_policy_db_format["name"] = runtime_policy_name
+    # TODO: This was required to ensure e2e CI tests pass
```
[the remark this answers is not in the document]
mbestavros: I think that `TODO` was left in from when it was implemented by @stefanberger (I think?)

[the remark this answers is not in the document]
stefanberger:
```
commit 89067d731eb155f6eb85c9e7103317216cd58f48
Author: Marcio Silva <marcio.a.silva@ibm.com>
Date:   Mon Nov 14 14:50:09 2022 -0500
```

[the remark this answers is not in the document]
mbestavros: Whoops, @maugustosilva then! Apologies.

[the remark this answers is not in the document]
maugustosilva: The subsequent four lines
```
    if ima_policy == "{}":
        ima_policy_db_format["ima_policy"] = None
    else:
```

Basically, there was a test on the RH CI which checked for "ima_policy" to be boolean `None`, and thus failed when the "ima_policy" was just an empty JSON ("{}")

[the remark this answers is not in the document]
mpeters: @maugustosilva so is this something we should fix on those tests? Or should we leave the conversion to `None`?

keylime/migrations/versions/8c0f8ded1f90_convert_allowlists_to_ima_policies.py
mbestavros: As discussed in the Keylime video meeting today, let's hold off on database changes for a separate PR.

test/test_ima_verification.py
mbestavros: As discussed in the team meeting, we aren't maintaining any legacy tenant <-> verifier APIs, so we shouldn't need to maintain backwards-compatible tests either.

test/test_restful.py
mbestavros: As discussed in the team meeting today, we aren't maintaining any legacy tenant <-> verifier APIs, so we shouldn't need to maintain backwards-compatible tests either.

keylime/cli/policies.py
```diff
@@ -25,12 +28,14 @@ def enforce_pcrs(tpm_policy, protected_pcrs, pcr_use):
             sys.exit(1)
 
 
-def process_allowlist(args):
+def process_policy(args):
```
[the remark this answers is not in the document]
mbestavros: @mpeters At present, this function does also deal with `tpm_policy` and `mb_refstate` as well, so maybe it's most descriptive to leave it as is? A later refactor could pull those bits out into a separate function, and then it could be renamed.

keylime/cli/options.py
```diff
@@ -25,25 +25,33 @@ def extract_password(pwstring):
 
 
 def get_opts_error(args) -> Tuple[bool, str]:
+    if args.command == "addallowlist" and not (
```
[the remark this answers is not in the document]
mbestavros: I agree with the naming change, but we probably want to do the same strategy as with `--allowlist` where we provide new options and then deprecate the old ones later.

docs/rest_apis.rst
```diff
@@ -201,6 +205,66 @@ Cloud verifier (CV)
     Stop cv polling on `agent_id`, but don’t delete (for an already started `agent_id`).
     This will make the agent verification fail.
 
+
+.. http:post::  /v2.1/allowlists/{ima_policy_name:string}
```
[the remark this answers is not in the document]
mbestavros: While agreed in principle, as discussed in the team meeting today, we should hold off until a separate PR, since this is big/disruptive enough as is and the verifier APIs are not user-facing.

keylime/ima/ima.py
```diff
@@ -441,262 +422,146 @@ def process_measurement_list(
...
-                # matching checksums for identical policies.
-                ima_policy_bytes = ima_policy.replace(meta_timestamp, "").encode()
+def verify_runtime_policy(
+    runtime_policy: bytes,
+    runtime_policy_key: Optional[bytes] = None,
+    runtime_policy_sig: Optional[bytes] = None,
+    verify_sig: Optional[bool] = True,
+) -> None:
+    """
+    Verify that a runtime policy is valid. If provided IMA policy has a detached signature, verify the signature.
```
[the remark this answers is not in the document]
mbestavros: Good catch - fixed in my latest push.

keylime/cli/options.py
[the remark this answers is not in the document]
mbestavros: @mpeters This is now in place in my latest push.

keylime/cli/policies.py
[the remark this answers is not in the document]
mpeters: Nevermind then, I didn't catch that it also processed those other policy things.

packit-ci.fmf
mpeters: @mbestavros and @kkaarreell what is the general process for merges like this that require a different testing branch? Do we merge this change first and then merge the changes for `RedHat-SP-Security`? If so, we'll need to make sure we do that it quick succession so CI isn't broken. Or is it best to merge those changes first and then update this PR to no longer point to the branch? If that latter is possible I think it would be preferrable.

kkaarreell: We can merge test update first and then restore ref in this PR. I am just not at work right now.

kkaarreell: However, Mark's test PR has failing tests ATM

mpeters: Right, we're not ready to merge this yet (and I would like to cut a new incremental release before this anyway) I was just asking about the process when we get to it.

keylime.conf
```diff
@@ -338,19 +338,19 @@ severity_policy = [{"event_id": ".*", "severity_label": "emergency"}]
 tomtou_errors = False
 
 # Durable Attestation is currently marked as an experimental feature
-# In order to enable Durable Attestation, an "adapter" for a Persistent data Store 
-# (time-series like database) needs to be specified. Some example adapters can be 
-# found under "da/examples" so, for instance 
+# In order to enable Durable Attestation, an "adapter" for a Persistent data Store
```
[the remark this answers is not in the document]
mbestavros: Removed in my latest push.

keylime.conf
```diff
@@ -477,10 +477,6 @@ tpm_policy = {"22":["0000000000000000000000000000000000000001","0000000000000000
 # SHA1sum path_to_file
 ima_allowlist = allowlist.txt
 
-# Specify a file containing paths that should be ignored by IMA.
```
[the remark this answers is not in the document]
mbestavros: @mpeters FWIW, this (as well as the similar "ima_allowlist" entry) have not been used in code for a while, so their removal shouldn't impact functionality.

packit-ci.fmf
mbestavros: @kkaarreell I've just kicked a retest which should work.

mbestavros:
@mpeters I've addressed most of your review and squashed the PR down to a single commit. The only outstanding issue is, I believe, the config file version and migration.

Is there anything else outstanding that would be blocking an eventual merge?

mbestavros:
@stefanberger @kkaarreell Apologies for the review notifications - apparently you can only have six reviewers on a PR, and re-requesting a seventh kicks someone else off.

stefanberger:
> @stefanberger @kkaarreell Apologies for the review notifications - apparently you can only have six reviewers on a PR, and re-requesting a seventh kicks someone else off.

FYI: I'll be out for a while...

keylime.conf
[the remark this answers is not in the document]
mbestavros: @mpeters I've had a look at the existing migration files and can't find any reference to either `ima_allowlist` or `ima_excludelist` in any of the current materials. In other words, once the unified `keylime.conf` is migrated to the separate config files, neither of those fields are carried over - they won't exist in any of the newly-created config files. I've also checked this against my dev environment's config files - they're not there.

Seems like these were just completely unused options, and we can safely remove them without trouble.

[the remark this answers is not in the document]
THS-on: Yeah, I think we dropped those default options a while ago.

mpeters:
@mbestavros Everything looks good to me now. It seems there are test failures related to infrastructure, not your changes, but we should wait until there's a clean run before merging. Don't worry too much if the actual merger needs to wait until you get back.

keylime/cloud_verifier_tornado.py
```diff
@@ -609,43 +595,58 @@ def post(self):
...
+                        runtime_policy = json.dumps(ima.EMPTY_RUNTIME_POLICY)
+
+                    if runtime_policy:
+                        runtime_policy_sig = json_body.get("runtime_policy_sig")
+                        runtime_policy_key = json_body.get("runtime_policy_key")
+
+                        if runtime_policy_key and runtime_policy_sig:
+                            runtime_policy_key = base64.b64decode(runtime_policy_key)
+                            runtime_policy_sig = base64.b64decode(runtime_policy_sig)
+
```
stefanberger, line 609: Had problems with typecheckers changing variable types... here from `str` to `bytes`.

keylime/cloud_verifier_tornado.py
```diff
@@ -888,42 +890,50 @@ def post(self):
...
-                return
+        runtime_policy = base64.b64decode(json_body.get("runtime_policy")).decode()
+
+        runtime_policy_key = json_body.get("runtime_policy_key")
+        runtime_policy_sig = json_body.get("runtime_policy_sig")
+
+        if runtime_policy_key and runtime_policy_sig:
+            runtime_policy_key = base64.b64decode(runtime_policy_key)
+            runtime_policy_sig = base64.b64decode(runtime_policy_sig)
+
```
stefanberger, line 905: same comment here -- file is probably not currently type-checked ...

keylime/ima/ima.py
```diff
@@ -414,7 +395,7 @@ def _process_measurement_list(
 def process_measurement_list(
     agentAttestState: AgentAttestState,
     lines: List[str],
-    lists: Optional[Union[str, Dict[str, Any]]] = None,
+    runtime_policy: Optional[Dict[str, Any]] = None,
```
stefanberger, line 398: Thanks! 💯

keylime/ima/ima.py
```diff
@@ -441,262 +422,146 @@ def process_measurement_list(
...
-# Does not process exclusion lists.
-def read_allowlist(
+# Read IMA policy files from disk, validate signatures and checksums, and prepare for sending.
+def read_runtime_policy(
     alist: Optional[str] = None,
     checksum: Optional[str] = "",
     al_sig_file: Optional[str] = None,
     al_key_file: Optional[str] = None,
-) -> Dict[str, Any]:
+) -> Tuple[bytes, bytes, bytes]:
```
stefanberger, line 431: I hope the annotations made it easier to detect callers...

keylime/tenant.py
```diff
@@ -1156,32 +1162,50 @@ def do_verify(self):
                 continue
             break
 
-    def do_add_allowlist(self, args):
-        if "allowlist_name" not in args or not args["allowlist_name"]:
-            raise UserError("allowlist_name is required to add an allowlist")
+    def do_add_runtime_policy(self, args):
+        if "runtime_policy_name" not in args or not args["runtime_policy_name"]:
```
stefanberger, line 1166: if args.get("runtime_policy_name") is None:

keylime/tpm/tpm_main.py
```diff
@@ -1225,8 +1226,8 @@ def check_quote(
         if tpm_policy is None:
             tpm_policy = {}
 
-        if allowlist is None:
-            allowlist = {}
+        if ima_policy is None:
+            ima_policy = ima.EMPTY_RUNTIME_POLICY
```
stefanberger, line 1230: ima_policy- > runtime_policy ?

keylime/tpm/tpm_main.py
```diff
@@ -1215,7 +1216,7 @@ def check_quote(
         aikTpmFromRegistrar: str,
         tpm_policy: Optional[Union[str, Dict]] = None,
         ima_measurement_list: Optional[str] = None,
-        allowlist: Optional[Union[str, Dict[str, Any]]] = None,
+        ima_policy: Optional[Dict[str, Any]] = None,
```
stefanberger, line 1219: ima_policy- > runtime_policy ?

keylime/tpm/tpm_abstract.py
```diff
@@ -343,7 +343,7 @@ def check_pcrs(
                     agentAttestState,
                     pcrs_dict[config.IMA_PCR],
                     ima_measurement_list,
-                    allowlist,
+                    ima_policy,
```
stefanberger, line 346: ima_policy- > runtime_policy ?

test/test_ima_verification.py
```diff
@@ -15,11 +15,11 @@
 
 # BEGIN TEST DATA
 
-ALLOWLIST = {
+IMA_POLICY_TEST = {
```
stefanberger, line 18: RUNTIME_POLICY_TEST ?

stefanberger: I had the impression we got rid of ima_policy word within Keylime or did I misunderstand?

mbestavros: @stefanberger Not entirely. There are actually two separate concepts that were previously both referred to as "IMA policies":

- the kernel IMA policy present at `/etc/ima/ima-policy`
- the user-provided Keylime policy that determines file paths and hashes to check against (which this PR overhauls)

To avoid confusion, we decided earlier on in the PR's lifecycle to rename the Keylime user policy to "runtime policies". The highest priority is to change the terminology in any place where it might be user-facing; internal Keylime references aren't as important. I'm also not necessarily sure if some of the internal Keylime references (especially deeper in the code like in `tpm_main.py` and `tpm_abstract.py`) might be referring to the kernel IMA policy instead of the Keylime user policy. CC @mpeters - any insight? Does the Keylime verifier utilize the kernel IMA policy at all?

All that said, I've renamed the "ima_policy" terminology in the test files, since that's certainly an appropriate renaming. Happy to rename others, if they make sense.

mbestavros: Update: per a discussion offline with @mpeters, Keylime likely doesn't interact with the kernel IMA policy directly, so it should be safe to rename these references to `runtime_policy`.

keylime/cloud_verifier_tornado.py
mbestavros: @stefanberger Would you mind clarifying what the issue is a little more? I'm not super familiar with typecheckers yet.

Is it that I need to create separate variables for the `bytes`-decoded key and sig, instead of reusing the existing `str` variables?

mbestavros:
@stefanberger I've addressed your review - and thanks to @mpeters' clarification, I've also gone ahead and updated a bunch of stray `ima_policy` and `allowlist` references to `runtime_policy`.

This includes renaming the `create_allowlist.sh` script to `create_runtime_policy.sh`. @mpeters how do you feel about this rename? I've changed all references to that script in the Keylime codebase, but would existing users be relying on that script?

test/test_ima_verification.py
stefanberger: > Does the Keylime verifier utilize the kernel IMA policy at all?

Afaik Keylime knows nothing about the IMA policy at all.

keylime/cloud_verifier_tornado.py
stefanberger: Sorry, for some reason I didn't seem to get email notifications about your messages here. I would have reacted a bit faster.

What typecheckers don't seem to like is if variables are first of one type and then they change to another one. So, what you did looks good.

mbestavros:
As a note, the currently-failing lint should be fixed by this PR: https://github.com/keylime/keylime/pull/1267

THS-on:
> This includes renaming the create_allowlist.sh script to create_runtime_policy.sh. @mpeters how do you feel about this rename? I've changed all references to that script in the Keylime codebase, but would existing users be relying on that script?

Not @mpeters, but it is mainly for example purposes, so I think it is fine to rename. 

@mbestavros can you split up the commit into the following and cleanup the commit messages:
* Base changes
* renaming of the create_allowlist.sh script
* packit changes (those are dropped before merging)

I now tested the PR manually and found the following things:
* the `--signature-verification-*` options are not marked as deprecated in `keylime_tenant -h` output
* Converting a policy with IMA keys does not work (The allowlist is one with a single entry)  [1] 

Otherwise it seems to work great!

We are very close to merging this and ideally we get this in the master branch in the next week, so that we can make a final 6.X release before start removing the agent from the code.

[1]:
```
[root@fedora ~]# keylime_convert_runtime_policy -v pubkey_evm.pem -o test.json --allowlist allowlist 
allowlist is not JSON-formatted; attempting to convert to IMA policy from flat file format
  File "/usr/local/bin/keylime_convert_runtime_policy", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.10/site-packages/keylime/cmd/convert_runtime_policy.py", line 211, in main
    policy_out = update_runtime_policy(
  File "/usr/local/lib/python3.10/site-packages/keylime/cmd/convert_runtime_policy.py", line 153, in update_runtime_policy
    keyring = file_signatures.ImaKeyring().from_string(policy["ima_sign_verification_keys"])
KeyError: 'ima_sign_verification_keys'
```

mpeters:
I agree with @THS-on, `create_allowlist.sh` is more of an example utility rather than a supported contract. It's not like the other programs that start with `keylime_`, etc. So renaming it should be fine, in fact it's already been renamed before :)

THS-on (approved):
I've now completed my manual testing and it LGTM!

stefanberger:
I tested this now with signed files requiring a signature verification key and also the acceptance of an `ima-buf` entry in the IMA log that is holding a key.

I used a flat file list like this here and converted it to a policy.
```
# cat whitelist.txt
[...]
5ae7268c592dc940cf4a9c88231bc287baec6a968396bc9f427cd9e8cf40c9cc  /usr/bin/troff
a9e6589e14ec90a31aff52d1c5077af9afcb67e247c2d97b11363e0cabed1c41  %keyring:_ima

# keylime_convert_runtime_policy -a whitelist.txt -e exclude.txt -o runtime_policy -v ./verificationkey.der
```

The conversion from a flat file worked and the resulting signature verification worked [better than before with the previous JSON format].

THS-on:
> The lint complain about the "implementing the abstract class Context" seems wrong.

@aplanas  yes this looks like a false positive in pylint. Context is definitely not an abstract class.

stefanberger:
What do we do about the `scripts/create_policy` script?

mbestavros:
> What do we do about the `scripts/create_policy` script?

Whoa, didn't even know that script existed!

The output from that script can (theoretically) be fed into the `convert_runtime_policy.sh` script and then to the new policy architecture, but it would be nice to have a unified script for creating new runtime policies from raw data. There may be some functionality the old script has that the conversion script doesn't (things like ima-buf support etc).

stefanberger:
> What do we do about the `scripts/create_policy` script?

I don't know of any other tool that gives more control over 

> > What do we do about the `scripts/create_policy` script?
> 
> @stefanberger Whoa, didn't even know that script existed!

Never mind.
> 
> The output from that script can (theoretically) be fed into the `keylime/cmd/convert_runtime_policy.py` script and then to the new policy architecture, but it would be nice to have a unified script for creating new runtime policies from raw data. There may be some functionality the old script has that the conversion script doesn't (things like ima-buf support etc).

In one way or another we need to give users control over the various fields of the JSON policy so they at least can add things to it. Besides that it allows you to read the ascii_runtime_measurements and create a policy from it, which may be a bit more strict  that running sha256sum across the filesystem...  I am currently converting it to create the latest policy format from this series.

aplanas:
> which may be a bit more strict that running sha256sum across the filesystem..

As a complement to this, I am converting my [imadata](https://github.com/aplanas/imadata) script to be used as an alternative generator of the IMA hashes from the repository information. I will like to integrate this into keylime

mpeters:
@mbestavros Just need to rebase and fix the conflict and this should be good to merge.

mbestavros:
@mpeters rebased!

Do we want to drop the final commit pointing to updated Packit tests? How do we handle that merge?

mpeters:
> Do we want to drop the final commit pointing to updated Packit tests? How do we handle that merge?

Are the packit tests ready to go? @kkaarreell how have we done this in the past to make a smooth merge?

kkaarreell:
> > Do we want to drop the final commit pointing to updated Packit tests? How do we handle that merge?
> 
> Are the packit tests ready to go? @kkaarreell how have we done this in the past to make a smooth merge?

Tests are ready. We also need to drop changes made in [plans/main.fmf](https://github.com/RedHat-SP-Security/keylime-tests/pull/256/files#diff-5ae9adacc9ce4ac4f6d80f8c06e2f4a4967a272f41bc5f2643ed946c4077f685) and then we can merge it right away.

mpeters (approved):

mbestavros:
@mpeters I just dropped the commit that changed e2e tests. Should be ready to go now.

mbestavros:
@mpeters Tests all pass now!

PR #1148, opened 2022-10-31
Remove semantic-release action to stop erroneous releases

mbestavros:


mpeters (approved):

sergio-correia:
There is a newer version of it, `ycjimmy/semantic-release-action@v2`; I wonder if we should try it before removing it altogether.

maugustosilva (approved):

Issue #1146, opened 2022-10-28
Verifier allowlist implementation bug causes double insert sqlalchemy errors in extreme circumstances


galmasi:
Is your issue a feature request? If so, please raise it as an [enhancement](https://github.com/keylime/enhancements)

# Environment
 * OS / version: Ubuntu 22.04
 * Processor architecture: x86_64
 * TPM Manufacturer: virtual (irrelevant)
 * Keylime version: 6.5.1+

# Description 

In unusual circumstances the current implementation of the keylime allowlist database causes sql failures by attempting to insert entries into the sql database with duplicated indices.

This problem was found "in the wild" on a large scaling cluster where transient mysql access failures are known to happen. However, we have managed to reproduce the problem by causing a mysql connection failure explicitly, then attempting to recover from it.

Generally speaking, to produce the error reliably, one have a running a keylime verifier attached to a mysql database and an ongoing attestation process. No need for IMA to be enabled.

Next, trigger the sequence of events by killing mysql connections while attestation is ongoing. This produces an (expected) attestation failure due to loss of connectivity to mysql.

The actual problem occurs during an attempt to restore attestation. This involves restarting the keylime verifier and attempting to re-add the agent for attestation. The last step produces an sql failure ("duplicate entry") that could be prevented by coding allowlist operations with care.

## Expected behavior vs. actual behavior 

Expected behavior would be that even if an agent stopped attesting amidst a transient mysql failure, re-adding the agent should not cause duplicate database entries. Actual behavior is that it *does*.

## Root cause analysis

The problem seems to be centered around line 568 of `cloud_verifier_tornado.py`, where no check is made whether a `VerifierAllowList` entry exists for an agent before one is inserted. I expect that the fix is trivial -- force-remove the old entry or avoid overwriting it without causing the verifier to throw an exception aborting the `add` process.

## Steps to reproduce problem

1. Set up a keylime verifier running against a  mysql database (important not to use sqlite)
2. Turn on attestation one an agent -- in our case we did not turn on IMA at all, merely IDENT attestation.
3. Get into the database with mysql or mysqladmin, and KILL ALL CONNECTIONS (to simulate a network failure)
4. Check the verifier -- attestation should haved failed with an SQLAlchemy error
5. Restart the verifier to force it to re-establish connectivity to the database.
6. Turn off attestation; turn attestation back on (with keylime_tenant -c delete ; keylime_tenant -c add
7. Expect to see new failures caused by duplicate insertion of agent's allow list.

[...]

PR #1150, opened 2022-11-02
Fix for writing the same allow list twice during agent activation

galmasi:
Addresses Issue #1146 -- failing tenant add with IMA in unusual circumstances (i.e. unreliable mysql connections, or when attempting to run the same `tenant -c add` command simultaneously from two different tenants).

### DIFF
--- a/keylime/cloud_verifier_tornado.py
+++ b/keylime/cloud_verifier_tornado.py
@@ -401,7 +401,9 @@ class AgentsHandler(BaseHandler):
         agents requests require a json block sent in the body
         """
         session = get_session()
-        try:
+        # TODO: exception handling needs fixing
+        # Maybe handle exceptions with if/else if/else blocks ... simple and avoids nesting
+        try:  # pylint: disable=too-many-nested-blocks
             rest_params = web_util.get_restful_params(self.request.uri)
             if rest_params is None:
                 web_util.echo_json_response(self, 405, "Not Implemented: Use /agents/ interface")
--- a/keylime/cloud_verifier_tornado.py
+++ b/keylime/cloud_verifier_tornado.py
@@ -570,12 +570,18 @@ class AgentsHandler(BaseHandler):
                             ima_policy_stored = (
                                 session.query(VerifierAllowlist).filter_by(name=ima_policy_name).one_or_none()
                             )
+                        except SQLAlchemyError as e:
+                            logger.error(
+                                "SQLAlchemy Error while retrieving stored ima policy for agent ID %s: %s", agent_id, e
+                            )
+                            raise
+                        try:
                             if ima_policy_stored is None:
                                 ima_policy_stored = VerifierAllowlist(**ima_policy_db_format)
                                 session.add(ima_policy_stored)
                                 session.commit()
                         except SQLAlchemyError as e:
-                            logger.error("SQLAlchemy Error: %s", e)
+                            logger.error("SQLAlchemy Error while updating ima policy for agent ID %s: %s", agent_id, e)
                             raise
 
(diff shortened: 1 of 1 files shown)


THS-on:
For the `too-many-nested-statements` you can just add an ignore comment. The other changes LGTM, but some one else should also look at this before merging.

aplanas:
I agree with @THS-on . Add the ignore and a TODO to mark it. The fix in the first commit is direct, and the refactoring because of the complain is hiding it.

You can later split the rest in a different commit and remove the ignore. I will review also the second one (but I think that I get how you addressed it)

galmasi:
I was looking at the "nested" lint issue and found that the error condition handling in this function is all over the place -- some if statements defer failures to the end of the function, others return immediately.

I saw an opportunity for fixing the indentation issue by front-loading all failures and leaving the rest of the code unindented. But this results in 100+ lines changed for what was originally a 2 line PR.

In the end, it's a choice between a minimalistic *semantic* fix and a fairly large *stylistic* fix. You get to choose. If you don't like the latest evolution of the PR I start from scratch with the 2 lines that fix the actual problem.

galmasi:
FWIW even the 2 line semantic fix leaves out the fact that some variables are potentially undefined, depending on the path taken through the code. I would urge a serious review of this function (`cloud_verifier_tornado::post`) because there be dragons there. But I didn't want to broach the larger subject in this PR.

galmasi:
Al right @aplanas @THS-on I hear you. Going back to the 2-line fix with a lint exception soon(ish). Don't merge anything yet :)

THS-on:
@galmasi I agree that we should refactor this method to make it cleaner. So it  still makes sense to keep your commits and just put them into another PR. Ideally we get https://github.com/keylime/keylime/pull/1139 merged before which also addresses some of the issues.

mbestavros (changes requested):
We already perform a check for allowlists that exist in the DB to prevent overwriting, but you're right that it doesn't catch the case of an unnamed allowlist. So this additional check makes sense.

Could you add error handling like the other `session.query` calls? There's some code above this with examples.

galmasi:
I think the feeling among the reviewers was to minimize the disruption with this PR -- it's already had too long a life -- and postpone complex fixes. 

Without troubling this PR any further, @maugustosilva @mdrocco and myself are finding issues related to operation of IMA at scale and against databases != sqlite. We don't want to hurry and make things worse. We are collecting data for opening a larger issue that can serve as a basis for fixes.

keylime/cloud_verifier_tornado.py
```diff
@@ -565,7 +565,9 @@ def post(self):
                         ima_policy_db_format["name"] = ima_policy_name
                         ima_policy_db_format["ima_policy"] = ima_policy
                         try:
-                            ima_policy_stored = session.query(VerifierAllowlist).filter_by(name=ima_policy_name).one_or_none()
+                            ima_policy_stored = (
+                                session.query(VerifierAllowlist).filter_by(name=ima_policy_name).one_or_none()
+                            )
```
stefanberger, line 572: you may want to squash this commit into the previous one ...

stefanberger:
LGTM

stefanberger:
To get rid of the test failure you will probably need to rebase on master.

aplanas:
Can you drop the merge commit and make a rebase? I think the history would be a bit more clear : ))

Otherwise LGTM

galmasi:
@aplanas  Dropped the merge and rebased instead.

aplanas (approved):

THS-on (approved):

mpeters (approved):

PR #1151, opened 2022-11-04
ima: Fix log evaluation on quick-succession execution of scripts

stefanberger:
In case the attested-to host quickly executes files measured by IMA we may run into the case that the keylime agent retrieved the state of the PCR at 'state n' but then IMA appended the log with several entries leading to a log representing 'state n + x' (with x>=1), which may not just be the previously anticipated single additional entry (state n+1). Therefore, remove the check for the number of entries in the log and always compare the running_hash that iterative attestation was resumed with against the provided PCR value from 'state n'.


THS-on (approved):

aplanas (approved):
Very fast fixing the typo : ))

PR #1155, opened 2022-11-04
Very limited code fixup/cleanup on keylime_tenant CLI

maugustosilva:
This PR starts the cleanup and fixup of several small issues on `keylime_tenant` code. First of all, make sure all "jsonification" of resquest's responses happen at the static method `_jsonify_response` (renamed from `_print_json_response`) which can now optionally print the response. Second make sure this new function is invoked as often as possible throughout the CLI code, with obvious exceptions being documented. Third, the old `do_cv` method is renamed to `do_cvadd`, in order to improve the code's understandability and maintainability.

This by no means fixes all the problems and issues on `keylime_tenant`, but it is an incremental step in the right direction.

### DIFF
--- a/keylime/tenant.py
+++ b/keylime/tenant.py
@@ -530,7 +530,7 @@ class Tenant:
                 logger.debug("ek_check output: %s", line.strip())
         return True
 
-    def do_cv(self):
+    def do_cvadd(self):
         """Initiate v, agent_id and ip and initiate the cloudinit sequence"""
         b64_v = base64.b64encode(self.V).decode("utf-8")
         logger.debug("b64_v: %s", b64_v)
@@ -569,12 +569,14 @@ class Tenant:
             logger.error("Verifier at %s with Port %s timed out.", self.verifier_ip, self.verifier_port)
             sys.exit()
 
+        response_json = Tenant._jsonify_response(response, print_response=False, raise_except=True)
+
         if response.status_code == 409:
             # this is a conflict, need to update or delete it
-            Tenant._print_json_response(response)
+            print(response_json)
             sys.exit()
         elif response.status_code != 200:
-            keylime_logging.log_http_response(logger, logging.ERROR, response.json())
+            keylime_logging.log_http_response(logger, logging.ERROR, response_json)
             logger.error(
                 "POST command response: %s Unexpected response from Cloud Verifier: %s",
                 response.status_code,
@@ -589,16 +591,18 @@ class Tenant:
 
         response = do_cvstatus.get((f"/v{self.api_version}/agents/{self.agent_uuid}"), timeout=self.request_timeout)
 
+        response_json = Tenant._jsonify_response(response, print_response=False, raise_except=True)
+
         if response.status_code == 503:
             logger.error(
                 "Cannot connect to Verifier at %s with Port %s. Connection refused.",
                 self.verifier_ip,
                 self.verifier_port,
             )
-            return response.json()
+            return response_json
         if response.status_code == 504:
             logger.error("Verifier at %s with Port %s timed out.", self.verifier_ip, self.verifier_port)
-            return response.json()
+            return response_json
         if response.status_code == 404:
             logger.info(
                 "Verifier at %s with Port %s does not have agent %s.",
@@ -606,19 +610,18 @@ class Tenant:
                 self.verifier_port,
                 self.agent_uuid,
             )
-            return response.json()
+            return response_json
+
         if response.status_code == 200:
-            response = response.json()
+            res = response_json.pop("results")
+            response_json["results"] = {self.agent_uuid: res}
 
-            res = response.pop("results")
-            response["results"] = {self.agent_uuid: res}
+            operational_state = states.state_to_str(response_json["results"][self.agent_uuid]["operational_state"])
+            response_json["results"][self.agent_uuid]["operational_state"] = operational_state
 
-            operational_state = states.state_to_str(response["results"][self.agent_uuid]["operational_state"])
-            response["results"][self.agent_uuid]["operational_state"] = operational_state
+            logger.info("Agent Info:\n%s", json.dumps(response_json["results"]))
 
-            logger.info("Agent Info:\n%s", json.dumps(response["results"]))
-
-            return response
+            return response_json
 
         logger.info(
             "Status command response: %s. Unexpected response from Cloud Verifier %s on port %s. %s",
@@ -627,7 +630,7 @@ class Tenant:
             self.verifier_port,
             str(response),
         )
-        return response
+        return response_json
 
     def do_cvlist(self):
         """List all agent statuses in cloudverifier"""
@@ -640,16 +643,18 @@ class Tenant:
 
         response = do_cvstatus.get(f"/v{self.api_version}/agents/?verifier={verifier_id}", timeout=self.request_timeout)
 
+        response_json = Tenant._jsonify_response(response, print_response=False, raise_except=True)
+
         if response.status_code == 503:
             logger.error(
                 "Cannot connect to Verifier at %s with Port %s. Connection refused.",
                 self.verifier_ip,
                 self.verifier_port,
             )
-            return response.json()
+            return response_json
         if response.status_code == 504:
             logger.error("Verifier at %s with Port %s timed out.", self.verifier_ip, self.verifier_port)
-            return response.json()
+            return response_json
         if response.status_code == 404:
             logger.info(
                 "Verifier at %s with Port %s does not have agent %s.",
@@ -657,11 +662,10 @@ class Tenant:
                 self.verifier_port,
                 self.agent_uuid,
             )
-            return response.json()
+            return response_json
         if response.status_code == 200:
-            response = response.json()
 
-            logger.info('From verifier %s port %s retrieved: "%s"', self.verifier_ip, self.verifier_port, response)
+            logger.info('From verifier %s port %s retrieved: "%s"', self.verifier_ip, self.verifier_port, response_json)
 
             return response
 
@@ -687,16 +691,18 @@ class Tenant:
             f"/v{self.api_version}/agents/?bulk={True}&verifier={verifier_id}", timeout=self.request_timeout
         )
 
+        response_json = Tenant._jsonify_response(response, print_response=False)
+
         if response.status_code == 503:
             logger.error(
                 "Cannot connect to Verifier at %s with Port %s. Connection refused.",
                 self.verifier_ip,
                 self.verifier_port,
             )
-            return response.json()
+            return response_json
         if response.status_code == 504:
             logger.error("Verifier at %s with Port %s timed out.", self.verifier_ip, self.verifier_port)
-            return response.json()
+            return response_json
         if response.status_code == 404:
             logger.info(
                 "Verifier at %s with Port %s does not have agent %s.",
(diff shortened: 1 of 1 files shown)


keylime/tenant.py
```diff
@@ -589,6 +591,8 @@ def do_cvstatus(self):
 
         response = do_cvstatus.get((f"/v{self.api_version}/agents/{self.agent_uuid}"), timeout=self.request_timeout)
 
+        response_json = Tenant._jsonify_response(response)
+
```
stefanberger, line 595: there are two more `return response.json()` at lines 602 and 605 that probably should also be converted just like you converted line 613.

keylime/tenant.py
```diff
@@ -640,6 +643,8 @@ def do_cvlist(self):
 
         response = do_cvstatus.get(f"/v{self.api_version}/agents/?verifier={verifier_id}", timeout=self.request_timeout)
 
+        Tenant._jsonify_response(response, True)
```
stefanberger, line 646: Why not `response_json = Tenant._jsonify_response(response)` here as well and then also convert lines 654 and 657, 665, 669,  etc.

keylime/tenant.py
```diff
@@ -687,6 +692,9 @@ def do_cvbulkinfo(self):
             f"/v{self.api_version}/agents/?bulk={True}&verifier={verifier_id}", timeout=self.request_timeout
         )
 
+        # Bulkinfo can be quite large. Will not jsonify it without explicity need
+        # Tenant._jsonify_response(response, True)
```
stefanberger, line 696: lines 704, 707, 715, 717 all convert it to json.

For some odd reason the final return at line 734 doesn't seem to return a `response.json()` but a plain `response`. Maybe that case is never reached ? It's odd for sure if the function was to return  non-JSON there.

keylime/tenant.py
```diff
@@ -915,9 +930,9 @@ def do_cvstop(self):
             logger.error("Verifier at %s with Port %s timed out.", self.verifier_ip, self.verifier_port)
             sys.exit()
 
-        response_body = response.json()
```
stefanberger, line 918: move line 920 `response_json = Tenant._jsonify_response(response, True)` to here...

keylime/tenant.py
```diff
@@ -1162,26 +1177,28 @@ def do_add_allowlist(self, args):
...
         print(f"Show allowlist command response: {response.status_code}.")
-        Tenant._print_json_response(response)
+        Tenant._jsonify_response(response, True)
 
     @staticmethod
-    def _print_json_response(response):
+    def _jsonify_response(response, print_response=False):
         try:
             json_response = response.json()
         except ValueError:
```
stefanberger, line 1197: The difference to 'before' is that all `response.json()` you converted would have thrown a ValueError on bad input and now you change it to an empty map if there's an error . If you now raised the error in the except clause you would approximately get the same result again as before...

stefanberger:
To get rid of the test failure you will probably need to rebase on master.

keylime/tenant.py
maugustosilva: same, thank you

keylime/tenant.py
maugustosilva: This function is in a somewhat fluid state. Right now, this is one of the very few function which returns (stdout) a (properly formatted) JSON string which can be easily consumed by a tool such as `jq`. I would like to make this the standard, but will reserve it for another PR.

keylime/tenant.py
maugustosilva: Very good point. I am changing the behavior of the function to **optionally** raise an exception, thus keeping the usual behavior for the "other" methods. In the future, I will re-organize and standardize the behavior on all methods.

keylime/tenant.py
```diff
@@ -743,36 +748,39 @@ def do_cvdelete(self, verifier_check=True):
         do_cvdelete = RequestsClient(self.verifier_base_url, True, tls_context=self.tls_context)
         response = do_cvdelete.delete(f"/v{self.api_version}/agents/{self.agent_uuid}", timeout=self.request_timeout)
 
-        response = response.json()
+        response_json = Tenant._jsonify_response(response, True)
```
stefanberger, line 751: You want this printed here?

keylime/tenant.py
```diff
@@ -949,8 +961,7 @@ def do_quote(self):
                     do_quote = RequestsClient(cloudagent_base_url, tls_enabled=False)
                     response = do_quote.get(params, timeout=self.request_timeout)
 
-                print(response)
-                response_body = response.json()
+                response_json = Tenant._jsonify_response(response, False, True)
```
stefanberger, line 964: Here: no more printing ?

keylime/tenant.py
```diff
@@ -743,36 +748,39 @@ def do_cvdelete(self, verifier_check=True):
...
-        if response["code"] == 202:
+        if response_json["code"] == 202:
             deleted = False
             for _ in range(12):
                 get_cvdelete = RequestsClient(self.verifier_base_url, True, tls_context=self.tls_context)
                 response = get_cvdelete.get(
                     f"/v{self.api_version}/agents/{self.agent_uuid}", timeout=self.request_timeout
                 )
 
+                response_json = Tenant._jsonify_response(response, False, True)
```
stefanberger, line 771: Here you convert it a 2nd time...

keylime/tenant.py
stefanberger: This should probably throw an exception if conversion fails, so `(..., False, True)` ? A bit tricky this function now...

keylime/tenant.py
maugustosilva: Correct, no more printing here

maugustosilva: Trying to keep it behaving as it was before

keylime/tenant.py
maugustosilva: Yes, it should throw an exception, as it was before

keylime/tenant.py
maugustosilva: It is a new response. This section of the code is a total mess, and I intend to fix that in the next PR :-)

stefanberger: ah, yes, right.

keylime/tenant.py
stefanberger: was before printing: `print(response)`

maugustosilva: ah, very true... I was assuming this was a "leftover print", but you're right. Absent of clear evidence that it is indeed the case, we should retain the print out functionality

keylime/tenant.py
```diff
@@ -569,12 +569,14 @@ def do_cv(self):
             logger.error("Verifier at %s with Port %s timed out.", self.verifier_ip, self.verifier_port)
             sys.exit()
 
+        response_json = Tenant._jsonify_response(response, False, True)
+
         if response.status_code == 409:
             # this is a conflict, need to update or delete it
-            Tenant._print_json_response(response)
+            json.dumps(response_json)
```
stefanberger, line 576: `print(json_response);` ?

keylime/tenant.py
```diff
@@ -1162,26 +1174,31 @@ def do_add_allowlist(self, args):
         response = cv_client.post(
             f"/v{self.api_version}/allowlists/{self.ima_policy_name}", data=body, timeout=self.request_timeout
         )
-        Tenant._print_json_response(response)
+        Tenant._jsonify_response(response)
```
stefanberger, line 1177: This is kind of tricky to understand why we need it jsonify'd and it ends up printing it. Maybe don't print it in the function and do `print(Tenant._jsonify_response(response))` instead.

maugustosilva: My longer term plan is to get agreement from others to have every command output a JSON-formatted string that is easily consumable by tools such as `jq` (think of `do_cvstatus` as the model). However, I am trying to contain the number of changes on this PR without disrupting the existing functionality. Please consider that at some point `_jsonify_response` will json.dumps (maybe even with indent, to make it more human readable).

keylime/tenant.py
maugustosilva: Fair enough, keeping exactly the same functionality as before, I will change.

keylime/tenant.py
maugustosilva: Please note that I have an issue open with this purpose now https://github.com/keylime/keylime/issues/1159

galmasi (approved):
The renaming of functions is much appreciated.
Arguably there should be no print() calls in error handling code in keylime. However, it's a small sin; there are worse dragons in the code.

keylime/tenant.py
aplanas: Same

keylime/tenant.py
```diff
@@ -640,28 +643,29 @@ def do_cvlist(self):
 
         response = do_cvstatus.get(f"/v{self.api_version}/agents/?verifier={verifier_id}", timeout=self.request_timeout)
 
+        response_json = Tenant._jsonify_response(response, False, True)
```
aplanas, line 646: That is the 3rd call with reversed boolean parameters from the default.  Should be that a reversed defaults are better?

keylime/tenant.py
```diff
@@ -569,12 +569,14 @@ def do_cv(self):
             logger.error("Verifier at %s with Port %s timed out.", self.verifier_ip, self.verifier_port)
             sys.exit()
 
+        response_json = Tenant._jsonify_response(response, False, True)
```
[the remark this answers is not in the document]
maugustosilva: Good point, will do

keylime/tenant.py
maugustosilva: I elected to keep the "newest" calls, on the allowlist API (near the end of the file), as unmodified as possible.

keylime/tenant.py
```diff
@@ -743,36 +748,39 @@ def do_cvdelete(self, verifier_check=True):
         do_cvdelete = RequestsClient(self.verifier_base_url, True, tls_context=self.tls_context)
         response = do_cvdelete.delete(f"/v{self.api_version}/agents/{self.agent_uuid}", timeout=self.request_timeout)
 
-        response = response.json()
+        response_json = Tenant._jsonify_response(response, False, True)
```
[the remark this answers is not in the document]
maugustosilva: At this point in time, the goal is to simply mimic the behavior of `.json()`

keylime/tenant.py
```diff
@@ -1162,26 +1174,31 @@ def do_add_allowlist(self, args):
...
-        Tenant._print_json_response(response)
+        Tenant._jsonify_response(response)
 
     @staticmethod
-    def _print_json_response(response):
+    def _jsonify_response(response, print_response=True, raise_except=False):
         try:
             json_response = response.json()
-        except ValueError:
+        except ValueError as e:
```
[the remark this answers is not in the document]
maugustosilva: Here is the crux of the matter: `tenant.py` is not in a good shape in my evaluation, and there will be more cleanup and fixup needed, including what will be required by https://github.com/keylime/keylime/issues/1158 and https://github.com/keylime/keylime/issues/1159 

I consider this to be a transitional and incremental PR, and that is why I have elected to not implement all that is needed at once. As it is, the PR lays the foundation for several of the changes, standardizes the treatment of request's response into a single static method, all without changing the current behavior.

mpeters (approved):

PR #1156, opened 2022-11-07
Move EK cert verification to cert_utils

sergio-correia:
We had partially addressed the issue we encountered when parsing some
certificates with python-cryptography by writing cert_utils module that
provided a single helper to parse the pubkey from a certificate.

However, we still were doing the EK validation in tpm_main using
python-cryptography, which means we would fall into the same parsing
problem again, since during the validation it would read all the certs
in the tpm cert store to check if it is signing the presented EK.

We address this issue by moving the EK cert verification to cert_utils:
we use the same approach as before, i.e. we parse the cert with pyasn1
when python-cryptography fails, and then use python-cryptography to do
the actual signature verification, as before.

By moving the method to cert_utils, it also becomes simpler to test it,
so in this commit we more tests to verify the methods work as expected.

Additionally, the updated EK verification is also capable of handling
ECDSA signatures

### DIFF
--- a/keylime/cert_utils.py
+++ b/keylime/cert_utils.py
@@ -1,4 +1,6 @@
 import io
+import os.path
+import subprocess
 import sys
 
 from cryptography import exceptions as crypto_exceptions
@@ -102,3 +104,40 @@ def verify_ek(ekcert, tpm_cert_store=config.get("tenant", "tpm_cert_store")):
 
     logger.error("No Root CA matched EK Certificate")
     return False
+
+
+def verify_ek_script(script, env, cwd):
+    if script is None:
+        logger.warning("External check script (%s) not specified", script)
+        return False
+
+    script_path = os.path.abspath(script)
+    if not os.path.isfile(script_path):
+        if cwd is None or not os.path.isfile(os.path.abspath(os.path.join(cwd, script))):
+            logger.warning("External check script (%s) not found; please make sure its path is correct", script)
+            return False
+        script_path = os.path.abspath(os.path.join(cwd, script))
+
+    try:
+        proc = subprocess.run(
+            [script_path],
+            env=env,
+            shell=False,
+            cwd=cwd,
+            stderr=subprocess.STDOUT,
+            stdout=subprocess.PIPE,
+            check=False,
+        )
+        if proc.returncode != 0:
+            errmsg = ""
+            if proc.stdout is not None:
+                errmsg = proc.stdout.decode("utf-8")
+            logger.error("External check script failed to validate EK: %s", errmsg)
+            return False
+        logger.debug("External check script successfully to validated EK")
+        if proc.stdout is not None:
+            logger.info("ek_check output: %s", proc.stdout.decode("utf-8"))
+    except subprocess.CalledProcessError as e:
+        logger.error("Error while trying to run external check script to validate EK: %s", e)
+        return False
+    return True
--- a/keylime/cert_utils.py
+++ b/keylime/cert_utils.py
@@ -1,13 +1,104 @@
-from cryptography.hazmat.primitives.serialization import load_der_public_key
-from pyasn1.codec.der import decoder, encoder
-from pyasn1_modules import rfc2459
+import io
+import sys
 
+from cryptography import exceptions as crypto_exceptions
+from cryptography import x509
+from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives.asymmetric import ec, padding
+from cryptography.hazmat.primitives.asymmetric.ec import EllipticCurvePublicKey
+from cryptography.hazmat.primitives.asymmetric.rsa import RSAPublicKey
+from pyasn1.codec.der import decoder, encoder
+from pyasn1_modules import pem, rfc2459
+
+from keylime import config, keylime_logging, tpm_ek_ca
 
 # Issue #944 -- python-cryptography won't parse malformed certs,
 # such as some Nuvoton ones we have encountered in the field.
 # Unfortunately, we still have to deal with such certs anyway.
-# Let's read the EK cert with pyasn1 instead of python-cryptography.
-def read_x509_der_cert_pubkey(der_cert_data):
-    """Returns the public key of a DER-encoded X.509 certificate"""
-    der509 = decoder.decode(der_cert_data, asn1Spec=rfc2459.Certificate())[0]
-    return load_der_public_key(encoder.encode(der509["tbsCertificate"]["subjectPublicKeyInfo"]))
+
+# Here we provide some helpers that use pyasn1 to parse the certificates
+# when parsing them with python-cryptography fails, and in this case, we
+# try to read the parsed certificate again into python-cryptograhy.
+
+logger = keylime_logging.init_logging("cert_utils")
+
+
+def x509_der_cert(der_cert_data: bytes):
+    """Load an x509 certificate provided in DER format
+    :param der_cert_data: the DER bytes of the certificate
+    :type der_cert_data: bytes
+    :returns: cryptography.x509.Certificate
+    """
+    try:
+        return x509.load_der_x509_certificate(data=der_cert_data, backend=default_backend())
+    except Exception as e:
+        logger.warning("Failed to parse DER data with python-cryptography: %s", e)
+        pyasn1_cert = decoder.decode(der_cert_data, asn1Spec=rfc2459.Certificate())[0]
+        return x509.load_der_x509_certificate(data=encoder.encode(pyasn1_cert), backend=default_backend())
+
+
+def x509_pem_cert(pem_cert_data: str):
+    """Load an x509 certificate provided in PEM format
+    :param pem_cert_data: the base-64 encoded PEM certificate
+    :type pem_cert_data: str
+    :returns: cryptography.x509.Certificate
+    """
+    try:
+        return x509.load_pem_x509_certificate(data=pem_cert_data.encode("utf-8"), backend=default_backend())
+    except Exception as e:
+        logger.warning("Failed to parse PEM data with python-cryptography: %s", e)
+        # Let's read the DER bytes from the base-64 PEM.
+        der_data = pem.readPemFromFile(io.StringIO(pem_cert_data))
+        # Now we can load it as we do in x509_der_cert().
+        pyasn1_cert = decoder.decode(der_data, asn1Spec=rfc2459.Certificate())[0]
+        return x509.load_der_x509_certificate(data=encoder.encode(pyasn1_cert), backend=default_backend())
+
+
+def verify_ek(ekcert, tpm_cert_store=config.get("tenant", "tpm_cert_store")):
+    """Verify that the provided EK certificate is signed by a trusted root
+    :param ekcert: The Endorsement Key certificate in DER format
+    :returns: True if the certificate can be verified, False otherwise
+    """
+    try:
+        trusted_certs = tpm_ek_ca.cert_loader(tpm_cert_store)
+    except Exception as e:
+        logger.warning("Error loading trusted certificates from the TPM cert store: %s", e)
+        return False
+
+    try:
+        ek509 = x509_der_cert(ekcert)
+        for cert_file, pem_cert in trusted_certs.items():
+            signcert = x509_pem_cert(pem_cert)
+            if ek509.issuer != signcert.subject:
+                continue
+
+            signcert_pubkey = signcert.public_key()
+            try:
+                if isinstance(signcert_pubkey, RSAPublicKey):
+                    signcert_pubkey.verify(
+                        ek509.signature,
+                        ek509.tbs_certificate_bytes,
+                        padding.PKCS1v15(),
+                        ek509.signature_hash_algorithm,
+                    )
+                elif isinstance(signcert_pubkey, EllipticCurvePublicKey):
+                    signcert_pubkey.verify(
+                        ek509.signature,
+                        ek509.tbs_certificate_bytes,
+                        ec.ECDSA(ek509.signature_hash_algorithm),
+                    )
+                else:
+                    logger.warning("Unsupported public key type: %s", type(signcert_pubkey))
+                    continue
+            except crypto_exceptions.InvalidSignature:
+                continue
+
+            logger.debug("EK cert matched cert: %s", cert_file)
+            return True
+    except Exception as e:
+        # Log the exception so we don't lose the raw message
+        logger.exception(e)
+        raise Exception("Error processing ek/ekcert. Does this TPM have a valid EK?").with_traceback(sys.exc_info()[2])
+
+    logger.error("No Root CA matched EK Certificate")
+    return False
--- a/keylime/tenant.py
+++ b/keylime/tenant.py
@@ -5,7 +5,6 @@ import io
 import json
 import logging
 import os
-import subprocess
 import sys
 import tempfile
 import time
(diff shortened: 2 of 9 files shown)


.github/workflows/test.yml
```diff
@@ -25,6 +25,8 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@v2
+    - name: Install build dependencies
+      run: sudo dnf -y install openssl
```
THS-on, line 29: `run_tests.sh` installs other build dependencies. It is probably better to add it there: https://github.com/keylime/keylime/blob/master/test/run_tests.sh

keylime/cert_utils.py
```diff
@@ -1,13 +1,133 @@
...
+
+logger = keylime_logging.init_logging("cert_utils")
+
+
+class UserError(Exception):
+    pass
+
+
+def x509_der_cert(der_cert_data):
+    if isinstance(der_cert_data, str):
```
THS-on, line 33: Can you add type hints that both bytes and str is accepted?

keylime/cert_utils.py
```diff
@@ -1,13 +1,133 @@
...
+        if cwd is None or not os.path.isfile(os.path.abspath(os.path.join(cwd, script))):
+            logger.warning("External check script (%s) not found; please make sure its path is correct", script)
+            return False
+        script_path = os.path.abspath(os.path.join(cwd, script))
+
+    try:
+        proc = subprocess.run(
+            [script_path],
+            env=env,
+            shell=True,
```
THS-on, line 115: This is dangerous in general and even if the old code had that we can it with this refactor.
I think the script is never called with any options, so we don't need that.

keylime/tenant.py
```diff
@@ -516,19 +515,11 @@ def validate_tpm_quote(self, public_key, quote, hash_alg):
...
-            logger.debug("External check script successfully to validated EK")
-            while True:
-                line = proc.stdout.readline().decode()
-                if line == "":
-                    break
-                logger.debug("ek_check output: %s", line.strip())
-        return True
+
+        # Define the TPM cert store for the external script.
+        env["TPM_CERT_STORE"] = config.get("tenant", "tpm_cert_store")
```
THS-on, line 520: This key than needs to be also added to the config (in the comment what variables are set).

keylime/cert_utils.py
```diff
@@ -1,13 +1,133 @@
...
+
+
+def x509_pem_cert(pem_cert_data):
+    if isinstance(pem_cert_data, (bytes, bytearray)):
+        pem_cert_data = pem_cert_data.decode("utf-8")
+    try:
+        cert = x509.load_pem_x509_certificate(data=pem_cert_data, backend=default_backend())
+        return cert
+    except Exception:
+        return x509_der_cert(pem.readPemFromFile(io.StringIO(pem_cert_data)))
```
aplanas, line 50: That confuses me, if is a call to read a PEM, why do we try a DER here? This make the operation non-symmetric (a call to read a PEM can try a DER, but no vice versa)

If symmetry is required, IMHO is a bit more clear if:
* The DER function try the pyasm1 decoder after the conversion to bytes, capturing the `pyasn1.error.SubstrateUnderrunError` exception to see if it valid or not, and calling the load_der_... only one
* Drop in the DER function the conversion to bytes (see later)
* In the PEM function only try the load_pem_ ... call
* Drop in the PEM function the conversion from bytes
* Create a new `x509_cert` that order the calls to PEM / DER in the order that you suit, and do the conversion from / to bytes
* Use only this function to load the certificates, and never call the PEM / DER versions outside this new function

scripts/ek-openssl-verify
```diff
@@ -0,0 +1,96 @@
+#!/bin/sh
```
THS-on, line 1: Can you add maybe also a reference to this script in the documentation? Thanks for adding an example!

keylime/cert_utils.py
```diff
@@ -1,13 +1,133 @@
...
-    return load_der_public_key(encoder.encode(der509["tbsCertificate"]["subjectPublicKeyInfo"]))
+
+# Here we provide some helpers that use pyasn1 to parse the certificates
+# when parsing them with python-cryptography fails, and in this case, we
+# try to read the parsed certificate again into python-cryptograhy.
+
+logger = keylime_logging.init_logging("cert_utils")
+
+
+class UserError(Exception):
```
THS-on, line 28: we have `UserError` defined multiple times in Keylime it is probably a good idea to move it into https://github.com/keylime/keylime/blob/master/keylime/common/exception.py

THS-on (changes requested):
This refactor LGTM, only a few comments.

Where it makes sense also add type hints, to make it easier to check this code with mypy in the future.

keylime/cert_utils.py
```diff
@@ -1,13 +1,133 @@
...
+                    )
+                else:
+                    logger.warning("Unsupported public key type: %s", type(signcert_pubkey))
+                    continue
+            except crypto_exceptions.InvalidSignature:
+                continue
+
+            logger.debug("EK cert matched cert: %s", cert_file)
+            return True
+    except Exception as e:
```
aplanas, line 90: This can hide problems in the certification load under a misleading error message.

The `try` should be moved early, surrounding only `tpm_ek_ca.cert_loader`, because as it is now it is also catching any exception in the PEM / DER loads, that does not indicate any issue with the TPM.

keylime/cert_utils.py
```diff
@@ -1,13 +1,90 @@
...
+# try to read the parsed certificate again into python-cryptograhy.
+
+logger = keylime_logging.init_logging("cert_utils")
+
+
+def x509_der_cert(der_cert_data):
+    if isinstance(der_cert_data, str):
+        der_cert_data = der_cert_data.encode("utf-8")
+    try:
+        cert = cert = x509.load_der_x509_certificate(data=der_cert_data, backend=default_backend())
```
stefanberger, line 36: typo ?

keylime/cert_utils.py
```diff
@@ -1,13 +1,90 @@
...
+        return cert
+    except Exception:
+        pyasn1_cert = decoder.decode(der_cert_data, asn1Spec=rfc2459.Certificate())[0]
+        return x509.load_der_x509_certificate(data=encoder.encode(pyasn1_cert), backend=default_backend())
+
+
+def x509_pem_cert(pem_cert_data):
+    if isinstance(pem_cert_data, (bytes, bytearray)):
+        pem_cert_data = pem_cert_data.decode("utf-8")
```
stefanberger, line 45: maybe it should be documented what types of data can be passed into this function. Ideally it would just be one datatype that 'pem_cert_data' could be...

keylime/cert_utils.py
```diff
@@ -1,13 +1,90 @@
...
+
+# Here we provide some helpers that use pyasn1 to parse the certificates
+# when parsing them with python-cryptography fails, and in this case, we
+# try to read the parsed certificate again into python-cryptograhy.
+
+logger = keylime_logging.init_logging("cert_utils")
+
+
+def x509_der_cert(der_cert_data):
+    if isinstance(der_cert_data, str):
```
stefanberger, line 33: Same comment about data type for der_cert_data here ...

keylime/cert_utils.py
```diff
@@ -1,13 +1,133 @@
...
+        if proc.returncode != 0:
+            errmsg = ""
+            if proc.stdout is not None:
+                errmsg = proc.stdout.decode("utf-8")
+            logger.info("External check script failed to validate EK: %s", errmsg)
+            return False
+        logger.debug("External check script successfully to validated EK")
+        if proc.stdout is not None:
+            logger.info("ek_check output: %s", proc.stdout.decode("utf-8"))
+    except Exception as e:
```
aplanas, line 130: there should be an specific exception to indicate an error with the subproces

keylime/cert_utils.py
```diff
@@ -88,3 +94,40 @@ def verify_ek(ekcert, tpm_cert_store=config.get("tenant", "tpm_cert_store")):
 
     logger.error("No Root CA matched EK Certificate")
     return False
+
+
+def verify_ek_script(script, env, cwd):
+    if script is None:
+        logger.warning("External check script (%s) not specified", script)
```
stefanberger, line 101: Not sure what the current direction is for keylime but consider the f-strings with {script} ?

aplanas: > consider the f-strings

That makes me wonder if we want this for logs or not. The logic behind `log.level("str", params)` (note that `%` is not used) is that interpolation only happens iff the call is made.  With f-strings we move out from lazy interpolation.

Not sure that is relevant, tho.

scripts/ek-openssl-verify
```diff
@@ -0,0 +1,96 @@
...
+
+# This script can be used as the `ek_check_script' (tenant configuration),
+# to attempt to verify a provided EK_CERT via env var using openssl.
+
+# EK       - contains a PEM encoded version of the public EK
+# EK_CERT  - contains a DER encoded EK certificate if one is available.
```
stefanberger, line 7: contains a base64 DER-encoded ...

keylime/cert_utils.py
sergio-correia: Done.

keylime/tenant.py
sergio-correia: Done.

keylime/cert_utils.py
sergio-correia: I will do this in a separate PR. I removed this definition from here, in any case.

keylime/cert_utils.py
sergio-correia: Right, fixed (removed the variable, as you suggested in another comment)

keylime/cert_utils.py
```diff
@@ -1,13 +1,133 @@
...
+
+class UserError(Exception):
+    pass
+
+
+def x509_der_cert(der_cert_data):
+    if isinstance(der_cert_data, str):
+        der_cert_data = der_cert_data.encode("utf-8")
+    try:
+        cert = cert = x509.load_der_x509_certificate(data=der_cert_data, backend=default_backend())
```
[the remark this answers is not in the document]
sergio-correia: Done.

keylime/cert_utils.py
sergio-correia: I changed to be in line with your comment. Let me know it if is better now, please.

keylime/cert_utils.py
sergio-correia: Done, please re-check.

scripts/ek-openssl-verify
```diff
@@ -0,0 +1,96 @@
...
+  [ -e "${c}" ] || continue
+  log "Checking if ${c} is the issuer of EK cert..."
+  if openssl verify -partial_chain -CAfile "${c}" "${EK_CERT_FILE_PEM}" \
+                                           >>"${LOG}" 2>>"${LOG}"; then
+    log "${EK_CERT} successfully verified by $(basename "${c}")" _
+    exit 0
+  fi
+done
+
+die "EK signature did not match cerficates from TPM cert store"
```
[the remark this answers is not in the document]
sergio-correia: Fixed, thanks.

keylime/cert_utils.py
```diff
@@ -1,13 +1,90 @@
...
+        pyasn1_cert = decoder.decode(der_cert_data, asn1Spec=rfc2459.Certificate())[0]
+        return x509.load_der_x509_certificate(data=encoder.encode(pyasn1_cert), backend=default_backend())
+
+
+def x509_pem_cert(pem_cert_data):
+    if isinstance(pem_cert_data, (bytes, bytearray)):
+        pem_cert_data = pem_cert_data.decode("utf-8")
+    try:
+        cert = x509.load_pem_x509_certificate(data=pem_cert_data, backend=default_backend())
+        return cert
```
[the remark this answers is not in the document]
sergio-correia: Done.

scripts/ek-openssl-verify
sergio-correia: Fixed.

keylime/cert_utils.py
sergio-correia: I added the type hints, but for now I changed to be in line with comments from Stefan, so `x509_der_cert()` current accepts `bytes` and `x509_pem_cert()` accepts `str`. Please let me know your thoughts.

keylime/cert_utils.py
sergio-correia: Yeah, I tend to agree with this. I changed `x509_pem_cert()` to accept `str` and `x509_der_cert()` to accept `bytes`, for now.

keylime/cert_utils.py
sergio-correia: PEM here is the DER data base64 encoded, with the added header and footer (e.g. `----BEGIN CERT...`), so once we call `pem.readPemFromFile()` we have the actual bytes, which we can parse with the method that parses the DER data. I changed it slightly and added a comment, to try to clarify it a bit, but I am not sure I addressed your concern fully, so if you could recheck I would appreciate.

keylime/cert_utils.py
sergio-correia: Fixed.

mpeters:
This looks good to me but I'll wait until @aplanas @stefanberger and @THS-on can re-comment after you most recent changes. Would be good to get this fixed for 6.5.2

THS-on (approved):

keylime/cert_utils.py
```diff
@@ -1,13 +1,143 @@
...
+
+# Here we provide some helpers that use pyasn1 to parse the certificates
+# when parsing them with python-cryptography fails, and in this case, we
+# try to read the parsed certificate again into python-cryptograhy.
+
+logger = keylime_logging.init_logging("cert_utils")
+
+
+def x509_der_cert(der_cert_data: bytes):
+    """Load a x509 certificate provided in DER format
```
stefanberger, line 29: a -> an

stefanberger:
LGTM

keylime/cert_utils.py
```diff
@@ -1,13 +1,143 @@
...
+    try:
+        return x509.load_der_x509_certificate(data=der_cert_data, backend=default_backend())
+    except Exception as e:
+        logger.warning("Failed to parse DER data with python-cryptography: %s", e)
+        pyasn1_cert = decoder.decode(der_cert_data, asn1Spec=rfc2459.Certificate())[0]
+        return x509.load_der_x509_certificate(data=encoder.encode(pyasn1_cert), backend=default_backend())
+
+
+def x509_pem_cert(pem_cert_data: str):
+    """Load a x509 certificate provided in PEM format
```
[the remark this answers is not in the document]
sergio-correia: Fixed.

keylime/cert_utils.py
sergio-correia: Fixed.

aplanas (approved):

THS-on:
@kkaarreell @ansasaki the measured boot unit test fails on centos 9. Any idea why?

THS-on:
> It fails in Centos 8 stream. From what I saw here quickly, the tpm2-tools version there (4.1.1) does not have tpm2_eventlog, so the test fails.

Yeah you are right, so the test can just add a skip if the the TPM version is below 4.1.1

kkaarreell:
Correct, in E2E tests we are disabling measured boot tests for C8S.

ansasaki:
I'll create a PR to skip the measured boot log parsing for old tpm2-tools versions

THS-on:
@ansasaki thank you. @sergio-correia I'll merge this PR now, because the failure has nothing to do with this PR.

Issue #1106, opened 2022-09-19
Improved error message for when IMA emulator is not started with correct hash alg flag


lkatalin:
# Environment
 * OS / version: Fedora 36 (VM)
 * Processor architecture: x86_64
 * TPM Manufacturer: swtpm
 * Keylime version: latest HEAD

# Description 
I set up swtpm like so:
```
modprobe tpm_vtpm_proxy
mkdir /var/vtpm
swtpm_setup --tpm2 \
          --tpmstate /var/vtpm \
          --createek --decryption --create-ek-cert \
          --create-platform-cert \
          --display \
swtpm chardev --tpm2 --vtpm-proxy --tpmstate dir=/var/vtpm -d

export TPM2TOOLS_TCTI="device:/dev/tpmrm0"
export TCTI="device:/dev/tpmrm0"
```

Then when starting the IMA emulator, it gives this error:
```
[root@fedora f36-keylime]# keylime_ima_emulator
Reading configuration from ['/etc/keylime/logging.conf']
2022-09-19 10:58:39.879 - keylime.tpm - INFO - TPM2-TOOLS Version: 5.2
Traceback (most recent call last):
  File "/usr/local/bin/keylime_ima_emulator", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.10/site-packages/keylime/cmd/ima_emulator_adapter.py", line 68, in main
    pcr_val = tpm_instance.readPCR(config.IMA_PCR, pcr_hash_alg)
  File "/usr/local/lib/python3.10/site-packages/keylime/tpm/tpm_main.py", line 1355, in readPCR
    return f"{jsonout[hash_alg][pcrval]:0{alg_size}x}"
TypeError: 'NoneType' object is not subscriptable
```

The Python version is 3.10.6.
This does not happen on the F34 VM, which runs Python 3.9.13

[This line](https://github.com/keylime/keylime/blob/master/keylime/tpm/tpm_main.py#L1355) is the culprit.

lkatalin:
@THS-on `keylime_ima_emulator -a sha256` actually fixes the problem! Thanks. `tpm2_pcrread` has data under `sha256` but not `sha1`. I did not realize I need to specify a parameter to the emulator.

lkatalin:
Actually I would like to reopen this as I think an improved error message would help here.

PR #1163, opened 2022-11-08
Fix issue #1106 by handling reading a PCR from an unsupported PCR bank

stefanberger:
This PR fixes issue #1106 by handling the failure when trying to read a PCR from an unsupported PCR bank

### DIFF
--- a/keylime/tpm/tpm_main.py
+++ b/keylime/tpm/tpm_main.py
@@ -1358,7 +1358,7 @@ class tpm(tpm_abstract.AbstractTPM):
         if jsonout is None:
             raise Exception("Could not read YAML output of tpm2_pcrread.")
 
-        if hash_alg not in jsonout:
+        if not jsonout.get(hash_alg):
             raise Exception(f"Invalid hashing algorithm '{hash_alg}' for reading PCR number {pcrval}.")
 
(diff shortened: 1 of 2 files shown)


stefanberger:
So now the printout is:
```
# keylime_ima_emulator
2022-11-08 16:59:02.328 - keylime.tpm - INFO - TPM2-TOOLS Version: 5.2
Error: Invalid hashing algorithm 'sha1' for reading PCR number 10.
```

keylime/tpm/tpm_main.py
```diff
@@ -1358,7 +1358,7 @@ def readPCR(self, pcrval, hash_alg=None):
         if jsonout is None:
             raise Exception("Could not read YAML output of tpm2_pcrread.")
 
-        if hash_alg not in jsonout:
+        if not jsonout.get(hash_alg):
```
aplanas, line 1361: I see, this is a limitation of the conversion, that keep stuff like `{"sha1": None, ...}`

aplanas (approved):

keylime/tpm/tpm_main.py
stefanberger: Right. And if you ask whether "sha1" is in the map, then it will say "yes".

PR #1165, opened 2022-11-09
ima: Accept x509 certificates if no Subject Key Identifier is available

stefanberger:
Also accept x509 certificates if they do not have a Subject Key Identifier (SKID). Previously only certificates with SKID were accepted because the SKID is required for the key to be used by the kernel keyrings and keylime would only accept a key that could possibly also have served for IMA signature verification in the kernel.

Return None in case no keyidv2 could be determined either due to missing extensions or missing SKID and warn the use about it.

Signed-off-by Stefan Berger <stefanb@linux.ibm.com>

### DIFF
--- a/keylime/ima/file_signatures.py
+++ b/keylime/ima/file_signatures.py
@@ -11,6 +11,7 @@ from cryptography.hazmat.primitives.asymmetric import ec, padding
 from cryptography.hazmat.primitives.asymmetric.ec import EllipticCurvePublicKey
 from cryptography.hazmat.primitives.asymmetric.rsa import RSAPublicKey
 from cryptography.hazmat.primitives.asymmetric.utils import Prehashed
+from cryptography.x509.extensions import ExtensionNotFound
 
 from keylime import keylime_logging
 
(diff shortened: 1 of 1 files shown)


mpeters (approved):

mpeters (changes requested):
Looks good, but needs to fix linting failures and the commit needs a `--signoff` in git.

stefanberger:
> Looks good, but needs to fix linting failures and the commit needs a `--signoff` in git.

Will push once tests are through.

keylime/ima/file_signatures.py
```diff
@@ -392,15 +393,18 @@ def _get_pubkey_from_pem_private_key(filedata, backend):
...
+    """Get the keyidv2 from the cert's Subject Key Identifier (SKID) if available."""
+    if cert.extensions:
+        try:
+            skid = cert.extensions.get_extension_for_oid(x509.oid.ExtensionOID.SUBJECT_KEY_IDENTIFIER)
+            if skid and skid.value and len(skid.value.digest) >= 4:
+                keyidv2 = int.from_bytes(skid.value.digest[-4:], "big")
+                logger.debug("Extracted keyidv2 from cert: 0x%08x", keyidv2)
+                return keyidv2
+        except ExtensionNotFound as ex:
+            pass
```
aplanas, line 405: LGTM but now that `ex` is not referenced, do you want to log the exception?

aplanas (approved):

sergio-correia (approved):

mpeters (approved):

PR #1166, opened 2022-11-09
Test binary measured boot event log parsing

ansasaki:
Add a test to `test_tpm.py` to test measured boot event log parsing.

Signed-off-by: Anderson Toshiyuki Sasaki <ansasaki@redhat.com>


test/test_tpm.py
```diff
@@ -91,6 +93,22 @@ def test_get_tpm_manufacturer_challenges(self):
             response = self.tpm.get_tpm_manufacturer(output=test["challenge"])
             self.assertEqual(test["response"], response)
 
+    # test parsing binary measured boot event log
```
aplanas, line 96: If you put that as a doc string, unittest will reference it if the test fails

ansasaki: Right, will use doc strings for the descriptions

aplanas (approved):

mpeters (approved):

PR #1168, opened 2022-11-09
installer: Enable installation on RHEL-9

stefanberger:
RHEL-9 has the tpm2-tools package available and therefore does not need to build it.

### DIFF
--- a/installer.sh
+++ b/installer.sh
@@ -91,19 +91,32 @@ case "$ID" in
                 NEED_BUILD_TOOLS=1
                 CENTOS7_TSS_FLAGS="--enable-esapi=no --disable-doxygen-doc"
             ;;
-            8*)
+            8*|9*)
                 echo "${ID} ${VERSION_ID} selected."
                 PACKAGE_MGR=$(command -v dnf)
                 NEED_EPEL=1
                 PYTHON_PREIN="python3 python3-devel python3-setuptools python3-pip"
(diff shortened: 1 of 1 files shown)


installer.sh
```diff
@@ -91,19 +91,32 @@ case "$ID" in
...
-                NEED_BUILD_TOOLS=1
-                NEED_PYTHON_DIR=1
+                case "${VERSION_ID}" in
+                8*)
+                    NEED_BUILD_TOOLS=1
+                    NEED_PYTHON_DIR=1
+                ;;
+                9*)
+                    NEED_BUILD_TOOLS=0
+                    NEED_PYTHON_DIR=0
```
aplanas, line 116: `NEED_PYTHON_DIR=1`? This will create `/usr/local/lib/python3.6/site-packages/` later, that seems unrelated with `tpm2-tools`

stefanberger: `NEED_PYTHON_DIR=1` was there before for RHEL 8.x, so I am not changing this fro RHEL 8.x.

aplanas: Is it OK to change it on 9?

THS-on: I think it is fine to change it, because I `NEED_PYTHON_DIR` was specific to Python 3.6 and RHEL 9 ships with 3.9

THS-on (approved):

installer.sh
aplanas: ACK!

aplanas (approved):

PR #1169, opened 2022-11-10
Check tpm2-tools before running create_mb_refstate

ansasaki:
Check if the tpm2-tools version is recent enough to have the tpm_eventlog tool available before trying to run the create_mb_refstate script.

Also skip the test for measured boot log parsing test in test_tpm.py if the tpm2-tools version is too old.

### DIFF
--- a/scripts/create_mb_refstate
+++ b/scripts/create_mb_refstate
@@ -12,6 +12,8 @@ import json
 import re
 import sys
 
+from packaging.version import Version
+
 from keylime.tpm import tpm_main
 
 
@@ -124,6 +126,12 @@ def get_mok(events):
 
 
 def main():
+    tpm = tpm_main.tpm()
+
+    if Version(tpm.tools_version) < Version("4.2"):
+        print(f"tpm2_eventlog is not available: tpm2-tools version {tpm.tools_version} < 4.2")
+        sys.exit(1)
+
     parser = argparse.ArgumentParser()
     parser.add_argument(
         "eventlog_file",
@@ -136,7 +144,6 @@ def main():
     )
     args = parser.parse_args()
     log_bin = args.eventlog_file.read()
-    tpm = tpm_main.tpm()
 
     failure, log_data = tpm.parse_binary_bootlog(log_bin)
     if failure or not log_data:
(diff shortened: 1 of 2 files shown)


scripts/create_mb_refstate
```diff
@@ -124,6 +126,12 @@ def get_mok(events):
 
 
 def main():
+    tpm = tpm_main.tpm()
+
+    if Version(tpm.tools_version) < Version("4.2"):
+        print(f"tpm2_eventlog is not available (min tpm2-tools version is 4.2)")
```
aplanas, line 132: it is not an f-string

scripts/create_mb_refstate
```diff
@@ -124,6 +126,12 @@ def get_mok(events):
...
+        return
```
aplanas, line 133: Should it not signalize a retcode != 0? (`sys.exit(1)`, for example)

scripts/create_mb_refstate
```diff
@@ -136,9 +144,17 @@ def main():
...
     args = parser.parse_args()
     log_bin = args.eventlog_file.read()
-    tpm = tpm_main.tpm()
-    _, log_data = tpm.parse_binary_bootlog(log_bin)
-    events = log_data["events"]
+
+    failure, log_data = tpm.parse_binary_bootlog(log_bin)
+    if failure or not log_data:
+        print(f"Parsing of binary boot measurements failed with: {list(map(lambda x: x.context, failure.events))}")
+        return
```
aplanas, line 151: same

scripts/create_mb_refstate
```diff
@@ -136,9 +144,17 @@ def main():
...
+
+    events = log_data.get("events")
+    if not events:
+        print("No events found on binary boot measurements log")
+        return
```
aplanas, line 156: here too?

scripts/create_mb_refstate
ansasaki: Yes, you are right, I'll fix these

THS-on (approved):

aplanas (approved):

Issue #1167, opened 2022-11-09
IMA scalability is currently limited


maugustosilva:
# Environment
 * OS / version: Ubuntu 22.04, but not relevant
 * Processor architecture: amd64, irrelevant
 * TPM Manufacturer: virtual, irrelevant
 * Keylime version: 6.5.1+
 * Important: database connection is MySQL, with large (100+) number of agents; IMA is turned on, and using "version 1" allowlists.

# Description

While running scaling experiments in our virtual cluster, we have detected multiple serious problems with the current IMA implementation. We are planning to describe the general problem here, then issue a series of PRs to patch the problems we *can* patch.

We have a production scenario with the following boundary conditions:
1) Our nodes have "real-world production-ready" filesystems, with a large number of executables/libraries (tens of thousands)
2) We are forced to use a "version 1" IMA allowlist (policy), with a large number (again, tens of thousands) <checksum>,<path> pairs. This currently results in a roughly 20 MiB file size
3) Given we are including the `boot_aggregate` on the allowlist, we have a scenario where each individual node will have its own allowlist.

In a nutshell, the center of the problem is that the IMA implementation of keylime makes no allowances handling allowlists on this scenario.

Specifically:

* On the `verifier` each process in charge of `agent` verification has to hit the database twice for each verification: once to get the `agent` data from `verifiermain` table, a second time to get the allowlist contents from `allowlists` table. For *every individual attestation* (which by the default happens once each 2 seconds), this process is repeated.
* In addition to that, API calls to ascertain the status of `agent` currently do pull the whole allowlist contents from the `allowlists` table, applies a `len()` call to it and then populate a single attribute called `allowlist_len`

Our scaling test runs 1500 agents running copies of Ubuntu 22.04, running "identity + mesaured boot" attestation (one quote every 2 seconds for every `agent`). This works perfectly fine, requiring 9 `verifiers` each handling ~ 180 agents.

When IMA is turned on, using the ~ 20 MiB "version 1" allowlist previously discussed, attesting each `agent` requires 10 MiBps of exclusive bandwidth from the `verifier` doing the attestation to the database server (MySQL in our case). In theory this would exhaust our 100Gbps network connection after 1000 deployed agents. However, we don't even come close to that. Our monitoring system issues API calls via the CLI `keylime_tenant -c bulkinfo` to assess the status of the system every so often, and after about 100 tenants have been activated, `bulkinfo` is no longer operational: the download time of the massed allowlists exceeds the REST timeout.

[...]

PR #1172, opened 2022-11-10
cloud_verifier: This is the first PR to address the scalability problems uncovered in #1167, starting with `agent` status.

maugustosilva:
While requesting `agent` status, either individually or in bulk (`bulkinfo`), there is simply no need for loading anything from the `allowlists` table on the database, especialmente if the leght of the allowlist for a given agent is non-trivial (e.g., tens of megabytes).

In addition to that, two attributes for the `agent` - `allowlist_len` and `mb_refstate_len` - in addition of being of dubious utility, were being calculated in an incorrect manner. These were dutifully removed.

A meaningful attribute for mesured boot reference states - `has_mb_refstate` is already implemented on this PR.

In a subsquent PR we will find a way to introduce meaningful attribute for IMA allowlists (i.e., `has_ima_allowlist`, or something to that effect), **without** requiring a full load from the database table.

### DIFF
--- a/keylime/cloud_verifier_common.py
+++ b/keylime/cloud_verifier_common.py
@@ -237,27 +237,20 @@ def prepare_get_quote(agent):
 
 def process_get_status(agent):
     agent_id = agent.agent_id
-    allowlist_json = json.loads(agent.ima_policy.ima_policy)
-    if isinstance(allowlist_json, dict) and "allowlist" in allowlist_json:
-        al_len = len(allowlist_json["allowlist"])
-    else:
-        al_len = 0
 
+    has_mb_refstate = 0
     try:
         mb_refstate = json.loads(agent.mb_refstate)
+        if mb_refstate and mb_refstate.keys():
+            has_mb_refstate = 1
     except Exception as e:
         logger.warning(
-            'Non-fatal problem ocurred while attempting to evaluate agent %s attribute "mb_refstate" (%s). Will just consider the value of this attribute to be "None"',
+            'Non-fatal problem ocurred while attempting to evaluate agent %s attribute "mb_refstate" (%s). Will just consider the value of this attribute as empty',
             agent_id,
             e.args,
         )
-        mb_refstate = None
         logger.debug('The contents of the agent %s attribute "mb_refstate" are %s', agent_id, agent.mb_refstate)
 
-    if isinstance(mb_refstate, dict) and "mb_refstate" in mb_refstate:
-        mb_refstate_len = len(mb_refstate["mb_refstate"])
-    else:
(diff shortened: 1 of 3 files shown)


keylime/cloud_verifier_tornado.py
```diff
@@ -278,7 +278,7 @@ def get(self):
             try:
                 agent = (
                     session.query(VerfierMain)
-                    .options(joinedload(VerfierMain.ima_policy))
+                    .options(noload(VerfierMain.ima_policy))
```
THS-on, line 281: This makes it so that the attribute is always None, right?

In this case it should be fine, but maybe add an comment either here or in the commit messages why noload is used.

keylime/cloud_verifier_tornado.py
```diff
@@ -278,7 +278,7 @@ def get(self):
             try:
                 agent = (
                     session.query(VerfierMain)
-                    .options(joinedload(VerfierMain.ima_policy))
```
mpeters, line 281: If we aren't using any of the fields in the `ima_policy` joined table then why do a join at all?

THS-on:
Is it possible to change the first line of the commit so that it is clear what parts of keylime are affected and what was changed?
The why and how then can be included in the rest of the commit message. Similar to this: https://github.com/keylime/keylime/commit/4a7242d85f11fc4f392cdb1712fcdb6399e389e5

We do have a guideline [here](https://github.com/keylime/keylime/blob/master/CONTRIBUTING.md#commit-message-guidelines), but never really enforced it.

keylime/cloud_verifier_common.py
```diff
@@ -237,36 +237,28 @@ def prepare_get_quote(agent):
...
     response = {
         "operational_state": agent.operational_state,
         "v": agent.v,
         "ip": agent.ip,
         "port": agent.port,
         "tpm_policy": agent.tpm_policy,
         "meta_data": agent.meta_data,
-        "allowlist_len": al_len,
-        "mb_refstate_len": mb_refstate_len,
+        "has_mb_refstate": has_mb_refstate,
```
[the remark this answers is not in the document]
maugustosilva: good catch, I totally forgot to update it, done

keylime/cloud_verifier_tornado.py
maugustosilva: Added a comment explaining why it was **temporarily** set as `noload`

keylime/cloud_verifier_tornado.py
maugustosilva: Right we are no longer using `joinedload` there

mpeters (approved):

PR #1173, opened 2022-11-10
chore: non-obvious type fixes not concerning IMA

mdrocco:
This PR provides a batch of type fixes that are:
- non-trivial from the perspective of code behavior
- not concerning IMA (I will create a separate PR for those)

In my local testing setup, they run clean all tests including regression.

PR #1176, opened 2022-11-11
tpm_main: fix ek creation for tpm2-tools versions > 4.2

THS-on:
I missed that statement when I added the check for tpm2-tools version 5.4.

### DIFF
--- a/keylime/tpm/tpm_main.py
+++ b/keylime/tpm/tpm_main.py
@@ -289,7 +289,8 @@ class tpm(tpm_abstract.AbstractTPM):
                     "-P",
                     owner_pw,
                 ]
-            elif self.tools_version == "4.2":
+            else:
+                # version 4.2 or later
                 command = [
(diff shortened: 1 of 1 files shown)


keylime/tpm/tpm_main.py
```diff
@@ -289,7 +289,8 @@ def __create_ek(self, asym_alg=None):
                     "-P",
                     owner_pw,
                 ]
-            elif self.tools_version == "4.2":
+            else:
+                # version 4.2 or later
```
aplanas, line 293: Conflict with #1173

aplanas (approved):

sergio-correia (approved):

keylime/tpm/tpm_main.py
THS-on: Yes, the fix on the other branch would cause the agent to crash when tpm2-tools version is > 4.2

PR #1177, opened 2022-11-11
allowlist: Add an AllowlistDBProxy acting as a proxy and cache for the allowlis…

stefanberger:
…t table

Implement an AllowlistDBProxy that acts as a proxy and cache to the
allowlists DB table by holding the VerifierAllowlist DB entries in
a map whose keys are the IDs of the allowlists. Whenever one needs a
VerfierAllowlist DB entry one queries a AllowlistDBProxy singleton for an
entry with that allowlist ID. If the entry is not currently cached in
the map it will be loaded from the DB and otherwise every few seconds the
currently cached entry will be checked for an update and the latest version
of the VerfierAllowlist object will be returned. Every once-in-while
(60 seconds) all entries that have expired will be purged in order to
clear up memory of either deleted or unreferenced VerifierAllowlist DB
entries.

Resolves: https://github.com/keylime/keylime/issues/1167

### DIFF
--- a/keylime/allowlistdbproxy.py
+++ b/keylime/allowlistdbproxy.py
@@ -1,11 +1,11 @@
 import time
 from datetime import datetime
-from threading import Event, Lock
 from typing import Dict, Optional
 
 from sqlalchemy.exc import SQLAlchemyError
 from sqlalchemy.orm import Session
 from sqlalchemy.orm.exc import NoResultFound
+from tornado.locks import Event, Lock
 
 from keylime import keylime_logging
 from keylime.db.verifier_db import VerifierAllowlist
@@ -88,10 +88,10 @@ class AllowlistEntry:
 
         return verifier_allowlist_list[0].modified != self.allowlist_modified  # type: ignore
 
-    def check_update(self) -> bool:
+    async def check_update(self) -> bool:
         """Check for an update to the AllowlistEntry in the DB"""
 
-        with self.update_lock:
+        async with self.update_lock:
             now = time.monotonic()
             # We have to be really careful with this update_timeout. If attestation fails
             # and an update of the allowlist comes very quickly this update_timeout may
@@ -163,29 +163,29 @@ class AllowlistDBProxy:
         self.loader_map_lock = Lock()
         self.loader_map = {}
 
-    def __purge(self) -> None:
+    async def __purge(self) -> None:
         """If purge timeout occurred remove all expired entries"""
 
-        with self.purge_lock:
+        async with self.purge_lock:
             now = time.monotonic()
             if self.last_purge + self.purge_timeout < now:
                 self.last_purge = now
             else:
                 return
 
-        with self.map_lock:
+        async with self.map_lock:
             for allowlist_id in list(self.map.keys()):
                 allowlist_entry = self.map[allowlist_id]
                 if allowlist_entry.is_expired(self.purge_timeout):
                     del self.map[allowlist_id]
 
-    def __load_allowlist(self, allowlist_id: int) -> Optional[AllowlistEntry]:
+    async def __load_allowlist(self, allowlist_id: int) -> Optional[AllowlistEntry]:
         """Load an allowlist given its ID and put it into the map and return
         the loaded entry. Ensure that only one thread loads a specific entry
         and all other ones wanting to load the same entry wait for the first
         thread to have it loaded."""
 
-        with self.loader_map_lock:
+        async with self.loader_map_lock:
             wait_event = self.loader_map.get(allowlist_id)
             if not wait_event:
                 # First thread to load the entry
@@ -197,15 +197,15 @@ class AllowlistDBProxy:
         if wait_event:
             # Wait for other thread to notify us of loaded object
             wait_event.wait()
-            with self.map_lock:
+            async with self.map_lock:
                 return self.map.get(allowlist_id)
 
         allowlist_entry = AllowlistEntry.from_db(self.session, allowlist_id)
         if allowlist_entry:
-            with self.map_lock:
+            async with self.map_lock:
                 self.map[allowlist_id] = allowlist_entry
 
-        with self.loader_map_lock:
+        async with self.loader_map_lock:
             del self.loader_map[allowlist_id]
             # Notify all waiters
             assert notify_event  # pyright
@@ -213,26 +213,26 @@ class AllowlistDBProxy:
 
         return allowlist_entry
 
-    def __get_allowlist(self, allowlist_id: int) -> Optional[AllowlistEntry]:
+    async def __get_allowlist(self, allowlist_id: int) -> Optional[AllowlistEntry]:
         """Get an allowlist given its ID"""
 
-        with self.map_lock:
+        async with self.map_lock:
             allowlist_entry = self.map.get(allowlist_id)
 
         if allowlist_entry:
-            if not allowlist_entry.check_update():
+            if not await allowlist_entry.check_update():
                 # allowlist_entry likely disappeared from DB
                 del self.map[allowlist_id]
                 return None
             return allowlist_entry
 
-        return self.__load_allowlist(allowlist_id)
+        return await self.__load_allowlist(allowlist_id)
 
-    def get_allowlist(self, allowlist_id: int) -> Optional[AllowlistEntry]:
+    async def get_allowlist(self, allowlist_id: int) -> Optional[AllowlistEntry]:
         """Get an allowlist given its ID; cleanup stale entries"""
 
-        allowlist_entry = self.__get_allowlist(allowlist_id)
+        allowlist_entry = await self.__get_allowlist(allowlist_id)
 
-        self.__purge()
+        await self.__purge()
 
         return allowlist_entry
--- a/keylime/allowlistdbproxy.py
+++ b/keylime/allowlistdbproxy.py
@@ -1,0 +1,238 @@
+import time
+from datetime import datetime
+from threading import Event, Lock
+from typing import Dict, Optional
+
+from sqlalchemy.exc import SQLAlchemyError
+from sqlalchemy.orm import Session
+from sqlalchemy.orm.exc import NoResultFound
+
+from keylime import keylime_logging
+from keylime.db.verifier_db import VerifierAllowlist
+
+logger = keylime_logging.init_logging("allowlistdbproxy")
+
+
+class AllowlistEntry:
+    """AllowlistEntry represents a named allowlist in the DB"""
+
+    session: Session
+
+    update_lock: Lock
+    last_update_check: float
+    update_timeout: float
+
+    allowlist_modified: datetime
+    allowlist_id: int
+    verifier_allowlist_db: VerifierAllowlist
+    verifier_allowlist: VerifierAllowlist
+
+    @staticmethod
+    def from_db(session: Session, allowlist_id: int) -> Optional["AllowlistEntry"]:
+        """Create an AllowlistEntry object from an entry in the DB"""
+        try:
+            verifier_allowlist_db = session.query(VerifierAllowlist).filter_by(id=allowlist_id).one()
+        except NoResultFound:
+            return None
+        except SQLAlchemyError as e:
+            logger.error("SQLAlchemy Error: %s", e)
+            return None
+        return AllowlistEntry(session, allowlist_id, verifier_allowlist_db.modified, verifier_allowlist_db)
+
+    def __init__(
+        self,
+        session: Session,
+        allowlist_id: int,
+        modified: datetime,
+        verifier_allowlist_db: VerifierAllowlist,
+        update_timeout: int = 0,
+    ) -> None:
+        """constructor"""
+        self.session = session
+
+        self.update_lock = Lock()
+        self.last_update_check = time.monotonic()
+        self.update_timeout = update_timeout
+
+        self.allowlist_modified = modified
+        self.allowlist_id = allowlist_id
+        self.verifier_allowlist_db = verifier_allowlist_db
+        self.verifier_allowlist = AllowlistEntry.__copy_verifier_allowlist(self.verifier_allowlist_db)
+
+    @staticmethod
+    def __copy_verifier_allowlist(tocopy: VerifierAllowlist) -> VerifierAllowlist:
+        """Make a copy of the verifier_allowlist_db"""
+        # This is the object we will give to the user; it's basically a read-only object without
+        # access to the DB and should prevent that the user of the object sees changing fields
+        # while we refresh() the object from the DB
+        verifier_allowlist = VerifierAllowlist()
+        verifier_allowlist.id = tocopy.id
+        verifier_allowlist.name = tocopy.name
+        verifier_allowlist.ima_policy = tocopy.ima_policy
+        verifier_allowlist.tpm_policy = tocopy.tpm_policy
+        verifier_allowlist.modified = tocopy.modified
+
+        return verifier_allowlist
+
+    def __was_modified(self) -> bool:
+        """Check whether the modified timestamp of the allowlist in the DB has changed"""
+
+        # Use raw SQL since formulating an SQLAlchemy statement may not work on older versions
+        # This returns a ResultProxy on older versions, CursorResult on newer ones, the
+        # fetchall() makes this a list of Row objects
+        verifier_allowlist_list = self.session.execute(
+            f"SELECT modified FROM allowlists WHERE id=={self.allowlist_id}"
+        ).fetchall()
+        if len(verifier_allowlist_list) == 0:
+            raise NoResultFound()
+
+        return verifier_allowlist_list[0].modified != self.allowlist_modified  # type: ignore
+
+    def check_update(self) -> bool:
+        """Check for an update to the AllowlistEntry in the DB"""
+
+        with self.update_lock:
+            now = time.monotonic()
+            # We have to be really careful with this update_timeout. If attestation fails
+            # and an update of the allowlist comes very quickly this update_timeout may
+            # cause the new policy not to be loaded and cause an unexpected attestation
+            # failure... it may be better to always poll the DB.
+            if self.last_update_check + self.update_timeout < now:
+                try:
+                    if self.__was_modified():
+                        self.session.refresh(self.verifier_allowlist_db)
+                        self.allowlist_modified = self.verifier_allowlist_db.modified  # type: ignore
+
+                        self.verifier_allowlist = AllowlistEntry.__copy_verifier_allowlist(self.verifier_allowlist_db)
+                except NoResultFound:
+                    return False
+                except SQLAlchemyError as e:
+                    logger.error("SQLAlchemy Error: %s", e)
+                    return False
+
+                self.last_update_check = now
+
+            return True
+
+    def is_expired(self, timeout: float) -> bool:
+        """Check whether the object was not updated for 'timeout' seconds"""
+        return self.last_update_check + timeout < time.monotonic()
+
+
+class AllowlistDBProxy:
+    """AllowlistDBProxy is used as a layer in front of the allowlist in the DB"""
+
+    instance = None
+
+    session: Session
+
+    map_lock: Lock
+    map: Dict[int, AllowlistEntry]
+
+    purge_lock: Lock
+    last_purge: float
+    purge_timeout: float
+
+    loader_map_lock: Lock
+    loader_map: Dict[int, Event]
+
+    @staticmethod
+    def create_instance(session: Session) -> None:
+        """Create the AllowlistDBProxy singleton"""
+        if not AllowlistDBProxy.instance:
+            AllowlistDBProxy.instance = AllowlistDBProxy(session)
+
+    @staticmethod
+    def get_instance() -> "AllowlistDBProxy":
+        """Return the singleton AllowlistDBProxy"""
+        assert AllowlistDBProxy.instance
+        return AllowlistDBProxy.instance
+
+    def __init__(self, session: Session, purge_timeout: int = 60) -> None:
+        """constructor"""
+
+        self.session = session
+
+        self.map_lock = Lock()
+        self.map = {}
+
+        self.purge_lock = Lock()
+        self.last_purge = 0
+        self.purge_timeout = purge_timeout
+
+        self.loader_map_lock = Lock()
+        self.loader_map = {}
+
+    def __purge(self) -> None:
+        """If purge timeout occurred remove all expired entries"""
+
+        with self.purge_lock:
+            now = time.monotonic()
+            if self.last_purge + self.purge_timeout < now:
+                self.last_purge = now
+            else:
+                return
+
+        with self.map_lock:
+            for allowlist_id in list(self.map.keys()):
+                allowlist_entry = self.map[allowlist_id]
+                if allowlist_entry.is_expired(self.purge_timeout):
+                    del self.map[allowlist_id]
+
+    def __load_allowlist(self, allowlist_id: int) -> Optional[AllowlistEntry]:
+        """Load an allowlist given its ID and put it into the map and return
+        the loaded entry. Ensure that only one thread loads a specific entry
+        and all other ones wanting to load the same entry wait for the first
+        thread to have it loaded."""
+
+        with self.loader_map_lock:
+            wait_event = self.loader_map.get(allowlist_id)
+            if not wait_event:
+                # First thread to load the entry
+                notify_event = Event()
+                self.loader_map[allowlist_id] = notify_event
+            else:
+                notify_event = None
+
+        if wait_event:
+            # Wait for other thread to notify us of loaded object
+            wait_event.wait()
+            with self.map_lock:
+                return self.map.get(allowlist_id)
+
+        allowlist_entry = AllowlistEntry.from_db(self.session, allowlist_id)
+        if allowlist_entry:
+            with self.map_lock:
+                self.map[allowlist_id] = allowlist_entry
+
+        with self.loader_map_lock:
+            del self.loader_map[allowlist_id]
+            # Notify all waiters
+            assert notify_event  # pyright
+            notify_event.set()
+
+        return allowlist_entry
+
+    def __get_allowlist(self, allowlist_id: int) -> Optional[AllowlistEntry]:
+        """Get an allowlist given its ID"""
+
+        with self.map_lock:
+            allowlist_entry = self.map.get(allowlist_id)
+
+        if allowlist_entry:
+            if not allowlist_entry.check_update():
+                # allowlist_entry likely disappeared from DB
+                del self.map[allowlist_id]
+                return None
+            return allowlist_entry
+
+        return self.__load_allowlist(allowlist_id)
+
+    def get_allowlist(self, allowlist_id: int) -> Optional[AllowlistEntry]:
+        """Get an allowlist given its ID; cleanup stale entries"""
+
+        allowlist_entry = self.__get_allowlist(allowlist_id)
+
+        self.__purge()
+
+        return allowlist_entry
--- a/keylime/cloud_verifier_tornado.py
+++ b/keylime/cloud_verifier_tornado.py
@@ -15,7 +15,7 @@ import tornado.process
 import tornado.web
 from sqlalchemy.engine import Engine
 from sqlalchemy.exc import SQLAlchemyError
-from sqlalchemy.orm import Session, joinedload
+from sqlalchemy.orm import Session, joinedload, noload
 from sqlalchemy.orm.exc import NoResultFound
 
 from keylime import api_version as keylime_api_version
@@ -29,6 +29,7 @@ from keylime import (
     web_util,
 )
 from keylime.agentstates import AgentAttestState, AgentAttestStates
+from keylime.allowlistdbproxy import AllowlistDBProxy
 from keylime.common import retry, states, validators
 from keylime.da import record
 from keylime.db.keylime_db import DBEngineManager, SessionManager
@@ -62,6 +63,9 @@ def get_session() -> Session:
 
 def get_AgentAttestStates() -> AgentAttestStates:
     return AgentAttestStates.get_instance()
+
+
+AllowlistDBProxy.create_instance(get_session())
 
 
 # The "exclude_db" dict values are removed from the response before adding the dict to the DB
@@ -1112,12 +1116,14 @@ async def process_agent(
         main_agent_operational_state = agent["operational_state"]
         stored_agent = None
         try:
+            # Never load the ima_policy as part of loading VerifierMain
             stored_agent = (
                 session.query(VerfierMain)
-                .options(joinedload(VerfierMain.ima_policy))
+                .options(noload(VerfierMain.ima_policy))
                 .filter_by(agent_id=str(agent["agent_id"]))
                 .first()
             )
+            assert not stored_agent.ima_policy
         except SQLAlchemyError as e:
             logger.error("SQLAlchemy Error for agent ID %s: %s", agent["agent_id"], e)
 
@@ -1184,8 +1190,14 @@ async def process_agent(
         except SQLAlchemyError as e:
             logger.error("SQLAlchemy Error for agent ID %s: %s", agent["agent_id"], e)
 
+        allowlist_id = stored_agent.ima_policy_id
+        allowlist_entry = AllowlistDBProxy.get_instance().get_allowlist(allowlist_id)
+        if not allowlist_entry:
+            logger.error("Could not get an allowlist for agent %s", stored_agent.agent_id)
+            return
+
         # Load agent's IMA policy
-        ima_policy = stored_agent.ima_policy
+        ima_policy = allowlist_entry.verifier_allowlist
 
         # If agent was in a failed state we check if we either stop polling
         # or just add it again to the event loop
--- a/keylime/cloud_verifier_tornado.py
+++ b/keylime/cloud_verifier_tornado.py
@@ -1191,7 +1191,7 @@ async def process_agent(
             logger.error("SQLAlchemy Error for agent ID %s: %s", agent["agent_id"], e)
 
         allowlist_id = stored_agent.ima_policy_id
-        allowlist_entry = AllowlistDBProxy.get_instance().get_allowlist(allowlist_id)
+        allowlist_entry = await AllowlistDBProxy.get_instance().get_allowlist(allowlist_id)
         if not allowlist_entry:
             logger.error("Could not get an allowlist for agent %s", stored_agent.agent_id)
             return
--- a/keylime/db/verifier_db.py
+++ b/keylime/db/verifier_db.py
@@ -1,4 +1,6 @@
-from sqlalchemy import Column, ForeignKey, Integer, LargeBinary, PickleType, String, Text, schema
+import datetime
+
+from sqlalchemy import TIMESTAMP, Column, ForeignKey, Integer, LargeBinary, PickleType, String, Text, schema
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.orm import relationship
 
@@ -61,3 +63,4 @@ class VerifierAllowlist(Base):
     generator = Column(Integer)
     tpm_policy = Column(Text())
     ima_policy = Column(Text().with_variant(Text(429400000), "mysql"))
+    modified = Column(TIMESTAMP, nullable=False, default=datetime.datetime.now, onupdate=datetime.datetime.now)
--- a/keylime/migrations/versions/bc9b95b3a2f4_add_modified_to_allowlist.py
+++ b/keylime/migrations/versions/bc9b95b3a2f4_add_modified_to_allowlist.py
@@ -1,0 +1,42 @@
+"""empty message
+
+Revision ID: bc9b95b3a2f4
+Revises: 039322ea079b
+Create Date: 2022-11-23 08:37:37.600030
+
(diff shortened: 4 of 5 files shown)


keylime/allowlistdbproxy.py
```diff
@@ -0,0 +1,128 @@
...
+        self.allowlist_name = allowlist_name
+        self.last_update_check = time.time()
+        self.update_timeout = update_timeout
+        self.checksum = checksum
+        self.verifier_allowlist = verifier_allowlist
+
+    def __has_hash_changed(self):
+        """Check whether the unique identifier in the DB has changed"""
+        # FIXME: change to .hash once available
+        statement = select(VerifierAllowlist.id).filter_by(name=self.allowlist_name)
```
stefanberger, line 42: This line seems to cause issued in the Fedora 34 container where SQLAlchemy package 1.3.22 is likely installed (https://rpmfind.net/linux/RPM/fedora/34/x86_64/p/python3-sqlalchemy-1.3.22-2.fc34.x86_64.html). More recent versions of SQLAlchemy support this statement.

```
2022-11-11 15:34:23.191 - keylime.verifier - ERROR - Operator 'getitem' is not supported on this expression
  File "/usr/local/lib/python3.9/site-packages/keylime/cloud_verifier_tornado.py", line 1165, in process_agent
    allowlist_entry = allowlist_db_proxy.get_allowlist(allowlist_name)
  File "/usr/local/lib/python3.9/site-packages/keylime/allowlistdbproxy.py", line 117, in get_allowlist
    if not allowlist_entry.check_update():
  File "/usr/local/lib/python3.9/site-packages/keylime/allowlistdbproxy.py", line 54, in check_update
    if self.__has_hash_changed():
  File "/usr/local/lib/python3.9/site-packages/keylime/allowlistdbproxy.py", line 42, in __has_hash_changed
    statement = select(VerifierAllowlist.id).filter_by(name=self.allowlist_name)
```

stefanberger: The above is works on RHEL-9 with `SQLAlchemy==1.4.43` and following the tests runs it probably also works with the version in Fedora 35 (1.4.23, https://fedora.pkgs.org/35/fedora-x86_64/python-sqlalchemy-doc-1.4.23-1.fc35.noarch.rpm.html)

keylime/allowlistdbproxy.py
```diff
@@ -0,0 +1,177 @@
...
+from sqlalchemy.exc import SQLAlchemyError
+from sqlalchemy.orm.exc import NoResultFound
+
+from keylime import keylime_logging
+from keylime.db.verifier_db import VerifierAllowlist
+
+logger = keylime_logging.init_logging("allowlistdbproxy")
+
+
+class AllowlistEntry:
```
aplanas, line 14: I like this approach. Did you research a more generic approach? `functools` has `lru_cache` that it is not supporting `ttl`, but I think that should be fairly easy to add.

Having some kind generic reusable memoization class can help also in different scenarios in keylime.  This class can be used as a decorator for specific methods, and later can be added persistence too (one example can be [2])

[1] https://docs.python.org/3/library/functools.html#functools.lru_cache
[2] https://github.com/openSUSE/openSUSE-release-tools/blob/master/osclib/memoize.py

stefanberger: The `lru_cache` may be a bit older than the `cache` python version-wise.

From what I see we need to constantly poll the DB for an update of AllowlistVerifier table so we don't really have a constant cache but have to explicitly go out to the DB every time (or every 0.1s of a non-update) .

```
            # We have to be really careful with this update_timeout. If attestation fails
            # and an update of the allowlist comes very quickly this update_timeout may
            # cause the new policy not to be loaded and cause an unexpected attestation
            # failure... it may be better to always poll the DB.
```


```
 @functools.lru_cache(user_function)
 @functools.lru_cache(maxsize=128, typed=False)

    Decorator to wrap a function with a memoizing callable that saves up to the maxsize most recent calls. It can save time when an expensive or I/O bound function is periodically called with the same arguments.
```

lease-recently used doesn't sound right. I would need something that times out unused elements but not something like a limited-size LRU queue.

aplanas: > The `lru_cache` may be a bit older than the `cache` python version-wise.

Right. `cache` is expressed as a `lru_cache` with different defaults, so can be more handy. IIRC is just a two lines decorator.
 
> lease-recently used doesn't sound right. I would need something that times out unused elements but not something like a limited-size LRU queue.

Yes. That was I was referring as a "different invalidation than a TTL". As commented before, a time-to-live parameter should be fairly easy to implement on top of `lru_cache`.

The second link of my first comment is a different decorator (not using `lru_cache`, as the original code comes from 2.7) that it is implementing a TTL invalidator.

In any case, I think that we can move on : ))

galmasi:
We have been reviewing the code and the comments with @maugustosilva . 

1. The "weakness" of the current implementation, as we see it, is that it trades sqlalchemy bandwidth against computational bandwidth. I have not measured (yet) the time it takes to calculate a digest for a 20MB allowlist, but I'm persuaded that it is *substantial*. @maugustosilva is working on a small (few lines) PR that adds said digest to the `VerifierAllowlist` schema, saving ourselves the trouble of recalculating it.

2. We much prefer the approach suggested by @aplanas -- the LRU cache, and using python native features as much as possible in preference to code we write ourselves. @stefanberger please consider the idea.

3. Obviously, as the code matures here, we would like to _test_ it for scalability right away, even before it is merged -- there is no point in this fix unless it does what it's supposed to (i.e. allow flat scaling of attestation with IMA).

stefanberger:
> 1. The "weakness" of the current implementation, as we see it, is that it trades sqlalchemy bandwidth against computational bandwidth. I have not measured (yet) the time it takes to calculate a digest for a 20MB allowlist, but I'm persuaded that it is _substantial_. @maugustosilva is working on a small (few lines) PR that adds said digest to the `VerifierAllowlist` schema, saving ourselves the trouble of recalculating it.

From the perspective of this patch we don't need a digest. All we need is an identifier that is updated when the entry in the VerifierAllowlist is updated that should cause a reload/`refresh()` of the entry.

> 
>     2. We much prefer the approach suggested by @aplanas -- the LRU cache, and using python native features as much as possible in preference to code we write ourselves. @stefanberger please consider the idea.


 See above conversations about applicability of LRU caches. IMO it doesn't apply here.

> 
>     3. Obviously, as the code matures here, we would like to _test_ it for scalability right away, even before it is merged -- there is no point in this fix unless it does what it's supposed to (i.e. allow flat scaling of attestation with IMA).

Go ahead. **However, consider that the unique update identifier for an update to an entry in the DB is currently missing and every time an entry in the DB is reloaded. I did it like this for now to pass the current `-c update` tests.** FIXMEs in the code indicate what will need to be changed once the unique update identifier comes along.

```
    def __has_checksum_changed(self):
        """Check whether the checksum of the allowlist in the DB has changed"""

        verifier_allowlist = self.session.execute(self.query_checksum_stmt).one()

        # FIXME: Don't always return True
        if verifier_allowlist.id != self.allowlist_checksum:
            return True
        return True
```

galmasi:
Hmmm I think @maugustosilva 's proposed PR is *exactly* designed to implement the missing code in your PR with trivial amounts of plumbing.

I don't think allowlist "unique identifiers" are reliable indicators in the current implementation. FWIW the default behavior for allowlist names is to be the UUID of the agent they belong to, and they keep their names even when one changes them. 

The best way IMHO to identify is an allowlist is through a digest. It's portable, and it only needs to be computed once. 

I reviewed Python `functools` and yes, I'm ready to move on too :)

stefanberger:
> Hmmm I think @maugustosilva 's proposed PR is _exactly_ designed to implement the missing code in your PR with trivial amounts of plumbing.

I will rebease this PR once @maugustosilva is viable/available or has been merged.

> 
> I don't think allowlist "unique identifiers" are reliable indicators in the current implementation. FWIW the default behavior for allowlist names is to be the UUID of the agent they belong to, and they keep their names even when one changes them.
> 

Once you use a named allowlist it's NOT the UUID anymore. You have to use the stored_agent's ima_policy_id to get the proper name, which may be the UUID but can also be a name allowlist.
```
        allowlist_id = stored_agent.ima_policy_id
```

> The best way IMHO to identify is an allowlist is through a digest. It's portable, and it only needs to be computed once.

Sure but that's going to have to be an update to the existing code. But don't forget about named allowlists. I don't see that this needs to change since it's working just fine right now.

> 
> I reviewed Python `functools` and yes, I'm ready to move on too :)

:-)

stefanberger:
> The best way IMHO to identify is an allowlist is through a digest. It's portable, and it only needs to be computed once.

If a named allowlist is in use by several agent and now you update this named allowlist you will have to update all agents that reference this allowlist IFF you reference it by its digest. IMO, the digest can serve as some sort of heavy checksum, which from the perspective of this PR doesn't need to be anything so heavy, but could be a simple increasing number.

keylime/allowlistdbproxy.py
stefanberger: .where() is supported by older versions

keylime/allowlistdbproxy.py
```diff
@@ -0,0 +1,198 @@
...
+        verifier_allowlist.name = tocopy.name
+        verifier_allowlist.ima_policy = tocopy.ima_policy
+        verifier_allowlist.tpm_policy = tocopy.tpm_policy
+
+        return verifier_allowlist
+
+    def __has_checksum_changed(self):
+        """Check whether the checksum of the allowlist in the DB has changed"""
+
+        verifier_allowlist = self.session.execute(self.query_checksum_stmt).one()
```
stefanberger, line 62: This line here will put the biggest stress on the database but select only one field (checksum) for each allowlist.  If we have N clients this would query the table N times, once for each attestation, if no timeout is being used. If we use a timeout between queries, such as 0.1s for example, all those clients asking for the same allowlist within 0.1s would be bunched together after the first one queried for the checksum. This would reduce the stress on the DB. However, if during that 0.1s the allowlist is updated and an attestation is run on a system that just used a file that wasn't part of the old allowlist then we get an attestation failure. With no timeout we wouldn't risk any attestation failures... I am not sure how bad such a query for N clients possibly asking for N different entries in the table to get their checksums would be. Does anyone know?

keylime/cloud_verifier_tornado.py
```diff
@@ -1170,8 +1173,12 @@ async def process_agent(agent, new_operational_state, failure=Failure(Component.
         except SQLAlchemyError as e:
             logger.error("SQLAlchemy Error for agent ID %s: %s", agent["agent_id"], e)
 
+        allowlist_db_proxy = AllowlistDBProxy.get_instance(session)
+        allowlist_id = stored_agent.ima_policy_id
+        allowlist_entry = allowlist_db_proxy.get_allowlist(allowlist_id)
```
stefanberger, line 1178: Still need to check for `None` here ....

keylime/allowlistdbproxy.py
```diff
@@ -0,0 +1,247 @@
...
+        return verifier_allowlist
+
+    def __has_checksum_changed(self) -> bool:
+        """Check whether the checksum of the allowlist in the DB has changed"""
+        rnd = int.from_bytes(random.randbytes(4), byteorder="little")
+
+        verifier_allowlist = self.session.execute(
+            f"SELECT checksum,{rnd} FROM allowlists WHERE id={self.allowlist_id}",
+            execution_options={"compiled_cache": None},
+        ).one()
```
stefanberger, line 91: I have been entertaining the testfarm quite a bit. For some reason the test runs don't seem to always see the updated checksum in the `-c update` test. I don't see this problem locally but I don't run the same type of database setup locally. My guess it has something to with caching but I am trying to eliminate caching. Maybe there's a path in the code that doesn't update the checksum but why would this not always happen. Odd.

keylime/allowlistdbproxy.py
```diff
@@ -103,6 +110,12 @@ def check_update(self) -> bool:
                         self.allowlist_checksum = self.verifier_allowlist_db.checksum  # type: ignore
 
                         self.verifier_allowlist = AllowlistEntry.__copy_verifier_allowlist(self.verifier_allowlist_db)
+                    else:
+                        # Refresh anyway...
+                        logger.info(f"ima_policy BEFORE: {self.verifier_allowlist_db.ima_policy}")
+                        self.session.refresh(self.verifier_allowlist_db)
+                        logger.info(f"ima_policy AFTER : {self.verifier_allowlist_db.ima_policy}")
```
stefanberger, line 117: It does seem to be the case that the checksum did not get updated but the ima_policy changed: @maugustosilva 
```
Nov 17 00:24:37 ip-172-31-19-1.us-east-2.compute.internal keylime_verifier[59508]: 2022-11-17 00:24:37.660 - keylime.allowlistdbproxy - INFO - ima_policy BEFORE: {"allowlist": {"meta": {"version": 5, "checksum": "d8d581d93b1421ec6874b84369e880eb1a7c1275ad5f95b13faa6696e8637066", "timestamp": "2022-11-17 00:24:24.121668", "generator": "keylime-legacy-format-upgrade"}, "release": 0, "hashes": {"boot_aggregate": ["0000000000000000000000000000000000000000000000000000000000000000"]}, "keyrings": {}, "ima": {"ignored_keyrings": [], "log_hash_alg": "sha1", "dm_policy": null}, "ima-buf": {}}, "exclude": ["/afs(/.*)?", "/bin(/.*)?", "/boot(/.*)?", "/dev(/.*)?", "/etc(/.*)?", "/home(/.*)?", "/lib(/.*)?", "/lib64(/.*)?", "/media(/.*)?", "/mnt(/.*)?", "/opt(/.*)?", "/proc(/.*)?", "/root(/.*)?", "/run(/.*)?", "/sbin(/.*)?", "/srv(/.*)?", "/sys(/.*)?", "/tmp(/.*)?", "/usr(/.*)?", "/var(/.*)?", "/sysroot/etc/fstab", "/dracut-state.sh", "/keylime-tests/basic-attestation-on-localhost-pbI7Q(/.*)?", "/keylime-tests/basic-attestation-on-localhost-s01Ko(/.*)?", "/keylime-tests/basic-attestation-with-custom-certificates-R1GzJ(/.*)?", "/keylime-tests/basic-attestation-with-ima-signatures-y49Ml(/.*)?", "/keylime-tests/basic-attestation-without-mtls-6ovPZ(/.*)?", "/keylime-tests/basic-attestation-with-unpriviledged-agent-9xbL6(/.*)?", "/keylime-tests/install-rpm-with-ima-signature-S6TAh(/.*)?", "/keylime-tests/keylime-non-default-ports-uLO6I(/.*)?"]}
Nov 17 00:24:37 ip-172-31-19-1.us-east-2.compute.internal keylime_verifier[59508]: 2022-11-17 00:24:37.661 - keylime.allowlistdbproxy - INFO - ima_policy AFTER : {"allowlist": {"meta": {"version": 5, "checksum": "d8d581d93b1421ec6874b84369e880eb1a7c1275ad5f95b13faa6696e8637066", "timestamp": "2022-11-17 00:24:37.592173", "generator": "keylime-legacy-format-upgrade"}, "release": 0, "hashes": {"boot_aggregate": ["0000000000000000000000000000000000000000000000000000000000000000"]}, "keyrings": {}, "ima": {"ignored_keyrings": [], "log_hash_alg": "sha1", "dm_policy": null}, "ima-buf": {}}, "exclude": ["/afs(/.*)?", "/bin(/.*)?", "/boot(/.*)?", "/dev(/.*)?", "/etc(/.*)?", "/home(/.*)?", "/lib(/.*)?", "/lib64(/.*)?", "/media(/.*)?", "/mnt(/.*)?", "/opt(/.*)?", "/proc(/.*)?", "/root(/.*)?", "/run(/.*)?", "/sbin(/.*)?", "/srv(/.*)?", "/sys(/.*)?", "/tmp(/.*)?", "/usr(/.*)?", "/var(/.*)?", "/sysroot/etc/fstab", "/dracut-state.sh", "/keylime-tests/basic-attestation-on-localhost-pbI7Q(/.*)?", "/keylime-tests/basic-attestation-on-localhost-s01Ko(/.*)?", "/keylime-tests/basic-attestation-with-custom-certificates-R1GzJ(/.*)?", "/keylime-tests/basic-attestation-with-ima-signatures-y49Ml(/.*)?", "/keylime-tests/basic-attestation-without-mtls-6ovPZ(/.*)?", "/keylime-tests/basic-attestation-with-unpriviledged-agent-9xbL6(/.*)?", "/keylime-tests/install-rpm-with-ima-signature-S6TAh(/.*)?", "/keylime-tests/keylime-non-default-ports-uLO6I(/.*)?", "/keylime-tests/keylime_tenant-commands-on-localhost-zFCBQ(/.*)?"]}
```

stefanberger: The failure occurred in this test run: https://artifacts.dev.testing-farm.io/88e306a5-2d73-4b59-ab93-05df9bdf733d/

stefanberger:
The user has to provide a checksum for the allowlist ? Why does the tool not compute the checksum?

```
    parser.add_argument(
        "--allowlist-checksum",
        action="store",
        dest="allowlist_checksum",
        default=None,
        help="Specify the SHA-256 checksum of an allowlist",
    )
```

keylime/allowlistdbproxy.py
maugustosilva: The checksum is always calculated over the `allowlist` only. If any other aspect of the IMA policy, such as the `excludelist` was changed, a new checksum is (currently) not recalculated.

maugustosilva:
**If** a checksum is passed as a CLI parameter, then it is compared with the `calculated_checksum` on the `read_allowlist` function (`ima.py`). If it is not, then latter becomes the `checksum` parameter on the "IMA policy bundle"

stefanberger:
@maugustosilva It's possible that a bug I introduced has something to do with it (PR #1192 1192) that the checksum doesn't get updated.

maugustosilva:
> @maugustosilva It's possible that a bug I introduced has something to do with it (PR #1192 1192) that the checksum doesn't get updated.

I have merged it now.

stefanberger:
> I have merged it now.

Thanks. I rebased this PR on it and we still have the same issue with the not-updated checksum in `-c updated`: https://artifacts.dev.testing-farm.io/c18d0716-4a2a-4a49-b7a1-9d6205cc2132/

```
Nov 17 17:43:13 ip-172-31-20-219.us-east-2.compute.internal keylime_verifier[52693]: 2022-11-17 17:43:13.529 - keylime.allowlistdbproxy - INFO - ima_policy BEFORE: {"allowlist": {"meta": {"version": 5, "checksum": "d8d581d93b1421ec6874b84369e880eb1a7c1275ad5f95b13faa6696e8637066", "timestamp": "2022-11-17 17:42:53.264297", "generator": "keylime-legacy-format-upgrade"}, "release": 0, "hashes": {"boot_aggregate": ["0000000000000000000000000000000000000000000000000000000000000000"]}, "keyrings": {}, "ima": {"ignored_keyrings": [], "log_hash_alg": "sha1", "dm_policy": null}, "ima-buf": {}}, "exclude": ["/bin(/.*)?", "/boot(/.*)?", "/dev(/.*)?", "/etc(/.*)?", "/home(/.*)?", "/lib(/.*)?", "/lib64(/.*)?", "/media(/.*)?", "/mnt(/.*)?", "/opt(/.*)?", "/proc(/.*)?", "/root(/.*)?", "/run(/.*)?", "/sbin(/.*)?", "/srv(/.*)?", "/sys(/.*)?", "/tmp(/.*)?", "/usr(/.*)?", "/var(/.*)?", "/sysroot/etc/fstab", "/dracut-state.sh", "/keylime-tests/basic-attestation-on-localhost-B0BmN(/.*)?", "/keylime-tests/basic-attestation-on-localhost-oo1X4(/.*)?", "/keylime-tests/basic-attestation-with-custom-certificates-3xjDq(/.*)?", "/keylime-tests/basic-attestation-with-ima-signatures-rMpOg(/.*)?", "/keylime-tests/basic-attestation-without-mtls-rBeXc(/.*)?", "/keylime-tests/basic-attestation-with-unpriviledged-agent-Dr8w4(/.*)?", "/keylime-tests/keylime-non-default-ports-aHbWW(/.*)?"]}
Nov 17 17:43:13 ip-172-31-20-219.us-east-2.compute.internal keylime_verifier[52693]: 2022-11-17 17:43:13.530 - keylime.allowlistdbproxy - INFO - ima_policy AFTER : {"allowlist": {"meta": {"version": 5, "checksum": "d8d581d93b1421ec6874b84369e880eb1a7c1275ad5f95b13faa6696e8637066", "timestamp": "2022-11-17 17:43:13.402651", "generator": "keylime-legacy-format-upgrade"}, "release": 0, "hashes": {"boot_aggregate": ["0000000000000000000000000000000000000000000000000000000000000000"]}, "keyrings": {}, "ima": {"ignored_keyrings": [], "log_hash_alg": "sha1", "dm_policy": null}, "ima-buf": {}}, "exclude": ["/bin(/.*)?", "/boot(/.*)?", "/dev(/.*)?", "/etc(/.*)?", "/home(/.*)?", "/lib(/.*)?", "/lib64(/.*)?", "/media(/.*)?", "/mnt(/.*)?", "/opt(/.*)?", "/proc(/.*)?", "/root(/.*)?", "/run(/.*)?", "/sbin(/.*)?", "/srv(/.*)?", "/sys(/.*)?", "/tmp(/.*)?", "/usr(/.*)?", "/var(/.*)?", "/sysroot/etc/fstab", "/dracut-state.sh", "/keylime-tests/basic-attestation-on-localhost-B0BmN(/.*)?", "/keylime-tests/basic-attestation-on-localhost-oo1X4(/.*)?", "/keylime-tests/basic-attestation-with-custom-certificates-3xjDq(/.*)?", "/keylime-tests/basic-attestation-with-ima-signatures-rMpOg(/.*)?", "/keylime-tests/basic-attestation-without-mtls-rBeXc(/.*)?", "/keylime-tests/basic-attestation-with-unpriviledged-agent-Dr8w4(/.*)?", "/keylime-tests/keylime-non-default-ports-aHbWW(/.*)?", "/keylime-tests/keylime_tenant-commands-on-localhost-CFcz0(/.*)?"]}
```


There must be a code path that looses the updated checksum or just reuses the old one...

stefanberger:
@maugustosilva  Ok, so there's nothing wrong with the checksum except it's useless for the purpose of detecting changes to the whole ima_policy since it's only over the allowlist and not the exclude list. Ideally it would be over the whole ima_policy.

maugustosilva:
> @maugustosilva Ok, so there's nothing wrong with the checksum except it's useless for the purpose of detecting changes to the whole ima_policy since it's only over the allowlist and not the exclude list. Ideally it would be over the whole ima_policy.

I agree, we should calculate the checksum over exclude lists too... I will open a PR dedicated to this purpose.

stefanberger:
> I agree, we should calculate the checksum over exclude lists too... I will open a PR dedicated to this purpose.

Will this change the semantics of the checksum ?
I introduced a timestamp `modified` now to detect updates of the allowlists table.  This also works module some issues with older versions of SQLAlchemy() on CentOS 8.
```
Nov 18 15:22:04 ip-172-31-18-25.us-east-2.compute.internal keylime_verifier[21038]:   File "/usr/local/lib/python3.6/site-packages/keylime-6.5.2-py3.6.egg/keylime/cloud_verifier_tornado.py", line 1178, in process_agent
Nov 18 15:22:04 ip-172-31-18-25.us-east-2.compute.internal keylime_verifier[21038]:     allowlist_entry = allowlist_db_proxy.get_allowlist(allowlist_id)
Nov 18 15:22:04 ip-172-31-18-25.us-east-2.compute.internal keylime_verifier[21038]:   File "/usr/local/lib/python3.6/site-packages/keylime-6.5.2-py3.6.egg/keylime/allowlistdbproxy.py", line 231, in get_allowlist
Nov 18 15:22:04 ip-172-31-18-25.us-east-2.compute.internal keylime_verifier[21038]:     allowlist_entry = self.__get_allowlist(allowlist_id)
Nov 18 15:22:04 ip-172-31-18-25.us-east-2.compute.internal keylime_verifier[21038]:   File "/usr/local/lib/python3.6/site-packages/keylime-6.5.2-py3.6.egg/keylime/allowlistdbproxy.py", line 226, in __get_allowlist
Nov 18 15:22:04 ip-172-31-18-25.us-east-2.compute.internal keylime_verifier[21038]:     return self.__load_allowlist(allowlist_id)
Nov 18 15:22:04 ip-172-31-18-25.us-east-2.compute.internal keylime_verifier[21038]:   File "/usr/local/lib/python3.6/site-packages/keylime-6.5.2-py3.6.egg/keylime/allowlistdbproxy.py", line 200, in __load_allowlist
Nov 18 15:22:04 ip-172-31-18-25.us-east-2.compute.internal keylime_verifier[21038]:     allowlist_entry = AllowlistEntry.from_db(self.session, allowlist_id)
Nov 18 15:22:04 ip-172-31-18-25.us-east-2.compute.internal keylime_verifier[21038]:   File "/usr/local/lib/python3.6/site-packages/keylime-6.5.2-py3.6.egg/keylime/allowlistdbproxy.py", line 44, in from_db
Nov 18 15:22:04 ip-172-31-18-25.us-east-2.compute.internal keylime_verifier[21038]:     return AllowlistEntry(session, allowlist_id, verifier_allowlist_db.modified, verifier_allowlist_db)
Nov 18 15:22:04 ip-172-31-18-25.us-east-2.compute.internal keylime_verifier[21038]:   File "/usr/local/lib/python3.6/site-packages/keylime-6.5.2-py3.6.egg/keylime/allowlistdbproxy.py", line 67, in __init__
Nov 18 15:22:04 ip-172-31-18-25.us-east-2.compute.internal keylime_verifier[21038]:     self.query_modified_stmt = select(VerifierAllowlist.modified).where(VerifierAllowlist.id == self.allowlist_id)
Nov 18 15:22:04 ip-172-31-18-25.us-east-2.compute.internal keylime_verifier[21038]:   File "<string>", line 2, in select
Nov 18 15:22:04 ip-172-31-18-25.us-east-2.compute.internal keylime_verifier[21038]:   File "<string>", line 2, in __init__
Nov 18 15:22:04 ip-172-31-18-25.us-east-2.compute.internal keylime_verifier[21038]:   File "/usr/lib64/python3.6/site-packages/sqlalchemy/util/deprecations.py", line 130, in warned
Nov 18 15:22:04 ip-172-31-18-25.us-east-2.compute.internal keylime_verifier[21038]:     return fn(*args, **kwargs)
Nov 18 15:22:04 ip-172-31-18-25.us-east-2.compute.internal keylime_verifier[21038]:   File "/usr/lib64/python3.6/site-packages/sqlalchemy/sql/selectable.py", line 2929, in __init__
Nov 18 15:22:04 ip-172-31-18-25.us-east-2.compute.internal keylime_verifier[21038]:     for c in columns:
Nov 18 15:22:04 ip-172-31-18-25.us-east-2.compute.internal keylime_verifier[21038]:   File "/usr/lib64/python3.6/site-packages/sqlalchemy/sql/operators.py", line 432, in __getitem__
Nov 18 15:22:04 ip-172-31-18-25.us-east-2.compute.internal keylime_verifier[21038]:     return self.operate(getitem, index)
Nov 18 15:22:04 ip-172-31-18-25.us-east-2.compute.internal keylime_verifier[21038]:   File "/usr/lib64/python3.6/site-packages/sqlalchemy/orm/attributes.py", line 211, in operate
Nov 18 15:22:04 ip-172-31-18-25.us-east-2.compute.internal keylime_verifier[21038]:     return op(self.comparator, *other, **kwargs)
Nov 18 15:22:04 ip-172-31-18-25.us-east-2.compute.internal keylime_verifier[21038]:   File "/usr/lib64/python3.6/site-packages/sqlalchemy/sql/operators.py", line 432, in __getitem__
Nov 18 15:22:04 ip-172-31-18-25.us-east-2.compute.internal keylime_verifier[21038]:     return self.operate(getitem, index)
Nov 18 15:22:04 ip-172-31-18-25.us-east-2.compute.internal keylime_verifier[21038]:   File "/usr/lib64/python3.6/site-packages/sqlalchemy/orm/properties.py", line 315, in operate
Nov 18 15:22:04 ip-172-31-18-25.us-east-2.compute.internal keylime_verifier[21038]:     return op(self.__clause_element__(), *other, **kwargs)
Nov 18 15:22:04 ip-172-31-18-25.us-east-2.compute.internal keylime_verifier[21038]:   File "/usr/lib64/python3.6/site-packages/sqlalchemy/sql/operators.py", line 432, in __getitem__
Nov 18 15:22:04 ip-172-31-18-25.us-east-2.compute.internal keylime_verifier[21038]:     return self.operate(getitem, index)
Nov 18 15:22:04 ip-172-31-18-25.us-east-2.compute.internal keylime_verifier[21038]:   File "/usr/lib64/python3.6/site-packages/sqlalchemy/sql/elements.py", line 727, in operate
Nov 18 15:22:04 ip-172-31-18-25.us-east-2.compute.internal keylime_verifier[21038]:     return op(self.comparator, *other, **kwargs)
Nov 18 15:22:04 ip-172-31-18-25.us-east-2.compute.internal keylime_verifier[21038]:   File "/usr/lib64/python3.6/site-packages/sqlalchemy/sql/operators.py", line 432, in __getitem__
Nov 18 15:22:04 ip-172-31-18-25.us-east-2.compute.internal keylime_verifier[21038]:     return self.operate(getitem, index)
Nov 18 15:22:04 ip-172-31-18-25.us-east-2.compute.internal keylime_verifier[21038]:   File "<string>", line 1, in <lambda>
Nov 18 15:22:04 ip-172-31-18-25.us-east-2.compute.internal keylime_verifier[21038]:   File "/usr/lib64/python3.6/site-packages/sqlalchemy/sql/type_api.py", line 67, in operate
Nov 18 15:22:04 ip-172-31-18-25.us-east-2.compute.internal keylime_verifier[21038]:     return o[0](self.expr, op, *(other + o[1:]), **kwargs)
Nov 18 15:22:04 ip-172-31-18-25.us-east-2.compute.internal keylime_verifier[21038]:   File "/usr/lib64/python3.6/site-packages/sqlalchemy/sql/default_comparator.py", line 232, in _getitem_impl
Nov 18 15:22:04 ip-172-31-18-25.us-east-2.compute.internal keylime_verifier[21038]:     _unsupported_impl(expr, op, other, **kw)
Nov 18 15:22:04 ip-172-31-18-25.us-east-2.compute.internal keylime_verifier[21038]:   File "/usr/lib64/python3.6/site-packages/sqlalchemy/sql/default_comparator.py", line 237, in _unsupported_impl
Nov 18 15:22:04 ip-172-31-18-25.us-east-2.compute.internal keylime_verifier[21038]:     "Operator '%s' is not supported on " "this expression" % op.__name__
Nov 18 15:22:04 ip-172-31-18-25.us-east-2.compute.internal keylime_verifier[21038]: NotImplementedError: Operator 'getitem' is not supported on this expression
```

maugustosilva:
Maybe. One of the current functionalities is the comparison between the calculated checksum and one **optionally** provided. 

It is entirely possible to contain the modification in such a way where, checksums for both allowlist and excludelist are always calculated, a checksum, **if provided** is checked against the former (allowlist only) and, if not provided, the calculated checksum becomes the checksum of both checksums. 

I will document all that in an issue and give others the opportunity to comment.

stefanberger:
@galmasi I expect this code to pass all the tests now (spanning all the different versions of SQLAlchemy in use). I think it would be worth giving this a try now for the scalability aspect.

stefanberger:
... actually I may change this here again and instrument the call sites where the allowlists table is written to and have the DBProxy notified about the update this way rather than it polling on the table's modified or checksum fields.

THS-on:
@stefanberger I think the notify code won't work because the verifier uses multiprocessing, not multi-threading.
This has mainly to do that multi-threading in Python is mostly useful for blocking IO because we have a global instruction lock.

stefanberger:
I reverted this PR here now to use polling and caching.

stefanberger:
> @stefanberger I think the notify code won't work because the verifier uses multiprocessing, not multi-threading.

You have a point there :-) I thought these were threads. 

I suppose that due to the usage of asyncio and the `async def` function declarations that we are using tornado in multi-threading mode so that at least the locking that I do around maps is necessary.

THS-on:
> I suppose that due to the usage of asyncio and the async def function declarations that we are using tornado in multi-threading mode so that at least the locking that I do around maps is necessary.

Yes, due to the the async runtime it still makes sense to implement locking, but I think we then should use the asyncio locks and not the ones from the threading library.

stefanberger:
> Yes, due to the the async runtime it still makes sense to implement locking, but I think we then should use the asyncio locks and not the ones from the threading library.

Also tornado has locks: https://www.tornadoweb.org/en/branch4.5/locks.html

stefanberger:
Do we have a multi-threading app or co-routines only? It sounds to me the threading library is a super-set and possibly safer.
```
Note that these primitives are not actually thread-safe and cannot be used in place of those from the standard library–they are meant to coordinate Tornado coroutines in a single-threaded app, not to protect shared objects in a multithreaded app.
```

mbestavros:
The idea of the allowlist proxy seems like a good one! Reducing the amount of database calls we need to make will be helpful. At a quick glance this all seems pretty good to me.

I'll jump in with a few comments here in response to a few things I've seen in the review thread:

- You may have already figured this out, but the `checksum` field within the allowlist metadata should NOT be used. That was a relic of older code that manipulated allowlists prior to transit to the cloud verifier. Now, the verifier assumes that the allowlist that it gets is unchanged from user input. This is necessary to support detached allowlist signature checks on the verifier.

My IMA policy overhaul work (which is hopefully very close to done, just working on updating e2e tests) drops this field entirely. If we want to retain checksum verification between tenant and verifier, we should make it an optional field in the payload so that the file can remain untouched. Otherwise, we should just calculate checksums on the verifier alone.

- I saw some discussion about allowlist UUIDs, and the fact that certain allowlists get named with an agent UUID - to give some background, this was to work around the scenario where a user adds an agent and an allowlist at the same time, without passing an allowlist name. The assumption at play here is that the user assumes that allowlist is tied to that particular agent, and that the allowlist shouldn't stick around after if the agent is deleted.

However, if the user DOES provide an allowlist name, that signals that they don't expect that allowlist to be tied to a particular agent, and should be independent.

The solutions proposed definitely feel less "hacky", so if we can get them working I'm all in favor.

stefanberger:
@mbestavros I don't need the `checksum` field but introduced a `modified` timestamp that is automatically set/updated when a new entry in the allowlists table is written or an existing one is updated. 

I am also not using the allowlist UUID or name anymore but I am indexing the map by the unique allowist Id that is a key in the allowlist table.

I think this PR should help relief unnecessary traffic from the DB. However, the fact that we have multi-processing (if it is active at all and I am not sure about this) instead of multi-threading, will not necessarily be able to provide the full benefit of caching allowlists table entries in memory but some may expire before they get used again IF the processes don't refresh them by accessing them.

mbestavros:
> I don't need the checksum field but introduced a modified timestamp that is automatically set/updated when a new entry in the allowlists table is written or an existing one is updated.

As long as the `modified` timestamp is separate from the policy file itself (i.e. we aren't manipulating policy files in transit or in storage), I think this is a good idea.

mpeters:
I like the improvements of using a timestamp and unique id from the db. Any thought to using a UUID to make it db agnostic? I can imagine some scenarios where orgs are migrated or merged and/or a db is changed from sqlite to mysql where having something external to the DB's incrementing counter could be useful. Not sure though.

In general though, I'm a bit hesitant to introduce a caching layer (as this seems to be) since cache invalidation is such a tricky thing. Also, do we know that this actually has a measured impact on performance at scale? And if so, how much?

stefanberger:
@mpeters What his patch now avoids is unnecessary loading of a whole allowlists entry that hasn't changed. The entry can be quite large:

```
class VerifierAllowlist(Base):
    __tablename__ = "allowlists"
    __table_args__ = (schema.UniqueConstraint("name", name="uniq_allowlists0name"),)
    id = Column(Integer, primary_key=True)
    agent = relationship("VerfierMain", back_populates="ima_policy")
    name = Column(String(255), nullable=False)
    checksum = Column(String(128))
    generator = Column(Integer)
    tpm_policy = Column(Text())
```

I don't have numbers and I don't have infrastructure or setups to get numbers, either.

stefanberger:
> I like the improvements of using a timestamp and unique id from the db. Any thought to using a UUID to make it db agnostic? I can imagine some scenarios where orgs are migrated or merged and/or a db is changed from sqlite to mysql where having something external to the DB's incrementing counter could be useful. Not sure though.

I am not sure what you mean. I admit that I have only tried the `TIMESTAMP` with sqlite. Per its definition it's not a server timestamp but happens on the client, so at least that should be DB-agnostic and I would think that TIMESTAMP type is commonly supported by databases these days.

`modified = Column(TIMESTAMP, nullable=False, default=datetime.datetime.now, onupdate=datetime.datetime.now)`

stefanberger:
Considering the size of the IMA policy `ima_policy = Column(Text().with_variant(Text(429400000), "mysql"))` we can probably assume that there's some line in the MB area where permanently reloading the policy has a performance impact and puts unnecessary stress on the DB if we have many systems that are being attested-to in short intervals. Also, if the size of this policy was to become excessive one could argue that it cannot all be kept in memory if there were multiple systems with each having an excessive-sized policy and you don't have enough RAM on the system. So one could now restrict the size of the cache and/or not cache any allowlists entries that are beyond a certain size and prefer loading the big ones every time. I am not sure what is better and whether it then solves the scalability problem.

Issue #1143, opened 2022-10-26
outstanding type issues


mdrocco:
This is a spinoff of #929 , following up the first batches of type fixes in various PRs. The following files are reported to contain type issues according to `pyright` with Python 3.10:

- [ ] `keylime/backport_dataclasses.py`
- [x]  `keylime/cloud_verifier_tornado.py`
- [ ]  `keylime/da/examples/file.py`
- [ ]  `keylime/da/examples/redis.py`
- [ ]  `keylime/da/examples/sqldb.py`
- [x]  `keylime/da/examples/tsa_rfc3161.py`
- [x]  `keylime/da/record.py`
- [ ]  `keylime/mba/elchecking/example.py`
- [x]  `keylime/ima/ima_dm.py`
- [x]  `keylime/keylime_agent.py`
- [ ]  `keylime/migrations/versions/*`
- [x]  `keylime/revocation_actions/update_crl.py`
- [ ]  `test/data/template-invalid-adjust/2.0/adjust.py`

mdrocco:
Here is a new one: possibly unbounded variable in the verifier's code.

https://github.com/keylime/keylime/blob/8e417f5b99f0dea0f9f36e26ab15549fcd3e1844/keylime/cloud_verifier_tornado.py#L608

I tried the obvious fix but it breaks CI tests:

https://github.com/mdrocco/keylime/blob/ccfd29a2942eed2bed86628d7fadc44bd86b6810/keylime/cloud_verifier_tornado.py#L576-L577

mdrocco:
Update: PR #1186 fixes the reported problems in `ima/ima_dm.py`, I am removing the section of the issue about that file.

mdrocco:
Update: PR #1194 fixes the reported problems in `ima/file_signatures.py` and `json.py`, I am removing the section of the issue about those files.

mdrocco:
I am restructuring this issue to keep track of all the files with outstanding type issues, considering the existing PRs.

mdrocco:
update: upgrading Lark from 1.1.3 to 1.1.4 fixed the issue below:

```console
/home2/mdrocco/git/keylime/keylime/ima/ima_dm.py
  /home2/mdrocco/git/keylime/keylime/ima/ima_dm.py:16:18 - error: "Lark" is not exported from module "lark"
    Import from "lark.lark" instead (reportPrivateImportUsage)
[...]

PR #1180, opened 2022-11-14
type fixes concerning IMA

mdrocco:
This PR provides a batch of type fixes that are:

- almost trivial from the perspective of code behavior
- concerning IMA

In my local testing setup, I have one test failing that I am investigating with @maugustosilva. This PR should be considered temporary until I am done debugging.


keylime/ima/ast.py
```diff
@@ -29,11 +29,15 @@
 
 
 def get_START_HASH(hash_alg: Hash) -> bytes:
-    return codecs.decode(b"0" * (hash_alg.get_size() // 4), "hex")
+    hsize = hash_alg.get_size()
```
stefanberger, line 32: You are doing this to make pyright happy, right? It complains that `hsize` could be `None`? A hash whose size could be `None`...  hm... ? I am surprised it doesn't complain that one couldn't called `get_size()` on  a `NoneType`.

mdrocco: I am not following you on the "missing" complaint: what do you mean calling `get_size()` on  a `NoneType`? Anyway yes, I am "fixing" the assumption that `hash_alg.get_size()` is not a `NoneType`, as that would break the `//` application type-wise.

aplanas: I think that was the second time that we discussed this same point. Seems that the reason is this:

https://github.com/keylime/keylime/blob/5a7c5e3d3dde1d35fec688323744fe9530e585b8/keylime/common/algorithms.py#L48

_HASH_SIZE is a dict and the size is checked as a query on it. I would be in favor of not making it optional, and change the class in a way that if the hash family is not recognized, fail.

But for this PR, that is about adding type hints without changing semantic, I think that what @mdrocco did it is OK

mdrocco: thanks @aplanas for clarifying, in fact I saw the same pattern here and there while applying type fixes/hints, we might want to take an extra step and remove some of those `Optional` as further refinement, but for the sake of minimalism I would not include that in this PR.

keylime/ima/ima.py
```diff
@@ -287,6 +288,7 @@ def _process_measurement_list(
     if not is_valid:
         # This should not happen as the exclude list has already been validated
         # by the verifier before acceping it. This is a safety net just in case.
+        assert err_msg is not None
```
stefanberger, line 291: I now put in PR #1185 to address this particular issue that we have here due to redundancy and we can avoid the assert if we want to.

stefanberger: Now that PR #1185 has been merged you should be able to remove this assert. 

The issue with the asserts is that that they may in some cases hide issues that we can fix but in other cases pyright simply doesn't understand more complex code and it may be necessary to add them then.

mdrocco:
Thanks @stefanberger for reviewing this. I like your suggestion of splitting into single-purpose commits, I will apply that once I am done investigating the CI failure.

keylime/ima/ima.py
```diff
@@ -233,7 +233,8 @@ def _validate_ima_buf(
             failure = _validate_ima_ng(exclude_regex, allowlist, digest, path, hash_types="keyrings")
             if not failure:
                 # Add the key only now that it's validated (no failure)
-                ima_keyrings.add_pubkey_to_keyring(pubkey, path.name, keyidv2=keyidv2)
+                if ima_keyrings is not None:
+                    ima_keyrings.add_pubkey_to_keyring(pubkey, path.name, keyidv2=keyidv2)
```
stefanberger, line 237: Fixes:   /home/stefanb/keylime/keylime/ima/ima.py:312:90 - error: Argument of type "Unknown | None" cannot be assigned to parameter "ima_keyrings" of type "ImaKeyrings function "_validate_ima_buf" (reportGeneralTypeIssues)

LGTM.

keylime/ima/ima.py
```diff
@@ -352,8 +354,8 @@ def _process_measurement_list(
                         agentAttestState.set_ima_dm_state(dm_validator.state_dump())
 
             # Keep old functionality for writing the parsed files with hashes into a file
-            if m2w is not None and (type(entry.mode) in [ast.Ima, ast.ImaNg, ast.ImaSig]):
-                hash_value = codecs.encode(entry.mode.digest.bytes, "hex")
+            if m2w is not None and (isinstance(entry.mode, (ast.Ima, ast.ImaNg, ast.ImaSig))):
+                hash_value = codecs.encode(entry.mode.bytes(), "hex")
```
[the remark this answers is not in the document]
THS-on: I think I tested originally, but when I refactored the AST code and missed changing this part. So yes this code is currently not covered by any test.

keylime/ima/ast.py
stefanberger: > But for this PR, that is about adding type hints without changing semantic, I think that what @mdrocco did it is OK

As long as we remember that the added asserts are hiding the issue...

aplanas: An issue can be created, indeed

keylime/ima/ima_dm.py
```diff
@@ -392,7 +393,7 @@ def _check_attr(attr: Optional[Union[int, str, bool]], reference_value: Optional
     if reference_value is None:
         return True
 
-    if attr is None and reference_value is not None:
+    if attr is None:
```
stefanberger, line 396: LGTM considering above already has `if reference_value is None:`

keylime/ima/ima_dm.py
```diff
@@ -293,7 +294,7 @@ def validate_table_load(self, event: "LoadEvent", match_key, digest) -> Failure:
         # Check "num_targets"
         # Note that we get actually could get multiple lines, but this does not happen for our use cases so we
         # treat it as a failure.
-        if event.device_metadata.num_targets != len(event.targets) or used_policy["table_load"][entry] != len(
+        if event.device_metadata.num_targets != len(event.targets) or used_policy["table_load"]["num_targets"] != len(
```
stefanberger, line 297: Probably was assuming that entry leaves the loop above with "num_targets". Tricky. LGTM.

keylime/ima/ima_dm.py
```diff
@@ -189,12 +189,13 @@ def validate_device_resume(self, event: "ResumeEvent", match_key) -> Failure:
             return failure
 
         # Check if the table hash is consistent
+        assert device_state.active_table_hash is not None
```
stefanberger, line 192: Would defer.

keylime/ima/ima_dm.py
```diff
@@ -272,7 +273,7 @@ def validate_table_load(self, event: "LoadEvent", match_key, digest) -> Failure:
                 used_policy_name = policy_name
                 break
 
-        if used_policy_name is None:
+        if used_policy is None or used_policy_name is None:
```
stefanberger, line 276: LGTM

maugustosilva (approved):

maugustosilva:
Please rebase, squash the commits and LGTM

mdrocco:
thanks @maugustosilva, actually I just un-squashed into unitary commits, as @stefanberger requested for the sake of simplifying the review process. Once everybody is happy (including the CI, hopefully), I will squash back into a single commit.

keylime/ima/ast.py
stefanberger: PR #1189 should resolve this and the issue below.

keylime/cli/policies.py
```diff
@@ -53,7 +53,7 @@ def process_allowlist(args):
         ima_sign_verification_keys = tenant_keyring.to_string()
 
     # Read command-line path string allowlist
-    al_data = None
+    al_data = {"excllist": []}
```
stefanberger, line 56: LGTM

keylime/cloud_verifier_tornado.py
```diff
@@ -573,6 +573,8 @@ def post(self):
                         logger.info("Allowlist data not provided with request! Using default empty allowlist.")
                         ima_policy = json.dumps(ima.process_ima_policy(ima.EMPTY_ALLOWLIST, []))
 
+                    # Although legit in theory, the following initialization breaks the CI tests
+                    # ima_policy_stored = None
```
stefanberger, line 577: I don't see that this one has to be considered to be initialized since it's only used under the `if ima_policy`. I would drop this patch.

keylime/ima/ast.py
```diff
@@ -374,7 +374,9 @@ def __init__(
             raise ParserError(f"No parser for mode {tokens[2]} implemented.")
         self.mode = mode(tokens[3])
         self._bytes = self.mode.bytes()
-        self.pcr_template_hash = self._pcr_hash_alg.hash(self._bytes)
+        pcr_hash_alg_ = self._pcr_hash_alg.hash(self._bytes)
+        assert pcr_hash_alg_ is not None
```
stefanberger, line 378: Likely resolved with PR #1189 -- at least I don't see it anymore when running `pyright keylime/ima/ast.py`.

The good part is probably the change above to get rid of Optional.

keylime/ima/ima.py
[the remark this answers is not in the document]
stefanberger: Entry class has a `mode` field that has a bytes() method, but I am not sure whether that is what is accessed here because there is also a Digest class ...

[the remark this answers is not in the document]
THS-on: You are right we want the digest of the file not the content of the IMA entry. So I think `entry.mode.digest.hash` should be the correct field.

stefanberger:
@mdrocco At the end of this series you could add this file here: scripts/check_static.sh

```
#!/usr/bin/env bash

for PYRIGHT in pyright; do
        if [ -n "$(type -P ${PYRIGHT})" ]; then
                break
        fi
done

if [ -z "$(type -P ${PYRIGHT})" ]; then
        echo "pyright is required"
        exit 1
fi

${PYRIGHT} --version

${PYRIGHT} \
        $(find ./keylime ./test -name '*.py' \
          ! -name 'convert_ima_policy.py' \
          ! -name 'ima_emulator_adapter.py' \
          ! -name 'policies.py' \
          ! -name 'cloud_verifier_tornado.py' \
          ! -name 'file_signatures.py' \
          ! -name 'ima.py' \
          ! -name 'ima_dm.py' \
          ! -name 'ima_test.py' \
          ! -name 'json.py' \
          ! -name 'oldtest.py' \
          ! -name 'registrar_common.py' \
          ! -name 'revocation_notifier.py' \
          ! -name 'test_ca_util.py' \
          ! -name 'test_crypto.py' \
          ! -name 'test_tpm.py' \
          ! -name 'test_validators.py' \
          ! -name 'test_restful.py' \
          ! -name 'test_ima_dm.py' \
          ! -name 'test_ima_verification.py' \
          ! -name 'test_policy_conversion.py' \
          ! -name 'test_ca_impl_openssl.py' \
          ! -path './keylime/revocation_actions/update_crl.py' \
          ! -path './keylime/da/record.py' \
          ! -path './keylime/da/examples/*.py')

exit $?
```

Maybe the changes in this series already resolve all issues in specific files and you could remove them from the above list in that file. The goal would be to remove issue in one file after another while not allowing new issues to come into the repo.

keylime/cloud_verifier_tornado.py
mdrocco: Unfortunately, that is not true, in fact the type checker is (obviously) complaining about the variable being possibly unbound: 

https://github.com/mdrocco/keylime/blob/chore-fix-type-issues-ima/keylime/cloud_verifier_tornado.py#L610

keylime/ima/ast.py
mdrocco: I dropped the assert.

keylime/ima/ima.py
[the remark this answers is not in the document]
mdrocco: done

keylime/ima/ast.py
mdrocco: I dropped the assert

keylime/cli/policies.py
```diff
@@ -70,7 +70,7 @@ def process_allowlist(args):
             raise UserError(str(ima_e)) from ima_e
 
     # Read command-line path string IMA exclude list
-    excl_data = None
+    excl_data = []
```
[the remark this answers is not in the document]
mdrocco: done

mdrocco:
@stefanberger about your latest comment, I did not get which file you suggest adding to that list: `ast.py` or/and something else?

keylime/cloud_verifier_tornado.py
mdrocco: I think this really highlights some hidden problem(s) in the code, as that initialization looks sound to me. Should I open a separate issue to unclog this PR or we want to figure this out before going on?

stefanberger: From my PoV, go ahead and file a separate issue.

stefanberger: Why is this initialization before the if clause necessary? I only see it being used further under the `if`:
```

                        try:
                            ima_policy_stored = (
                                session.query(VerifierAllowlist).filter_by(name=ima_policy_name).one_or_none()
                            )
                        except SQLAlchemyError as e:
                            logger.error(
                                "SQLAlchemy Error while retrieving stored ima policy for agent ID %s: %s", agent_id, e
                            )
                            raise
                        try:
                            if ima_policy_stored is None:
                                ima_policy_stored = VerifierAllowlist(**ima_policy_db_format)
                                session.add(ima_policy_stored)
                                session.commit()
                        except SQLAlchemyError as e:
                            logger.error("SQLAlchemy Error while updating ima policy for agent ID %s: %s", agent_id, e)
                            raise
```

mdrocco: as I pointed out above, it is also referenced outside the `if`: https://github.com/keylime/keylime/blob/8e417f5b99f0dea0f9f36e26ab15549fcd3e1844/keylime/cloud_verifier_tornado.py#L608

stefanberger: right ...

keylime/ima/ima_dm.py
mdrocco: What about this? https://github.com/keylime/keylime/pull/1180/commits/0b02ca3b4dbd6ebdefd0f5ef4dd7da7d79a39aac

stefanberger:
> @stefanberger about your latest comment, I did not get which file you suggest adding to that list: `ast.py` or/and something else?

I had only tried the script with the current master repository. What should happen is that we remove files from the filter-list so that over time more and more files will be checked with pyright. So maybe your PR already has one file completely fixed for pyright and you could remove it from above list, that's what I meant.

keylime/ima/ima_dm.py
mdrocco: I think it is an internal error, although my understanding is we want to avoid throwing exceptions from the verifier.

mdrocco:
I see. I can do that, but I don't understand why we need another mechanism to include/exclude files from pyright's radius, as that is already specified in `pyproject.toml`, which is fully honored by pyright.

stefanberger:
> I see. I can do that, but I don't understand why we need another mechanism to include/exclude files from pyright's radius, as that is already specified in `pyproject.toml`, which is fully honored by pyright.


I see, then forget about this script. pyright on the command line today throws all these errors with these filters from pyproject.toml applied.

A good thing would be to enable pyright for PRs as well but then you would have to crank up that pyproject.toml file filter for now to only cover those files that are completely fixed and PRs don't bring in any new issues in files that are already clean...  while work would be going on to reduce the file filter

maugustosilva:
This "full pyright enablement", with an additional script (if needed) should happen on a separate PR, IMO

mdrocco:
I agree with @maugustosilva for a number of reasons, including that after recent updates to `master` (which I rebased this PR on) I see pyright raising several new type issues, that are obviously out of the scope of this IMA-centric PR.

keylime/cloud_verifier_tornado.py
mdrocco: I updated the existing issue, please take a look: https://github.com/keylime/keylime/issues/1143#issuecomment-1318795728

keylime/ima/ima_dm.py
THS-on: If I remember correctly that the active_table_hash is None should never happen, so if it does it is an internal error.

>  I think it is an internal error, although my understanding is we want to avoid throwing exceptions from the verifier.

Yes correct, but this masks an potential error. The current solution is fine for now, but open also an issue so that we can investigate this further.

mdrocco: I propose a more radical approach: https://github.com/keylime/keylime/commit/9f1d01eda80b71095d685437e63335028a44fe32

mdrocco: Since @stefanberger sounds happy about the radical approach, I will call it resolved unless @THS-on has any counterargument.

mdrocco:
I think this PR is ready to be merged. As soon as one qualified review is there, I will rebase and squash.

keylime/ima/ima_dm.py
THS-on: Looking at the code it seems that the active_table_hash currently never none, so the more radical approach should be fine.

keylime/ima/ima_dm.py
```diff
@@ -52,7 +52,7 @@ class DeviceState:
     policy_name: str
     active: bool
     valid_state: bool
-    active_table_hash: Optional[ast.Digest]
+    active_table_hash: ast.Digest
```
stefanberger, line 55: LGTM

mdrocco:
rebased and squashed, ready to go!

stefanberger:
> rebased and squashed, ready to go!

let's go!  ;-)

keylime/cloud_verifier_tornado.py
```diff
@@ -573,6 +573,8 @@ def post(self):
                         logger.info("Allowlist data not provided with request! Using default empty allowlist.")
                         ima_policy = json.dumps(ima.process_ima_policy(ima.EMPTY_ALLOWLIST, []))
 
+                    # Although legit in theory, the following initialization breaks the CI tests
```
mpeters, line 576: Instead of putting this here maybe open a ticket about it? I don't like leaving commented out code in and will most likely be forgotten.

mpeters (changes requested):
Minor nit

maugustosilva:
As soon as tests finish, I will merge

keylime/cloud_verifier_tornado.py
maugustosilva: very good point

stefanberger:
I have python 3.9.14 locally and I would not be able to remove ima_dm.py from the filtered-out list:

```
$ pyright
No configuration file found.
pyproject.toml file found at /home/stefanb/keylime.
Loading pyproject.toml file at /home/stefanb/keylime/pyproject.toml
Assuming Python version 3.9
Assuming Python platform Linux
Auto-excluding **/node_modules
Auto-excluding **/__pycache__
Auto-excluding **/.*
stubPath /home/stefanb/keylime/typings is not a valid directory.
Searching for source files
Found 111 source files
pyright 1.1.280
/home/stefanb/keylime/keylime/ima/ima_dm.py
  /home/stefanb/keylime/keylime/ima/ima_dm.py:100:56 - error: Argument of type "ParseTree | Unknown" cannot be assigned to parameter "event" of type "LoadEvent" in function "validate_table_load"
    Type "ParseTree | Unknown" cannot be assigned to type "LoadEvent"
      "ParseTree" is incompatible with "LoadEvent" (reportGeneralTypeIssues)
  /home/stefanb/keylime/keylime/ima/ima_dm.py:102:59 - error: Argument of type "ParseTree | Unknown" cannot be assigned to parameter "event" of type "ResumeEvent" in function "validate_device_resume"
    Type "ParseTree | Unknown" cannot be assigned to type "ResumeEvent"
      "ParseTree" is incompatible with "ResumeEvent" (reportGeneralTypeIssues)
  /home/stefanb/keylime/keylime/ima/ima_dm.py:104:59 - error: Argument of type "ParseTree | Unknown" cannot be assigned to parameter "event" of type "RemoveEvent" in function "validate_device_remove"
    Type "ParseTree | Unknown" cannot be assigned to type "RemoveEvent"
      "ParseTree" is incompatible with "RemoveEvent" (reportGeneralTypeIssues)
  /home/stefanb/keylime/keylime/ima/ima_dm.py:106:59 - error: Argument of type "ParseTree | Unknown" cannot be assigned to parameter "event" of type "RenameEvent" in function "validate_device_rename"
    Type "ParseTree | Unknown" cannot be assigned to type "RenameEvent"
      "ParseTree" is incompatible with "RenameEvent" (reportGeneralTypeIssues)
  /home/stefanb/keylime/keylime/ima/ima_dm.py:108:57 - error: Argument of type "ParseTree | Unknown" cannot be assigned to parameter "event" of type "ClearEvent" in function "validate_table_clear"
    Type "ParseTree | Unknown" cannot be assigned to type "ClearEvent"
      "ParseTree" is incompatible with "ClearEvent" (reportGeneralTypeIssues)
  /home/stefanb/keylime/keylime/ima/ima_dm.py:110:59 - error: Argument of type "ParseTree | Unknown" cannot be assigned to parameter "event" of type "UpdateEvent" in function "validate_target_update"
    Type "ParseTree | Unknown" cannot be assigned to type "UpdateEvent"
      "ParseTree" is incompatible with "UpdateEvent" (reportGeneralTypeIssues)
6 errors, 0 warnings, 0 informations
Completed in 5.392sec

```

maugustosilva:
@mpeters "PR code coverage" keeps failing in way that clearly shows a infrastructure problem. Do you have a guideline on what should be done on this case (which does include "merge if this is the only testing failing")?

keylime/cloud_verifier_tornado.py
mdrocco: good point, and I already have a ticket for that, I will simply remove the comment.

mdrocco: done

mdrocco:
@stefanberger weird that I do not see that issue with Python 3.10:

```console
$ pyright
WARNING: there is a new pyright version available (v1.1.279 -> v1.1.280).
Please install the new version or set PYRIGHT_PYTHON_FORCE_VERSION to `latest`

No configuration file found.
pyproject.toml file found at /storage_sda/mdrocco/git/keylime.
Loading pyproject.toml file at /storage_sda/mdrocco/git/keylime/pyproject.toml
Assuming Python version 3.10
Assuming Python platform Linux
Auto-excluding **/node_modules
Auto-excluding **/__pycache__
Auto-excluding **/.*
stubPath /storage_sda/mdrocco/git/keylime/typings is not a valid directory.
Searching for source files
Found 111 source files
pyright 1.1.279
0 errors, 0 warnings, 0 informations
Completed in 9.083sec
```

stefanberger:
> @stefanberger weird that I do not see that issue with Python 3.10:

the test server runs Python 3.6.15and doesn't see it either. Phew.

THS-on (approved):

THS-on:
> @mpeters "PR code coverage" keeps failing in way that clearly shows a infrastructure problem. Do you have a guideline on what should be done on this case (which does include "merge if this is the only testing failing")?

In the past we merged it if the test was failing because the code coverage is more a nice to have thing.


@mpeters can you approve this, so that it can be merged?

mpeters (approved):

mpeters:
> In the past we merged it if the test was failing because the code coverage is more a nice to have thing.

Agreed. Code coverage shouldn't block merges. I wish there was a way to have it kick off during PRs but not block merges, but last time I checked there wasn't.

PR #1181, opened 2022-11-14
ima: this PR adds checksums for allowlists as a separate column on the DB

maugustosilva:
When an user provides an allowlist, a checksum is already automatically calculated (used for comparison when a checksum is also provided, ignored if not). 

### DIFF
--- a/keylime/cloud_verifier_tornado.py
+++ b/keylime/cloud_verifier_tornado.py
@@ -580,11 +580,11 @@ class AgentsHandler(BaseHandler):
                             logger.warning(err_msg)
                             return
 
-                        ima_policy_db_format = {}
                         if not ima_policy_name:
                             ima_policy_name = agent_id
-                        ima_policy_db_format["name"] = ima_policy_name
-                        ima_policy_db_format["ima_policy"] = ima_policy
+
+                        ima_policy_db_format = ima.ima_policy_db_contents(ima_policy_name, ima_policy)
+
                         try:
                             ima_policy_stored = (
                                 session.query(VerifierAllowlist).filter_by(name=ima_policy_name).one_or_none()
@@ -848,9 +848,9 @@ class AllowlistHandler(BaseHandler):
             logger.warning("POST returning 400 response. Expected non zero content length.")
             return
 
-        ima_policy_db_format = {}
+        ima_policy = "{}"
+
         json_body = json.loads(self.request.body)
-        ima_policy_db_format["name"] = allowlist_name
 
         ima_policy_bundle = json.loads(json_body.get("ima_policy_bundle"))
         if ima_policy_bundle:
@@ -864,11 +864,10 @@ class AllowlistHandler(BaseHandler):
                 web_util.echo_json_response(self, e.code, e.message)
                 logger.warning(e.message)
                 return
-            ima_policy_db_format["ima_policy"] = ima_policy
 
         tpm_policy = json_body.get("tpm_policy")
-        if tpm_policy:
-            ima_policy_db_format["tpm_policy"] = tpm_policy
+
+        ima_policy_db_format = ima.ima_policy_db_contents(allowlist_name, ima_policy, tpm_policy)
 
         session = get_session()
         # don't allow overwritting
--- a/keylime/db/verifier_db.py
+++ b/keylime/db/verifier_db.py
@@ -57,5 +57,6 @@ class VerifierAllowlist(Base):
     id = Column(Integer, primary_key=True)
     agent = relationship("VerfierMain", back_populates="ima_policy")
     name = Column(String(255), nullable=False)
+    checksum = Column(String(128))
     tpm_policy = Column(Text())
     ima_policy = Column(Text().with_variant(Text(429400000), "mysql"))
--- a/keylime/ima/ima.py
+++ b/keylime/ima/ima.py
@@ -33,6 +33,7 @@ ALLOWLIST_CURRENT_VERSION = 5
 EMPTY_ALLOWLIST = {
     "meta": {
         "version": ALLOWLIST_CURRENT_VERSION,
+        "checksum": "1ab36a57ad793caba756b01fc142f296cbfc3d24088fe6d66448f3bb6ebadeee",
     },
     "release": 0,
     "hashes": {},
@@ -507,7 +508,9 @@ def read_allowlist(alist=None, checksum="", al_sig_file=None, al_key_file=None):
             raise Exception(
                 f"Checksum of allowlist does not match! Expected {checksum}, Calculated {calculated_checksum}"
             )
-
+    else:
+        logger.debug("Allowlist did not have a checksum provided, will add calculated checksum to bundle")
+        checksum = calculated_checksum
     # Return artifacts as a JSON bundle, ready for sending
     bundle = {
         "ima_policy": base64.b64encode(alist_bytes).decode(),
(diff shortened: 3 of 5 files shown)


maugustosilva:
@stefanberger FYI, it should allow us to use `checksum` (calculated at the time of the `allowlist` creation) as the key/hash for the proxy/cache

stefanberger:
I'll look at your patch and probably will adapt mine based on it...

keylime/ima/ima.py
```diff
@@ -610,6 +615,30 @@ def canonicalize_allowlist(alist_bytes, checksum=""):
     return alist
 
 
+def ima_policy_db_contents(ima_policy_name: str, ima_policy: str, tpm_policy: bool = False) -> dict:
```
stefanberger, line 618: Is tpm_policy just a bool (rather than a str)? It seems to be `text` in the DB table...

keylime/ima/ima.py
```diff
@@ -610,6 +615,30 @@ def canonicalize_allowlist(alist_bytes, checksum=""):
...
+    "Assembles an ima policy dictionary to be written on the database"
```
stefanberger, line 619: Function comments have to start and end with 3 `"""`.

keylime/ima/ima.py
```diff
@@ -610,6 +615,30 @@ def canonicalize_allowlist(alist_bytes, checksum=""):
...
+    ima_policy_db_format = {}
+    ima_policy_db_format["name"] = ima_policy_name
+    ima_policy_db_format["ima_policy"] = ima_policy
```
stefanberger, line 622: Following the test failure it should probably be sth like:

```
if ima_policy:
```

keylime/ima/ima.py
maugustosilva: oops

keylime/ima/ima.py
maugustosilva: It is `str`, right

keylime/cloud_verifier_tornado.py
```diff
@@ -848,6 +848,8 @@ def post(self):
             logger.warning("POST returning 400 response. Expected non zero content length.")
             return
 
+        ima_policy = json.dumps("{}")
```
stefanberger, line 851: I don't think this is right.
```
>>> from keylime import json
>>> json.dumps("{}")
'"{}"'
>>> v=json.dumps("{}")
>>> v
'"{}"'
```

  Maybe it should be `json.loads("{}")` but why go via json?

keylime/db/verifier_db.py
```diff
@@ -57,5 +57,7 @@ class VerifierAllowlist(Base):
     id = Column(Integer, primary_key=True)
     agent = relationship("VerfierMain", back_populates="ima_policy")
     name = Column(String(255), nullable=False)
+    generator = Column(String(256))
```
aplanas, line 60: Maybe normalize it is too much, because will create also a new JOIN, but I think that this can be moved to an Integer, and user a Python Enum to represent the set of valid values:

https://docs.python.org/3/library/enum.html

keylime/db/verifier_db.py
```diff
@@ -57,5 +57,7 @@ class VerifierAllowlist(Base):
...
+    checksum = Column(String(128))
```
aplanas, line 61: can we rename it to document the hash used? (I guess that it is sha256)

keylime/ima/ima.py
```diff
@@ -572,6 +576,7 @@ def canonicalize_allowlist(alist_bytes, checksum=""):
         else:
             logger.debug("Allowlist does not specify a version. Assuming current version %s", ALLOWLIST_CURRENT_VERSION)
 
+        alist["meta"]["generator"] = "keylime-allowlist-compatible-format"
```
aplanas, line 579: The advantage of using an Enum is that here and in like 36 will be hard to make a typo. And searching over integers will be faster (if this becomes a bottleneck we can create an index, that will fit in memory)

keylime/db/verifier_db.py
maugustosilva: not sure if we want to embed the choice of checksum on the column name

keylime/db/verifier_db.py
maugustosilva: I don't oppose the use of `Enum` but I would rather have it on a separate PR then... I will do it

keylime/cloud_verifier_tornado.py
maugustosilva: a bizarre mistake, evidently... should be simply a string

keylime/db/verifier_db.py
mpeters: I'm fine not having the hash type used in the name if we think it might change. But if it changes can we really support both? Queries against the data will have to use the right algorithm beforehand... but I guess it can send multiple values in the query (say sha256 or sha512).

keylime/migrations/versions/2fbc0fb8fa4d_add_checksum_to_allowlist.py
```diff
@@ -0,0 +1,39 @@
...
+def upgrade_registrar():
+    pass
+
+
+def downgrade_registrar():
+    pass
+
+
+def upgrade_cloud_verifier():
+    op.add_column("allowlists", sa.Column("checksum", sa.String(128)))
```
mpeters, line 35: Can we add a index on this column? That will force mysql (and postgres) will keep it in memory if it will fit and queries against it will be very fast, which should help with scalability issues.

test/test_ima_verification.py
```diff
@@ -281,7 +281,6 @@ def test_read_allowlist(self):
         self.assertEqual(
             al_data["meta"]["generator"], "keylime-legacy-format-upgrade", "AllowList metadata generator is correct"
         )
-        self.assertNotIn("checksum", al_data["meta"], "AllowList metadata no checksum")
```
mpeters, line 284: You removed the test that checked for a checksum, how about adding a test to verify the checksum?

mpeters (changes requested):

keylime/migrations/versions/2fbc0fb8fa4d_add_checksum_to_allowlist.py
maugustosilva: very good point, will do

test/test_ima_verification.py
maugustosilva: fair enough

mpeters (approved):

keylime/ima/ima.py
```diff
@@ -610,6 +613,30 @@ def canonicalize_allowlist(alist_bytes, checksum=""):
...
+def ima_policy_db_contents(ima_policy_name: str, ima_policy: str, tpm_policy: str = "") -> dict:
+    """Assembles an ima policy dictionary to be written on the database"""
+    ima_policy_db_format = {}
+    ima_policy_db_format["name"] = ima_policy_name
+    # TODO: This was required to ensure e2e CI tests pass
+    if ima_policy == "{}":
+        ima_policy_db_format["ima_policy"] = None
+    else:
+        ima_policy_db_format["ima_policy"] = ima_policy
+    ima_policy = json.loads(ima_policy)
```
stefanberger, line 625: shouldn't this be ima_policy_json ?

test/test_ima_verification.py
mbestavros: Just a note, checksums within allowlist objects will no longer be supported with the IMA policy overhaul (because referring to a file's own checksum within that same file will change the checksum, making it useless).

mbestavros:
Apologies for not having the time to weigh in on this (I've been traveling since Sunday), but in general this seems like a good implementation, so it would have had my approval as well :+1: 

It is worth noting that although this code implements support for checksums within allowlist objects, that will no longer be supported with the IMA policy overhaul (because referring to a file's own checksum within that same file will change the checksum, making it useless). However, since there's code to handle the case of not having a `checksum` field on the policy, I think this will still be fine.

keylime/ima/ima.py
```diff
@@ -610,6 +613,30 @@ def canonicalize_allowlist(alist_bytes, checksum=""):
     return alist
 
 
+def ima_policy_db_contents(ima_policy_name: str, ima_policy: str, tpm_policy: str = "") -> dict:
+    """Assembles an ima policy dictionary to be written on the database"""
+    ima_policy_db_format = {}
+    ima_policy_db_format["name"] = ima_policy_name
+    # TODO: This was required to ensure e2e CI tests pass
+    if ima_policy == "{}":
```
stefanberger, line 621: if len(ima_policy_json) == 0:

keylime/ima/ima.py
aplanas: Ah I think that @stefanberger has a point:

```
ima_policy_db_format = {"name": ima_policy_name, "ima_policy": None}
ima_policy = json.loads(ima_policy_json)
if ima_policy:
    ima_policy_db_format["ima_policy"] = ima_policy_json
```

That should be more correct, as there are multiple representation of an empty policy when you compare the JSON string, but there is only one (not the `len`) when you compare the `dict`

PR #1184, opened 2022-11-15
verifier: Implement updateruntimepolicy to update an existing runtime policy

stefanberger:
This PR extends the keylime tenant tool with the updateruntimepolicy command to enable a user to update an existing named runtime policy. IMA Attestation is NOT reset but simply continues with the new policy.

### DIFF
--- a/keylime/cloud_verifier_tornado.py
+++ b/keylime/cloud_verifier_tornado.py
@@ -783,7 +783,7 @@ class AllowlistHandler(BaseHandler):
         runtime_policy_name = rest_params["allowlists"]
         if runtime_policy_name is None:
             web_util.echo_json_response(self, 400, "Invalid URL")
-            if method != "POST":
+            if method not in ["POST", "PUT"]:
                 logger.warning("%s returning 400 response: %s", method, self.request.path)
             return None
 
@@ -946,7 +946,44 @@ class AllowlistHandler(BaseHandler):
         logger.info("POST returning 201")
 
     def put(self):
-        web_util.echo_json_response(self, 400, "Allowlist handler: PUT Not Implemented")
+        """Update an allowlist
+
+        PUT /allowlists/{name}
+        body: {"tpm_policy": {..} ...
+        """
+
+        runtime_policy_name = self.__validate_input("PUT")
+        if runtime_policy_name is None:
+            return
+
+        runtime_policy_db_format = self.__get_runtime_policy_db_format(runtime_policy_name)
+        if not runtime_policy_db_format:
+            return
+
+        session = get_session()
+        # don't allow creating a new policy
+        try:
+            runtime_policy_count = session.query(VerifierAllowlist).filter_by(name=runtime_policy_name).count()
+            if runtime_policy_count != 1:
+                web_util.echo_json_response(
+                    self, 409, f"Runtime policy with name {runtime_policy_name} does not already exist"
+                )
+                logger.warning("Runtime policy with name %s does not already exist", runtime_policy_name)
+                return
+        except SQLAlchemyError as e:
+            logger.error("SQLAlchemy Error: %s", e)
+            raise
+
+        try:
+            # Update the named runtime policy
+            session.query(VerifierAllowlist).filter_by(name=runtime_policy_name).update(runtime_policy_db_format)
+            session.commit()
+        except SQLAlchemyError as e:
+            logger.error("SQLAlchemy Error: %s", e)
+            raise
+
+        web_util.echo_json_response(self, 201)
+        logger.info("PUT returning 201")
 
     def data_received(self, chunk):
         raise NotImplementedError()
--- a/keylime/cloud_verifier_tornado.py
+++ b/keylime/cloud_verifier_tornado.py
@@ -770,24 +770,32 @@ class AllowlistHandler(BaseHandler):
     def head(self):
         web_util.echo_json_response(self, 400, "Allowlist handler: HEAD Not Implemented")
 
+    def __validate_input(self, method: str) -> Optional[str]:
+        """Validate the input"""
+        rest_params = web_util.get_restful_params(self.request.uri)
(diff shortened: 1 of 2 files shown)


mbestavros:
I'm just coming back from a week of travel, but I'd also note that the IMA policy overhaul is near-complete, and it would probably make sense to introduce this API after that has landed: https://github.com/keylime/keylime/pull/1147

stefanberger:
> I'm just coming back from a week of travel, but I'd also note that the IMA policy overhaul is near-complete, and it would probably make sense to introduce this API after that has landed: #1147

Sounds good.

mpeters:
@stefanberger just wanted to bring to your attention that @mbestavros 's work on the policy overhaul is done, so would be nice to have this feature completed on top of that.

stefanberger:
Here's a test script that works for me. It creates a named policy and has the local system monitored with this named policy. Then a test script is created and the policy is updated before the test script is run.

NOTE: Some delays in this script are necessary for the verification not to fail, in particular the `updateruntimepolicy` needs to be followed by a `sleep` and then one can run the modified test prg. If the sleep is missing then it will most likely still use the old policy and things are going to fail... It could be related to some caching issues with the database.

```
#!/usr/bin/env bash
#set -x

testfile="${PWD}/test-$(date +%s)"
touch "${testfile}"

gawk '{print substr($4,8)"  "$5}' \
        < /sys/kernel/security/ima/ascii_runtime_measurements \
        > allowlist

# clean up from previous run
keylime_tenant \
        -t 127.0.0.1 \
        -r 127.0.0.1 \
        --uuid d432fbb3-d2f1-4a97-9ef7-75bd81c00000 \
        -c delete \
        &>/dev/null

keylime_tenant \
        -t 127.0.0.1 \
        -r 127.0.0.1 \
        --runtime-policy-name myrp \
        -c deleteruntimepolicy \
        &>/dev/null

# Need the cert due to signatures on my system
keylime_convert_runtime_policy \
        -a allowlist \
        -o ./rp.json \
        -v ./releasecert.der || exit 1

# add named runtime policy
keylime_tenant \
        -t 127.0.0.1 \
        -r 127.0.0.1 \
        --runtime-policy ./rp.json \
        --runtime-policy-name myrp \
        -c addruntimepolicy || exit 1

# start monitoring local system
keylime_tenant \
        -t 127.0.0.1 \
        -r 127.0.0.1 \
        -f "${testfile}" \
        --uuid d432fbb3-d2f1-4a97-9ef7-75bd81c00000 \
        --runtime-policy-name myrp \
        -c add || exit 1

# Let it start monitoring the system
sleep 3

# Create the test file
cat > "${testfile}" << _EOF_
#!/usr/bin/env bash
echo a\\
_EOF_

chmod 755 "${testfile}"

sha256sum "${testfile}" >> allowlist

# Before running this file update the named policy
keylime_convert_runtime_policy \
        -a allowlist \
        -o ./rp.json \
        -v ./releasecert.der || exit 1

# add named runtime policy
keylime_tenant \
        -t 127.0.0.1 \
        -r 127.0.0.1 \
        --runtime-policy ./rp.json \
        --runtime-policy-name myrp \
        -c updateruntimepolicy || exit 1

# Let it start monitoring the system
sleep 1

echo "=========================="
echo "Running ${testfile} (hash '$(sha256sum ${testfile} |cut -f1 -d " ")) -- keylime should not report error"

#run the executable -- keylime should still be monitoring the system
${testfile}

sleep 5

echo "Modifying ${testfile} & updating runtime policy -- keylime should not report error"

# Append to testfile and run it again after updating the runtime policy
echo -n "b" >> "${testfile}"

sha256sum "${testfile}" >> allowlist

# Before running this file update the named policy
keylime_convert_runtime_policy \
        -a allowlist \
        -o rp.json \
        -v ./releasecert.der

# add named runtime policy
keylime_tenant \
        -t 127.0.0.1 \
        -r 127.0.0.1 \
        --runtime-policy ./rp.json \
        --runtime-policy-name myrp \
        -c updateruntimepolicy

# We MUST give some time for the policy update to propagate...
sleep 2

#run the executable -- keylime should still be monitoring the system
${testfile}

sleep 5

echo -n "c" >> "${testfile}"
echo "=========================="
echo "Modifying ${testfile} (hash '$(sha256sum ${testfile} |cut -f1 -d " ")) but NOT updating runtime policy -- keylime should report error"
${testfile}
```

stefanberger:
The typical expected failure message with the script above should be something like this (hashes always the same, filename different):

```
2023-01-25 16:15:14.795 - keylime.ima - WARNING - Hashes for file /root/test-1674681290 don't match cdd6cd5272d5ffefe676fa76bf7c5648ea5fedcb6d773208cd44db433fd0c8a8 not in ['06843c42ebe14fdb900ab7572c0bbcf80e1dec83d1742e609bf9862dae9bca5d', '7f5acc8899cd50f772c82dda1b97a8a03529bc613e37c01ce5b9a12927a4eafa']
```

stefanberger:
I can change the `sleep 2` before the execution of the test program `${testfile}` to a `sleep 1`. It's odd that this is necessary at all because on the policy-update side inside the verifier we have presumably a synchronous DB update:
```
        try:
            # Update the named runtime policy
            session.query(VerifierAllowlist).filter_by(name=runtime_policy_name).update(runtime_policy_db_format)
            session.commit()
        except SQLAlchemyError as e:
            logger.error("SQLAlchemy Error: %s", e)
            raise
```

The `commit()` should make this synchronous relative to any database reader, but it seems like a database reader still sees the old policy when querying within a short amount of time because I cannot change the `sleep ` to a `sleep 0` because then the updated test program would run too quickly after the policy update.

stefanberger:
A `session.flush()` after the `session.commit()` doesn't help it. So I think this is as good as it gets.

aplanas (approved):
LGTM. The `vars` use is very ugly but very unrelated to the PR

mpeters (approved):

PR #1185, opened 2022-11-16
Remove redundant return parameters from functions

stefanberger:
Remove a redundant 'valid' return parameter from functions that also return an error message when the result is not valid and return None otherwise.

### DIFF
--- a/keylime/cloud_verifier_common.py
+++ b/keylime/cloud_verifier_common.py
@@ -313,14 +313,8 @@ def validate_ima_policy_data(agent_data):
     lists = json.loads(agent_data)
 
     # Validate exlude list contains valid regular expressions
-    is_valid, _, err_msg_from_validator = validators.valid_exclude_list(lists.get("exclude"))
-    if not is_valid:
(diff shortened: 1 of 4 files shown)


keylime/cloud_verifier_common.py
```diff
@@ -313,14 +313,8 @@ def validate_ima_policy_data(agent_data):
...
-            err_msg_base = ""
-        err_msg = err_msg_base + "Exclude list regex is misformatted. Please correct the issue and try again."
-    else:
-        err_msg = err_msg_from_validator
+    _, err_msg_from_validator = validators.valid_exclude_list(lists.get("exclude"))
+    if err_msg_from_validator:
+        err_msg_from_validator += " Exclude list regex is misformatted. Please correct the issue and try again."
 
-    return is_valid, err_msg
+    return err_msg_from_validator is None, err_msg_from_validator
```
stefanberger, line 320: This would be yet another redundancy to fix.

THS-on (approved):

aplanas (approved):

PR #1186, opened 2022-11-16
ima_dm: add type checks and hints

THS-on:
This adds type hints to all of the functions in ima_dm. I'm currently waiting for https://github.com/keylime/keylime/pull/1180 to finalize this PR.

### DIFF
--- a/keylime/ima/ima_dm.py
+++ b/keylime/ima/ima_dm.py
@@ -9,10 +9,12 @@ import re
 import sys
 from collections import ChainMap
 from itertools import chain
-from typing import Dict, List, Optional, Union
+from typing import Any, Callable, Dict, List, Optional, Union
 
 import lark
-from lark import Lark, Transformer, v_args
+from lark.exceptions import LarkError
+from lark.lark import Lark
+from lark.visitors import Transformer, v_args
 from packaging import version
 
 from keylime.common.algorithms import Hash
@@ -25,26 +27,73 @@ if sys.version_info >= (3, 7):
 else:
     from keylime.backport_dataclasses import dataclass
 
+if sys.version_info >= (3, 8):
+    from typing import Literal, TypedDict
+else:
+    from typing_extensions import Literal, TypedDict
 
-class DeepChainMap(ChainMap):
+
+class DeepChainMap(ChainMap):  # type: ignore[type-arg]
     """
     Variant of ChainMap that handles updates for nested ChainMaps
     Based on the example in https://docs.python.org/3/library/collections.html#collections.ChainMap
     """
 
-    def __setitem__(self, key, value):
+    def __setitem__(self, key: Any, value: Any) -> None:
         for mapping in self.maps:
             if key in mapping:
                 mapping[key] = value
                 return
         super().__setitem__(key, value)
 
-    def __delitem__(self, key):
+    def __delitem__(self, key: Any) -> None:
         for mapping in self.maps:
             if key in mapping:
                 del mapping[key]
                 return
         raise KeyError(key)
+
+
+RuleAttributeType = Optional[Union[int, str, bool]]
+
+# Special Type for the match key.
+# Note that mypy currently does not detect that the variable is a literal,
+# so we still need to add a type ignore
+MatchKeyType = Literal["name", "uuid"]
+
+
+class DeviceRenameRule(TypedDict):
+    valid_name: RuleAttributeType
+    valid_uuid: RuleAttributeType
+
+
+class DeviceRemoveRule(TypedDict):
+    allow_removal: bool
+
+
+class TableLoadRule(TypedDict):
+    allow_multiple_loads: bool
+    name: str
+    uuid: str
+    major: int
+    minor: int
+    minor_count: int
+    num_targets: int
+    targets: List[Dict[str, RuleAttributeType]]
+
+
+class Rule(TypedDict):
+    required: bool
+    device_resume_required: bool
+    device_rename: DeviceRenameRule
+    device_remove: DeviceRemoveRule
+    allow_clear: bool
+    table_load: TableLoadRule
+
+
+class Policies(TypedDict):
+    match_on: MatchKeyType
+    rules: Dict[str, Rule]
 
 
 @dataclass
@@ -67,10 +116,10 @@ class DmIMAValidator:
         "dm_device_rename",
         "dm_target_update",
     ]
-    policies: dict
+    policies: Policies
     devices: Dict[str, DeviceState]
 
-    def __init__(self, policies) -> None:
+    def __init__(self, policies: Policies) -> None:
         self.policies = policies
         self.devices = {}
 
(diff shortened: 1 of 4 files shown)


keylime/ima/ima_dm.py
```diff
@@ -7,9 +7,10 @@
 import pickle
 import re
 import sys
+import typing
 from collections import ChainMap
 from itertools import chain
-from typing import Dict, List, Optional, Union
+from typing import Any, Callable, Dict, List, Optional, TypeAlias, TypedDict, Union
```
aplanas, line 13:
```
from typing import Any, Callable, Dict, List, Optional, TypeAlias, TypedDict, Union
ImportError: cannot import name 'TypeAlias' from 'typing' (/usr/lib64/python3.9/typing.py)
```

`New in version 3.10.` according to doc

THS-on: Yeah I missed that this feature is only supported in Python 3.10. Removed it for now.

THS-on:
`TypedDict` is also a Python 3.8 feature. How well is the `typing-extensions` package supported in distributions? 
@aplanas do have any SLES version that uses python 3.7?

aplanas:
> `TypedDict` is also a Python 3.8 feature. How well is the `typing-extensions` package supported in distributions? @aplanas do have any SLES version that uses python 3.7

Right. For the next version we will have 3.10, and `typing-extensions` is already present.

THS-on:
@kkaarreell @aplanas @ansasaki  what option do you prefer:

* Keep the patch as it is and add `typing-extensions` as a dependency
* Ship a backport of the TypedDict in Keylime
* Use a generic Dict type annotation

kkaarreell:
We are not shipping keylime for RHEL-8 so if installing python-typing-extensions would do the job then we can install it from EPEL8. You can test that by installing the package in an extra step after
https://github.com/keylime/keylime/blob/master/packit-ci.fmf#L67
We should probably also discuss if there is still a value in running tests on C8S.

aplanas:
> If we want to support Python 3.6 still, probably yes otherwise not really.

Do we have a process to drop old Python versions? For example, release a Keylime 7 based on Python 3.10 (that has almost 4 years of live ahead[1])

With a process like that we can pin in the enterprise distro the last supported Keylime branch (6.5 in this case).

A process like that will open the question about how many Keylime branches are supported by the project, so when a security fix happens in Keylime N, it is also backported to the N-1, N-2 ... branches.

[1] https://endoflife.date/python

THS-on:
> Do we have a process to drop old Python versions?

Not yet, I think it is probably a good idea to raise this in the Slack channel.

keylime/ima/ima_dm.py
```diff
@@ -904,11 +959,24 @@ def mirror_mirror_device_data(children):
...
+        ("dm_table_load", LoadEvent),
+        ("dm_device_resume", ResumeEvent),
+        ("dm_device_remove", RemoveEvent),
+        ("dm_table_clear", ClearEvent),
+        ("dm_device_rename", RenameEvent),
+        ("dm_target_update", UpdateEvent),
+    ]:
+        raise TypeError(f"{event} was parsed as: {type(event)}")
+
+    assert isinstance(out, (LoadEvent, ResumeEvent, RemoveEvent, ClearEvent, RenameEvent, UpdateEvent))
```
stefanberger, line 981: The above should have this already covered...

THS-on: Yes, but unfortunately mypy does not find it. I could use cast, but I don't know if this is the better solution.

stefanberger: Could you move the type checking code into a function of its own and also apply it to those checks in `validate()`? I think that function could be called from both places? As long as pyright/mypy is happy afterwards...

THS-on: @stefanberger I'll try to find a way to make this more elegant. The issue I don't think there is an elegant way to associate input values with the output type. @aplanas do you have an idea to solve this?

aplanas: No idea!

Only a suggestion: if we create a function `def _from_str_to_even(event: str) -> EventTypes:` could this help?

THS-on: It is probably a good idea to create that function, but I don't think we can get rid of the asserts in the validate function...

stefanberger:
I am also running into the missing `TypedDict` issue: pr #1254 

One way to work around this would be to try to import TypedDict and if it works we could use it, otherwise define the new datatype to be the fallbacks of Dict[Any, Any] or Dict[str, Any] or similar. Once we get rid of older versions of python we could get rid of the try ... except sequences then. Though having this with later versions of python that we work with would really help.

keylime/ima/ima_dm.py
THS-on: @aplanas what do you think of the idea of introducing functions of the style `parse_LoadEvent(data: str) -> LoadEvent:`? The advantage is that we have type safety, the disadvantage is that we introduce 6 new functions that basically are doing the same thing.

THS-on: Ok, then we will keep the asserts for now. Maybe in the future we find a better way to solve this.

THS-on:
@kkaarreell it currently point to my fork of the tests that installs the typing-extensions module on RHEL. So when this gets merged this needs to be updated.

kkaarreell:
> @kkaarreell it currently point to my fork of the tests that installs the typing-extensions module on RHEL. So when this gets merged this needs to be updated.

python3-typing-extensions v3.7.4.3 are available in EPEL both for C8S and C9S. I think I would rather install it from EPEL than from pip. I will prepare PR to test it.

packit-ci.fmf
```diff
@@ -15,7 +15,7 @@
 
   discover:
     how: fmf
-    url: https://github.com/RedHat-SP-Security/keylime-tests
+    url: https://github.com/THS-on/keylime-tests
```
mpeters, line 18: What needs to change in the RH-SP-S tests for this to get merged?

mpeters: nvm

kkaarreell:
> > @kkaarreell it currently point to my fork of the tests that installs the typing-extensions module on RHEL. So when this gets merged this needs to be updated.
> 
> python3-typing-extensions v3.7.4.3 are available in EPEL both for C8S and C9S. I think I would rather install it from EPEL than from pip. I will prepare PR to test it.

In my PR https://github.com/RedHat-SP-Security/keylime-tests/pull/276 I can see the same failure on C8S as the one present here
```
  File "/usr/local/bin/keylime_tenant", line 11, in <module>
    load_entry_point('keylime==6.5.2', 'console_scripts', 'keylime_tenant')()
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 476, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2700, in load_entry_point
    return ep.load()
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2318, in load
    return self.resolve()
  File "/usr/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2324, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/usr/local/lib/python3.6/site-packages/keylime-6.5.2-py3.6.egg/keylime/cmd/tenant.py", line 3, in <module>
  File "/usr/local/lib/python3.6/site-packages/keylime-6.5.2-py3.6.egg/keylime/tenant.py", line 20, in <module>
    from keylime.cli import options, policies
  File "/usr/local/lib/python3.6/site-packages/keylime-6.5.2-py3.6.egg/keylime/cli/policies.py", line 4, in <module>
    from keylime.ima import file_signatures, ima
  File "/usr/local/lib/python3.6/site-packages/keylime-6.5.2-py3.6.egg/keylime/ima/ima.py", line 17, in <module>
    from keylime.ima import ast, file_signatures, ima_dm
  File "/usr/local/lib/python3.6/site-packages/keylime-6.5.2-py3.6.egg/keylime/ima/ima_dm.py", line 36, in <module>
    class DeepChainMap(ChainMap[Any, Any]):  # pylint: disable=unsubscriptable-object
TypeError: 'ABCMeta' object is not subscriptable
```

requirements.txt
```diff
@@ -20,3 +20,5 @@ pyasn1-modules>=0.2.1 # BSD
 jinja2>=3.0.0 # BSD
 # Keylime requires also gpg (LGPLv2+).
 # Those bindings must match the local GPG version and therefore this package is not installed via PyPI.
+# The typing-extensions module is currently required for Python versions < 3.8
```
kkaarreell, line 23: Does it mean it won't be needed for C9S with Python 3.9.14, just for C8S?

stefanberger: It depends on what features one wants from it.Some of the newer features are only available in 3.10 I read.

requirements.txt
```diff
@@ -20,3 +20,5 @@ pyasn1-modules>=0.2.1 # BSD
...
+typing-extensions>=3.7.4 #PSF-2.0
```
aplanas, line 24: Based on https://pip.pypa.io/en/stable/reference/requirement-specifiers/#examples

```
typing-extensions>=3.7.4 ; python_version < '3.8' #PSF-2.0
```

requirements.txt
THS-on: We are currently using Literal and TypedDict. If we use something else we can just bump the requirement.

requirements.txt
THS-on: Done

THS-on:
@kkaarreell this should be fixed now.

aplanas (approved):

THS-on:
Hmm now it does not find the module. I have no idea why, because it should be installed.

THS-on:
> Thank you @THS-on , tests are passing in my PR also for C8S.

Ok, I now dropped the commit that temporarily changed it to my branch.

kkaarreell:
> > Thank you @THS-on , tests are passing in my PR also for C8S.
> 
> Ok, I now dropped the commit that temporarily changed it to my branch.

I have merged my PR so you can rerun Packit tests against the main branch.

mpeters (approved):

PR #1189, opened 2022-11-16
Have the Hash class raise a ValueError also on unsupported algorithms

stefanberger:
When trying to instantiate the Hash class using a call like Hash("sha1") one would get a Hash object back. If one was to instantiate Hash("bad"), it would raise a ValueError. With this change the constructor will also raise a ValueError if the hash is not supported by the underlying crypto library, such as may be the case for sm3_256. Therefore, no object would exist for sm3_256 and we wouldn't be able to call the get_size() method that previously also could return None (in case sm3_256 wasn't supported by the crypto library).

### DIFF
--- a/keylime/common/algorithms.py
+++ b/keylime/common/algorithms.py
@@ -1,6 +1,6 @@
 import enum
 import hashlib
-from typing import Any, List, Optional
+from typing import Any, List
 
 
 def is_accepted(algorithm: str, accepted: List[Any]) -> bool:
@@ -12,12 +12,33 @@ def is_accepted(algorithm: str, accepted: List[Any]) -> bool:
     return algorithm in accepted
 
 
+def _hashit(algorithm: str, data: bytes) -> bytes:
+    if algorithm == "sha1":
+        return hashlib.sha1(data).digest()
+    if algorithm == "sha256":
+        return hashlib.sha256(data).digest()
+    if algorithm == "sha384":
+        return hashlib.sha384(data).digest()
+    if algorithm == "sha512":
+        return hashlib.sha512(data).digest()
+    if algorithm == "sm3_256":
+        # SM3 is not guaranteed to be there
+        return hashlib.new("sm3", data).digest()
+
+    raise ValueError(f"Unsupported hash algorithm {algorithm}")
+
+
 class Hash(str, enum.Enum):
     SHA1 = "sha1"
     SHA256 = "sha256"
     SHA384 = "sha384"
     SHA512 = "sha512"
     SM3_256 = "sm3_256"
+
+    def __init__(self, *args):  # pylint: disable=unused-argument
+        super().__init__()
+        # Test hash to raise ValueError for unsupported hashes
+        _hashit(self.value, b"")
 
     @staticmethod
     def is_recognized(algorithm: str) -> bool:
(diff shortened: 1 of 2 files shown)


stefanberger:
> Change LGTM. Can you fix the style check by running black or replacing the single quotes with double qoutes?

Yes, right... done

stefanberger:
I need to add another method to the class `is_supported()` because `is_recognized()` is not good enough for this here or adapt `Is_recognized()` to return false if the hash is not supported:

```
./keylime/tpm/tpm_main.py:119:            elif details["hash"] == 1 and algorithms.Hash.is_recognized(algorithm):
./keylime/tenant.py:1005:        ) or not algorithms.Hash.is_recognized(hash_alg):
./keylime/cloud_verifier_common.py:100    if not algorithms.is_accepted(hash_alg, agent["accept_tpm_hash_algs"]) or not algorithms.Hash.is_recognized(
```

stefanberger:
`is_recognized` already filters. So this PR should be good.

THS-on (approved):

keylime/common/algorithms.py
```diff
@@ -12,13 +12,34 @@ def is_accepted(algorithm: str, accepted: List[Any]) -> bool:
     return algorithm in accepted
 
 
+def _hashit(algorithm: str, data: bytes) -> bytes:
```
aplanas, line 15: I do not think that this should be extracted from the class, and the main reason is that it is making worse a problem that the old class has: the list of supported hashes: is it replicated all over the code

keylime/common/algorithms.py
```diff
@@ -12,13 +12,34 @@ def is_accepted(algorithm: str, accepted: List[Any]) -> bool:
...
+    if algorithm == "sha1":
```
aplanas, line 16: One step in the right direction should use `Hash.SHA1`.

But ideally there will be a hidden table with a mapping between the hash name, the hash ID and the hash function, so adding or removing a hash family is editing this table.  The rest of the operation are like `if algorithm in self.supported_hash`, or `return self.supported_hash[algorithm](data).digest()`

keylime/common/algorithms_test.py
```diff
@@ -0,0 +1,29 @@
+import unittest
+
+from keylime.common.algorithms import Hash
+
+
+class TestHash(unittest.TestCase):
```
aplanas, line 6: Thanks a lot!

keylime/common/algorithms.py
stefanberger: I had to extract it to be able to call this function from the constructor since I could neither call a @staticfunction nor a @classsfunction in the constructor. Nobody should use this function directly.

keylime/common/algorithms.py
stefanberger: Again, this is a consequence of having to extract the function from the class to be able to call it from the constructor. I didn't like it either, but if you can fix it in another way...

keylime/common/algorithms.py
stefanberger: Actually, that was wrong. I can revert the extraction. PR soon.

What does not work is using Hash.SHA1 etc when calling the private function from the constructor:

```
  File "/home/stefanb/keylime/keylime/common/algorithms.py", line 36, in __hash
    if self.value == Hash.SHA1:
```

aplanas: I will try to move a bit in this direction:  https://github.com/keylime/keylime/pull/1191

The idea is that to add a new hash algorithm, only the enum needs to be extended

PR #1190, opened 2022-11-17
common: Move _hashit function back into Hash class

stefanberger:
Move the _hashit function back into the Hash class as hash().


stefanberger:
@aplanas Here's the partial revert.

PR #1191, opened 2022-11-17
algorithms: simplify the Hash class

aplanas:


keylime/common/algorithms.py
```diff
@@ -12,47 +12,30 @@ def is_accepted(algorithm: str, accepted: List[Any]) -> bool:
...
         try:
             Hash(algorithm)
-            return True
         except ValueError:
             return False
+        return True
 
     def hash(self, data: bytes) -> bytes:
-        return _hashit(self.value, data)
+        return hashlib.new(self.value, data).digest()
```
stefanberger, line 35: Does this work for `sm3` ? It looks like it would need "sm3" string but we seem to have the sm3_256 string. 

Also, do we still get a value error when someone creates an sm3_256 hash but the underlying crypto library doesn't support it? We don't want this to happen nor is_recognized() to pass and fail much later in the code on hashing.

keylime/common/algorithms.py
```diff
@@ -12,47 +12,30 @@ def is_accepted(algorithm: str, accepted: List[Any]) -> bool:
...
-    raise ValueError(f"Unsupported hash algorithm {algorithm}")
-
-
 class Hash(str, enum.Enum):
     SHA1 = "sha1"
     SHA256 = "sha256"
     SHA384 = "sha384"
     SHA512 = "sha512"
-    SM3_256 = "sm3_256"
+    SM3_256 = "sm3"
```
stefanberger, line 20: At least now it's the same as what the Linux kernel uses: https://elixir.bootlin.com/linux/latest/source/crypto/hash_info.c#L29

I wonder whether this change now will make a difference for anyone previously having used Keylime with `sm3_256` string "somewhere" in the config.

keylime/common/algorithms.py
```diff
@@ -12,47 +12,30 @@ def is_accepted(algorithm: str, accepted: List[Any]) -> bool:
...
 
-    def __init__(self, *args):  # pylint: disable=unused-argument
-        super().__init__()
-        # Test hash to raise ValueError for unsupported hashes
-        _hashit(self.value, b"")
+    @classmethod
+    def from_algorithm(cls, algorithm: str) -> "Hash":
+        return cls(algorithm)
```
stefanberger, line 24: So you don't want to also break right in the constructor when someone grabs the Hash("sm3") and the underlying crypto lib doesn't support it? Hash("bad") throws a ValueError and Hash("sm3") should throw one too IMO if the crypto lib doesn't support it because it looks like the usage of this class is to call `is_recognized()` first to see whether it will work.

keylime/common/algorithms.py
aplanas: Right. A `grep` did not show anything in the codebase, and if it comes from a configuration file it will trigger now the ValueError exception (and the fix should be directy)

I was inclined to remove this `sm3` algorithm, or extend it with `sm4` too.

THS-on: The name `sm3_256` is also used by tpm2_tools and we pass the string value to it, so we cannot just change it to `sm3`

stefanberger: Worth writing this into the file...

aplanas: Reverted the name and add some documentation

keylime/common/algorithms.py
```diff
@@ -12,47 +12,34 @@ def is_accepted(algorithm: str, accepted: List[Any]) -> bool:
...
         try:
-            Hash(algorithm)
-            return True
+            Hash(algorithm).get_size()
         except ValueError:
             return False
+        return True
+
+    def __hashfn(self, data: bytes) -> Any:
+        return hashlib.new(self.value, data)
```
THS-on, line 36: This will fail with `sm3_256` because in openssl it is only called `sm3`.

aplanas: Right, I extracted the function just to address this issue : )

Done!

THS-on (approved):

PR #1192, opened 2022-11-17
Some fixes to validate_ima_policy_data

stefanberger:
This PR fixes a recently introduced bug (by me) where the return parameters of the function were adjusted to only return an error message instead of the redundant is_valid parameter.

### DIFF
--- a/keylime/cloud_verifier_common.py
+++ b/keylime/cloud_verifier_common.py
@@ -305,12 +305,12 @@ def prepare_error(agent, msgtype="revocation", event=None):
     return tosend
 
 
-def validate_ima_policy_data(agent_data):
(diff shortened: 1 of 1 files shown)


aplanas (approved):

THS-on (approved):

maugustosilva:
Hmmm... I've ended up with a somewhat confusing commit message while squashing. Should I revert @mpeters ?

mpeters:
It seems a bit redundant and long, but not inaccurate, right? I'm inclined to just leave it.

Btw, we shouldn't ever do a force push on the main branch as that will mess up everyone who has a fork or checkout.

PR #1195, opened 2022-11-17
Run 'pyright' and 'black --diff' as part of make check

stefanberger:
This PR extends tox to run 'pyright' and 'black --diff' when running 'make check'

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 .PHONY: check check-rebuild
 check:
-	tox -vv -epylint -epyright -eblack -eisort
+	tox -epylint -epyright -eblack -eisort
 
 check-rebuild:
 	tox -r -vv -epylint -epyright -eblack -eisort
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 .PHONY: check check-rebuild
 check:
-	tox -vv -epylint -epyright
+	tox -vv -epylint -epyright -eblack
 
 check-rebuild:
-	tox -r -vv -epylint -epyright
+	tox -r -vv -epylint -epyright -eblack
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 .PHONY: check check-rebuild
 check:
-	tox -vv -epylint
+	tox -vv -epylint -epyright
 
 check-rebuild:
-	tox -r -vv -epylint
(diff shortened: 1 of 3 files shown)


THS-on:
We currently run black and isort using pre-commit and pylint using tox (make check). 
pre-commit allows us to enforce the style checks pretty easily and that everyone uses the same version, the question is on how to keep that in sync with if we add the other ones to the make check script.

THS-on:
I think adding pyright to the make check script is fine.

stefanberger:
> We currently run black and isort using pre-commit and pylint using tox (make check).
> pre-commit allows us to enforce the style checks pretty easily and that everyone uses the same version, the question is on how to keep that in sync with if we add the other ones to the make check script.

I run `make check` locally all the time and it would be good to have all these checks run in one command before pushing.

When pyright runs via tox it adheres to the file list from pyproject. So maybe the script/check_codestyle.sh could go away if we can pass a similar list of files to pylint via pyproject? Would that help?

THS-on:
> I run make check locally all the time and it would be good to have all these checks run in one command before pushing.

I agree. Maybe we can do the following:
* add isort and black to tox
* move the pre-commit hooks to use the tox environment. I would only run isort and tox because pylint is pretty slow. (I heavily use the pre-commit hooks)
* find a way to pin versions and conditionally rebuild the tox environment

> So maybe the script/check_codestyle.sh could go away if we can pass a similar list of files to pylint via pyproject? Would that help?

That would be nice to have all the ignored files only in the pyproject configuration.

stefanberger:
> That would be nice to have all the ignored files only in the pyproject configuration.

- pylint doesn't support pyproject from what I could find. If someone wants to clean more files for pylint he just needs to go into scipts/check_codestyle.sh. Config is in pylintrc. So one just has to know where to look for pylint but there's no duplication of config or so.
- pyright supports pyproject and we have all the filtered-out files there. No duplication in config.
- black is just a simple command line invocation
- isort is also just a simple command line invocation. When run from tox it does look into pyproject file and should take profile and line-length from there.

I have `make check` make no modifications to files but only report the errors, primarily of course for black and isort that could do in-place changes. So, CI/CD could just run `make check` as well.

THS-on (approved):

PR #1196, opened 2022-11-21
pyright: Fix a 2nd batch of issues reported by pyright

stefanberger:
This PR fixes a 2nd batch of issues reported by pyright:

- Raise a ValueError on unsupported key types so that applications using a non-RSA or non-ECDSA don't use the key; this helps to subsequently add restrictive type annotations to ima/file_signatures.py
- Add type annotations to ima/file_signatures.py
- Add type annotations to cmd/convert_ima_policy.py
- Enable pyright checking on files that are clean

### DIFF
--- a/keylime/cli/policies.py
+++ b/keylime/cli/policies.py
@@ -46,7 +46,10 @@ def process_allowlist(args):
         # Add all IMA file signing verification keys to a keyring
         tenant_keyring = file_signatures.ImaKeyring()
         for filename in args["ima_sign_verification_keys"]:
-            pubkey, keyidv2 = file_signatures.get_pubkey_from_file(filename)
+            try:
+                pubkey, keyidv2 = file_signatures.get_pubkey_from_file(filename)
+            except ValueError as e:
+                raise UserError(f"File '{filename}' does not have a supported key: {e}") from e
             if not pubkey:
                 raise UserError(f"File '{filename}' is not a file with a key")
             tenant_keyring.add_pubkey(pubkey, keyidv2)
--- a/keylime/cmd/convert_ima_policy.py
+++ b/keylime/cmd/convert_ima_policy.py
@@ -140,10 +140,13 @@ def update_ima_policy(policy, excludelist_path=None, verification_keys=None):
     if verification_keys:
         keyring = file_signatures.ImaKeyring().from_string(policy["ima_sign_verification_keys"])
         for key in verification_keys:
-            pubkey, keyidv2 = file_signatures.get_pubkey_from_file(key)
-            if not pubkey:
-                print(f"File '{key}' is not a file with a key")
-            keyring.add_pubkey(pubkey, keyidv2)
+            try:
+                pubkey, keyidv2 = file_signatures.get_pubkey_from_file(key)
+                if not pubkey:
+                    print(f"File '{key}' is not a file with a key")
+                keyring.add_pubkey(pubkey, keyidv2)
+            except ValueError as e:
+                print(f"File '{key}' does not have a supported key: {e}")
         verification_key_list = json.dumps(keyring.to_string())
 
     policy["excludes"] += excl_list
--- a/keylime/cmd/convert_ima_policy.py
+++ b/keylime/cmd/convert_ima_policy.py
@@ -4,8 +4,11 @@ import datetime
 import json
 import sys
 from os.path import basename
+from typing import Any, Dict, List, Optional
 
 from keylime.ima import file_signatures, ima
+
+PolicyDict = Dict[str, Any]
 
 # pylint: disable=pointless-string-statement
 """
@@ -35,7 +38,7 @@ python convert_ima_policy.py -h
 """
 
 # Creates an IMA policy from provided legacy allowlist.
-def convert_legacy_allowlist(allowlist_path):
+def convert_legacy_allowlist(allowlist_path: str) -> PolicyDict:
     with open(allowlist_path, "r", encoding="utf8") as f:
         alist_raw = f.read()
 
@@ -53,7 +56,7 @@ def convert_legacy_allowlist(allowlist_path):
 
 
 # Converts JSON-format allowlist to JSON-format IMA policy
-def _convert_json_allowlist(alist_json):
+def _convert_json_allowlist(alist_json: PolicyDict) -> PolicyDict:
     ima_policy = copy.deepcopy(ima.EMPTY_IMA_POLICY)
     ima_policy["meta"]["timestamp"] = str(datetime.datetime.now())
     ima_policy["meta"]["generator"] = "convert_ima_policy.py"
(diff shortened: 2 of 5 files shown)


keylime/cli/policies.py
```diff
@@ -46,7 +46,10 @@ def process_allowlist(args):
         # Add all IMA file signing verification keys to a keyring
         tenant_keyring = file_signatures.ImaKeyring()
         for filename in args["ima_sign_verification_keys"]:
-            pubkey, keyidv2 = file_signatures.get_pubkey_from_file(filename)
+            try:
+                pubkey, keyidv2 = file_signatures.get_pubkey_from_file(filename)
+            except ValueError as e:
+                raise UserError(f"File '{filename}' does not have a supported key: %{e}") from e
```
aplanas, line 52: `%`?

keylime/cmd/convert_ima_policy.py
```diff
@@ -109,12 +110,14 @@ def _convert_flat_format_allowlist(alist_raw):
...
-def update_ima_policy(policy, excludelist_path=None, verification_keys=None):
+def update_ima_policy(
+    policy: dict, excludelist_path: Optional[str] = None, verification_keys: Optional[list[str]] = None
+) -> dict:
     if policy["meta"]["version"] < ima.IMA_POLICY_CURRENT_VERSION:
         print(
             f"Provided policy has version {policy['meta']['version']}; latest policy has version {ima.IMA_POLICY_CURRENT_VERSION}. Updating to latest version."
         )
-        updated_policy = copy.deepcopy(ima.IMA_POLICY_CURRENT_VERSION)
+        updated_policy: dict = copy.deepcopy(policy)
```
aplanas, line 120: IIUC it should be `ima.IMA_POLICY_CURRENT_VERSION`.  Seems to be that the idea is to iterate over the expected keys (from the default template), and build from there.

keylime/cmd/convert_ima_policy.py
```diff
@@ -180,6 +190,12 @@ def main():
     elif args.ima_policy:
         with open(args.ima_policy, "r", encoding="utf8") as f:
             policy = json.load(f)
+        if not isinstance(policy, dict):
```
aplanas, line 193: Can this happen?

stefanberger: It can probably happen if you construct a file that contains maybe a list for example.

keylime/cmd/convert_ima_policy.py
stefanberger: It looked wrong to me and pyright  says:
```
/home/stefanb/keylime/keylime/cmd/convert_ima_policy.py
  /home/stefanb/keylime/keylime/cmd/convert_ima_policy.py:120:46 - error: Argument of type "Literal[1]" cannot be assigned to parameter "x" of type "_T@deepcopy" in function "deepcopy"
    Type "int" cannot be assigned to type "dict[Unknown, Unknown]"
      "int" is incompatible with "dict[Unknown, Unknown]" (reportGeneralTypeIssues)
```

keylime/cmd/convert_ima_policy.py
aplanas: Indeed!

keylime/cli/policies.py
stefanberger: Thanks.

keylime/cmd/convert_ima_policy.py
aplanas: I see, maybe was `ima.EMPTY_IMA_POLICY` at some time. I am pretty sure that it should not be `policy`, that is the object that we are transforming. 

cc: @mbestavros

THS-on: If you look at line 130 there is where the copying happens, so I think @aplanas is right. I think that line 130 this should be a deep copy and we need to make sure that we return a fully copy of the policy and do not modify the old once accidentally.

stefanberger: Line 130 looks good to me. I agree though that the above then should be `copy.deepcopy(ima.EMPTY_IMA_POLICY)`

THS-on: If `policy[key]` is a dict we only copy the reference. I don't know if that is intentional.

mbestavros: @aplanas @stefanberger @THS-on That line was definitely a mistake - it's supposed to deepcopy `ima.EMPTY_IMA_POLICY`. The `IMA_POLICY_CURRENT_VERSION` field is just a version number, hence the type error.

keylime/cmd/convert_ima_policy.py
```diff
@@ -180,6 +190,12 @@ def main():
...
+            print(f"The policy in file {args.ima_policy} must be a dictionary")
+            sys.exit(1)
+    else:
+        policy = {}  # pyright
```
mbestavros, line 197: Maybe this should be another `copy.deepcopy(ima.EMPTY_IMA_POLICY)` instead? That way we always end up with a valid-looking policy instead of an empty list.

mbestavros (approved):
The IMA policy conversion script changes seem mostly common-sense to me. I'll defer judgment on `file_signatures.py` to someone with more knowledge of that part of the code.

keylime/cmd/convert_ima_policy.py
stefanberger: Let me actually make this an assert(False). This case cannot happen due to this check here:

```
    if (bool(args.allowlist) and not bool(args.ima_policy)) and (not bool(args.allowlist) and bool(args.ima_policy)):
        print("Either --allowlist or --ima_policy is required!")
        sys.exit(1)
```

mbestavros: Yep, I think that was intended when I wrote the code. If there's a more efficient way to evaluate an xor over those two arguments, I welcome it :)

stefanberger: I fixed it now... I think, along with adding a case with the deepcopy.

mbestavros: > So maybe we do need to force the user to provide either allowlist or ima_policy but not both.

That was my intention when I xor'd `args.allowlist` and `args.ima_policy` as a check.

aplanas (approved):

THS-on:
@stefanberger can you rebase on the latest master?

stefanberger:
> @stefanberger can you rebase on the latest master?

Done. Also revised some `Dict`'s to make them more concrete...

keylime/ima/file_signatures.py
```diff
@@ -125,7 +130,7 @@ def _get_keyidv2(pubkey):
         keydigest = digest.finalize()
         return int.from_bytes(keydigest[16:], "big")
 
-    def add_pubkey(self, pubkey, keyidv2):
+    def add_pubkey(self, pubkey: SupportedKeyTypes, keyidv2: Union[int, None]) -> None:
```
THS-on, line 133: Instead of `Union[int, None]` you can use `Optional[int]`. This can be changed in a few places.

THS-on (changes requested):
One small thin regarding type hints. The rest LGTM.

keylime/ima/file_signatures.py
stefanberger: Done.

THS-on (approved):

PR #1198, opened 2022-11-21
Calculates the checksum for the whole IMA policy on the verifier

maugustosilva:
A previous PR (#1181) introduced `checksum` as a separate column on the DB for the `allowlists` table. The initial implementation focused on the calculating the checksum for the allowlist, piggybacking already existing code on the `tenant`. However, this is not enough to detect full changes on the "IMA policy" (which will cover also the "exclude list"). This PR now ensures that a checksum is calculated for the whole policy, on the `verifier`.

### DIFF
--- a/keylime/ima/ima.py
+++ b/keylime/ima/ima.py
@@ -42,7 +42,7 @@ EMPTY_ALLOWLIST = {
     "meta": {
         "version": ALLOWLIST_CURRENT_VERSION,
         "generator": IMA_POLICY_GENERATOR.EmptyAllowList,
-        "checksum": "1ab36a57ad793caba756b01fc142f296cbfc3d24088fe6d66448f3bb6ebadeee",
+        "checksum": "",
     },
     "release": 0,
     "hashes": {},
@@ -638,18 +638,15 @@ def ima_policy_db_contents(ima_policy_name: str, ima_policy: str, tpm_policy: st
         ima_policy_db_format["ima_policy"] = None
     else:
         ima_policy_db_format["ima_policy"] = ima_policy
-    ima_policy_json = json.loads(ima_policy)
-    if "allowlist" in ima_policy_json:
-        if "meta" in ima_policy_json["allowlist"]:
-            if "checksum" in ima_policy_json["allowlist"]["meta"]:
-                ima_policy_db_format["checksum"] = ima_policy_json["allowlist"]["meta"]["checksum"]
-            else:
-                alist_bytes = json.dumps(copy.deepcopy(EMPTY_ALLOWLIST)).encode()
-                sha256 = hashlib.sha256()
-                sha256.update(alist_bytes)
-                ima_policy_db_format["checksum"] = sha256.hexdigest()
-            if "generator" in ima_policy_json["allowlist"]["meta"]:
-                ima_policy_db_format["generator"] = ima_policy_json["allowlist"]["meta"]["generator"]
+
+    ima_policy_bytes = ima_policy.encode()
+    ima_policy_db_format["checksum"] = hashlib.sha256(ima_policy_bytes).hexdigest()
+
+    ima_policy_dict = json.loads(ima_policy)
+    if "allowlist" in ima_policy_dict:
+        if "meta" in ima_policy_dict["allowlist"]:
+            if "generator" in ima_policy_dict["allowlist"]["meta"]:
+                ima_policy_db_format["generator"] = ima_policy_dict["allowlist"]["meta"]["generator"]
             else:
                 ima_policy_db_format["generator"] = IMA_POLICY_GENERATOR.Unknown
 


keylime/ima/ima.py
```diff
@@ -33,7 +33,6 @@
 EMPTY_ALLOWLIST = {
     "meta": {
         "version": ALLOWLIST_CURRENT_VERSION,
-        "checksum": "1ab36a57ad793caba756b01fc142f296cbfc3d24088fe6d66448f3bb6ebadeee",
```
aplanas, line 36: I think that makes sense to leave the field empty: `"checksum": "",`

This can help documentation (expected schema)

keylime/ima/ima.py
```diff
@@ -624,16 +623,12 @@ def ima_policy_db_contents(ima_policy_name: str, ima_policy: str, tpm_policy: st
...
-            else:
-                alist_bytes = json.dumps(copy.deepcopy(EMPTY_ALLOWLIST)).encode()
-                sha256 = hashlib.sha256()
-                sha256.update(alist_bytes)
-                ima_policy_db_format["checksum"] = sha256.hexdigest()
+
+    ima_policy_bytes = ima_policy.encode()
+    sha256 = hashlib.sha256()
+    sha256.update(ima_policy_bytes)
+    ima_policy_db_format["checksum"] = sha256.hexdigest()
```
aplanas, line 630:
```suggestion
    ima_policy_bytes = ima_policy.encode()
    ima_policy_db_format["checksum"] = hashlib.sha256(ima_policy_bytes).hexdigest()
```

keylime/ima/ima.py
aplanas: So there is this tool, cmd/convert_ima_policy.py, that copy this `dict` and transform it. As today `checksum` is not touched, but seems to be that uses this as a template, and I guess that a correct transformation can touch this parameter in the future.

keylime/ima/ima.py
```diff
@@ -624,16 +623,12 @@ def ima_policy_db_contents(ima_policy_name: str, ima_policy: str, tpm_policy: st
...
-            else:
-                alist_bytes = json.dumps(copy.deepcopy(EMPTY_ALLOWLIST)).encode()
-                sha256 = hashlib.sha256()
-                sha256.update(alist_bytes)
-                ima_policy_db_format["checksum"] = sha256.hexdigest()
+
+    alist_bytes = ima_policy.encode()
+    sha256 = hashlib.sha256()
+    sha256.update(alist_bytes)
+    ima_policy_db_format["checksum"] = sha256.hexdigest()
```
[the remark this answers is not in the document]
maugustosilva: done

keylime/ima/ima.py
maugustosilva: done

maugustosilva:
FYI @stefanberger This PR should make sure that **any** changes on IMA policy get properly detected by calculating a new checksum at the `verifier` side.

stefanberger:
Per my latest changed to PR #1177 the DB Proxy is not polling on the database for changes to an entry anymore but is now completely driven by notifications when updates to the allowlists table are done in the verifier. So at least from this perspective we wouldn't need a change to the checksum anymore.

galmasi:
@stefanberger _without having looked at your code_, I am suspicious about relying on notifications. Because in our scaled deployments there are many verifiers, and only one of them handles allowlist changes; but all the others need to be notified about such changes. IMHO communicating through the database is inevitable.

That being said, I may be wrong about your implementation -- please clarify.

galmasi:
And just to announce it here, this PR has been tested in isolation on the scale cluster, and is quite effective in actually reducing bandwidth to the SQL database. By itself it does not resolve the entire problem, of course, but it at least allows status and bulk status requests to avoid overloading the DB.

maugustosilva:
Just a quick clarification to @galmasi 's comment: what we tested at scale was a combination of this PR (#1198) plus #1200

keylime/ima/ima.py
mbestavros: We should not be putting `checksum` within the policy itself, IMO. Editing policies after creation to add the checksum in will break signing/verifying steps - it's crucial that the file provided to Keylime at the beginning remains the same, and is never modified.

If we want to add checksums, it should happen entirely on the verifier.

stefanberger:
> @stefanberger _without having looked at your code_, I am suspicious about relying on notifications. Because in our scaled deployments there are many verifiers, and only one of them handles allowlist changes; but all the others need to be notified about such changes. IMHO communicating through the database is inevitable.

Notifications correspond to interrupts, which is better than polling.

How did you guys solve this now? IMO we would need some sort of messaging system that verifiers are connected to and get notifications about changes to the allowlists/ima_policy because otherwise the only way they find out about updates in the database is by continuously polling the DB, which I would think isn't good, either.

So I guess there setups that run multiple verifiers behind a load balancer already?

keylime/ima/ima.py
mbestavros: @aplanas Because of the above reasoning, the IMA policy overhaul will not use the `checksum` field in the future.

galmasi:
@stefanberger the current baked in assumption in keylime is that verifiers communicate through a shared mysql database and **nothing else**. Communicating through mysql is not by itself a scaling problem. If anything, the database server _is_ the scale proof backbone of the system; inventing anything else to do its job is just a bad idea for a million different reasons, including but not limited to complexity and technical debt.

keylime/ima/ima.py
maugustosilva: Yes, there are **two** `checksums`. Regrettably, my previous PR #1181 was conflating the two, but this is now fixed on this one. The checksum that matters is the one calculated **entirely on the verifier**

maugustosilva: So, if IMA policy overhaul stops using the `checksum` key within `meta`, it will **not** affect anything on this PR.

stefanberger:
> Communicating through mysql is not by itself a scaling problem. If anything, the database server _is_ the scale proof backbone of the system; inventing anything else to do its job is just a bad idea for a million different reasons, including but not 

One can relief (even more) stress from the DB by avoiding having to poll the DB for updates but that requires a notification system when the updates are happening, which in turn does require additional architectural support to notify all those that must know about updates to a table. The 'additional architectural support' applies only IFF there are indeed setups with multiple clients (verifiers) using a single database.

> limited to complexity and technical debt.

Possibly, yes.

mpeters:
For indexed based reads mysql is extremely fast and scalable. I would much rather use it than try to implement our own notification/cached based system until proven otherwise.

stefanberger:
> For indexed based reads mysql is extremely fast and scalable. I would much rather use it than try to implement our own notification/cached based system until proven otherwise.

@mpeters All I can do is point you to my implementation in PR #1177 that tries to avoid the polling and relies on caching and notification but doesn't handle setups where multiple verifiers use one DB (which I don't know whether that actually exists or would work).

mpeters:
>but doesn't handle setups where multiple verifiers use one DB (which I don't know whether that actually exists or would work).

Using a single database for mulitple verifiers is something that does work today, although with caveats.

As we work on a push model vs a pull model (for one-time boots, IoT and edge, etc) it will get more and more important that a host isn't tied to a single verifier and could either use several behind a load balancer or a list that it uses to do client-side load-balancing/fail-over, which is a very common pattern for large scale web deployments. I would be really hesitant to introduce any new code that makes this even harder in the future.

mpeters (approved):

PR #1200, opened 2022-11-21
The verifier can selectively load only a subset of columns from the `allowlist` table.

maugustosilva:
With this PR, the `verifier` can load only a few colmuns from the `allowlist` table, giving it the ability to selectively decide if the full row needs to be loaded (very useful for implementing a proxy or cache).

In addition to that, a new column, `generator` is introduced to reflect the fact that an IMA policy was supplied by an user, versus automatically generated as "empty" from within the verifier.

### DIFF
--- a/keylime/ima/ima.py
+++ b/keylime/ima/ima.py
@@ -21,6 +21,14 @@ logger = keylime_logging.init_logging("ima")
 # The version of the allowlist format that is supported by this keylime release
 ALLOWLIST_CURRENT_VERSION = 5
 
+
+class IMA_POLICY_GENERATOR:
+    Unknown = 0
+    EmptyAllowList = 1
+    CompatibleAllowList = 2
+    LegacyAllowList = 3
+
+
 # A correctly-formatted empty allowlist.
 #
 # IMA allowlists of versions older than 5 will not have the "log_hash_alg"
@@ -33,6 +41,7 @@ ALLOWLIST_CURRENT_VERSION = 5
 EMPTY_ALLOWLIST = {
     "meta": {
         "version": ALLOWLIST_CURRENT_VERSION,
+        "generator": IMA_POLICY_GENERATOR.EmptyAllowList,
         "checksum": "1ab36a57ad793caba756b01fc142f296cbfc3d24088fe6d66448f3bb6ebadeee",
     },
     "release": 0,
@@ -581,13 +590,14 @@ def canonicalize_allowlist(alist_bytes, checksum=""):
         else:
             logger.debug("Allowlist does not specify a version. Assuming current version %s", ALLOWLIST_CURRENT_VERSION)
 
+        alist["meta"]["generator"] = IMA_POLICY_GENERATOR.CompatibleAllowList
     else:
         # convert legacy format into new structured format
         logger.debug("Converting legacy allowlist format to JSON")
 
         alist = copy.deepcopy(EMPTY_ALLOWLIST)
         alist["meta"]["timestamp"] = str(datetime.datetime.now())
-        alist["meta"]["generator"] = "keylime-legacy-format-upgrade"
+        alist["meta"]["generator"] = IMA_POLICY_GENERATOR.LegacyAllowList
         if checksum:
             alist["meta"]["checksum"] = checksum
 
@@ -638,6 +648,11 @@ def ima_policy_db_contents(ima_policy_name: str, ima_policy: str, tpm_policy: st
                 sha256 = hashlib.sha256()
                 sha256.update(alist_bytes)
                 ima_policy_db_format["checksum"] = sha256.hexdigest()
+            if "generator" in ima_policy_json["allowlist"]["meta"]:
+                ima_policy_db_format["generator"] = ima_policy_json["allowlist"]["meta"]["generator"]
+            else:
+                ima_policy_db_format["generator"] = IMA_POLICY_GENERATOR.Unknown
+
     if tpm_policy:
         ima_policy_db_format["tpm_policy"] = tpm_policy
     return ima_policy_db_format
--- a/docs/rest_apis.rst
+++ b/docs/rest_apis.rst
@@ -73,6 +73,7 @@ Cloud verifier (CV)
             "vtpm_policy": "{\"23\": [\"ffffffffffffffffffffffffffffffffffffffff\", \"0000000000000000000000000000000000000000\"], \"15\": [\"0000000000000000000000000000000000000000\"], \"mask\": \"0x808000\"}",
             "meta_data": "{}",
             "has_mb_refstate": 0,
+            "has_ima_policy": 0,
             "accept_tpm_hash_algs": [
               "sha512",
               "sha384",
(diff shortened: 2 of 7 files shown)


mpeters (changes requested):

stefanberger:
From you r commit:

```
With this PR, the `verifier` can load only a few colmuns from the
`allowlist` table, giving it the ability to selectively decide if the
full row needs to be loaded (very useful for implementing a proxy or
cache).
```

Regarding **full row**: One could also introduce a column `modified` that covers a full row's update and not just indicates that those components covered by the checksum changed.

I had this here in my PR while it was still polling:

```
commit fbf936b64d35a39b9386bf2c03a64ad178e6e643
Author: Stefan Berger <stefanb@linux.ibm.com>
Date:   Thu Nov 17 19:31:52 2022 -0500

    verifier: Add timestamp 'modified' column to VerifierAllowlist

    Add a timestamp 'modified' column to the VerifierAllowlist that
    must not be NULL and when adding the column it is initialized with
    a default date. The column is automatially set to the current
    time on insert and also to the current time on updates.

diff --git a/keylime/db/verifier_db.py b/keylime/db/verifier_db.py
index d677814..b5b5ea8 100644
--- a/keylime/db/verifier_db.py
+++ b/keylime/db/verifier_db.py
@@ -1,4 +1,6 @@
-from sqlalchemy import Column, ForeignKey, Integer, LargeBinary, PickleType, String, Text, schema
+import datetime
+
+from sqlalchemy import TIMESTAMP, Column, ForeignKey, Integer, LargeBinary, PickleType, String, Text, schema
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.orm import relationship

@@ -60,3 +62,4 @@ class VerifierAllowlist(Base):
     checksum = Column(String(128))
     tpm_policy = Column(Text())
+    modified = Column(TIMESTAMP, nullable=False, default=datetime.datetime.now, onupdate=datetime.datetime.now)
diff --git a/keylime/migrations/versions/90e4f63b79aa_add_modified_to_allowlist.py b/keylime/migrations/versions/90e4f63b79aa_add_modified_to_allowlist.py
new file mode 100644
index 0000000..50bc650
--- /dev/null
+++ b/keylime/migrations/versions/90e4f63b79aa_add_modified_to_allowlist.py
@@ -0,0 +1,44 @@
+"""add_modified_to_allowlist
+
+Revision ID: 90e4f63b79aa
+Revises: 2fbc0fb8fa4d
+Create Date: 2022-11-17 22:22:55.602069
+
+"""
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "90e4f63b79aa"
+down_revision = "2fbc0fb8fa4d"
+branch_labels = None
+depends_on = None
+
+
+def upgrade(engine_name):
+    globals()[f"upgrade_{engine_name}"]()
+
+
+def downgrade(engine_name):
+    globals()[f"downgrade_{engine_name}"]()
+
+
+def upgrade_registrar():
+    pass
+
+
+def downgrade_registrar():
+    pass
+
+
+    # sqlite only lets us put a "0" into modified column when adding column
+    # Replace it with a true timestamp in 2nd step
+    op.add_column(
+        "allowlists",
+        sa.Column("modified", sa.TIMESTAMP, nullable=False, server_default="2000-01-01 00:00:00"),
+    )
+
+
+def downgrade_cloud_verifier():
+    op.drop_column("allowlists", "modified")
```

keylime/ima/ima.py
```diff
@@ -33,6 +34,7 @@
 EMPTY_ALLOWLIST = {
     "meta": {
         "version": ALLOWLIST_CURRENT_VERSION,
+        "generator": 1,
```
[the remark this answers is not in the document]
aplanas: Agreed. I am not sure if this is a "version" or a boolean type

keylime/ima/ima.py
```diff
@@ -55,6 +57,11 @@
     "verification-keys": "",
 }
 
+IMA_POLICY_GENERATOR = Enum(
+    "generator",
+    ["unknown", "keylime-ima-empty-allowlist", "keylime-allowlist-compatible-format", "keylime-legacy-format-upgrade"],
+)
+
```
aplanas, line 64:
```python
class Generator:
      Unknown = 0
      EmptyAllowList = 1
      CompatibleAllowList = 2
      LegacyAllowList = 3
```

keylime/ima/ima.py
```diff
@@ -577,13 +584,14 @@ def canonicalize_allowlist(alist_bytes, checksum=""):
         else:
             logger.debug("Allowlist does not specify a version. Assuming current version %s", ALLOWLIST_CURRENT_VERSION)
 
+        alist["meta"]["generator"] = 2
```
aplanas, line 587: `alist["meta"]["generator"] = Generator.CompatibleAllowList`

keylime/ima/ima.py
```diff
@@ -577,13 +584,14 @@ def canonicalize_allowlist(alist_bytes, checksum=""):
...
     else:
         # convert legacy format into new structured format
         logger.debug("Converting legacy allowlist format to JSON")
 
         alist = copy.deepcopy(EMPTY_ALLOWLIST)
         alist["meta"]["timestamp"] = str(datetime.datetime.now())
-        alist["meta"]["generator"] = "keylime-legacy-format-upgrade"
+        alist["meta"]["generator"] = 3
```
aplanas, line 594: `alist["meta"]["generator"] = Generator.LegacyAllowList`

keylime/ima/ima.py
```diff
@@ -627,13 +635,11 @@ def ima_policy_db_contents(ima_policy_name: str, ima_policy: str, tpm_policy: st
...
-            if "checksum" in ima_policy["allowlist"]["meta"]:
-                ima_policy_db_format["checksum"] = ima_policy["allowlist"]["meta"]["checksum"]
+            if "generator" in ima_policy["allowlist"]["meta"]:
+                ima_policy_db_format["generator"] = ima_policy["allowlist"]["meta"]["generator"]
             else:
-                alist_bytes = json.dumps(copy.deepcopy(EMPTY_ALLOWLIST)).encode()
-                sha256 = hashlib.sha256()
-                sha256.update(alist_bytes)
-                ima_policy_db_format["checksum"] = sha256.hexdigest()
+                ima_policy_db_format["generator"] = 0
```
aplanas, line 641: `ima_policy_db_format["generator"] = Generator.Unknown`

keylime/ima/ima.py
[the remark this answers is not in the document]
maugustosilva: fixed

keylime/ima/ima.py
maugustosilva: done

keylime/ima/ima.py
maugustosilva: done

keylime/ima/ima.py
maugustosilva: done

keylime/ima/ima.py
maugustosilva: done

mpeters (approved):

PR #1205, opened 2022-11-22
chore: cleaning up pyproject

mdrocco:
cleaning up pyproject.toml

Removing leftover entry from pyright exclude list: keylime/ca_impl_openssl.py

Signed-off-by: Maurizio Drocco <maurizio.drocco@ibm.com>

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,6 +28,5 @@ ignore = [
     "keylime/da/examples/file.py",
     "keylime/da/record.py",
     "keylime/cloud_verifier_tornado.py",
-    "keylime/ca_impl_openssl.py",
 ]
 reportMissingImports = false


THS-on:
@stefanberger you added this to make it pass `make check`, right?

stefanberger:
I had seen the error and then it went away. I had copied and pasted the wrong stuff into the commit text and removed this now.

The problem is a mismatch between key type from PEM files and key types suitable for signing an x509 cert:

load_pem_private_key: https://github.com/pyca/cryptography/blob/01687d63a5cb4475d84feae959bec6dea5fce3f1/src/cryptography/hazmat/primitives/serialization/base.py#L15

key types: https://github.com/pyca/cryptography/blob/01687d63a5cb4475d84feae959bec6dea5fce3f1/src/cryptography/hazmat/primitives/asymmetric/types.py#L31-L40

vs.

key types allowed to sign an x509: https://github.com/pyca/cryptography/blob/01687d63a5cb4475d84feae959bec6dea5fce3f1/src/cryptography/hazmat/primitives/asymmetric/types.py#L41-L49

THS-on (approved):
@stefanberger so we can remove this file from the ignore list correct?

aplanas (approved):

mpeters (approved):

PR #1206, opened 2022-11-22
pyright: Fix a pyright issue in ca_impl_openssl

stefanberger:
Filter-out unsupported private key types and raise a ValueError before trying to sign with the private key. Apply a filter of all those keys that could have possibly been used for signing. Any other type of key would have already thrown an exception since it was not possible to sign an x509 certificate with it.

### DIFF
--- a/keylime/ca_impl_openssl.py
+++ b/keylime/ca_impl_openssl.py
@@ -4,6 +4,11 @@ from cryptography import x509
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import hashes
 from cryptography.hazmat.primitives.asymmetric import rsa
+from cryptography.hazmat.primitives.asymmetric.dsa import DSAPrivateKey
+from cryptography.hazmat.primitives.asymmetric.ec import EllipticCurvePrivateKey
+from cryptography.hazmat.primitives.asymmetric.ed448 import Ed448PrivateKey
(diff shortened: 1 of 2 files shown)


keylime/ca_impl_openssl.py
```diff
@@ -188,5 +193,9 @@ def gencrl(serials, cert: str, ca_pk: str):
         cert_builder = cert_builder.revocation_date(date_now)
         builder = builder.add_revoked_certificate(cert_builder.build())
 
+    if not isinstance(
+        priv_key, (EllipticCurvePrivateKey, RSAPrivateKey, DSAPrivateKey, Ed448PrivateKey, Ed25519PrivateKey)
+    ):
+        raise ValueError(f"Unsupported key type {type(priv_key).__name__}")
```
THS-on, line 199: Have you checked on how the callers are handling this ValueError?

aplanas: If I read the code correctly, no one is capturing this, so the user will see the exception and can fix the issue

THS-on (approved):

stefanberger:
Thanks for merging. The problem was actually not even reported anymore by pyright and I am not sure why it just disappeared. Anyway, it's unlikely that someone uses these exotic keys from a PEM file that cannot be used for signing an x509 certificate. Previously I would have expected the code to throw an exception from deep down in the crypto library, now it throws it 'earlier.'

PR #1208, opened 2022-11-28
config: remove fallback config

THS-on:
The fallback config used the old format which no longer works.

### DIFF
--- a/keylime/config.py
+++ b/keylime/config.py
@@ -4,15 +4,6 @@ import os
 import os.path
 from typing import Optional
 
-# resources in importlib is a Python 3.7 feature, so we disable
-# fallback support if we cannot import this module
-try:
-    from importlib import resources
-
-    _CONFIG_FALLBACK = True
-except ImportError:
-    _CONFIG_FALLBACK = False
-
 import yaml
 
 try:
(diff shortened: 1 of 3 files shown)


aplanas (approved):

maugustosilva:
We are currently relying on @ansasaki script (`convert_config.py`) to deploy KeyLime on our QA/Prod environments, which requires a `keylime.conf` as input.

It is my understanding that PR #1133 would implement a clean alternative to have `keylime.conf` removed (by admitting a "null input" and building the multiple configuration files from the defaults on jinja templates), which does not seem to be case with this PR.

If my understanding is correct, I would argue in favor of holding the merge of this PR until the aforementioned other is merged first.

aplanas:
> We are currently relying on @ansasaki script (`convert_config.py`) to deploy KeyLime on our QA/Prod environments, which requires a `keylime.conf` as input.

Am I reading the code wrongly? I think that this PR it is not removing `keylime.conf`, but the link introduced in https://github.com/keylime/keylime/pull/873, and this PR is a partial reverse of it.

THS-on:
@aplanas is correct. This is a revert of that part, because the code is currently broken.

mpeters (approved):

PR #1210, opened 2022-11-28
codestyle: Add missing type annotations to ima/ima.py and address issues

stefanberger:
Type-annotation all functions in ima/ima.py and fix some potential issues that pyright detected on the way.

### DIFF
--- a/keylime/ima/ast.py
+++ b/keylime/ima/ast.py
@@ -150,7 +150,7 @@ class Digest:
     """
 
     hash: bytes
-    algorithm: Optional[str] = None
+    algorithm: str
     legacy: bool = False
 
     def __init__(self, digest: str, legacy: bool = False):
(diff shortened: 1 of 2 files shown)


stefanberger:
pyright with python 3.6 has an issue with re.Pattern:
```
pylint 2.13.9
astroid 2.11.7
Python 3.6.15 (default, Mar  3 2022, 00:00:00) 
[GCC 11.2.1 20220127 (Red Hat 11.2.1-9)]
************* Module keylime.ima.ima
keylime/ima/ima.py:10:0: E0611: No name 'Pattern' in module 're' (no-name-in-module)
```

I am going to have to drop annotating the regex since annotation with re.Pattern also didn't work with python 3.6.

keylime/ima/ima.py
```diff
@@ -197,6 +212,11 @@ def _validate_ima_sig(
             logger.debug("IMA: ignoring excluded path %s", path.name)
             return failure
 
+        if not digest.algorithm:
```
THS-on, line 214: I think this we set the `algorithm` always in the `Digest` class. The algorithm is never set in the old `d` template.

stefanberger: Then that should probably be done in the Digest class then. The problem is this here:

```
class Digest:
    """
    Class for types "d" and "d-ng" with and without algorithm
    """

    hash: bytes
    algorithm: Optional[str] = None
[...]
```

It must not be Optional...

stefanberger: I now added another patch to the front of this series that fixes this issue.

aplanas (approved):

keylime/ima/ast.py
```diff
@@ -150,7 +150,7 @@ class Digest:
     """
 
     hash: bytes
-    algorithm: Optional[str] = None
+    algorithm: str = "sha1"
```
THS-on, line 153: Do we need to default initialize it?

keylime/ima/ima.py
THS-on: Yes, the provided patch fixes that thanks!

THS-on (approved):

PR #1214, opened 2022-11-29
Enable mypy in CI and add full type annotation for config and ima AST

THS-on:
- tox: add mypy support
- json: allow ignore comment to be parsed by mypy
- ima/ast: add missing type hints
- config: add missing type hints
- ci: add mypy to CI

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -1,2 +1,10 @@
 dbus-python
-jinja2
+# packages required for mypy
+sqlalchemy-stubs
+typed-ast
+types-atomicwrites
+types-python-dateutil
+types-PyYAML
+types-redis
+types-requests
+types-urllib3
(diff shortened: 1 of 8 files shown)


stefanberger:
LGTM

aplanas (approved):

stefanberger:
> I needed to relax the type for the `yaml_to_dict(.. )` function, so that pyright does not complain.

After I saw your changes I was wondering how you got past pyright...

THS-on:
> After I saw your changes I was wondering how you got past pyright...

I changed the type hint for `arry` to `Any`. This is not ideal, but I couldn't figure out another way to to make pyright happy.

aplanas (approved):

test-requirements.txt
```diff
@@ -1,2 +1,10 @@
 dbus-python
-jinja2
```
mpeters, line 2: Was this used during the config conversion? And we're no longer doing that during tests? Do we need a test for that specifically?

mpeters (approved):
One small question, but doesn't hold up the approval

PR #1215, opened 2022-11-29
codestyle: Add type annotations to cmd_exec.py

stefanberger:

### DIFF
--- a/keylime/cmd_exec.py
+++ b/keylime/cmd_exec.py
@@ -1,23 +1,34 @@
 import os
 import subprocess
 import time
+from typing import Any, Dict, List, Optional, Sequence, Tuple, Union, cast
 
 EXIT_SUCESS = 0
 
+EnvType = Dict[str, str]
 
-def _execute(cmd, env=None, **kwargs):
+
+def _execute(cmd: Sequence[str], env: Optional[EnvType] = None, **kwargs) -> Tuple[bytes, bytes, int]:
     with subprocess.Popen(cmd, env=env, stdout=subprocess.PIPE, stderr=subprocess.PIPE, **kwargs) as proc:
         out, err = proc.communicate()
-        return out, err, proc.returncode
+        # All callers assume to receive a list of bytes back; none of them uses 'text mode'
+        return cast(bytes, out), cast(bytes, err), proc.returncode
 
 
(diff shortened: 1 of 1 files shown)


keylime/cmd_exec.py
```diff
@@ -1,23 +1,37 @@
...
 
 EXIT_SUCESS = 0
 
+EnvType = Dict[str, str]
 
-def _execute(cmd, env=None, **kwargs):
+
+def _execute(
+    cmd: Sequence[str], env: Optional[EnvType] = None, **kwargs
+) -> Tuple[Union[str, bytes], Union[str, bytes], int]:
```
aplanas, line 13: I think that all the communications are expected to be as strings. Maybe I am wrong, but if we pass `text=True` we can remove the Union

stefanberger: Another one:
```
                        cmd = ("openssl", "crl", "-inform", "der", "-in", "cacrl.der", "-text", "-noout")
                        retout = cmd_exec.run(cmd)["retout"]
                        for line in retout:
                            line = line.strip()
                            if line.startswith(b"Next Update:"):
```

And all callers from tpm.py also currently get `bytes` back. So it be `bytes` then.

aplanas: Yes, that also would work. I suggested the str because all are doing decode, that I guess can be moved early. But I am fine with the reverse logic (produce bytes and decode as str later)

aplanas (approved):

keylime/cmd_exec.py
```diff
@@ -1,31 +1,42 @@
...
+EnvType = Dict[str, str]
 
-def _execute(cmd, env=None, **kwargs):
+
+def _execute(cmd: Sequence[str], env: Optional[EnvType] = None, **kwargs) -> Tuple[bytes, bytes, int]:
     with subprocess.Popen(cmd, env=env, stdout=subprocess.PIPE, stderr=subprocess.PIPE, **kwargs) as proc:
         out, err = proc.communicate()
-        return out, err, proc.returncode
+        # All callers assume to receive a list of bytes back
+        return cast(bytes, out), cast(bytes, err), proc.returncode
```
THS-on, line 15: This only holds when text mode is never set to True. We should add a comment indicating that.

stefanberger: Added that text mode is not being used to the comment now.

THS-on (approved):

PR #1218, opened 2022-11-29
Cleanup: Remove ancient test case and a dead function parameter

stefanberger:
This PR removes what seems to be an ancient and dead test case in ima.py using a `main` function.
Once this test case has been removed the m2w parameter can also be removed from process_measurement_list.


stefanberger:
If someone is still using this test case let me know and I drop this PR but will type-annotate the m2w parameter.

THS-on (approved):
I think it can be removed.

aplanas (approved):

PR #1219, opened 2022-11-29
code-style: Type annotate various function in keylime/cmd

stefanberger:
In this PR various low-level functions and command line tools in keylime/cmd are type annotated.


keylime/api_version.py
```diff
@@ -8,11 +9,11 @@
...
 
 
-def current_version():
+def current_version() -> str:
     return CURRENT_VERSION
 
 
-def latest_minor_version(v):
+def latest_minor_version(v: str) -> str:
     v_obj = version.parse(str(v))
```
stefanberger, line 17: This str(v) would probably enable floats and ints as well but so far callers only seem to pass in strings ...

aplanas: Once added the type maybe is safe to drop it call

keylime/api_version.py
```diff
@@ -23,21 +24,21 @@ def latest_minor_version(v):
...
 
 
-def all_versions():
+def all_versions() -> List[str]:
     return VERSIONS.copy()
 
 
-def is_supported_version(v):
+def is_supported_version(v: str) -> bool:
     v_obj = version.parse(str(v))
```
aplanas, line 32: same here

keylime/api_version.py
```diff
@@ -23,21 +24,21 @@ def latest_minor_version(v):
...
     return v_obj.base_version in VERSIONS
 
 
-def is_deprecated_version(v):
+def is_deprecated_version(v: str) -> bool:
     v_obj = version.parse(str(v))
```
aplanas, line 37: seems that here too

keylime/api_version.py
```diff
@@ -23,21 +24,21 @@ def latest_minor_version(v):
...
     return is_supported_version(v) and v_obj.base_version in DEPRECATED_VERSIONS
 
 
-def normalize_version(v):
+def normalize_version(v: str) -> str:
     v = str(v)
```
aplanas, line 42: here, and later there are more. I vote to remove all of them and see what is broken

keylime/api_version.py
stefanberger: Added another patch now and also type-annotated also constants in this file now.

stefanberger: The unittests failed due to floats and ints being used there. So, for now I adjusted the types to include floats and ints and didn't change the test cases.

aplanas (approved):

mpeters (approved):

PR #1220, opened 2022-11-30
DO NOT MERGE

stefanberger:


maugustosilva:
The `checksum` is used on a solution for the IMA scalability problem we have been testing internally (no PR open yet)… furthermore, calculating a `checksum` internally on the `verifier` is, on itself, a harmless and cheap operation, one which will not preclude the adoption of any alternative solution (such as one based on SQL timestamps) in the future.

maugustosilva (changes requested):
Leave the `verifier` with ability to calculate an IMA policy `checksum`

stefanberger:
> Leave the `verifier` with ability to calculate an IMA policy `checksum`

This PR was only for running the tests, nothing else.

PR #1227, opened 2022-12-02
codestyle: Set the type of trusted server_ca_type in tenant.py to be a List[str] and initalize it with empty list

stefanberger:
The tenant's trusted_server_ca is not a boolean but a List[str]. Initialize it with and empty list and not a boolean.

Clarify that the trusted_ca return parameter from get_tls_options() is a List[str].


keylime/web_util.py
```diff
@@ -252,17 +253,18 @@ def get_tls_options(component, is_client=False, logger=None):
     # Peer certificate verification is enabled by default
     verify_peer_certificate = True
 
-    trusted_ca = config.get(component, ca_option)
-    if not trusted_ca:
+    trusted_ca: List[str]
```
THS-on, line 256: What does this exactly do in python? Because in my understanding this does not create a variable with the name trusted_ca.

stefanberger: I dropped this line now.

THS-on (approved):

PR #1230, opened 2022-12-02
cmd: removal of keylime_migrations_apply

THS-on:
Migrations are normally applied either manually or automatically during startup of
the verifier or registrar, so the option to run the migrations via a standalone
command is no longer required.


aplanas (approved):

mpeters (approved):
I agree. I originally implemented this because I thought there might be a need for some complicated setups to have more control over when migrations were applied as part of their deployment process, but even if that were the case, there's no need to make it a separate command. They'd just start up the verifier/registrar when needed.

PR #1232, opened 2022-12-02
codestyle: Type-annotate web_util.py

stefanberger:
Type-annotate web_util.py. The following type annotation forces the casting of rest_params["api_version"] to a str. We can do this since we know that "api_version" has been set in get_restful_params.

def get_restful_params(urlstring: str) -> Dict[str, Union[str, None]]:

### DIFF
--- a/keylime/web_util.py
+++ b/keylime/web_util.py
@@ -6,14 +6,17 @@ import sys
 import tempfile
 import urllib.parse
 from http.server import BaseHTTPRequestHandler
+from logging import Logger
+from typing import Any, Dict, List, Optional, Tuple, Union
 
 import tornado.web
 
 from keylime import api_version as keylime_api_version
 from keylime import ca_util, config, json, secure_mount
+from keylime.api_version import VersionType
 
 
-def get_tls_dir(component):
+def get_tls_dir(component: str) -> str:
     # Get the values from the configuration file
     tls_dir = config.get(component, "tls_dir")
 
@@ -43,7 +46,7 @@ def get_tls_dir(component):
     return tls_dir
 
 
-def init_tls_dir(component, logger=None):
+def init_tls_dir(component: str, logger: Optional[Logger] = None) -> str:
     """
     Init the TLS directory, generating keys and certificates if requested
     """
@@ -144,15 +147,15 @@ def init_tls_dir(component, logger=None):
 
 
 def generate_tls_context(
-    certificate,
-    private_key,
-    trusted_ca,
-    private_key_password=None,
-    verify_peer_cert=True,
-    is_client=False,
-    ca_cert_string=None,
-    logger=None,
-):
+    certificate: Optional[str],
+    private_key: Optional[str],
+    trusted_ca: List[str],
+    private_key_password: Optional[str] = None,
+    verify_peer_cert: bool = True,
+    is_client: bool = False,
+    ca_cert_string: Optional[str] = None,
+    logger: Optional[Logger] = None,
+) -> ssl.SSLContext:
     """
     Generate the TLS context
 
@@ -227,7 +230,9 @@ def generate_tls_context(
     return context
 
 
-def get_tls_options(component, is_client=False, logger=None):
+def get_tls_options(
+    component: str, is_client: bool = False, logger: Optional[Logger] = None
+) -> Tuple[Tuple[Optional[str], Optional[str], List[str], Optional[str]], bool]:
     """
     Get the TLS key and certificates to use for the given component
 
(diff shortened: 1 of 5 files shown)


keylime/web_util.py
```diff
@@ -440,11 +449,11 @@ def validate_api_version(handler, version, logger):
...
 
 
-def _list_to_dict(alist):
+def _list_to_dict(alist: List[str]) -> Dict[str, Union[str, None]]:
     """Convert list into dictionary via grouping [k0,v0,k1,v1,...]"""
     params = {}
     i = 0
     while i < len(alist):
-        params[alist[i]] = alist[i + 1] if (i + 1) < len(alist) else None
+        params[alist[i]] = alist[i + 1] if (i + 1) < len(alist) else None  # Must keep NONE here
```
stefanberger, line 457: It's unfortunate that there is this `None` here. An empty string "" **may be** equivalent an would help to make the return parameter a `Dict[str, str]` but then we do have issues with things like these here (in registrar_common.py for example):

```
        rest_params = web_util.get_restful_params(self.path)

        [...]

        agent_id = rest_params["agents"]

        if agent_id is not None:
```

keylime/web_util.py
```diff
@@ -406,14 +415,14 @@ def echo_json_response(handler, code, status=None, results=None):
...
-def get_restful_params(urlstring):
+def get_restful_params(urlstring: str) -> Dict[str, Union[str, None]]:
     """Returns a dictionary of paired RESTful URI parameters"""
     parsed_path = urllib.parse.urlsplit(urlstring.strip("/"))
     query_params = urllib.parse.parse_qsl(parsed_path.query)
     path_tokens = parsed_path.path.split("/")
 
     # If first token looks like an API version, validate it and make sure it's supported
-    api_version = 0
+    api_version = "0"
```
stefanberger, line 425: This avoids `Dict[str, Union[str, int, None]]`

stefanberger: I had to adjust the test case because of this change.

keylime/web_util.py
```diff
@@ -406,14 +415,14 @@ def echo_json_response(handler, code, status=None, results=None):
     return False
 
 
-def get_restful_params(urlstring):
+def get_restful_params(urlstring: str) -> Dict[str, Union[str, None]]:
```
aplanas, line 418: Not sure, but maybe `Dict[str, Optional[str]]`

keylime/web_util.py
aplanas: I personally never do a `if var is [not] None`, but `if [not] var`. Maybe this helps here as for one side the code will be equivalent and the type can be forced.

keylime/web_util.py
stefanberger: right. I did this kind of on purpose in this case to make the `None` visible.

keylime/web_util.py
stefanberger: I would put it on a todo list because I think this kind of change requires a bit more 'care'.

stefanberger:
> @stefanberger once the annotations pass also mypy you can add the following to the mypy.ini:
> 
> ```
> [mypy-keylime.web_util]
> ignore_errors = False
> ```

That's only going to be possible once other low level functions have been annotated.


```
mypy run-test: commands[0] | mypy keylime/
keylime/web_util.py:41: error: Call to untyped function "get_secdir" in typed context  [no-untyped-call]
keylime/web_util.py:86: error: Call to untyped function "setpassword" in typed context  [no-untyped-call]
keylime/web_util.py:99: error: Call to untyped function "cmd_init" in typed context  [no-untyped-call]
keylime/web_util.py:118: error: Call to untyped function "cmd_mkcert" in typed context  [no-untyped-call]
keylime/web_util.py:133: error: Call to untyped function "cmd_mkcert" in typed context  [no-untyped-call]
keylime/web_util.py:280: error: Incompatible types in assignment (expression has type "None", variable has type "str")  [assignment]
keylime/web_util.py:294: error: Incompatible types in assignment (expression has type "None", variable has type "str")  [assignment]
keylime/web_util.py:310: error: Incompatible types in assignment (expression has type "None", variable has type "str")  [assignment]
keylime/web_util.py:400: error: Incompatible types in assignment (expression has type "bytes", variable has type "str")  [assignment]
keylime/web_util.py:406: error: Argument 1 to "write" of "IO" has incompatible type "str"; expected "bytes"  [arg-type]
```

aplanas (approved):

THS-on (approved):

PR #1233, opened 2022-12-03
codestyle: Fix some issues with type annotiations in the verifier

stefanberger:
This PR fixes some issue with type annotations in the verifier.

### DIFF
--- a/keylime/cloud_verifier_tornado.py
+++ b/keylime/cloud_verifier_tornado.py
@@ -34,7 +34,7 @@ from keylime.da import record
 from keylime.db.keylime_db import DBEngineManager, SessionManager
 from keylime.db.verifier_db import VerfierMain, VerifierAllowlist
 from keylime.elchecking import policies
-from keylime.failure import MAX_SEVERITY_LABEL, Component, Failure, SeverityLabel, set_severity_config
+from keylime.failure import MAX_SEVERITY_LABEL, Component, Event, Failure, set_severity_config
 from keylime.ima import ima
 
 logger = keylime_logging.init_logging("verifier")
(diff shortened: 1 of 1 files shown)


keylime/cloud_verifier_tornado.py
```diff
@@ -691,10 +691,10 @@ def put(self):
                 return
 
             if "reactivate" in rest_params:
-                if not isinstance(db_agent, dict):
-                    agent = _from_db_obj(db_agent)
+                if not isinstance(db_agent, dict):  # FIXME: should always be true...
```
stefanberger, line 694: This `if not isinstance(db_agent, dict)` check can probably go since the conversion is always need and this should just be `agent = cast(Dict[str, Any], _from_db_obj(db_agent))`

keylime/cloud_verifier_tornado.py
```diff
@@ -680,19 +680,19 @@ def put(self):
...
-            if agent is None:
+            if db_agent is None:
                 web_util.echo_json_response(self, 404, "agent id not found")
                 logger.info("PUT returning 404 response. agent id: %s not found.", agent_id)
                 return
 
             if "reactivate" in rest_params:
-                if not isinstance(agent, dict):
-                    agent = _from_db_obj(agent)
+                agent = cast(Dict[str, Any], _from_db_obj(db_agent))
```
aplanas, line 694: What is the complain here? `_from_db_obj` seems to be `Dict[str, str]` and `agent` values seems to be used as `str` here except the `is None`. 

Can be that `_from_db_obj` type should be `Dict[str, Optional[str]]` and this cast can be dropped?

stefanberger: The complaint is that `_from_db_obj` returns a `Dict[str, str]` but this then doesn't accept the `SSLContext | None` from this call here a few lines below:
```
agent["ssl_context"] = web_util.generate_agent_tls_context(
                        "verifier", agent["mtls_cert"], logger=logger
                    )
```

and the `int` added by this call:
```
agent["operational_state"] = states.START
```

keylime/cloud_verifier_tornado.py
stefanberger: I removed the isinstance check since it's always the case that it's not a dict.

aplanas (approved):

PR #1236, opened 2022-12-05
codestyle: Add missing annotations to crypto.py and add to mypy

stefanberger:


keylime/crypto.py
```diff
@@ -71,6 +71,7 @@ def get_public_key(private_key: RSAPrivateKey) -> RSAPublicKey:
 
 def rsa_sign(key: RSAPrivateKey, message: bytes, paddingt: str = "internal") -> bytes:
     """RSA sign message"""
+    _padding: Union[padding.PSS, padding.PKCS1v15]
```
stefanberger, line 74: @THS-on It doesn't have an effect in python but makes mypy happy...

aplanas (approved):

THS-on (approved):
The rawhide failure has nothing to do with this change.

Issue #1207, opened 2022-11-28
Keylime server logs created with not secured permissions


zkolpet:
# Environment
 * OS / version: Ubuntu22.04
 * Processor architecture: N/A
 * TPM Manufacturer: N/A
 * Keylime version: 6.5.2

# Description 
verifier.log and registrar.log created with wrong permisssions 644. These are not secured enough permissions. All logs should be created with permissions 640.

## Expected behavior vs. actual behavior 
Expected: verifier and registrar logs are created with 640 permissions

## Steps to reproduce problem

# Relevant logs

Attach any relevant log files that can help to debug your issue.

THS-on:
I think the issue here is that we set the umask in the main method after the logging module already created the files.
See here: 
https://github.com/keylime/keylime/blob/5db2f7b94fe0bf36e8ca2e2218eedc159bb0a297/keylime/cloud_verifier_tornado.py#L1356

We can either call umask when initializing the logging or just move those calls out of the `main()` method and put them before calling anything else.

@aplanas @sergio-correia @ansasaki opinions on how we should solve that?

zkolpet:
I am fine with whatever solution as long as it works :)

aplanas:
> We can either call umask when initializing the logging or just move those calls out of the `main()` method and put them before calling anything else.
> 
> @aplanas @sergio-correia @ansasaki opinions on how we should solve that?

IMHO none. `umask` is a trigger bomb and we are already on top of it.

I inspected the code of Python logging module and there is no place to set the file permissions, but I think that a bit better approach is now when the logging configuration is done, just before, to be sure that we create the file and do a `os.chmod` unconditionally later.

A better approach is IMHO what `rust-keylime` is doing: write that in stdout! That way systemd will take care of it and make the service logs accessed by `journalctl` : ))

zkolpet:
Hi,

little workaround I did for now is what @aplanas mentioned above as "a bit better approach":

in keylime_logging.py module (last 3 lines in below code):

    if loggername in LOG_TO_FILE:
        logfilename = os.path.join(LOGDIR, f"{loggername}.log")
        if not os.path.exists(LOGDIR):
            os.makedirs(LOGDIR, 0o750)
        if not os.path.isfile(logfilename):
            with open(logfilename, "a"):
                os.chmod(logfilename, 0o640)

aplanas:
Do you want to make a PR?

THS-on:
[...]

Issue #1174, opened 2022-11-11
Verifier not logging into a file: /var/log/keylime/verifier.log


zkolpet:
# Environment
 * OS / version: Ubuntu 22.04
 * Processor architecture: N/A
 * TPM Manufacturer: N/A
 * Keylime version: 6.5.0

# Description 
Verifier component is not logging into a file /var/log/keylime/verifier.log when set so in verifier.conf.

config preview:
Destination for log output, in addition to console. If left empty, the log
output will only be printed to console (recommended for containers to avoid
filling data storage). The accepted values are:
'file': The log output will also be written to a file named after the
        component in '/var/log/keylime/verifier.log'
'stream': The log output will be written to a common file in
          'var/log/keylime/keylime-stream.log'
log_destination = file

This log file: /var/log/keylime/verifier.log is not simply created. I have the same settings for registrar and registrar is creating and logging into this log file just fine: /var/log/keylime/registrar.log.

## Expected behavior vs. actual behavior 
Expected:
Verifier component is logging into **/var/log/keylime/verifier.log** when **log_destination = file** set in verifier.conf

Actual:
Verifier component IS NOT logging into **/var/log/keylime/verifier.log** when **log_destination = file** set in verifier.conf

## Steps to reproduce problem
1. Start keylime server side
2. See that no /var/log/keylime/verifier.log is created.

# Relevant logs

Attach any relevant log files that can help to debug your issue.

THS-on:
Thanks for reporting. I have found the cause of issue: the log_destination option is in the revocation section, which makes this fail. See:
https://github.com/keylime/keylime/blob/56e951a851b9a12a58fa473da94d1c1bf1415676/scripts/templates/2.0/verifier.j2#L200-L207

On another note this only logs the entries produced directly by the verifier and not other components in Keylime used by the verifier (e.g. IMA).

zkolpet:
Fix confirmed. Moving the log section out of [revocations] section makes it work and /var/log/keylime/verifier.log is now properly generated.

Thank you

PR #1240, opened 2022-12-06
logging: remove option to log into separate file

THS-on:
The implementation had the issue that only the main loggers were added and that
the permissions were not set strict enough. Users should use the logging
provided by systemd instead.

Fixes: #1207
Fixes: #1174

### DIFF
--- a/keylime.conf
+++ b/keylime.conf
@@ -382,11 +382,6 @@ signed_attributes =
 # signature check before storing them in the database.
 require_allow_list_signatures = False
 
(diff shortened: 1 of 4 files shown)


maugustosilva (approved):

maugustosilva:
LGTM, no one is (and should not be) using "log to files" anymore. Whether deployed inside containers or as a systemd service, stdout suffices to generate the appropriate log.

aplanas (approved):
Thanks!

PR #1241, opened 2022-12-06
Add some files in keylime/tpm to mypy

stefanberger:
Fix type annotations and variables to avoid type-switching of variables in tpm2_objects.py and tpm2_object_test.py. Add the two files to mypy.

### DIFF
--- a/keylime/tpm/tpm2_objects.py
+++ b/keylime/tpm/tpm2_objects.py
@@ -1,6 +1,6 @@
 import hashlib
 import struct
-from typing import Tuple, Union
+from typing import Any, Tuple, Union, cast
 
 import cryptography.hazmat.primitives.asymmetric.ec as crypto_ec
 from cryptography.hazmat.backends import default_backend
@@ -14,7 +14,7 @@ from cryptography.hazmat.primitives.asymmetric.rsa import RSAPublicKey, RSAPubli
 pubkey_type = Union[RSAPublicKey, EllipticCurvePublicKey]
 
 
-def _pack_in_tpm2b(val):
+def _pack_in_tpm2b(val: bytes) -> bytes:
     return struct.pack(">H", len(val)) + val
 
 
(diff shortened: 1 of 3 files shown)


keylime/tpm/tpm2_objects.py
```diff
@@ -61,22 +61,30 @@ def _pack_in_tpm2b(val):
...
-    scheme_scheme = None
-    scheme_details = None
-
-    def __init__(self, sym_algorithm, sym_keybits, sym_mode, sym_details, scheme_scheme, scheme_details):
+    sym_algorithm: Optional[int] = None
+    sym_keybits: Optional[int] = None
+    sym_mode: Optional[int] = None
+    sym_details: Optional[int] = None
+    scheme_scheme: Optional[int] = None
+    scheme_details: Optional[int] = None
```
aplanas, line 69: IMHO this seems to be a confusion about class fields and instance fields? If so should be dropped.

aplanas (changes requested):

aplanas (approved):

keylime/tpm/tpm2_objects.py
stefanberger: These can at least serve as type specifiers if we modify them to something like this:
```
    sym_algorithm: int
    sym_keybits: int
    sym_mode: int
    sym_details: int
    scheme_scheme: int
    scheme_details: int
```

When I change sym_algorithm to be 'str' then mypy complains like this:

```
keylime/tpm/tpm2_objects.py:80: error: Incompatible types in assignment (expression has type "int", variable has type "str")  [assignment]
```

So there's a point in having these but initializing them is wrong, which makes them class variables.
I introduced such type specifiers in other parts as well without initializing them... and we already had them.

aplanas: Sure, I was pointing to the original code.

aplanas (approved):

mpeters (approved):

PR #1242, opened 2022-12-06
verifier: fix for IMA policy checksum calculation

maugustosilva:
The "allowlist" field on IMA policy (JSON) has, inside its "meta" subfield, a "timestamp" key,value pair (generated at the `verifier`). Due to this fact, whenever the `checksum` (over the whole policy) is calculated (also) at the `verifier`, the end result is different values for policies which are actually identical (minus the aforementioned `timestamp`).

While it could be possible to simply remove this timestamp, in order to prepare for future changes (such as PR #1147), this PR simply truncates the actual value of the timestamp, as it has no relation to the contents (e.g., "allowlist") of the IMA policy.

### DIFF
--- a/keylime/ima/ima.py
+++ b/keylime/ima/ima.py
@@ -656,6 +656,7 @@ def ima_policy_db_contents(ima_policy_name: str, ima_policy: str, tpm_policy: st
     """Assembles an ima policy dictionary to be written on the database"""
     ima_policy_db_format: Dict[str, Any] = {}
     ima_policy_db_format["name"] = ima_policy_name
+
     # TODO: This was required to ensure e2e CI tests pass
     if ima_policy == "{}":
         ima_policy_db_format["ima_policy"] = None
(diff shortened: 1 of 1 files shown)


keylime/ima/ima.py
```diff
@@ -675,8 +674,15 @@ def ima_policy_db_contents(ima_policy_name: str, ima_policy: str, tpm_policy: st
             else:
                 ima_policy_db_format["generator"] = IMA_POLICY_GENERATOR.Unknown
 
+            if "timestamp" in ima_policy_dict["allowlist"]["meta"]:
+                meta_timestamp = ima_policy_dict["allowlist"]["meta"]["timestamp"]
+                ima_policy_bytes = ima_policy.replace(meta_timestamp, "XXXX-XX-XX XX:XX:XX.XXXXXX").encode()
```
aplanas, line 679: The date format in json is a bit tricky, and I am not sure how is the string that represent the date. This suggest a template that can be different (but it is not relevant, as the change will be done in any case iif the date is still considered as a string in both representation)

My point is that instead of "XXX..." template use directly the empty string "", to avoid confusion.

aplanas (approved):
Only one nit, but makes sense to me

aplanas (approved):
LGTM. Maybe adding a comment in line 674 to express that the date needs to be removed to make the hash invariant?

THS-on (approved):

THS-on:
The tpm-tests error is adressed by #1248 and has nothing to do with this PR.

PR #1243, opened 2022-12-07
verifier: a (very simple) cache implementation for IMA policies (solves #1167)

maugustosilva:
This is a minimalistic implementation, using a python dictionary named GLOBAL_POLICY_CACHE (name inspired by GLOBAL_TPM_INSTANCE in cloud_verifier_common.py), indexed by agent UUID, to hold the contents (string) of whole ima_policy column as read from the `allowlists` table on the database. While theoretically not memory efficient, indexing the dictionary by agent UUID ensures only a single worker will write a key simultaneously, thus making explicit locks unnecessary.

This implementation was experimentally checked for scalability (more than 100 `agents` per `verifier` with ~ 25MB unique allowlist per `agent`, stored on a database shared across 10 `verifiers`) and also for efficient memory usage (orders of magnitude below the theoretical O(N*M)).

Two simple cleanup routines are also implemented: a) whenever an agent is deleted from a verifier, the corresponding key on the cache is deleted and b) if for any reason the IMA policy checksum changes (e.g., after PR #1184), the old <checksum,contents> key,value pair for a given agent UUID is also deleted.

Main goal here was to get at the “simplest solution that effectively solves the problem without breaking the ability to have shared databases”.

### DIFF
--- a/keylime/cloud_verifier_tornado.py
+++ b/keylime/cloud_verifier_tornado.py
@@ -39,6 +39,8 @@ from keylime.ima import ima
 
 logger = keylime_logging.init_logging("verifier")
 
+GLOBAL_POLICY_CACHE: Dict[str, Dict[str, str]] = {}
+
 set_severity_config(config.getlist("verifier", "severity_labels"), config.getlist("verifier", "severity_policy"))
 
 try:
@@ -123,6 +125,28 @@ def _from_db_obj(agent_db_obj: VerfierMain) -> Dict[str, str]:
         agent_dict[key] = val
 
     return agent_dict
+
+
+def verifier_read_policy_from_cache(stored_agent: VerfierMain) -> str:
+    if stored_agent.agent_id not in GLOBAL_POLICY_CACHE:
+        GLOBAL_POLICY_CACHE[stored_agent.agent_id] = {}
+
+    if stored_agent.ima_policy.checksum not in GLOBAL_POLICY_CACHE[stored_agent.agent_id]:
+        if GLOBAL_POLICY_CACHE[stored_agent.agent_id]:
+            # Perform a cleanup of the contents, IMA policy checksum changed
+            GLOBAL_POLICY_CACHE[stored_agent.agent_id] = {}
+
+        logger.debug(
+            "IMA policy with checksum %s, used by agent %s is not present on policy cache on this verifier, performing SQLAlchemy load",
+            stored_agent.ima_policy.checksum,
+            stored_agent.agent_id,
+        )
+        # Actually contacts the database and load the (large) ima_policy column for "allowlists" table
+        GLOBAL_POLICY_CACHE[stored_agent.agent_id][
+            stored_agent.ima_policy.checksum
+        ] = stored_agent.ima_policy.ima_policy
+
+    return GLOBAL_POLICY_CACHE[stored_agent.agent_id][stored_agent.ima_policy.checksum]
 
 
 def verifier_db_delete_agent(session: Session, agent_id: str) -> None:
@@ -398,6 +422,14 @@ class AgentsHandler(BaseHandler):
             web_util.echo_json_response(self, 404, "agent id associated to this verifier")
             logger.info("DELETE returning 404 response. agent id: %s not associated to this verifer.", agent_id)
             return
+
+        # Cleanup the cache when the agent is deleted. Do it early.
+        if agent_id in GLOBAL_POLICY_CACHE:
+            del GLOBAL_POLICY_CACHE[agent_id]
+            logger.debug(
+                "Cleaned up policy cache from all entries used by agent %s",
+                agent_id,
+            )
 
         op_state = agent.operational_state
         if op_state in (states.SAVED, states.FAILED, states.TERMINATED, states.TENANT_FAILED, states.INVALID_QUOTE):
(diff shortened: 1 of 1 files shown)


keylime/cloud_verifier_tornado.py
```diff
@@ -399,6 +423,14 @@ def delete(self):
             logger.info("DELETE returning 404 response. agent id: %s not associated to this verifer.", agent_id)
             return
 
+        # Cleanup the cache when the agent is deleted. Do it early.
+        if agent_id in GLOBAL_POLICY_CACHE:
+            del GLOBAL_POLICY_CACHE[agent_id]
```
stefanberger, line 428: Can you modify a map concurrently by multiple coroutines without having to lock?

aplanas: I was suspicious so I search a bit, but should not be strictly required for build-in data-types like set or dicts: https://docs.python.org/3/glossary.html#term-global-interpreter-lock

That is why I am a bit reluctant about the other PR. I counted 3 or 4 locks in one of the older reviews, that makes me clearly not smart enough to validate the code.

keylime/cloud_verifier_tornado.py
```diff
@@ -965,7 +997,7 @@ async def invoke_get_quote(
                 rmc.record_create(agent, json_response, ima_policy)
 
             failure = cloud_verifier_common.process_quote_response(
-                agent, ima_policy.ima_policy, json_response["results"], agentAttestState
```
mbestavros, line 968: Why was this `.ima_policy` removed? Should it also have a similar `verifier_read_policy_from_cache` call, like in `process_agent`?

mbestavros (commented):
This looks like it passes tests and doesn't change too much about the IMA codebase, so I have no objections in principle - especially if it improves DB performance. Just one small nit/question.

stefanberger:
This solution is definitely a lot simpler than what I am proposing in https://github.com/keylime/keylime/pull/1177

Here are some thoughts about what drove the implementation of PR #1177 dbprox that is acting as proxy and caching layer to the allowlists Db and what it should be able to do (short and long term) in my opinion (ymmv):
- short-term: the caching layer should transparently act as a proxy to the DB and handle updates of a allowlist entry before making the entry (object) available to the caller
- long-term: provide objects to the caller that holds transformed data rather than the raw DB row. We currently transform text/str to json/dict when we need it and possibly do this with shared allowlists many times per second, once for each caller. This shouldn't be necessary but the transformation should only be done once, either on the top level where it is done now (may need some locking of the object) or on the caching level immediately after reading the data
- encapsulate the logic into classes

galmasi:
I will post performance numbers later today. Preliminary results show that mysql network traffic drops to noise levels (as expected) and memory usage is growing linearly or sublinearly. 


We *may* be seeing an effect of python3 string deduplication, as the allow lists are stored as dictionaries and most/all dictionary entries across the agents are the same (in any sane keylime setup the number of distinct OS images is much, much smaller than the number of keylime agents). But I don't yet have authoritative figures, all I can say is that the current scaling setup (9 verifiers for 1480 agents) no longer crashes and/or floods the network when IMA is on.

aplanas:
> memory usage is growing linearly or sublinearly.

That can be alleviated with the good old `shelve` (https://docs.python.org/3/library/shelve.html) and make in O(1) in terms of memory and speed (with C a bit bigger for speed, as there is a local read)

keylime/cloud_verifier_tornado.py
```diff
@@ -399,6 +423,14 @@ def delete(self):
             logger.info("DELETE returning 404 response. agent id: %s not associated to this verifer.", agent_id)
             return
 
+        # Cleanup the cache when the agent is deleted. Do it early.
```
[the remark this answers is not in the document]
mpeters: I agree that a TTL is in general a better mechanism and is more deterministic and easier to explain.

keylime/cloud_verifier_tornado.py
maugustosilva: I realized, while writing the "Durable Attestation" (DR) code that the only thing needed from `ima_policy` DB object was the (potentially long) string from the `ima_policy` column on the `allowlists` table. Since `cloud_verifier_common.process_quote_response` is reused by DR, just passing a string made both codepaths significantly simpler.

keylime/cloud_verifier_tornado.py
```diff
@@ -39,6 +39,8 @@
 
 logger = keylime_logging.init_logging("verifier")
 
+GLOBAL_POLICY_CACHE = {}
```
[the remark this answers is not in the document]
maugustosilva: Fixed

keylime/cloud_verifier_tornado.py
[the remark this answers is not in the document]
maugustosilva: I certainly agree, but I propose we refine the invalidation mechanisms in subsequent PRs. This one at least fixes the problem described.

keylime/cloud_verifier_tornado.py
maugustosilva: That is what my own search on python documentation have shown (i.e., **not** strictly required for build-in data-types like set or dicts). Anyway, this was tested on a large(ish) deployment (low thousands of `agents`, tens of `verifiers`)

aplanas (approved):

mpeters (approved):

galmasi:
Since this is merged, appending the memory use information only as an afterthought. 

* Total RAM usage for ~ 1500 attested nodes with IMA attestation turned on (including all memory use on the verifiers) works out to maybe 90 GBytes of RAM, or 60 MBytes per node.
* Total RAM usage for the same amount of nodes running merely identity attestation (not IMA) works out to 13 GBytes of RAM total, or 8 MBytes per node (I'm not attaching that figure here).

The differential ("IMA only" RAM) works out to ~ 50 MBytes / node.
The actual IMA allow list works out to 190,000 entries or 20 MBytes total.

image

PR #1248, opened 2022-12-08
api_version: Catch InvalidVersion for packaging v22.0

stefanberger:
Starting with the packaging package v22.0 the parse function does not accept the 'vader' malformed version string anymore that one of our test cases passes but the function now throws an InvaidVersion exception when this string is given. Therefore, catch the InvalidVersion exception on all version.parse() calls and return an appropriate value.

Note: Older versions create a <LegacyVersion('vader')> object.

### DIFF
--- a/keylime/api_version.py
+++ b/keylime/api_version.py
@@ -17,7 +17,10 @@ def current_version() -> str:
 
 
 def latest_minor_version(v: VersionType) -> str:
-    v_obj = version.parse(str(v))
+    try:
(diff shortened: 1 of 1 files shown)


stefanberger:
Everybody will have to rebase their PRs onto this one.

mbestavros (approved):
I'm running into the same failure, and this looks like the fix.

THS-on (approved):

mpeters (approved):

PR #1253, opened 2022-12-09
tpm_abstract: move import of measured_boot into check_pcrs(..)

THS-on:
This is erquired because otherwise the registrar tries to load the verifier configuration.

### DIFF
--- a/keylime/tpm/tpm_abstract.py
+++ b/keylime/tpm/tpm_abstract.py
@@ -12,7 +12,7 @@ try:
 except ImportError:
     from yaml import SafeLoader, SafeDumper
 
-from keylime import config, crypto, json, keylime_logging, measured_boot
+from keylime import config, crypto, json, keylime_logging
 from keylime.agentstates import AgentAttestState
 from keylime.common import algorithms
 from keylime.common.algorithms import Hash
@@ -276,6 +276,9 @@ class AbstractTPM(metaclass=ABCMeta):
         hash_alg: Hash,
     ) -> Failure:
         failure = Failure(Component.PCR_VALIDATION)
+
+        # The measured_boot code loads verifier configuration automatically and therefore cannot be imported globally.
+        from keylime import measured_boot  # pylint: disable=import-outside-toplevel
 
         if isinstance(tpm_policy, str):
             tpm_policy_dict = json.loads(tpm_policy)


THS-on:
I noticed that because we import a custom policy that of course fails, because that code is not deployed on the registrar.

```
 Reading configuration from ['/etc/keylime/verifier.conf']
   File "/usr/local/bin/keylime_registrar", line 5, in <module>
     from keylime.cmd.registrar import main
   File "/usr/local/lib/python3.11/site-packages/keylime/cmd/registrar.py", line 1, in <module>
     from keylime import config, keylime_logging, registrar_common
   File "/usr/local/lib/python3.11/site-packages/keylime/registrar_common.py", line 23, in <module>
     from keylime.tpm.tpm_main import tpm
   File "/usr/local/lib/python3.11/site-packages/keylime/tpm/tpm_main.py", line 25, in <module>
     from keylime.tpm import tpm2_objects, tpm_abstract
   File "/usr/local/lib/python3.11/site-packages/keylime/tpm/tpm_abstract.py", line 15, in <module>
     from keylime import config, crypto, json, keylime_logging, measured_boot
   File "/usr/local/lib/python3.11/site-packages/keylime/measured_boot.py", line 8, in <module>
     from keylime.elchecking.policies import Policy, RefState
   File "/usr/local/lib/python3.11/site-packages/keylime/elchecking/__init__.py", line 9, in <module>
     from . import example
   File "/usr/local/lib/python3.11/site-packages/keylime/elchecking/example.py", line 4, in <module>
     from . import policies, tests
   File "/usr/local/lib/python3.11/site-packages/keylime/elchecking/policies.py", line 111, in <module>
     importlib.import_module(imp, __package__)
   File "/usr/lib64/python3.11/importlib/__init__.py", line 126, in import_module
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 ModuleNotFoundError: No module named 'test'

**Review by aplanas — COMMENTED** [review]
"required" in the commit message

**Comment by THS-on** [comment]
> "required" in the commit message

ups, fixed!

File: keylime/tpm/tpm_abstract.py [hunk 1]
```diff
@@ -277,6 +277,9 @@ def check_pcrs(
     ) -> Failure:
         failure = Failure(Component.PCR_VALIDATION)
 
+        # The measured_boot code loads verifier configuration automatically and therefore cannot be imported globally.
```
aplanas, line 280: I am OK, but this suggest a deeper problem in the architecture: how is possible that we depend in an abstract class on something that is in a config file?

The use of the `config` module should be restricted to the cli entry points, that can pass this object or bits of it when down into the tree. This way we resolve problems like this, when a side effect gets uncontrolled with a single `import`

aplanas (approved):

keylime/tpm/tpm_abstract.py
THS-on: > I am OK, but this suggest a deeper problem in the architecture: how is possible that we depend in an abstract class on something that is in a config file?

The main issue is that the registrar and verifier share the abstract TPM code. Once the agent is removed we can refactor it quite easily, because the feature sets used by the registrar and the verifier are disjunct.

sergio-correia (approved):

keylime/tpm/tpm_abstract.py
stefanberger: > how is possible that we depend in an abstract class on something that is in a config file?

The abstract class is having some methods implemented. They could/should probably be moved into tpm_main.py...

THS-on: I think the idea behind tpm_abstract.py was to share code between the TPM 1.2 and TPM 2.0 implementation. 
Now that we only support TPM 2.0 this split does not make much sense anymore. 

Ideally we split the TPM verification part in to its own module similar to the IMA code and build the main event loop so, that we can support different attestation mechanisms.

My idea is to do that before implementing the push model or removing the atomic quote requirement, but after the Python agent is gone, because that makes it easier to implement.

PR #1254, opened 2022-12-09
Introduce a RetDictType for return type of cmd_exec.run()

stefanberger:
Replace Dict[str, Any] with concrete RetDictType for the return type of cmd_exec.run() and propagate the type to all callers.

### DIFF
--- a/keylime/cmd_exec.py
+++ b/keylime/cmd_exec.py
@@ -1,11 +1,26 @@
 import os
 import subprocess
+import sys
 import time
-from typing import Any, Dict, List, Optional, Sequence, Tuple, Union, cast
+from typing import Dict, List, Optional, Sequence, Tuple, Union, cast
+
+if sys.version_info >= (3, 8):
+    from typing import TypedDict
+else:
+    from typing_extensions import TypedDict
+
 
 EXIT_SUCESS = 0
 
 EnvType = Dict[str, str]
+
+
+class RetDictType(TypedDict):
+    retout: List[bytes]
+    reterr: List[bytes]
+    code: int
+    fileouts: Dict[str, bytes]
+    timing: Dict[str, float]
 
 
 def _execute(cmd: Sequence[str], env: Optional[EnvType] = None, **kwargs) -> Tuple[bytes, bytes, int]:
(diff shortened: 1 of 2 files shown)


stefanberger:
Bummer...
```
Python 3.6.15 (default, Mar  3 2022, 00:00:00) 
[GCC 11.2.1 20220127 (Red Hat 11.2.1-9)]
************* Module keylime.cmd_exec
keylime/cmd_exec.py:4:0: E0611: No name 'TypedDict' in module 'typing' (no-name-in-module)
```

keylime/cmd_exec.py
```diff
@@ -1,19 +1,26 @@
...
+        retout: List[bytes]
+        reterr: List[bytes]
+        code: int
+        fileouts: Dict[str, bytes]
+        timing: Dict[str, float]
+
+except ImportError:
+    from typing import Any
+
+    RetDictType = Dict[str, Any]  # type: ignore
```
stefanberger, line 23: The disadvantage of this solution is that it may force us to add `cast()`s for this case here that are not necessary if `TypedDict` is available

aplanas: Seems that this was introduced in 3.8. Maybe add a note so when we change the minimal version we can drop this?

aplanas (approved):

keylime/cmd_exec.py
THS-on: Yes, I think we are going to add https://pypi.org/project/typing-extensions/ as a dependency.

stefanberger: > Seems that this was introduced in 3.8. Maybe add a note so when we change the minimal version we can drop this?

We could introduce a `cast37()` function that is necessary only for older versions of python and could help us remove it quickly.

> Yes, I think we are going to add https://pypi.org/project/typing-extensions/ as a dependency.

I am surprised that annotation extensions can all be done without extending the interpreter/compiler... So, sounds good and would probably make the idea of `cast37()` obsolete.

THS-on: > I am surprised that annotation extensions can all be done without extending the interpreter/compiler... So, sounds good and would probably make the idea of cast37() obsolete.

I think there are some limitations, but for your use cases this should work.

THS-on: I've updated the other PR to test it.

THS-on (approved):

mpeters (approved):

PR #1255, opened 2022-12-09
elchecking: add support for measured boot when SecureBoot is disabled

### LINKED ISSUE #1197: create_mb_refstate script is failing for measured boot log from Ubuntu machine (VM and also physical - HPE)
# Environment
 * OS / version: Ub22.04
 * Processor architecture: N/A
 * TPM Manufacturer: N/A
 * Keylime version: 6.5.2

# Description 
Running create_mb_refstate script with measured boot log from Ubuntu 22.04 machine (both virtual and physical - HPE) fails with same reason (see in relevant logs)

Expected:
measured_boot_reference_state.json file is properly created from input file (measured boot log)

## Steps to reproduce problem
1. run script as shown below in relevant logs section

# Relevant logs
Failure example:
server ~# ./create_mb_refstate /home/sysadmin/binary_bios_measurements measured_boot_reference_state.json
2022-11-15 14:47:40.901 - keylime.tpm - INFO - TPM2-TOOLS Version: 5.2
 File "/root/./create_mb_refstate", line 173, in <module>
   main()
 File "/root/./create_mb_refstate", line 167, in main
   **get_kernel(events),
 File "/root/./create_mb_refstate", line 87, in get_kernel
   **assert len(out) in [3, 4], "Expected 3 different UEFI applications to be booted (Shim, Grub, Linux)"
AssertionError: Expected 3 different UEFI applications to be booted (Shim, Grub, Linux)**
server ~#

THS-on:
In some setups SecureBoot is disabled. MeasuredBoot should still work in that
case. Most GRUB implementations use the SHIM_LOCK protocol to verify
signatures when SecureBoot is enabled which adds an
EV_EFI_BOOT_SERVICES_APPLICATION event for the vmlinuz image. This does not
happen when SecureBoot is disabled.

This introduces the flag "has_secureboot" to the example policy, when this
option is set to false instead of "kernel_authcode_sha256"
"vmlinuz_plain_sha256" is used for verifying the kernel.
Note that those two options are mutually exclusive.

create_mb_refstate is updated to detect if SecureBoot is disabled and can
generate valid refstates for those new options.

Fixes #1197

### DIFF
--- a/keylime/elchecking/example.py
+++ b/keylime/elchecking/example.py
@@ -67,6 +67,17 @@ allowed_kernel_test = tests.obj_test(
 
 allowed_kernel_list_test = tests.list_test(allowed_kernel_test)
 
+shim_authcode_sha256_no_secureboot = tests.obj_test(
+    shim_authcode_sha256=hex_test,
+    grub_authcode_sha256=hex_test,
+    vmlinuz_plain_sha256=hex_test,
+    initrd_plain_sha256=hex_test,
+    kernel_cmdline=tests.type_test(str),
+)
+
+
+allowed_kernel_list_test_no_secureboot = tests.list_test(shim_authcode_sha256_no_secureboot)
+
 
 class Example(policies.Policy):
     relevant_pcr_indices = frozenset(list(range(10)) + [14])
@@ -80,10 +91,15 @@ class Example(policies.Policy):
         if not isinstance(refstate, dict):
             raise Exception(f"Expected refstate to be a Python dict but instead got this Python value: {refstate!r}")
 
+        has_secureboot = refstate.get("has_secureboot", True)
+
         kernels = refstate.get("kernels")
         if not isinstance(kernels, list):
             raise Exception(f"refstate['kernels'] is {kernels!r} instead of a list")
-        allowed_kernel_list_test(kernels)
+        if has_secureboot:
+            allowed_kernel_list_test(kernels)
+        else:
+            allowed_kernel_list_test_no_secureboot(kernels)
 
         scrtm_and_bios_spec = refstate.get("scrtm_and_bios")
         if scrtm_and_bios_spec is None:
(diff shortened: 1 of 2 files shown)


keylime/elchecking/example.py
```diff
@@ -111,12 +127,20 @@ def refstate_to_test(self, refstate: policies.RefState) -> tests.Test:
         vd_authority = tests.VariableDispatch()
 
         def bsa_test(kernel: typing.Dict[str, str]) -> tests.Test:
+            if not has_secureboot:
```
aplanas, line 130: From where `has_secureboot` comes from? Maybe you want a new field `self.has_secureboot` set early in `refstate_to_test`?

keylime/elchecking/example.py
```diff
@@ -126,6 +150,13 @@ def bsa_test(kernel: typing.Dict[str, str]) -> tests.Test:
             ]
             return tests.Or(tests.TupleTest(*tt), tests.TupleTest(*tt2))
 
+        def vmlinuz_test(vmlinuz_digest: str) -> tests.Test:
+            # Only when SecureBoot is disabled validate the vmlinuz digest measured by GRUB. When SecureBoot is enabled
+            # this was already done using the tests for EV_EFI_BOOT_SERVICES_APPLICATION.
+            if has_secureboot:
```
aplanas, line 156: same

keylime/elchecking/example.py
```diff
@@ -158,6 +191,8 @@ def bsa_test(kernel: typing.Dict[str, str]) -> tests.Test:
         # tpm2-tools versions < 5.2 parsed the GUIDs wrong therefore we include a check for both here
         # For more information see: https://github.com/keylime/keylime/issues/1003
         secure_boot_test = tests.FieldTest("Enabled", tests.StringEqual("Yes"))
+        if not has_secureboot:
```
aplanas, line 194: here too?

keylime/elchecking/example.py
THS-on: It is set early in the `refstate_to_test` method:
https://github.com/keylime/keylime/blob/1266fde473f721b432dd090da421fe86cd6dc19b/keylime/elchecking/example.py#L94

aplanas: I saw that, but I did not realize that those are local function that close over this!

Sorry for the noise!

maugustosilva:
I believe this is the minimal (or very close to it) set of changes required to enable the use of Measured Boot in the case of disabled SecureBooot nodes (somewhat usual for VMs). 

I have tested it and everything works, so we should consider merging it.

However, I would like to remove the possibility of confusion caused by the `--without-secureboot` option, by adding a warning in case this flag is included on the invocation of `create_mb_refstate` on nodes were SecureBoot is enabled (to make clear to the use that this option will be ignored on this case).

THS-on:
> However, I would like to remove the possibility of confusion caused by the --without-secureboot option, by adding a warning in case this flag is included on the invocation of create_mb_refstate on nodes were SecureBoot is enabled (to make clear to the use that this option will be ignored on this case).

@maugustosilva done.

THS-on:
Ideally we wait for feedback from @zkolpet to check if it fully covers his use case, but I'm also fine with merging it early, so can be tested more easily.

mpeters (approved):

PR #1257, opened 2022-12-12
Fix the ima_sign_verification_keys initial datatype 

stefanberger:
…ings

The variable ima_sign_verification_keys is a string, so initialize
it with an empty string rather than an empty list.

### DIFF
--- a/keylime/ima/file_signatures.py
+++ b/keylime/ima/file_signatures.py
@@ -188,8 +188,8 @@ class ImaKeyring:
 
         default_be = backends.default_backend()
 
-        # An empty Db entry comes as a string '[]'. A valid DB entry as a string
-        # ith escaped quotes and needs to be loaded twice
+        # An empty Db entry comes as a string 'null' or previously '[]'. A valid DB entry
+        # is a string with escaped quotes and needs to be loaded twice
         obj = json.loads(stringrepr)
         if isinstance(obj, str):
             obj = json.loads(obj)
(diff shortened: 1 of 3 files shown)


stefanberger:
Getting better...

keylime/ima/file_signatures.py
```diff
@@ -184,21 +184,25 @@ def from_string(stringrepr: str) -> Optional["ImaKeyring"]:
...
         if isinstance(obj, str):
             obj = json.loads(obj)
         if not isinstance(obj, dict):
             return None
 
-        keyids = obj.get("keyids", [])
+        return ImaKeyring.from_json(obj)
+
+    @staticmethod
+    def from_json(obj: Dict[str, Union[List[int], List[str]]]) -> "ImaKeyring":
```
THS-on, line 198: I think `TypedDict` should work here again. I'll try to finish the other PR, so that we can start using it.

stefanberger: Let me pop this patch out and bring it again elsewhere later on. This split-off has time.

THS-on:
@stefanberger the include pattern that works with mypy is:
```
import sys
if sys.version_info >= (3, 8):
    from typing import Literal
else:
    from typing_extensions import Literal
```
trying something like this will not work:
```
try:
    from typing import Literal
except ImportError:
    from typing_extensions import Literal
```

Docs: https://mypy.readthedocs.io/en/stable/runtime_troubles.html#using-new-additions-to-the-typing-module

THS-on (approved):
Having `None` as a zero value is way better for dealing with REST APIs. Then we can potentially also skip None values in a request, but this for another PR.

mpeters (approved):

PR #1259, opened 2022-12-13
codestyle: Fix annotation of mb_measurement_data

stefanberger:

### DIFF
--- a/keylime/measured_boot.py
+++ b/keylime/measured_boot.py
@@ -5,6 +5,7 @@ import traceback
 from typing import Any, Dict, Optional, Set, Tuple
 
 from keylime import config, keylime_logging
+from keylime.elchecking import tests
 from keylime.elchecking.policies import Policy, RefState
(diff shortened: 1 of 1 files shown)


keylime/measured_boot.py
```diff
@@ -72,7 +73,7 @@ def evaluate_policy(
     mb_policy: Optional[Policy],
     mb_policy_name: str,
     mb_refstate_data: RefState,
-    mb_measurement_data: str,
+    mb_measurement_data: tests.Data,
```
stefanberger, line 76: This goes into this here `mb_policy.evaluate(mb_refstate_data, mb_measurement_data)` with signature `def evaluate(self, refstate: RefState, eventlog: tests.Data) -> str:`. A bit odd that our tools didn't detect this...

THS-on: It is because `str` is a strict subset of `tests.Data` so it works.

PR #1263, opened 2022-12-21
Annotate tpm_abstract.py and tpm_main.py and add to mypy

stefanberger:
Annotate tpm_abstract.py and tpm_main.py and add to mypy.

### DIFF
--- a/keylime/tpm/tpm_abstract.py
+++ b/keylime/tpm/tpm_abstract.py
@@ -2,7 +2,7 @@ import codecs
 import os
 import string
 from abc import ABCMeta, abstractmethod
-from typing import Any, Dict, List, Optional, Set, Union
+from typing import Any, Dict, List, Optional, Set, Tuple, Union, cast
 
 import yaml
 
@@ -10,17 +10,20 @@ try:
     from yaml import CSafeDumper as SafeDumper
     from yaml import CSafeLoader as SafeLoader
 except ImportError:
-    from yaml import SafeLoader, SafeDumper
+    from yaml import SafeLoader, SafeDumper  # type: ignore
 
 from keylime import config, crypto, json, keylime_logging, measured_boot
 from keylime.agentstates import AgentAttestState
 from keylime.common import algorithms
 from keylime.common.algorithms import Hash
+from keylime.elchecking.policies import RefState
 from keylime.failure import Component, Failure
 from keylime.ima import ima
 from keylime.ima.file_signatures import ImaKeyrings
 
 logger = keylime_logging.init_logging("tpm")
+
+TPMDataTypes = Union[int, str, bytes]
 
 
 class TPM_Utilities:
@@ -41,7 +44,7 @@ class TPM_Utilities:
 
     @staticmethod
     def readPolicy(configval: str) -> Dict[str, Any]:
-        policy = json.loads(configval)
+        policy: Dict[str, Any] = json.loads(configval)
 
         # compute PCR mask from tpm_policy
         mask = 0
@@ -74,8 +77,6 @@ class AbstractTPM(metaclass=ABCMeta):
     EMPTYMASK: str = "1"
     MAX_NONCE_SIZE: int = 64
 
-    TPMDataTypes = Union[int, str]
-
     need_hw_tpm: bool
     global_tpmdata: Optional[Dict[str, TPMDataTypes]]
     tpmrand_warned: bool
@@ -89,37 +90,37 @@ class AbstractTPM(metaclass=ABCMeta):
         self.global_tpmdata = None
         self.tpmrand_warned = False
         self.defaults = {}
-        self.defaults["hash"] = algorithms.Hash.SHA1
+        self.defaults["hash"] = Hash.SHA1
         self.defaults["encrypt"] = algorithms.Encrypt.RSA
         self.defaults["sign"] = algorithms.Sign.RSASSA
         self.supported = {}
 
     # tpm_initialize
     @abstractmethod
-    def flush_keys(self):
+    def flush_keys(self) -> None:
         pass
 
     @abstractmethod
-    def encryptAIK(self, uuid: str, ek_tpm: bytes, aik_tpm: bytes):
+    def encryptAIK(self, uuid: str, ek_tpm: bytes, aik_tpm: bytes) -> Optional[Tuple[bytes, str]]:
         pass
 
     @abstractmethod
-    def activate_identity(self, keyblob: bytes):
+    def activate_identity(self, keyblob: str) -> Optional[bytes]:
         pass
 
     @abstractmethod
-    def verify_ek(self, ekcert: bytes, tpm_cert_store: str):
+    def verify_ek(self, ekcert: bytes, tpm_cert_store: str) -> bool:
         pass
 
     @abstractmethod
-    def get_tpm_manufacturer(self, output: Optional[List[bytes]] = None):
+    def get_tpm_manufacturer(self, output: Optional[List[bytes]] = None) -> Optional[str]:
         pass
 
     @abstractmethod
-    def is_emulator(self):
+    def is_emulator(self) -> bool:
         pass
 
-    def warn_emulator(self):
+    def warn_emulator(self) -> None:
         if self.is_emulator():
             logger.warning("INSECURE: Keylime is using a software TPM emulator rather than a real hardware TPM.")
             logger.warning("INSECURE: The security of Keylime is currently NOT linked to a hardware root of trust.")
@@ -129,7 +130,7 @@ class AbstractTPM(metaclass=ABCMeta):
     def __read_tpm_data() -> Dict[str, TPMDataTypes]:
         if os.path.exists("tpmdata.yml"):
             with open("tpmdata.yml", "rb") as f:
-                return yaml.load(f, Loader=SafeLoader)
+                return cast(Dict[str, TPMDataTypes], yaml.load(f, Loader=SafeLoader))
         else:
             return {}
 
@@ -151,7 +152,9 @@ class AbstractTPM(metaclass=ABCMeta):
             self.__write_tpm_data()
 
     @abstractmethod
-    def tpm_init(self, self_activate: bool = False, config_pw: Optional[str] = None):
+    def tpm_init(
+        self, self_activate: bool = False, config_pw: Optional[str] = None
+    ) -> Tuple[Optional[bytes], Optional[bytes], Optional[bytes]]:
         pass
 
     # tpm_quote
@@ -163,7 +166,7 @@ class AbstractTPM(metaclass=ABCMeta):
         pcrmask: str = EMPTYMASK,
         hash_alg: Optional[str] = None,
         compress: bool = False,
-    ):
+    ) -> str:
         pass
 
     @abstractmethod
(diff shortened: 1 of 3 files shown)


keylime/tpm/tpm_abstract.py
```diff
@@ -89,37 +90,37 @@ def __init__(self, need_hw_tpm: bool = True) -> None:
         self.global_tpmdata = None
         self.tpmrand_warned = False
         self.defaults = {}
-        self.defaults["hash"] = algorithms.Hash.SHA1
+        self.defaults["hash"] = cast(algorithms.Hash, algorithms.Hash.SHA1)
```
stefanberger, line 93: Without the cast it thinks this is a string.

keylime/tpm/tpm_main.py
```diff
@@ -144,7 +145,7 @@ def __get_pcrs(self):
             raise Exception("get_tpm_algorithms failed with code " + str(code) + ": " + str(errout))
 
         retyaml = config.yaml_to_dict(output, logger=logger)
-        pcrs = {}
+        pcrs: Mapping[str, str] = collections.ChainMap()
```
stefanberger, line 148: Have to use Mapping here instead of ChainMap since ChainMap on python 3.6 doesn't seem to allow `[str, str]`.

keylime/tpm/tpm_main.py
```diff
@@ -849,15 +850,15 @@ def quoterepl(m):
             vendorStr = retyaml["TPM2_PT_VENDOR_STRING_1"]["value"]
         elif "TPM_PT_VENDOR_STRING_1" in retyaml:
             vendorStr = retyaml["TPM_PT_VENDOR_STRING_1"]["as string"].strip()
-        else:
-            vendorStr = ""
```
stefanberger, line 853: I added this previously but it's not necessary if we change the return value to `Optional[str]`

keylime/tpm/tpm_main.py
```diff
@@ -577,7 +578,7 @@ def __create_aik(
                     "unable to parse YAML output of tpm2_createak. Is your tpm2-tools installation up to date?"
                 )
             aik_tpm = retDict["fileouts"][akpubfile.name]
-            if aik_tpm == "":
+            if len(aik_tpm) == 0:
```
stefanberger, line 581: aik_tpm is `bytes` and not a `str`.

keylime/tpm/tpm_abstract.py
aplanas: `Hash` has been imported. You can drop the import, or drop the `algorithms.` part

In any case I wonder why it is the case. According to the doc[1] support of enum in mypy

[1] https://mypy.readthedocs.io/en/stable/literal_types.html#enums

keylime/tpm/tpm_main.py
```diff
@@ -477,7 +478,7 @@ def __create_aik(
                     cmd = ["tpm2_evictcontrol", "-A", "o", "-H", hex(cast(int, aik_handle)), "-P", owner_pw]
                     retDict = self.__run(cmd, raiseOnError=False)
                 else:
-                    cmd = ["tpm2_evictcontrol", "-C", "o", "-c", aik_handle, "-P", owner_pw]
+                    cmd = ["tpm2_evictcontrol", "-C", "o", "-c", cast(str, aik_handle), "-P", owner_pw]
```
aplanas, line 481: maybe do in line 449 a `aik_handle = str(self.get_tpm_metadata("aik_handle"))`

keylime/tpm/tpm_main.py
```diff
@@ -434,15 +435,15 @@ def __get_pub_ek(self) -> None:  # assumes that owner_pw is correct at this poin
...
+    ) -> None:
         if hash_alg is None:
             hash_alg = self.defaults["hash"]
         if asym_alg is None:
             asym_alg = self.defaults["encrypt"]
         if sign_alg is None:
             sign_alg = self.defaults["sign"]
 
-        owner_pw = self.get_tpm_metadata("owner_pw")
+        owner_pw = cast(str, self.get_tpm_metadata("owner_pw"))
```
aplanas, line 446: `owner_pw = str(self.get_tpm_metadata("owner_pw"))`?

keylime/tpm/tpm_main.py
aplanas: `if not aik_tpm:` then?

keylime/tpm/tpm_main.py
```diff
@@ -738,12 +739,12 @@ def activate_identity(self, keyblob: bytes) -> Optional[bytes]:
             secfd, secpath = tempfile.mkstemp(dir=secdir)
             sesspathfd, sesspath = tempfile.mkstemp(dir=secdir)
 
-            apw = self.get_tpm_metadata("aik_pw")
+            apw = cast(str, self.get_tpm_metadata("aik_pw"))
```
aplanas, line 742: `apw = str(self.get_tpm_metadata("aik_pw"))` ?

keylime/tpm/tpm_main.py
```diff
@@ -762,7 +763,7 @@ def activate_identity(self, keyblob: bytes) -> Optional[bytes]:
                 command = [
                     "tpm2_activatecredential",
                     "-c",
-                    aik_keyhandle,
+                    cast(str, aik_keyhandle),
```
aplanas, line 766: That seems unsafe, from one side we cast this variable as `int` and here as `str`. Cast does not change the type.

We need to agree on what we expect for `aik_keyhandle`, that seems to be `int` (there is a call to hex, that only works as if it is this type). That way we can say here `str(aik_keyhandle)` with confidence (and no casting).

keylime/tpm/tpm_main.py
```diff
@@ -913,7 +914,11 @@ def tpm_init(self, self_activate: bool = False, config_pw: Optional[str] = None)
         # if no AIK created, then create one
         self.__create_aik()
 
-        return self.get_tpm_metadata("ekcert"), self.get_tpm_metadata("ek_tpm"), self.get_tpm_metadata("aik_tpm")
+        return (
+            cast(Optional[bytes], self.get_tpm_metadata("ekcert")),
+            cast(Optional[bytes], self.get_tpm_metadata("ek_tpm")),
+            cast(Optional[bytes], self.get_tpm_metadata("aik_tpm")),
```
aplanas, line 920: every cast is a red flag, I am sure that here too

aplanas (changes requested):

keylime/tpm/tpm_main.py
stefanberger: This also works. Is this how we want to generally do it rather than using casts?

aplanas: The issue is that a `cast` does not change the type. I guess that what we want is a type-safe Python, and a `cast` does not help to achieve this goal. It is only a way to bypass the type checker[1]

[1] https://docs.python.org/3/library/typing.html#typing.cast

keylime/tpm/tpm_main.py
stefanberger: aik_handle can also be an int on older versions of the tools: `cmd = ["tpm2_evictcontrol", "-A", "o", "-H", hex(cast(int, aik_handle)), "-P", owner_pw]`

keylime/tpm/tpm_main.py
stefanberger: Would also work.

keylime/tpm/tpm_main.py
stefanberger: aik_handle needs to go onto the command line as a hex number for older tools and is a path to a file (iirc) on recent versions of the tools. I am going to replace this with int() and str() since it can be both. It may be also worth considering removal of support for older versions of the tools.

keylime/tpm/tpm_main.py
aplanas: I am not sure, the `get_tpm_metadata` return the value of a YAML, and AFAIK there is no `bytes` representation in YAML (there are arrays, dates, string, integers ...). So for sure what ever `get_tpm_metadata` it is returning it is not a `bytes`.

What I am falling to communicate is that this cast is not doing any conversion at all. You are pbly returning an object that is a string, but you are telling to the type checker that consider it as a `Optional[bytes]` for the type check about the return value.

If you read this carefully, this would help:

```python
>>> from typing import cast, Any
>>> a:Any = "1"
>>> __annotations__
{'a': typing.Any}
>>> b:int = cast(int, a)
>>> __annotations__
{'a': typing.Any, 'b': <class 'int'>}
>>> b
'1'
```

`b` is type `int` but we are storing a `str`. That is what is wrong with `cast`

stefanberger: We do get `bytes`, This here for example sets the "ekcert":

**https://github.com/keylime/keylime/blob/64e44d859608b69e03c970f30ff9841afe66752d/keylime/tpm/tpm_main.py#L910-L911**

https://github.com/keylime/keylime/blob/64e44d859608b69e03c970f30ff9841afe66752d/keylime/tpm/tpm_main.py#L1451

aplanas: > We do get bytes, This here for example sets the "ekcert":

Exactly, and this is why the `cast` is plain wrong.

Please, bear with me here : )))

Why are we doing all this type annotation work? Because we want an external tool to probe that the constrains imposed by the types are kept, so the invariant that they attach can be assumed. Basically here means that "ekcert" will be `None` or an array of bytes, so I can do for example a `base64.b64encode` without any conversion .

So the idea is that once variable and function signatures are annotated, the tool can follow the flow a make those proofs for us.

For the "ekcert" case we have the local variable `ekcert` with the correct type and the `tpm_init` return signature with the correct annotation. But the proof fails because the function `get_tpm_metadata` has a too generic return type, as has the `dict` internally represented.

Using a `cast` here invalidate the issues that pyright / mypy are pointing wrt to the `{set,get}_tpm_metadata`, the YAML document, or the internal `dict`. For example, I can make a mistake during a refactor and do in line 912 a `ekcert = "mytestcert"`, and `pyright` will not detect the change of type from bytes to string, hiding the exact type of bugs that we want to show.

For the specific `ekcert` case a better solution should be to return `ekcert` directly, without calling the metadata function, so we can connect the type of the variable with the return value, and drop one single cast. A better solution is to fix the metadata API and see if it can be fixed, or transform the `dict` into a `dataclass` with types.

The problem here is that each `cast` is an indication of a different bug that we are hiding, so will require a per-case fix.

AFAIK the tpm_* classes would be deeply refactored, so in this case (and in others) IMHO we should not go too far with the annotations, and wait until the refactor.

Also as a general idea, if there is no other alternative than a `cast`, an `assert` should be there too. But every `cast` is a red flag.

stefanberger: I 'fixed' this now without much refactoring, which should happen in a separate patch. For now I added asserts.

THS-on: When we remove the agent code (hopefully this month), most of the code in tpm_* will be also removed.

After the removal we should discuss on how to refactor the TPM validation and event loop code because both require massive changes to implement for example the push model idea or if we want to support other forms of attestation.

aplanas (approved):

Issue #1262, opened 2022-12-16
keylime_tenant tracebacks when verifier config is not available


kkaarreell:
Is your issue a feature request? If so, please raise it as an [enhancement](https://github.com/keylime/enhancements)

# Environment
 * OS / version: C9S
 * Processor architecture: x86_64
 * TPM Manufacturer: virtual swtpm
 * Keylime version: current upstream [4720332](https://github.com/keylime/keylime/commit/4720332ca9a9ad33e2b1ebea4e6baf89e1b8fcaa)

# Description 
keylime_tenant tracebacks when verifier configuration is not available. This is a regression introduced recently most likely by this MR
https://github.com/keylime/keylime/pull/1255/files

## Expected behavior vs. actual behavior 
keylime_tenant should not require config files of the verifier, registrar or agent

## Steps to reproduce problem
Run keylime_tenant cmf on a system with other config files deleted. E.g. when running all services on localhost
```
# rm -rf /etc/keylime/{agent,registrar,verifier}*
# keylime_tenant -v 127.0.0.1 -t 127.0.0.1 -u d432fbb3-d2f1-4a97-9ef7-75bd81c00000 --verify --allowlist allowlist.txt --exclude excludelist.txt --include payload-module --cert default -c add
```

# Relevant logs
```
Reading configuration from ['/etc/keylime/logging.conf']
Config file not found in ['/etc/keylime/verifier.conf', '/usr/etc/keylime/verifier.conf']. Please see /usr/lib/python3.9/site-packages/keylime/config.py for more details.
Traceback (most recent call last):
  File "/usr/bin/keylime_tenant", line 33, in <module>
    sys.exit(load_entry_point('keylime==6.5.2', 'console_scripts', 'keylime_tenant')())
  File "/usr/bin/keylime_tenant", line 25, in importlib_load_entry_point
    return next(matches).load()
  File "/usr/lib64/python3.9/importlib/metadata.py", line 86, in load
    module = import_module(match.group('module'))
  File "/usr/lib64/python3.9/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1030, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 680, in _load_unlocked
[...]

PR #1265, opened 2023-01-02
Fix that elchecking policy modules are loaded by agent or tenant

THS-on:
- Revert "tpm_abstract: move import of measured_boot into check_pcrs(..)"
- elchecking: load policy modules explicitly

Fixes: #1262

### DIFF
--- a/keylime/cmd/verifier.py
+++ b/keylime/cmd/verifier.py
@@ -1,5 +1,6 @@
 from keylime import cloud_verifier_tornado, config, keylime_logging
 from keylime.common.migrations import apply
+from keylime.elchecking.policies import load_policies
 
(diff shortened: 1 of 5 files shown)


keylime/cmd/verifier.py
```diff
@@ -1,5 +1,6 @@
 from keylime import cloud_verifier_tornado, config, keylime_logging
 from keylime.common.migrations import apply
+from keylime.elchecking.policies import load_policies
```
stefanberger, line 3: instaciating -> instantiating

maugustosilva (approved):

keylime/cmd/verifier.py
THS-on: Fixed

PR #1266, opened 2023-01-03
Add various files to mypy and invert mypy list to unchecked files

stefanberger:
This PR annotates files to be able to add them to mypy and also changes the list of files in mypy.ini to the list of unchecked files.

### DIFF
--- a/keylime/cmd_exec.py
+++ b/keylime/cmd_exec.py
@@ -2,7 +2,7 @@ import os
 import subprocess
 import sys
 import time
-from typing import Dict, List, Optional, Sequence, Tuple, Union, cast
+from typing import Any, Dict, List, Optional, Sequence, Tuple, Union, cast
 
 if sys.version_info >= (3, 8):
     from typing import TypedDict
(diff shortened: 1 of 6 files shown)


keylime/db/keylime_db.py
```diff
@@ -38,14 +41,16 @@ def make_engine(self, service):
 
         self.service = service
 
+        p_sz: Union[str, int]
+        m_ovfl: Union[str, int]
```
aplanas, line 45: Uhm this is only because there is no shadowing in Python. I do not know how far do we want to go with the type annotation, but IMHO a more correct approach it to decide for one type and do the conversion later in the code.

In any case I am OK with the current solution.

stefanberger: In this case I can fix the 5 and 10 to become strings...

stefanberger: Removed the type declarations and changed to "5" and "10".

keylime/cmd_exec.py
```diff
@@ -82,7 +82,7 @@ def run(
 
 # list_contains_substring checks whether a substring is contained in the given
 # list. The list may be the reterr from 'run' and contains bytes-like objects.
-def list_contains_substring(lst: List, substring: str) -> bool:
+def list_contains_substring(lst: List[bytes], substring: str) -> bool:
```
aplanas, line 85: The name, documentation and code suggest me `List[str]`. I propose to evaluate `List[str|bytes]`

aplanas (approved):

keylime/cmd_exec.py
stefanberger: I changed the two `cast()`s for the return values of `_execute()` to be forced to be `bytes` via `assert`s. This includes `reterr` then that must be 'bytes' or maybe 'None' (not covered, yet). Let's see how this turns out after the test cases run.

stefanberger: Should be good though:
```
import os, subprocess
proc=subprocess.Popen(['ls'], env=os.environ, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
out, err = proc.communicate()
type(err)
<class 'bytes'>
```

mpeters:
@aplanas are you happy with these changes now?

aplanas (approved):

aplanas:
@mpeters yes, sorry for the delay!

THS-on (approved):

PR #1267, opened 2023-01-04
pyright: Ignore a line in keylime/config.py due to newer pyright

stefanberger:
Have pyright ignore a line in the code that would otherwise cause the following issue with newer versions of pyright:

  .../keylime/keylime/config.py:23:26 - error: Argument of type "map[Any]" cannot be assigned to parameter "__iterable" of type "SupportsIter[_SupportsNextT@iter]" in function "iter"
    "map[Any]" is incompatible with protocol "SupportsIter[_SupportsNextT@iter]"
      TypeVar "_T_co@SupportsIter" is covariant
        Type "map[Any]" cannot be assigned to type "SupportsKeysAndGetItem[str, Any]"
          "map[Any]" is incompatible with protocol "SupportsKeysAndGetItem[str, Any]"
            "keys" is not present
            "__getitem__" is not present (reportGeneralTypeIssues)
  .../keylime/keylime/config.py:23:26 - error: Argument of type "map[Any]" cannot be assigned to parameter "__iterable" of type "_GetItemIterable[_T@iter]" in function "iter"
    "map[Any]" is incompatible with protocol "_GetItemIterable[_T@iter]"
      "__getitem__" is not present (reportGeneralTypeIssues)

### DIFF
--- a/keylime/config.py
+++ b/keylime/config.py
@@ -21,7 +21,7 @@ def convert(data: Any) -> Any:
     if isinstance(data, bytes):
         return data.decode()
     if isinstance(data, dict):
-        return dict(iter(map(convert, data.items())))
+        return dict(iter(map(convert, data.items())))  # pyright: ignore
     if isinstance(data, tuple):
         return tuple(map(convert, data))
     if isinstance(data, list):
--- a/keylime/config.py
+++ b/keylime/config.py
@@ -3,7 +3,7 @@ import logging
 import os
 import os.path
 from configparser import RawConfigParser
-from typing import Any, Dict, List, Optional, cast
+from typing import Any, Dict, Iterator, List, Optional, cast
 
 import yaml
 
@@ -21,7 +21,7 @@ def convert(data: Any) -> Any:
     if isinstance(data, bytes):
         return data.decode()
     if isinstance(data, dict):
-        return dict(iter(map(convert, data.items())))  # pyright: ignore
+        return dict(cast(Iterator[Any], iter(map(convert, data.items()))))
     if isinstance(data, tuple):
         return tuple(map(convert, data))
     if isinstance(data, list):


aplanas (approved):
What would be the alternative, doing a cast after the `isinstance()`?

mpeters:
Is it possible to fix the version of pyright we use for testing this?

THS-on:
> Is it possible to fix the version of pyright we use for testing this?

@mpeters we could but we should be using the latest version, so that it catches more issues. We are doing the same with pylint.

stefanberger:
> What would be the alternative, doing a cast after the `isinstance()`?

This here seems to work also:
```
return dict(cast(Iterator[Any], iter(map(convert, data.items()))))
```

stefanberger:
I pushed the version with the cast now as a patch on top of the previous solution. Let me know what you prefer and I'll squash the two solutions or remove the 2nd patch.

keylime/config.py
```diff
@@ -19,9 +19,7 @@ def convert(data: Any) -> Any:
     if isinstance(data, bytes):
         return data.decode()
     if isinstance(data, dict):
-        return dict(iter(map(convert, data.items())))  # pyright: ignore
-    if isinstance(data, tuple):
-        return tuple(map(convert, data))
+        return dict(cast(Iterator[Any], iter(map(convert, data.items()))))
```
aplanas, line 22: Did you drop the `tuple` case on purpose?

stefanberger: No, this was deleted by accident. Thanks.

stefanberger: Do we want to tolerate the cast here (not sure whether this is a bug or feature now of pyright) or go with ignoring the line? Or try yet something else?

aplanas: seems indeed as a bug to me? would make sense to keep it and create a bug report in pyright? Depending on the outcome we revert this, or address the issue differently.

stefanberger: It must be a bug related to running pyright under `tox`. When I revert the changes then I get no errors with this command line:

```
# pyright keylime/config.py
WARNING: there is a new pyright version available (v1.1.281 -> v1.1.287).
Please install the new version or set PYRIGHT_PYTHON_FORCE_VERSION to `latest`
[...]
0 errors, 0 warnings, 0 informations
Completed in 1.266sec
```

With `tox -epyright` the reported errors appear. Hm.

stefanberger: [delete previous posting. picked up old pyright version]

stefanberger: https://github.com/microsoft/pyright/issues/4428

stefanberger: Considering the response I got in the pyright project and the fact that they release a new version ever week (Monday), this problem will go away next week (latest) since it's a bug in pyright.

stefanberger: An anticipation of the new pyright release I am closing this PR now.

aplanas: Thanks for creating the upstream bug. This was really helpful!

PR #1269, opened 2023-01-05
codestyle: Annotate the keylime tenant tool and 2 dependencies

stefanberger:
This PR annotates the keylime tenant tool and 2 dependencies.

### DIFF
--- a/keylime/registrar_client.py
+++ b/keylime/registrar_client.py
@@ -1,15 +1,41 @@
 import logging
+import ssl
 import sys
+from typing import Any, Dict, Optional, Union
 
 from keylime import api_version as keylime_api_version
 from keylime import crypto, json, keylime_logging
 from keylime.requests_client import RequestsClient
 
+if sys.version_info >= (3, 8):
+    from typing import TypedDict
+else:
+    from typing_extensions import TypedDict
+
+if sys.version_info >= (3, 11):
+    from typing import NotRequired
+else:
+    from typing_extensions import NotRequired
+
+
+class RegistrarData(TypedDict):
+    ip: Optional[str]
+    port: Optional[str]
+    regcount: int
+    mtls_cert: Optional[str]
+    aik_tpm: str
+    ek_tpm: str
+    ekcert: Optional[str]
+    provider_keys: NotRequired[Dict[str, str]]
+
+
 logger = keylime_logging.init_logging("registrar_client")
 api_version = keylime_api_version.current_version()
 
 
-def getData(registrar_ip, registrar_port, agent_id, tls_context):
+def getData(
+    registrar_ip: str, registrar_port: str, agent_id: str, tls_context: Optional[ssl.SSLContext]
+) -> Optional[RegistrarData]:
     """
     Get the agent data from the registrar.
 
@@ -64,7 +90,20 @@ def getData(registrar_ip, registrar_port, agent_id, tls_context):
             logger.critical("Error: did not receive port from Registrar Server.")
             return None
 
-        return response_body["results"]
+        r = response_body["results"]
+        res: RegistrarData = {
+            "aik_tpm": r["aik_tpm"],
+            "regcount": r["regcount"],
+            "ek_tpm": r["ek_tpm"],
+            "ip": r["ip"],
+            "port": r["port"],
+            "mtls_cert": r.get("mtls_cert"),
+            "ekcert": r.get("ekcert"),
+        }
+        if "provider_keys" in r:
+            res["provider_keys"] = r["provider_keys"]
+
+        return res
 
     except AttributeError as e:
         if response and response.status_code == 503:
@@ -79,8 +118,16 @@ def getData(registrar_ip, registrar_port, agent_id, tls_context):
 
 
 def doRegisterAgent(
-    registrar_ip, registrar_port, agent_id, ek_tpm, ekcert, aik_tpm, mtls_cert=None, contact_ip=None, contact_port=None
-):
+    registrar_ip: str,
+    registrar_port: str,
+    agent_id: str,
+    ek_tpm: bytes,
+    ekcert: Optional[Union[bytes, str]],
+    aik_tpm: bytes,
+    mtls_cert: Optional[bytes] = None,
+    contact_ip: Optional[str] = None,
+    contact_port: Optional[str] = None,
+) -> Optional[str]:
     """
     Register the agent with the registrar
 
@@ -89,7 +136,7 @@ def doRegisterAgent(
     :returns: base64 encoded blob containing the aik_tpm name and a challenge. Is encrypted with ek_tpm.
     """
 
-    data = {
+    data: Dict[str, Any] = {
         "ekcert": ekcert,
         "aik_tpm": aik_tpm,
     }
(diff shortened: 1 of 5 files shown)


stefanberger:
This really is just a draft for now...

keylime/registrar_client.py
```diff
@@ -64,7 +68,7 @@ def getData(registrar_ip, registrar_port, agent_id, tls_context):
             logger.critical("Error: did not receive port from Registrar Server.")
             return None
 
-        return response_body["results"]
+        return cast(tenant.RegistrarData, response_body["results"])
```
aplanas, line 71: I think that this cast has the same problem. Did you try to write `response_body: tenant.RegistrarData = response.json()`?

But I guess that this would not work either. There is no guarantee that the json is storing a `RegistrarData`.  I think that good approach is to create a function that convert a dict into a RegistrarData, like a sort of type-safe dict.

For example, I found that `inspect` can be used: https://codereview.stackexchange.com/questions/200413/type-safe-json-deserialization

stefanberger: I pushed an update. Please have a look.

keylime/registrar_client.py
```diff
@@ -64,7 +68,20 @@ def getData(registrar_ip, registrar_port, agent_id, tls_context):
...
 
-        return response_body["results"]
+        r = response_body["results"]
+        res: tenant.RegistrarData = {
+            "aik_tpm": r["aik_tpm"],
+            "regcount": r["regcount"],
+            "ek_tpm": r["ek_tpm"],
+            "ip": r["ip"],
+            "port": r["port"],
+            "mtls_cert": r.get("mtls_cert"),
```
stefanberger, line 78: This line here in tenant.py makes me think that "mtls_cert" must be there but can be None:

https://github.com/keylime/keylime/blob/9dae00e519ba5f37cc7b8bfbebdb2ad6953f03bc/keylime/tenant.py#L951

However, the code block this is in is surrounded with a catcher for an `Exception` that could theoretically catch access to an unavailable dict key.

keylime/registrar_client.py
```diff
@@ -64,7 +68,20 @@ def getData(registrar_ip, registrar_port, agent_id, tls_context):
...
+            "ekcert": r.get("ekcert"),
```
stefanberger, line 79: From looking through tenant.py "ekcert" should always be there but may be None. However, this one is not checked in this function above.

keylime/registrar_client.py
```diff
@@ -131,7 +178,7 @@ def doRegisterAgent(
             logger.critical("Error: did not receive blob from Registrar Server: %s", response.status_code)
             return None
 
-        return response_body["results"]["blob"]
+        return cast(bytes, response_body["results"]["blob"])
```
aplanas, line 181: Uhm I am wondering why this cast. IIUC this comes from a `request.json()` that internally calls `loads`. If I read the code correctly it uses the Pyhon JSON module if `simplejson` is not installed. In my local installation, `simplejson` always return a `str` for `dumps`.

In any case I would return always `str`, as is more aligned with Python (for example, note that `dumps` returns a string, and `loads` cannot serialize `bytes`).

aplanas (changes requested):

keylime/registrar_client.py
stefanberger: Correct. This function has to return an `Optional[str]`. I now fixed the annotation of tpm_main's activate_identity() to also take a `str`.

keylime/registrar_client.py
```diff
@@ -64,7 +90,20 @@ def getData(registrar_ip, registrar_port, agent_id, tls_context):
             logger.critical("Error: did not receive port from Registrar Server.")
             return None
 
-        return response_body["results"]
+        r = response_body["results"]
+        res: RegistrarData = {
+            "aik_tpm": r["aik_tpm"],
```
aplanas, line 95: Is this working with pyright?

I was expecting more like: `"aik_tpm": bytes(r["aik_tpm"], encoding="utf-8"), "port": int(r["port"]), ...`

stefanberger: Yes, pyright passes.

aplanas: That is wild then!

A JSON should be considered mostly a `dict[str, Any]`, hidden in some C wraps. I guess that there is some thing that I do not understand in the subtyping / variance / contra-variance algorithm used in pyright (or less probably that in some time it will complain here).

Anyway :+1:

stefanberger: > That is wild then!

Indeed! I was surprised here as well that it take entries from a map where it doesn't know what the individual values' data types are and doesn't at least require a cast.

keylime/registrar_client.py
```diff
@@ -1,15 +1,41 @@
...
 
+if sys.version_info >= (3, 8):
+    from typing import TypedDict
+else:
+    from typing_extensions import TypedDict
+
+if sys.version_info >= (3, 11):
+    from typing import NotRequired
+else:
+    from typing_extensions import NotRequired
```
stefanberger, line 18: @kkaarreell : This one here now creates an issue on the test machines.

stefanberger: I have this here in requirements. Is there anything else that the test suite needs?

```
typing-extensions>=4.0.1 ; python_version < '3.11' #PSF-2.0
```

kkaarreell: Packit e2e tests are not using these dependencies but rather install them in 
https://github.com/RedHat-SP-Security/keylime-tests/blob/main/setup/install_upstream_keylime/test.sh#L40
as we prefer distribution packages. 
I need to check what do we have in current Fedora releases before fixing it.

stefanberger: It looks like Fedora 37 would have the right version of the package (v4.2.0) but on Fedora 36 it's only v3.10: https://pkgs.org/download/python3-typing-extensions

kkaarreell: I did required changes in https://github.com/RedHat-SP-Security/keylime-tests/pull/301 but tests are failing on Rawhide due to missing keylime_ima_emulator service. I need to investigate it, it's weird.

kkaarreell: Seems that the new systemd (intentionally?) changed the output of `systemctl is-enabled`. I have a patch available, it is being crunched in CI and I will merge it once it passes.

kkaarreell: Done, please try to rerun tests by adding a comment "/packit test"

kkaarreell: failure on testing-farm:fedora-36-x86_64  is something I am investigating right now. You can try to rerun it once or twice and it shoud pass (/packit retest-failed) or eventually ignore.

stefanberger: This time it passed.

maugustosilva (approved):

mypy.ini
```diff
@@ -44,9 +44,6 @@ ignore_errors = True
 [mypy-keylime.revocation_notifier]
 ignore_errors = True
 
-[mypy-keylime.tenant]
-ignore_errors = True
-
```
stefanberger, line 49: The registrar_client.py can also be removed. It probably got lost during rebasing.

aplanas (approved):

PR #1270, opened 2023-01-05
tenant: Remove code hashing a public key and using hash as UUID

stefanberger:
Remove the code that was trying to hash a public key and use the resulting hash as agent UUID. This most likely never worked since the uuid (which is supposed to be the content of a PEM public key file) is passed from the command line as string and hashlib.sha256() does not accept a string as an argument.


THS-on (approved):
LGTM

Issue #960, opened 2022-04-05
keylime_tenant initiated action on verifier timeouts occasionally


kkaarreell:
Is your an issue a feature request? If so, please raise it as an [enhancement](https://github.com/keylime/enhancements)

# Environment
 * OS / version: Fedora 35
 * Processor architecture: x86_64
 * TPM Manufacturer: swtpm
 * Keylime version: current upstream [3ab5531](https://github.com/keylime/keylime/commit/3ab55312006e54c32cb4ccb58b577c43ae0dd459)

# Description 
There are occasional failures in CI where keylime_tenant command timeouts.
For example in /functional/keylime_tenant-commands-on-localhost

```
::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
::   -c delete
::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::

:: [ 22:00:17 ] :: [  BEGIN   ] :: Running 'lime_keylime_tenant -c delete'
Reading configuration from ['/etc/keylime.conf']
2022-04-05 22:00:18.460 - keylime.tpm - INFO - TPM2-TOOLS Version: 5.2
2022-04-05 22:00:18.467 - keylime.tenant - INFO - Setting up client TLS in /var/lib/keylime/cv_ca
2022-04-05 22:00:18.467 - keylime.tenant - WARNING - Using default UUID d432fbb3-d2f1-4a97-9ef7-75bd81c00000
2022-04-05 22:00:18.507 - keylime.tenant - INFO - Agent Info:
[...]

PR #1271, opened 2023-01-10
tenant: reliable and consistent add/delete operations (fixes #1158)

### LINKED ISSUE #1158: Operations `add` and `delete` via `keylime_tenant` CLI need to produce consistent results

# Environment
* OS / version: Fedora Core 35
* TPM Manufacturer: swtpm
* Keylime version: 6.5.1

# Description 
When running the command `keylime_tenant -c add` in a large cluster (~ 1,500 `agents`), with a relatively long allowlist (~ 80,000 entries) and an external `mysql` database, it can be observed  that the API returns the correct response code, the command end with zero exit code, and yet, there are occasions when the actual entry pertaining the `agent`  has not yet been written to the database, causing a subsequent immediate `keylime_tenant -c cvstatus` execution to falsely report the `agent` as "not yet added to a `verifier`".

For both `keylime_tenant -c add` and ` keylime_tenant -c delete` (or `keylime_tenant -c update`), the CLI should return only after querying the actual status of the `agent` from the `verifier` via a new API call, and then ensuring that desired state has been reached. Currently, `keylime_tenant -c delete` already does it via a crude "for" loop, but this should be properly implemented for all the aforementioned commands, using a backoff algorithm.

## Steps to reproduce problem
So far, this has only detectable under the specific conditions listed at the description

# Relevant logs
NA

maugustosilva:
As discussed in issue #1158, there are real-world scenarios where "add" and/or "delete" operations via `keylime_tenant` CLI causes it to return before the data was written by `verifier` on the database. This causes inconsistent "status" (also via `keylime_tenant`) for a given `agent`, in case the CLI invoked within a small time window after "add"/"delete".

Given the fact the code for `keylime_tenant` was somewhat messy (e.g., a fixed 12-step loop with 0.4 seconds between attempts on "delete"), this new code uses the already existing "retry_interval", "exponential_backoff", and "max_retries" attributes to implement an ordered "status" polling on the `verifier` right after an "add" or "delete" operation, only returning when the desired state and the actual state effectivelly match (or produce an explicit error message in case it does not happen after multiple retries)

### DIFF
--- a/keylime/tenant.py
+++ b/keylime/tenant.py
@@ -103,6 +103,9 @@ class Tenant:
         self.api_version = keylime_api_version.current_version()
         self.enable_agent_mtls = config.getboolean("tenant", "enable_agent_mtls")
         self.request_timeout = config.getint("tenant", "request_timeout", fallback=60)
+        self.retry_interval = config.getfloat("tenant", "retry_interval")
+        self.exponential_backoff = config.getboolean("tenant", "exponential_backoff")
+        self.maxr = config.getint("tenant", "max_retries")
 
         logger.info("Setting up client TLS...")
         (cert, key, trusted_ca, key_password), verify_server_cert = web_util.get_tls_options(
@@ -582,6 +585,39 @@ class Tenant:
                 response.text,
             )
             sys.exit()
+        else:
+            numtries = 0
+            added = False
+
+            while not added:
+                reponse_json = self.do_cvstatus()
+                if reponse_json["code"] != 200:
+                    numtries += 1
+                    if numtries >= self.maxr:
+                        logger.error(
+                            "Agent ID %s still not added to the Verifier after %d tries",
+                            self.agent_uuid,
+                            numtries,
+                        )
+                        sys.exit()
+                    next_retry = retry.retry_time(self.exponential_backoff, self.retry_interval, numtries, logger)
+                    logger.info(
+                        "Agent ID %s still not added to the Verifier at try %d of %d , trying again in %s seconds...",
+                        self.agent_uuid,
+                        numtries,
+                        self.maxr,
+                        next_retry,
+                    )
+                    time.sleep(next_retry)
+                else:
+                    added = True
+
+            if added:
+                logger.info(
+                    "Agent ID %s added to the Verifier after %d tries",
+                    self.agent_uuid,
+                    numtries,
+                )
 
     def do_cvstatus(self):
         """Perform operational state look up for agent on the verifier"""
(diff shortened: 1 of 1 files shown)


THS-on:
The general code LGTM. I think "appear" is not ideal for the messages because it suggests that stuff can get lost somewhere. I'm for writing it in a more active form e.g.: "Agent is still not deleted from the verifier's database".

Some of the status messages changes, so that is the reason why the e2e test fail. @kkaarreell what is the best way to change those?

We should be consistent with the `sys.exit()` calls, because some of the `do_*` functions return a dict and others just exit.
I think we should move to a model of either returning data or throwing an error exception that then can be handled, but that is for another PR.

keylime/tenant.py
```diff
@@ -103,6 +103,9 @@ def __init__(self):
         self.api_version = keylime_api_version.current_version()
         self.enable_agent_mtls = config.getboolean("tenant", "enable_agent_mtls")
         self.request_timeout = config.getint("tenant", "request_timeout", fallback=60)
+        self.retry_interval = config.getfloat("tenant", "retry_interval")
+        self.exponential_backoff = config.getboolean("tenant", "exponential_backoff")
+        self.maxr = config.getint("tenant", "max_retries")
```
aplanas, line 108: Are those new options? Should we do a config migration? Add some defaults?

THS-on: There were already there, so no migration or default options needed: https://github.com/keylime/keylime/blob/9dae00e519ba5f37cc7b8bfbebdb2ad6953f03bc/templates/2.0/tenant.j2#L78

keylime/tenant.py
```diff
@@ -582,6 +585,39 @@ def do_cvadd(self):
...
+            numtries = 0
+            added = False
+
+            while not added:
+                reponse_json = self.do_cvstatus()
+                if reponse_json["code"] != 200:
+                    numtries += 1
+                    if numtries >= self.maxr:
+                        logger.error(
+                            'Agent ID %s still not appear as "added" on Verifier\'s database after %d tries',
```
mpeters, line 598: I agree with Thore about using "appear", it's too wishy-washy.

mpeters (changes requested):
Except for the wording on the user messages, this LGTM

keylime/tenant.py
maugustosilva: fixed :-)

keylime/tenant.py
```diff
@@ -582,6 +585,39 @@ def do_cvadd(self):
...
+            numtries = 0
+            added = False
+
+            while not added:
+                reponse_json = self.do_cvstatus()
+                if reponse_json["code"] != 200:
+                    numtries += 1
+                    if numtries >= self.maxr:
+                        logger.error(
+                            'Agent ID %s status is still not "added" on Verifier\'s database after %d tries',
```
mpeters, line 598: I still think this isn't direct enough. Maybe something like "Agent ID %s still not added to Verifier after %d tries", then repeat for the rest. We don't need to tell the user we keep track of it with a specific status value or how we track that internally, just the logical action of was it added or not.

mpeters (changes requested):

keylime/tenant.py
maugustosilva: Fixed now

THS-on (approved):
Before merging the tests should be still adjusted to the new messages, but otherwise LGTM. Thanks for adding that because during testing I also ran into those issues sometimes. 

Also it should also fix: https://github.com/keylime/keylime/issues/960

maugustosilva:
> Before merging the tests should be still adjusted to the new messages, but otherwise LGTM. Thanks for adding that because during testing I also ran into those issues sometimes.
> 
> Also it should also fix: #960

I have fixed the messages. Now I see the `rawhide` error that affects everything (@kkaarreell sent an email to devs, waiting for an answer) and an unrelated error on `centos-stram-9-x86_64

THS-on:
> I have fixed the messages. 

Yes I've only saw the old test run when I did the review. The other failing tests have nothing to do with this, so it is ready to merge. (I cannot merge it because @mpeters still needs to approve this, because he requested some changes)

mpeters (approved):

PR #1274, opened 2023-01-12
config: support override via environment variables

THS-on:
The variables have the following structure:
KEYLIME_{COMPONENT}_{OPTION}
or if there is a non default section:
KEYLIME_{COMPONENT}_{SECTION}_{OPTION}

All letters of the variable name need to be capitalized.


keylime/config.py
```diff
@@ -234,11 +234,23 @@ def get_config(component: str) -> RawConfigParser:
...
+
+
 def getlist(component: str, option: str, section: Optional[str] = None) -> List[Any]:
+    env_value = _get_env(component, option, section)
     if not section:
         section = component
 
-    read = get_config(component).get(section, option).strip('" ')
+    if env_value:
+        read = env_value.strip('" ')
```
mpeters, line 251: Since we're stripping out quotes and spaces in a couple places for env vars, should it be moved into `_get_env()`?

keylime/config.py
```diff
@@ -257,37 +269,59 @@ def getlist(component: str, option: str, section: Optional[str] = None) -> List[
...
     return get_config(component).getint(section, option, fallback=fallback)
 
 
 def getboolean(component: str, option: str, section: Optional[str] = None, fallback: bool = False) -> bool:
+    env_value = _get_env(component, option, section)
     if not section:
         section = component
 
+    if env_value:
+        if env_value.lower() not in RawConfigParser.BOOLEAN_STATES:
```
mpeters, line 299: Should we just do the `lower()` once and store the value? I know it's not a huge optimization, but I hate running things twice that we don't need to.

keylime/config.py
THS-on: We don't do it for the boolean, float and int conversions, so I don't think I can move into `_get_env()`.

mpeters: I guess I should have asked why we don't do it for those? I know spaces shouldn't really be there, but if people are going to accidentally quote strings, they might do it for ints/floats/booleans, especially if it's some automated config management thing happening.

THS-on: Good question. @ansasaki any inpunt on how this should be handled?

ansasaki: I think it would be less error prone to remove quotes/spaces from all strings directly obtained from the config file.

keylime/config.py
```diff
@@ -234,11 +234,23 @@ def get_config(component: str) -> RawConfigParser:
     return _config[component]
 
 
+def _get_env(component: str, option: str, section: Optional[str]) -> Optional[str]:
+    env_name = f"KEYLIME_{component.upper()}"
+    if section:
+        env_name += f"_{section.upper()}"
+    env_name += f"_{option.upper()}"
```
aplanas, line 241: nit:
```python3
opt_section = f"_{section}" if section else ""
env_name = f"KEYLIME_{component.upper()}{opt_section.upper()}_{option.upper()}"
```

aplanas (approved):
LGTM!

I personally would go in a different direction (not tested!!):

```python3
class EnvConfigParser(RawConfigParser):
    def _component(self):
        global _config
        # TODO: this is O(n), optimize it to memoize the component, as this
        # will be called in each `.get()`
        for k, v in _config.items():
            if v == self:
                return k
        # Non-reachable?
        raise ValueError("Component not found")

    def _env_name(self, section, option):
        opt_section = f"_{section}" if section else ""
        return f"KEYLIME_{self._component().upper()}{opt_section.upper()}_{option.upper()}"
        return os.environ.get(env_name, None)

    def get(self, section, option, *, raw=False, vars=None, fallback=_UNSET):
        return os.environ(self._env_name(section, option), None) or super().get(section, option, raw=raw, vars=vars, fallback=fallback)

    def has_option(self, section, option):
        return self._env_name(section, option) in os.environ or super().has_option(section, option)
```

Later change the types from `RawConfigParser` to `EnvConfigParser` and left the rest untouched.

But in any case this seems also OK!

THS-on:
@aplanas I don't know how good of an idea is it to override the `get(..)` method because it supports some special handling options.

mpeters (approved):

maugustosilva (approved):

maugustosilva:
@THS-on indeed, it was simpler (and more elegant) than I initially thought. Fully tested here internally, works as advertised.

LGTM

PR #1277, opened 2023-01-17
tenant: fix the exit code for `bulkinfo` operation

maugustosilva:
We have found a scenario (database connection dropped) where a wrong answer from the `verifier` still results in a zero exit code for `keylime_tenant -c bulkinfo`. We fix it by checking the REST response code and then raising an `UserError` exception (which should be used in place of `sys.exit()` on several places inside `tenant.py` anyway)

### DIFF
--- a/keylime/tenant.py
+++ b/keylime/tenant.py
@@ -692,24 +692,6 @@ class Tenant:
 
         response_json = Tenant._jsonify_response(response, print_response=False)
 
-        if response.status_code == 503:
-            logger.error(
-                "Cannot connect to Verifier at %s with Port %s. Connection refused.",
-                self.verifier_ip,
-                self.verifier_port,
-            )
-            return response_json
-        if response.status_code == 504:
-            logger.error("Verifier at %s with Port %s timed out.", self.verifier_ip, self.verifier_port)
-            return response_json
-        if response.status_code == 404:
-            logger.info(
-                "Verifier at %s with Port %s does not have agent %s.",
-                self.verifier_ip,
-                self.verifier_port,
-                self.agent_uuid,
-            )
-            return response_json
         if response.status_code == 200:
 
             for agent in response_json["results"].keys():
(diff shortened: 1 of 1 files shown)


keylime/tenant.py
```diff
@@ -692,16 +692,17 @@ def do_cvbulkinfo(self):
...
+        if response.status_code == 500:
+            raise UserError(f"Unexpected response from Verifier {self.verifier_ip} on port {self.verifier_port}")
+
         if response.status_code == 503:
-            logger.error(
-                "Cannot connect to Verifier at %s with Port %s. Connection refused.",
-                self.verifier_ip,
-                self.verifier_port,
+            raise UserError(
+                f"Cannot connect to Verifier at {self.verifier_ip} on port {self.verifier_port}. Connection refused."
```
stefanberger, line 700: An existing issue: this should 'service is unavailable' rather than 'connection refused', which I find misleading.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503

keylime/tenant.py
```diff
@@ -692,16 +692,17 @@ def do_cvbulkinfo(self):
 
         response_json = Tenant._jsonify_response(response, print_response=False)
 
+        if response.status_code == 500:
+            raise UserError(f"Unexpected response from Verifier {self.verifier_ip} on port {self.verifier_port}")
```
stefanberger, line 696: I looks like the logger.info at the bottom (line 725) should probably be converted to a catch-all that includes this 500...

keylime/tenant.py
maugustosilva: Good point, I did not change this message, as it was not part of the original PR, but it is a small fix.

keylime/tenant.py
maugustosilva: yup, done. Went "minimalistic" on the response code processing instead

THS-on (approved):

PR #1279, opened 2023-01-23
Update and extend experimental create_policy tool

stefanberger:
This PR updates and extends the experimental create_policy tool that one can use to add keys or hashes to a Keylime runtime policy.

### DIFF
--- a/scripts/create_policy
+++ b/scripts/create_policy
@@ -235,10 +235,12 @@ def main(argv):
 
             # Cherry-pick from base policy what is supported and merge into policy
             policy["digests"] = base_policy.get("digests", {})
+            policy["excludes"] = base_policy.get("excludes", [])
             policy["keyrings"] = base_policy.get("keyrings", {})
             policy["ima-buf"] = base_policy.get("ima-buf", {})
             ignored_keyrings = base_policy.get("ima", {}).get("ignored_keyrings", [])
             policy["ima"]["ignored_keyrings"] = ignored_keyrings
+            policy["verification-keys"] = base_policy.get("verification-keys", "")
         except (PermissionError, FileNotFoundError) as ex:
             print(f"An error occurred while loading the policy: {ex}")
             ret = 1
--- a/scripts/create_policy
+++ b/scripts/create_policy
@@ -10,6 +10,8 @@ import sys
 from cryptography import x509
 from cryptography.hazmat import backends
 
+from keylime.ima import ima
+
 IMA_MEASUREMENT_LIST = "/sys/kernel/security/ima/ascii_runtime_measurements"
 IGNORED_KEYRINGS = []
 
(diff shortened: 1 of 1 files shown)


THS-on:
The changes LGTM. Do we want to make this tool an official part of the Keylime commandline tools?

THS-on (approved):

scripts/create_policy
```diff
@@ -129,15 +131,46 @@ def process_ima_buf_in_measurement_list(
...
+        policy["verification-keys"] = verification_key_list
+
+    return policy
+
+
 def main(argv):
     """main"""
-    parser = argparse.ArgumentParser(argv[0])
+    parser = argparse.ArgumentParser(
+        prog=argv[0], description="This is an experimental tool for adding items to a Keylime's IMA runtime policy"
```
aplanas, line 166: `prog=argv[0]` can be dropped, it is taken by default

scripts/create_policy
```diff
@@ -177,18 +210,20 @@ def main(argv):
...
+        "-A",
+        "--add-ima-signature-verification-key",
+        action="append",
+        dest="ima_signature_keys",
+        default=[],
+        help="Add the given IMA signature verification key to the Keylime-internal 'tenant_keyring'; "
+        "the key should be an x509 certificate in DER or PEM format but may also be a public or private key file; "
+        "this option may be passed multiple times",
+    )
     args = parser.parse_args(argv[1:])
```
aplanas, line 223: The parameter `argv[1:]` can also be dropped

scripts/create_policy
```diff
@@ -129,15 +131,46 @@ def process_ima_buf_in_measurement_list(
     return keyrings_map, ima_buf_map, ret
 
 
+def process_signature_verification_keys(verification_keys, policy):
+    """Add the given keys (x509 certificates) to keyring"""
+
+    verification_key_list = None
+    if verification_keys:
+        if policy.get("verification-keys", None):
```
aplanas, line 139: If `policy` is a `dict`, `None` can be elided

aplanas (approved):

mpeters:
> The changes LGTM. Do we want to make this tool an official part of the Keylime commandline tools?

I think it's fine for it to still be experimental and not installed as part of keylime. Do we reference it in documentation anywhere?

scripts/create_policy
stefanberger: Fixed.

scripts/create_policy
stefanberger: Thanks. Fixed.

THS-on:
> Do we reference it in documentation anywhere?

@mpeters I don't think so, but I'm also currently rewriting the section on IMA because it is now out of date.

THS-on:
Failure has nothing to do with change, so I'll merge it.

Issue #1281, opened 2023-01-25
keylime_agent_secure.mount has a wrong filename


kkaarreell:
Is your issue a feature request? If so, please raise it as an [enhancement](https://github.com/keylime/enhancements)

# Environment
 * OS / version: Fedora Rawhide
 * Processor architecture: x86_64
 * Keylime version: current upstream ()

# Description 
`keylime_agent_secure.mount` is not correct name for a unit file mounting `/var/lib/keylime/secure`.
During system reboot one can see the following message in the console
```
systemd[1]: keylime_agent_secure.mount: Where= setting doesn't match unit name. Refusing.
```
The proper name for the unit file should be `var-lib-keylime-secure.mount`
This is explained in
https://unix.stackexchange.com/questions/283442/systemd-mount-fails-where-setting-doesnt-match-unit-name

## Expected behavior vs. actual behavior 
no error in a console, unit file has a correct filename

## Steps to reproduce problem
1. start the agent (so it mounts the secure mount)
2. reboot
3. watch console

PR #1282, opened 2023-01-25
Rename keylime_agent_secure.mount to comply policy

kkaarreell:
Filename keylime_agent_secure.mount does not comply systemd policy.

Per systemd.mount man page:

    Where=

    Takes an absolute path of a directory of the mount point.
    If the mount point does not exist at the time of mounting,
    it is created. This string must be reflected in the unit
    filename.

Correct name of a unit file mounting /var/lib/keylime/secure is var-lib-keylime-secure.mount


kkaarreell:
This resolves https://github.com/keylime/keylime/issues/1281

THS-on (approved):
LGTM

sergio-correia:
We should probably update the `installer.sh` script that lives under `services/`, as I think it uses the old unit name.

sergio-correia:
Please dismiss the previous comment; somehow I missed that the file in question was updated within this PR...

aplanas:
Ouch my bad. I think that I make wrong name at the beginning and I renamed that later in the package.

PR #1289, opened 2023-01-30
remove Python agent

THS-on:
Things that also will be removed:
*  IMA emulator: uses the old TPM code, Rust agent provides alternative implementation
* zeroMQ client code: was mainly used by the agent
* listen command in ca_util: removal was already agreed, uses zeroMQ client code
* secure_mount abstraction: was only used by the agent code


ansasaki:
@THS-on Thanks for working on this!

A point I'm concerned is about the removal of the agent from the templates used for config upgrade. I understand we don't want to keep things regarding the agent in this repo, but we need to provide a way for people to migrate from the old `keylime.conf` to the new split configuration format. For the rust agent, we rely on the `convert_config.py` and templates provided here.

For me, the options are to move the configuration upgrade stuff to another repo or rethink how the config upgrade works.

THS-on:
@ansasaki I agree with you that the removal from the templates in not a good idea yet. Maybe we just don't generate the agent.conf by default anymore?

stefanberger:
mypy.ini & pyproject.toml can have keylime_agent.py removed.

THS-on:
@kkaarreell I'm a little bit confused why only the `basic-attestation-on-localhost` fails and not the other ones. Looking at the logs it seems that the agent never receives the revocation message.

kkaarreell:
@THS-on This is not obvious from logs, unfortunately. Take a look at whole test plan. This test is scheduled 3 times and before the last iteration zeromq notifier is configured. This is not applicable to Rust agent so you should remove both the setup and 3rd test iteration.

THS-on:
@kkaarreell thank you! I now updated it, hope it now passes.

kkaarreell:
I believe we are not installing systemd unit files for the Rust agent in the installation task. So when you remove the ones for the Python agent there is nothing left. Agent starts through a direct call but e.g. verbose loging is not configured which explains log parsing errors. Could you add unit file installation to the installation task for the Rust agent? In fact we should probably do `make install` instead of copying individual files but we can do that later.

kkaarreell (commented):
Restoration of the agent revocation notifier in the test plan can be also removed along with the previous comment.

kkaarreell:
FYI, I am planning the rust agent installation task update in https://github.com/RedHat-SP-Security/keylime-tests/pull/319

THS-on:
@kkaarreell thanks. I'll update to them once they are merged

kkaarreell:
> @kkaarreell thanks. I'll update to them once they are merged

It is merged now.

THS-on:
@kkaarreell the centos 9 failure is because Keylime cannot find the `libefivar.so.1`. Which seems to be unrelated to this change.

kkaarreell:
> @kkaarreell the centos 9 failure is because Keylime cannot find the `libefivar.so.1`. Which seems to be unrelated to this change.

Yes, it will be addressed in https://github.com/RedHat-SP-Security/keylime-tests/pull/320.
Also, remove `centos-stream-8` from `.packit.yaml` because the Rust agent won't run there.

packit-ci.fmf
```diff
@@ -2,9 +2,12 @@
 
   summary: run keylime e2e tests
 
+  environment:
+    RUST_IMA_EMULATOR: 1
+
   context:
     swtpm: yes
-    agent: python
+    agent: rust
```
aplanas, line 10: maybe in a future commit this can be dropped, as there is only one

aplanas (approved):

packit-ci.fmf
kkaarreell: True. The removal would require test updates so better doing that in the future when change is already settled on keylime side.

maugustosilva (approved):

maugustosilva:
A much needed change, LGTM

PR #1291, opened 2023-01-31
codestyle: Type-annotate revocation_notifier.py and add to mypy

stefanberger:
Signed-off-by: Stefan Berger <stefanb@linux.ibm.com>

PR #1292, opened 2023-02-01
fixing linter non-W0719 issues

mdrocco:
Ignoring W0719 for now in `pylintrc`, as fixing all of them will be a separate substantial PR. See #1293


stefanberger:
I added all of these to PR #1291

mdrocco:
I see, then I was misled by 1291's title :) I am trying to push the minimal set of changes that would unlock other PRs, except for W0719 that would require some more work as I see tons of them. I will keep this unless 1291 comes first, worst case we did a little bit of replicated work. What you think?

maugustosilva (approved):

stefanberger:
I would just ignore W0719 altogether.

maugustosilva:
@mpeters LGTM this now, I propose, and we can revisit the (newly raised, due to newer pylint) W0719 issue later.

THS-on:
The fixes are merged with #1291

PR #1298, opened 2023-02-06
codestyle: Fully annotate cloud_verifier_common.py and add to mypy

stefanberger:


THS-on (approved):
LGTM

PR #1300, opened 2023-02-07
config: improve support for (log-based) debugging

maugustosilva:
This PR attempts to increase the ability to use logs (which can be ingested by an analytics stack, such as ELK) in order to debug startup. To this end, all `print` statements in config.py are replaced by calls to a (basic) logger (given we cannot use the keylime logger due to circular dependency), and a debug level log message is added whenever a configuration file option is overriden by an environment variable (ability introduced with PR #1274).

In addition to the above, the actual invocation of external commands (e.g., tpm2-tools) are also logged with debug statments.


maugustosilva:
Regarding the code scanning warning (sensitive data as clear text), here is what I see on the logs. Nothing really critical, but can drop it if anyone strenuously objects to it

on the `registrar` :

```
2023-02-07 23:38:39.020 - keylime.cmd_exec - DEBUG - Executing command "tpm2_startup --version"
2023-02-07 23:38:39.061 - keylime.cmd_exec - DEBUG - Executing command "tpm2_makecredential -T none -e /tmp/tmpl_gctmwd -s /tmp/tmp3a7qch76 -n 000b213081cbc542d05b1179dc0f4203dfc259758291cb6e785ec3ffdb9a850cf1c8 -o /tmp/tmpnhy89b3g"
```

on the `verifier` :

```
2023-02-07 23:42:23.724 - keylime.cmd_exec - DEBUG - Executing command "tpm2_checkquote -u /tmp/tmpyei7f0an -m /tmp/tmpjez1y41x -s /tmp/tmp4vm709a2 -f /tmp/tmp0olc9r9v -g sha256 -q 33474630715169416d57653677383649675a5245"
2023-02-07 23:42:23.743 - keylime.cmd_exec - DEBUG - Executing command "tpm2_print -t TPMS_ATTEST /tmp/tmpurcpd6ku"
2023-02-07 23:42:23.761 - keylime.cmd_exec - DEBUG - Executing command "tpm2_eventlog --eventlog-version=2 /tmp/tmp2yfiuo3u"
```

keylime/config.py
```diff
@@ -180,31 +183,27 @@ def get_config(component: str) -> RawConfigParser:
             if os.path.exists(CONFIG_ENV[component]):
                 if os.path.isfile(CONFIG_ENV[component]):
                     config_files = _config[component].read(CONFIG_ENV[component])
-                    print(f"Reading configuration from {config_files}")
+                    log_msg = f"Reading configuration from {config_files}"
+                    base_logger.debug(log_msg)
```
aplanas, line 187: nit: there is a difference between `base_logger.debug(f"Reading configuration from {config_files}")` and `base_logger.debug(f"Reading configuration from %s", config_files)`

In the first case the interpolation is always done, whatever the logger decides to output the string or not, but in the second the interpolation is done only if the string is emitted.

For this case the cost is almost the same, but as a general rule, for logs, we should always use the second form.

keylime/config.py
```diff
@@ -180,31 +183,27 @@ def get_config(component: str) -> RawConfigParser:
...
                     return _config[component]
 
-                print(
-                    f"Configuration file {CONFIG_ENV[component]} for"
-                    f"{component} set through environment variable is "
-                    "not a file, falling back to installed configuration"
-                )
+                log_msg = f"Configuration file {CONFIG_ENV[component]} for {component} set through environment variable is not a file, falling back to installed configuration"
+                base_logger.debug(log_msg)
```
aplanas, line 191: same (to avoid repetition I will skip the rest)

keylime/config.py
```diff
@@ -16,6 +16,9 @@
 
 from keylime.common.version import str_to_version
 
+logging.basicConfig(level=logging.DEBUG)
+base_logger = logging.getLogger("keylime.config")
```
aplanas, line 20: One solution to avoid the circular dependency is to drop / refactor `keylime_logger` and setup the logger in the first import to `config`.  Maybe we need to create an issue to do this, as will clear a bit the codebase.

keylime/config.py
maugustosilva: Changed all of it (except for the one in line 246, where pylint complained and wanted to keep a bit of post-interpolation beautification)

keylime/config.py
maugustosilva: same

keylime/config.py
maugustosilva: Seems to be the best course of action. Please note that once this is implemented the changes to `config.py` , thanks to this PR, will be minimal (basically replace a few occurrences of `base_logger`)

THS-on: I would change the default logging level to INFO, because that is was I would expect when nothing is configured.

Yeah we can probably put the `keylime_logging` code into `config`.

maugustosilva: Do you want me to change `base_logger` to `INFO` meanwhile?

THS-on: Yes

maugustosilva: done

aplanas (approved):

THS-on (approved):

PR #1302, opened 2023-02-08
tests: only keep test specific packages in test-requirements.txt

THS-on:


aplanas (approved):
I see. This makes more sense.

mpeters (approved):

PR #1303, opened 2023-02-08
Fix stray references to "IMA policies" in conversion script

mbestavros:
Note that this also changes a command line option, since that was incorrect as well.


sergio-correia (approved):

aplanas (approved):

THS-on:
Because it is only in the conversion script part, I think it is fine to change it without backwards compatibility.

PR #1287, opened 2023-01-30
Various documentation updates

THS-on:
Now from the correct branch

- docs: fix formatting of example IMA-policy
- docs: update theme to min 1.1.0
- docs: remove old development instructions, move dev conainter section
- docs: update REST APIs TLS documentation to match new default setup
- docs: add basic (m)TLS instructions to the installation guide
- docs: point newcomers to the design document
- docs: Update IMA instructions to new runtime policy format

PR #1305, opened 2023-02-10
Extract IMA hashes from local and remote repositories

aplanas:


aplanas:
Some improvements:

- Remove the Go idioms
- Promote the script as another Keylime cmdline tools

scripts/create_policy
```diff
@@ -160,11 +189,183 @@ def process_signature_verification_keys(verification_keys, policy):
...
 
 
+def analyze_rpm_pkg(pkg):
+    """Analyze a single RPM package."""
+    ts = rpm.TransactionSet()
+    ts.setVSFlags(rpm.RPMVSF_MASK_NOSIGNATURES | rpm.RPMVSF_MASK_NODIGESTS)
+
+    with open(pkg, "rb") as f:
+        hdr = ts.hdrFromFdno(f)
+    # TODO - How are symbolic links managed with IMA
```
stefanberger, line 199: if `ls` is a symbolic link to `busybox` then busybox will be logged an measured

scripts/create_policy
```diff
@@ -160,11 +189,183 @@ def process_signature_verification_keys(verification_keys, policy):
...
+    while True:
+        with tempfile.TemporaryFile() as f:
+            range_ = f"{len(blob)}-{len(blob) + chunk_size - 1}"
+            req = urllib.request.Request(url, headers={"Range": f"bytes={range_}"})
+            with urllib.request.urlopen(req) as resp:
+                blob += resp.read()
+
+            f.write(blob)
+            f.seek(0)
+
```
stefanberger, line 231: you should be able to call `analyze_rpm_pkg` on this package now ?

scripts/create_policy
```diff
@@ -242,19 +446,29 @@ def main():
...
-    # Add the digests map either from the allowlist, if given, or the IMA measurement list.
+    if (args.local_repo or args.remote_repo) and not HAS_RPM:
+        print('To analyze RPM repositories the "rpm" Python module is required', file=sys.stderr)
+        sys.exit(1)
+
+    # Add the digests map either from the allowlist, a repo, or the
+    # IMA measurement list.
     if args.allowlist:
-        policy["digests"], ret = process_flat_allowlist(args.allowlist, policy["digests"])
+        policy["digests"], ret = process_flat_allowlist(args.allowlist)
```
stefanberger, line 464: This is not the same as before... If the tool read some sort of base policy then this here should merge new values into it - at least that's what was the original intention. Now it's a replacement of the contents of the base policy. Both operations are probably valid, so this is a bit  'tricky'  on how to give users this type of control over it without the tool becoming confusing to use.

scripts/create_policy
```diff
@@ -242,19 +446,29 @@ def main():
...
+    elif args.local_repo:
+        policy["digests"], ret = analize_local_repo(args.local_repo)
+    elif args.remote_repo:
+        policy["digests"], ret = analize_remote_repo(args.remote_repo)
     elif not args.no_hashes:
-        policy["digests"], ret = get_hashes_from_measurement_list(args.ima_measurement_list, policy["digests"])
+        policy["digests"], ret = get_hashes_from_measurement_list(args.ima_measurement_list)
```
stefanberger, line 470: same comment here in regards to replacement versus merging ...

scripts/create_policy
```diff
@@ -40,14 +40,16 @@ def process_flat_allowlist(allowlist_file):
                     continue
                 pieces = line.split(None, 1)
                 if not len(pieces) == 2:
-                    print(f"Skipping line that was split into {len(pieces)} parts, " "expected 2: {line}")
```
stefanberger, line 42: commit text typo: emission & break

stefanberger: Otherwise this patch LGTM

scripts/create_policy
aplanas: Thanks, that simplify the problem. I will update the doc.

scripts/create_policy
aplanas: Reverting this then. Thanks!

scripts/create_policy
aplanas: Fixed

maugustosilva (approved):

maugustosilva:
Please squash the commits and LGTM

aplanas:
> Please squash the commits

I would prefer to not squash them. They are affecting different files, and represent different stages of the feature. But if there is a consensus then they can be squashed.

maugustosilva:
Well, the "standard operating procedure" for all PRs here is to have all commits squashed. I'd rather let you decide how to squash and how to work the unified commit message than do it via the "Squash and merge" button on the Web UI :-)

ansasaki:
I didn't know about this "always squash" policy. I personally prefer to have separate commits reasonably self contained that make specific and easy-to-review changes. It is a nightmare later to search through the commit history to find where a specific change was made if all commits are squashed.

maugustosilva:
Will have a short meeting with @mpeters and others maintainers soon, and will clarify this point. I have a preference for squashing (a self-contained change), but don't see it as crucial.

aplanas:
I was not aware of the policy neither. In fact I found it a mistake, and indeed there are recent PR that have multiple commits:

(6 commits, 2 weeks ago) https://github.com/keylime/keylime/pull/1291
(7 commits, 2 weeks ago) https://github.com/keylime/keylime/pull/1287

IMHO we should drop the policy, also because the mentioned reason exposed by @ansasaki, each commit can explain the rationale of the change, and conflating 8 commit are as not deeply related can be a pain when looking for an explanation.

Since long ago github also relate the commit into the pull request, so it is very easy when doing "blame" to get the single commit and from there the pull request that contains the extended discussion.

maugustosilva:
@aplanas aha, so it was not being enforced uniformly already! Ok, will have a quick chat (with @mpeters and others) and make sure we have a unified policy on it.

maugustosilva:
@aplanas @THS-on Ok, question settled. We might ask for it (squashing) but this will not be mandatory any longer. There are scenarios where having a PR with multiple commits indeed makes sense... if the reviewer decides that there are too many "fragmented" commits "polluting" the PR he can ask for a squashing.

aplanas:
Good news: this has been merged https://github.com/rpm-software-management/createrepo_c/pull/316 (later the metadata has been renamed to filelist-ext.xml)

This imply that the second algorithm for remote repositories can be adapted, and if this file is present the performance will be massively improved.

Issue #1067, opened 2022-08-03
Code improvement: refactor the use of temporary files in tpm_main.py


maugustosilva:
# Environment
 * OS / version: any 
 * Processor architecture: any
 * TPM Manufacturer: any
 * Keylime version: any

# Description 
The current use of temporary files in tpm_main.py is at once odd and inefficient. @aplanas and @THS-on have suggested some better, more pythonesque ways to rewrite it.

## Expected behavior vs. actual behavior 
Once the changes are in place, there will be no changes in behavior.

## Steps to reproduce problem
1.  Just inspect `tpm/tpm_main.py` in particular methods such as `_tpm2_checkquote` and `_tpm2_printquote` in `tpm` class

stefanberger:
I think this issue should either have its title adjusted since tpm_main.py does not contain temporary files anymore, or be closed.

maugustosilva:
@stefanberger you're right. Closing it now.

PR #1309, opened 2023-02-14
Refactoring the use of temporary files in tpm_main.py

ShubhikaBhardwaj:
Please fill in the fields below to submit a pull request. The more information
that is provided, the better.

**Fixes issue #1067  **:

Description of the changes being introduced by the pull request:

For Code improvement, refactoring the use of temporary files in tpm_main.py 
- Refactored methods such as _tpm2_checkquote and _tpm2_printquote in tpm class to follow PEP 8 Python standards
- Documenting these functions using pydocs

### DIFF
--- a/keylime/tpm/tpm_main.py
+++ b/keylime/tpm/tpm_main.py
@@ -1039,9 +1039,28 @@ class tpm(tpm_abstract.AbstractTPM):
 
         return "r" + quote
 
-    def __tpm2_checkquote(
-        self, pubaik: str, nonce: str, quoteFile: str, sigFile: str, pcrFile: str, hash_alg: Union[Hash, str]
+    def tpm2_checkquote(
+            self,
+            pubaik: str,
+            nonce: str,
+            quote_file: str,
+            sig_file: str,
+            pcr_file: str,
+            hash_alg: Union[Hash, str],
     ) -> cmd_exec.RetDictType:
+        """Executes the tpm2_checkquote command.
+
+        Args:
+            pubaik: The public AIK file.
+            nonce: The nonce to use.
+            quote_file: The quote file.
+            sig_file: The signature file.
+            pcr_file: The PCR file.
+            hash_alg: The hash algorithm to use.
+
+        Returns:
+            A dictionary containing the stdout and stderr of the command.
+        """
         nonce = bytes(nonce, encoding="utf8").hex()
         if self.tools_version == "3.2":
             command = [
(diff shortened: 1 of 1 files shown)


keylime/tpm/tpm_main.py
```diff
@@ -1039,51 +1039,77 @@ def create_quote(
...
+                pcr_file,
                 "-g",
                 hash_alg,
                 "-q",
                 nonce,
             ]
 
-        retDict = self.__run(command, lock=False)
-        return retDict
+        ret_dict = cmd_exec.run(command, lock=False)
```
aplanas, line 1098: that change it is not the same, the local `__run` should be indeed rewritten, but you are using parameters that make sense in `__run` but not in `run`, like `lock`

the rename of variables makes sense

keylime/tpm/tpm_main.py
```diff
@@ -1039,51 +1039,77 @@ def create_quote(
...
+        """Executes the tpm2_print command.
+
+        Args:
+            quote_file: The quote file to print.
+
+        Returns:
+            A dictionary containing the stdout and stderr of the command.
+        """
+        command = ["tpm2_print", "-t", "TPMS_ATTEST", quote_file]
+        ret_dict = cmd_exec.run(command, lock=False)
```
aplanas, line 1111: same

aplanas:
> **Fixes issue https://github.com/keylime/keylime/issues/1067 **

It is still not addressing the pointed issue?

aplanas:
Also: wellcome : )

keylime/tpm/tpm_main.py
stefanberger: > the rename of variables makes sense

It only makes sense if we all agree that we don't want CamelCase but underscore everywhere from now on.

aplanas: The agreement is mostly implicit, as we are following pep8:

https://peps.python.org/pep-0008/#function-and-variable-names

```
mixedCase is allowed only in contexts where that’s already the prevailing style (e.g. threading.py), to retain backwards compatibility.
```

keylime/tpm/tpm_main.py
```diff
@@ -1039,51 +1039,77 @@ def create_quote(
...
-        retDict = self.__run(command, lock=False)
-        return retDict
+        ret_dict = cmd_exec.run(command, lock=False)
+        return ret_dict
 
-    def __tpm2_printquote(self, quoteFile: str) -> cmd_exec.RetDictType:
-        command = ["tpm2_print", "-t", "TPMS_ATTEST", quoteFile]
-        retDict = self.__run(command, lock=False)
-        return retDict
+    def tpm2_printquote(self, quote_file: str) -> cmd_exec.RetDictType:
```
stefanberger, line 1101: You would have to adjust the callers as well.

keylime/tpm/tpm_main.py
```diff
@@ -1039,51 +1039,77 @@ def create_quote(
 
         return "r" + quote
 
-    def __tpm2_checkquote(
-        self, pubaik: str, nonce: str, quoteFile: str, sigFile: str, pcrFile: str, hash_alg: Union[Hash, str]
+    def tpm2_checkquote(
```
stefanberger, line 1042: You changed the function from a private to a public function. You would at least have to adjust all callers as well.

stefanberger:
Is this PR still alive? The title states 'Refactoring the use of temporary files in tpm_main.py' but it doesn't seem to address this particular issue at all...

ShubhikaBhardwaj:
Hi @stefanberger  
I am working on your suggestions.

stefanberger:
I think this PR and its related issue are obsoleted by the fact that we are getting rid of usage of tpm2-tools and temporary files in tpm_main.py won't be necessary anymore.

stefanberger:
Can we close this PR? At least all the code section that were supposed to be change have been moved out of tpm_main.py and are supposed to be replace with native implementations (rather than calling the tpm2 tools).

PR #1310, opened 2023-02-14
da: fixes for breakages on durable Attestation

maugustosilva:
The last couple of patches caused some breakages on Durable Attestation (DA), and given the absence of a proper CI test (will work on it after this PR is merged) it remained undetected. Here we have all the fixes needed to bring it back into operational state.

In addition to it, there are some small improvements on logging for the `check_pcrs` method, which benefits both DA and regular operations (basically, include agent ID on all messages, plus the actual name of measured boot policy used)

### DIFF
--- a/keylime/da/attest.py
+++ b/keylime/da/attest.py
@@ -1,9 +1,11 @@
 #!/usr/bin/python3
 import argparse
+import json
 from datetime import datetime
 
 from keylime import cloud_verifier_common, config, keylime_logging
 from keylime.da import record
+from keylime.elchecking.policies import load_policies
 
 logger = keylime_logging.init_logging("durable_attestation_fetch_and_replay")
 
@@ -36,6 +38,8 @@ def main() -> None:
     )
 
     args = parser.parse_args()
+
+    load_policies()
 
     rmcb = config.get("registrar", "durable_attestation_import", fallback="")
     rmc = record.get_record_mgt_class(rmcb)
@@ -96,7 +100,7 @@ def main() -> None:
         for attestation_record in attestation_record_list:
             agent = attestation_record["agent"]
             json_response = attestation_record["json_response"]
-            ima_policy = attestation_record["ima_policy"]
+            runtime_policy = json.loads(attestation_record["runtime_policy"])
 
             logger.info(
                 "----------- Attesting data (quote and logs from %s, captured by verifier %s (%s:%s)",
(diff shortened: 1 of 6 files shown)


THS-on:
@maugustosilva can you put the check_pcrs change into a separate commit?

@kkaarreell this changes the logging if a agent fails, so I think this is the reason that the test are failing. Can you adapt the regexes for that?

kkaarreell:
@THS-on I am on PTO and without a computer. @Koncpa can help eventually but ideally if you create a PR.
@Koncpa Verifier log message has beed extended with a part specifying the agent, I think we could make regexp check for that part optional for now and merge to main. In the future we will make that part in regexp mandatory.

maugustosilva:
@THS-on I thought about making it a separate PR, but the truth is the change in line 300 on `tpm_abstract.py` (the "Evaluating measured boot log") was pretty instrumental to detect the problem on durable attestation. 

The other changes there can go a separate PR if you insist, but given that CI will now have to be changed anyway (the regex), I would argue on keeping it all on the same PR. Acceptable?

THS-on:
> The other changes there can go a separate PR if you insist, but given that CI will now have to be changed anyway (the regex), I would argue on keeping it all on the same PR. Acceptable?

Yes, keeping it in the same PR is fine :)  I just would split it in two commits, so that if you look at the history you don't wonder what changes in check_pcrs have to do with DA.

maugustosilva:
Aren't we going to squash it before merge anyway? It was my understanding that it was the "standard operating procedure" for this repo.

THS-on:
> Aren't we going to squash it before merge anyway? It was my understanding that it was the "standard operating procedure" for this repo.

Not really, but I think we never discussed this explicitly the last years. I just squashed when they are the same logical wise and the commits are not cleanly separated, if they are two different things I think it makes to keep them as separate commits.

maugustosilva:
@kkaarreell @Koncpa opened a PR on `keylime-tests` https://github.com/RedHat-SP-Security/keylime-tests/pull/310

aplanas:
One naive question, that means that the old logs cannot be validated anymore?

maugustosilva:
@aplanas "Old" (the feature has barely 3 months, funny to even talk about "age" here :-) ) attestation records could still be replayed with a trivial, one-line change in `attest.py` (line 103). But given the feature is still experimental (we are not using in Production yet), I have decided to not add code to deal with it for now. 

But your point is well taken, we need to make sure attestation records can be used even in face of changes in the encoding/representation internally.

THS-on (approved):

aplanas (approved):

PR #1312, opened 2023-02-16
db: switch to pessimistic disconnect handling

THS-on:
If sessions are idle for a long time, the SQL server might no longer have that
connection open. This causes the Keylime to error when its tries to query the
database.

This enabled pool_pre_ping which checks if the connection is still active and
uses a new connection if not.

More details can be found here:
https://docs.sqlalchemy.org/en/20/core/pooling.html#disconnect-handling-pessimistic

### DIFF
--- a/keylime/db/keylime_db.py
+++ b/keylime/db/keylime_db.py
@@ -82,6 +82,7 @@ class DBEngineManager:
             if not url.count("sqlite:"):
                 engine_args["pool_size"] = int(p_sz)
                 engine_args["max_overflow"] = int(m_ovfl)
+                engine_args["pool_pre_ping"] = True
 
         # Enable DB debugging
         if config.DEBUG_DB and config.INSECURE_DEBUG:


THS-on:
This is currently marked as draft, because we are still testing if the issue is fully fixed.

keylime/db/keylime_db.py
```diff
@@ -82,6 +82,7 @@ def make_engine(self, service: str) -> Engine:
             if not url.count("sqlite:"):
                 engine_args["pool_size"] = int(p_sz)
                 engine_args["max_overflow"] = int(m_ovfl)
+                engine_args["pool_pre_ping"] = True
```
aplanas, line 85: Did you saw this issue?

In any case this see the less intrusive change, the other one seems to be setting the pool_recycle parameter and managing the exception per query

THS-on: Yes, we saw it on our test infrastructure happening. You see then errors like this:
````
sqlalchemy.exc.OperationalError: (pymysql.err.OperationalError) (2006, "MySQL server has gone away (ConnectionResetError(104, 'Connection reset by peer'))")
````

maugustosilva: This is a good idea. We have been facing the same troubles in our long-running large-scale testing. It will not solve all the problems*, but I am going to test it out just to make sure it does impose too much "additional overhead" on the DB connections (don't expect it does).

* (I believe @aplanas is right and we will need to harden our exception catching and - more crucially - recovery per query)

THS-on:
We no longer see that issue in our testbed with roughly idle times of up to 24 hours, so I think that it fixes issue for now (we should still refactor the DB access at some point).

aplanas (approved):

mpeters (approved):

PR #1313, opened 2023-02-17
Add timestamp of last successful attestation to verifier API (enhancement #74) and address some missing documentation

stringlytyped:
This pull request fulfills https://github.com/keylime/enhancements/issues/74 by adding a new database column (`last_successful_quote`) and exposing it through the verifier's REST API. Many thanks to @maugustosilva for his help with this [in the Slack channel](https://cloud-native.slack.com/archives/C01ARE2QUTZ/p1676370102354909).

Additionally, I've adding some descriptions to the API documentation for the `last_successful_quote` field and a couple other fields that were missing documentation (namely, `attestation_count` and `last_received_quote`).

See commit messages for further details.

### DIFF
--- a/docs/rest_apis.rst
+++ b/docs/rest_apis.rst
@@ -88,7 +88,10 @@ Cloud verifier (CV)
             "verifier_ip": "127.0.0.1",
             "verifier_port": 8881,
             "severity_level": 6,
-            "last_event_id": "qoute_validation.quote_validation"
+            "last_event_id": "qoute_validation.quote_validation",
+            "attestation_count": 240,
+            "last_received_quote": 1676644582,
+            "last_successful_attestation": 1676644462
           }
         }
 
@@ -114,6 +117,9 @@ Cloud verifier (CV)
     :>json int verifier_port: Port of the verifier that is used.
     :>json int severity_level: Severity level of the agent. Might be `null`. Levels are the numeric representation of the severity labels.
     :>json string last_event_id: ID of the last revocation event. Might be `null`.
+    :>json int attestation_count: Number of quotes received from the agent which have verified successfully.
+    :>json int last_received_quote: Timestamp of the last quote received from the agent irrespective of validity. A value of 0 indicates no quotes have been received. May be `null` after upgrading from a previous Keylime version.
+    :>json int last_successful_attestation: Timestamp of the last quote received from the agent which verified successfully. A value of 0 indicates no valid quotes have been received. May be `null` after upgrading from a previous Keylime version.
 
 
 .. http:post::  /v2.1/agents/{agent_id:UUID}
--- a/keylime/cloud_verifier_common.py
+++ b/keylime/cloud_verifier_common.py
@@ -204,7 +204,7 @@ def process_quote_response(
 
     if not failure:
         agent["attestation_count"] += 1
-
+        agent["last_successful_attestation"] = int(time.time())
         agent["tpm_clockinfo"] = agentAttestState.get_tpm_clockinfo().to_dict()
 
         # has public key changed? if so, clear out b64_encrypted_V, it is no longer valid
@@ -296,6 +296,7 @@ def process_get_status(agent: VerfierMain) -> Dict[str, Any]:
         "last_event_id": agent.last_event_id,
         "attestation_count": agent.attestation_count,
         "last_received_quote": agent.last_received_quote,
+        "last_successful_attestation": agent.last_successful_attestation,
     }
     return response
 
--- a/keylime/cloud_verifier_tornado.py
+++ b/keylime/cloud_verifier_tornado.py
@@ -115,6 +115,7 @@ def _from_db_obj(agent_db_obj: VerfierMain) -> Dict[str, Any]:
         "ak_tpm",
         "attestation_count",
         "last_received_quote",
+        "last_successful_attestation",
         "tpm_clockinfo",
     ]
     agent_dict = {}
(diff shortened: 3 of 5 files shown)


THS-on:
@stringlytyped thank you for the PR and welcome to Keylime! The changes in general LGTM, but I would name the option `last_successful_attestation` to make it more clear that it refers to the last successful attestation round.

docs/rest_apis.rst
```diff
@@ -114,6 +117,9 @@ Cloud verifier (CV)
     :>json int verifier_port: Port of the verifier that is used.
     :>json int severity_level: Severity level of the agent. Might be `null`. Levels are the numeric representation of the severity labels.
     :>json string last_event_id: ID of the last revocation event. Might be `null`.
+    :>json int attestation_count: Number of quotes received from the agent which have verified successfully.
+    :>json int last_received_quote: Timestamp of the last quote received from the agent irrespective of validity. A value of 0 indicates no quotes have been received. May be `null` after upgrading from a previous Keylime version.
+    :>json int last_successful_quote: Timestamp of the last quote received from the agent which verified successfully. A value of 0 indicates no valid quotes have been received. May be `null` after upgrading from a previous Keylime version.
```
aplanas, line 122: Why not to consolidate `null` into 0?

keylime/migrations/versions/f838d3cdeead_add_last_successful_quote.py
```diff
@@ -0,0 +1,41 @@
...
+def upgrade_registrar():
+    pass
+
+
+def downgrade_registrar():
+    pass
+
+
+def upgrade_cloud_verifier():
+    op.add_column("verifiermain", sa.Column("last_successful_quote", sa.Integer(), nullable=True))
```
aplanas, line 36: Maybe I am wrong, but verifier_db says that it should not be nullable, I wonder if we can drop this here and put some initial value (0)

Maybe even a similar change can be done for last_received_quote (or not, as at this point can be too late in the migration)

aplanas (commented):
Basically LGTM! Just a question about the nullability of the field

docs/rest_apis.rst
maugustosilva: Yeah, in this case, he just followed what I have done originally, and my concern was exactly the migration problems, especially with different SQL backends, especially in case of migrations.

keylime/migrations/versions/f838d3cdeead_add_last_successful_quote.py
maugustosilva: So, same problem. I remember an issue with at least one of the backends (MySQL, if memory serves) that lead me, after a quick chat with @mpeters to adopt the more laissez faire policy of set `nullable` to `True` and not setting a value there. I am not going to strongly advocate for it, but if we want to change, we might have to add in a migration for the other column too.

maugustosilva:
@stringlytyped very welcome addition, especially the API documentation (which I forgot to add for `attestation_count` and `last_successful_quote`). 

We are very close to a merge, but please change the column name to `last_successful_attestation` (@THS-on is correct on pointing out that this will avoid the eventual confusion about wether the `quote` was simply received by the `verifier` or was actually fully attested).

As for the `pylint` errors, you can either run `make check`, `./scripts/check_codestyle.sh` (for `pylint` only) or configure the per-commit hooks (which I recommend in case you plan to contribute more)

stringlytyped:
Thanks @THS-on, @aplanas and @maugustosilva for your feedback. It has taken me a moment to get your suggested changes implemented and to fix those formatting issues because I was struggling to get the formatters/linters/etc. working in my local environment. But I think that should be all good to go now, pending CI actions.

docs/rest_apis.rst
stringlytyped: That's right, I was just following the existing pattern. It does seem to me that there is a slight difference in semantics between "there has never been a successful attestation for this agent" (which results in `last_successful_attestation = 0`) and "there may have been a successful attestation for this agent but there is currently no information about when it occurred" (as would be the case after a keylime install is upgraded and the migration is run for the first time, resulting in  `last_successful_attestation = null`). But I'm not sure if this matters.

stringlytyped:
The only outstanding issue is on fedora-rawhide-x86_64. Looking at the logs, I cannot see how the failures could be due to my changes. Is there a chance this is related to [this bug](https://bugzilla.redhat.com/show_bug.cgi?id=2171376) and the upgrade in Fedora 38 of tpm2-tss and tpm2-tools to versions 4.0.1 and 5.4 respectively?

aplanas (approved):

THS-on:
> The only outstanding issue is on fedora-rawhide-x86_64. Looking at the logs, I cannot see how the failures could be due to my changes. Is there a chance this is related to [this bug](https://bugzilla.redhat.com/show_bug.cgi?id=2171376) and the upgrade in Fedora 38 of tpm2-tss and tpm2-tools to versions 4.0.1 and 5.4 respectively?

@stringlytyped  yes that failure is in the agent TPM code as you correctly identified and has nothing to do with this change.

Can you integrate the last fixup commit into the two first ones? After that this it should be ready to merge:)

stringlytyped:
> Can you integrate the last fixup commit into the two first ones? After that this it should be ready to merge:)

@THS-on Sure thing. Is it okay if I squash the commits (easier than rebasing manually) or would you prefer the changes to the code base in a separate commit from the updated docs?

THS-on:
@stringlytyped either one is fine for me because all the changes belong logical together.

maugustosilva (approved):

THS-on (approved):

maugustosilva:
As we discussed, the Fedora Rawhide failure is not related to your commit. Thank you again, will merge.

PR #1314, opened 2023-02-17
Implement DSSE signing/verification for Keylime runtime policies

mbestavros:
This pull request replaces Keylime's existing support for verifying detached signatures on runtime policies with a new system utilizing the DSSE signed envelope standard.

[DSSE](https://github.com/secure-systems-lab/dsse) aims to be a standard file format for encapsulating messages and signatures in the same file (i.e. an attached signature implementation). The spec is in use by several major projects within the supply chain security ecosystem, and implementing support for the spec within Keylime may position the project to better integrate with them down the line.

To add support for signing and verifying DSSE envelopes, this pull request includes an in-tree, [spec](https://github.com/secure-systems-lab/dsse/blob/master/protocol.md) compliant [library](https://github.com/mbestavros/dsse-lib), lightly modified from the [reference implementation](https://github.com/secure-systems-lab/dsse/tree/master/implementation) available on the DSSE repo. The library utilizes `cryptography` and nothing else, negating the need for any new dependencies. At the moment, ECDSA keys are the only supported key type.

A tool is provided to sign existing unsigned runtime policies and create DSSE envelopes for use with the new infrastructure, available under `keylime/cmd/sign_runtime_policy.py`.

Beyond the top-line support, this PR includes some neat additional features enabled by DSSE:

- While DSSE envelopes can be verified traditionally with an out-of-band ECDSA public key, the policy signing tool and verification library _also_ include support for **x509 signatures** embedded into the DSSE envelope (in particular, encoded as the `keyid`). This kind of envelope carries the public key in-band, meaning that those DSSE envelopes can _self-verify_ without any other materials.
- To complement this support, the Keylime verifier will also now cache any out-of-band ECDSA public keys at `/var/lib/keylime/signing_keys`.
- The combination of the above two features sets Keylime up for _continuous verification_ in the future: in other words, instead of only checking the DSSE signature when a runtime policy is added to the Keylime verifier, the verifier could check the signature _when the stored runtime policy is used_ as well.

This should serve as an implementation of https://github.com/keylime/enhancements/pull/89/.

Feedback welcome! I'd especially like feedback on lint/formatting in the DSSE library - I'm unsure of how best practices should fit in there.


mbestavros:
@aplanas Great questions.

> Can be possible to convince the DSSE group to release a version in PYPI that we can use?

We could inquire, but I lean towards no. For one thing, the reference implementation hasn't been updated since 2021, and the impression I get is that it wasn't intended to be shipped as a standalone package.

The other upstream option would be the `securesystemslib` project - its developers are working to enable DSSE support. I did investigate the possibility of using it in Keylime pretty thoroughly, and even built a prototype; the issue is that `securesystemslib` brings in a ton of new dependencies besides `cryptography`, including some Keylime worked to remove when packaging for RHEL. If we wanted to bring it in, there would be a lot of backend/dependency/packaging work needed, and that could take a long time.

Shipping an in-tree library sidesteps these issues for now, and over time Keylime can bring in upstream support.

> Also, can you EL5 to me why this should be used? We still need to validate other external elements that will be signed by a 3rd party, what is the advantage of using DSSE format to sign our assets?

Some of the reasoning was outlined in the [enhancement proposal](https://github.com/keylime/enhancements/blob/master/88-encapsulate-runtime-policies-in-dsse.md), but in short: attached signatures are less fragile than a detached scheme; and the DSSE standard is already in use by several notable supply chain security projects, including Sigstore. Enabling Keylime to understand DSSE could open up potential policy storage on Sigstore down the road with minimal effort.

I'll loop in @SantiagoTorres to talk more about this; as someone close to the standard, he can probably encapsulate the benefits better than I can.

keylime/cmd/sign_runtime_policy.py
```diff
@@ -0,0 +1,111 @@
...
+
+To view help and see all available options:
+
+```
+python sign_runtime_policy.py -h
+```
+"""
+
+
+def main() -> None:
```
stefanberger, line 36: add to seetup.cfg ?

keylime/cmd/sign_runtime_policy.py
```diff
@@ -0,0 +1,111 @@
...
+    parser.add_argument("-r", "--runtime_policy", help="Runtime policy file location", action="store")
+    parser.add_argument("-k", "--keyfile", help="Private key to sign policy with", action="store")
+    parser.add_argument(
+        "-p",
+        "--keypath",
+        help="Path to write created private key if -k is not specified",
+        action="store",
+        default="keylime-ecdsa-key",
+    )
+    parser.add_argument("-b", "--backend", help="DSSE backend to use", action="store", default="ecdsa")
```
stefanberger, line 47: "DSSE backed to use; either ecdsa or x509"

keylime/cmd/sign_runtime_policy.py
```diff
@@ -0,0 +1,111 @@
...
+
+    args = parser.parse_args()
+    if not args.runtime_policy:
+        print("Runtime policy is required!")
+        sys.exit(1)
+    elif args.backend not in ["ecdsa", "x509"]:
+        print("Unsupported DSSE backend!")
+        sys.exit(1)
+    elif args.keyfile and args.keypath:
+        print("Only one of -k and -p may be specified at once!")
```
stefanberger, line 63: This doesn't seem to work:
```
# /usr/local/bin/keylime_sign_runtime_policy -b ecdsa -o mypolicy-signed.json -k ecdsa-priv.pem -r mypolicy.json                                                                  Only one of -k and -p may be specified at once!
```

keylime/cmd/sign_runtime_policy.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+def main() -> None:
+    parser = ConversionParser()
+    parser.add_argument("-r", "--runtime_policy", help="Runtime policy file location", action="store")
+    parser.add_argument("-k", "--keyfile", help="Private key to sign policy with", action="store")
+    parser.add_argument(
+        "-p",
+        "--keypath",
+        help="Path to write created private key if -k is not specified",
```
stefanberger, line 43: It's actually the **filename** where the key is supposed to be written into.

keylime/cmd/sign_runtime_policy.py
mbestavros: Done!

keylime/cmd/sign_runtime_policy.py
mbestavros: Thanks, done!

keylime/cmd/sign_runtime_policy.py
mbestavros: Ah, thanks for the catch - because there was a default value for `--keypath`, that condition would trip if `--keyfile` was specified. It's been fixed.

keylime/cmd/sign_runtime_policy.py
mbestavros: Thanks!

keylime/tenant.py
```diff
@@ -1462,20 +1457,6 @@ def main() -> None:
...
-    )
-    parser.add_argument(
-        "--runtime-policy-sig-url",
-        action="store",
-        dest="runtime_policy_sig_url",
-        default=None,
-        help="Specify the URL of the remote signature file of a runtime policy",
-    )
     parser.add_argument(
         "--runtime-policy-sig-key",
```
stefanberger, line 1461: I think we should remove the signature verification key from the options and move it into the tenant configuration file. It's kind of odd to pass a policy as a parameter and voluntarily pass a key along with it. If the key goes into the configuration file it would become mandatory that all policies are signed IF a key is provided. If no key is provided we could just extract the payload and proceed (without signature verification). Also, currently one could just extract the payload from the json and convert it to a runtime policy and pass it in without key and so bypass the whole signature verification. Once it's in the tenant config file it could at least become mandatory that one one works with signed policies.

maugustosilva: If I understood it correctly, in an environment with diversity of nodes, we would end up with multiple configuration files just to deal with the "key" vs "no key" policies (and there is a variety of policies), correct? 
If yes, then I can safely say that, from an operations standpoint, I see this makes maintenance harder, not easier. We have been making a concerted effort to be able to work with configuration files with as little configuration as possible, with all customizations coming from either CLI options or environments variables.

maugustosilva:
Unless anyone strenuously disagree with the shipping of an in-tree library (I really don't see a better option at this time), LGTM.

maugustosilva (approved):

keylime/tenant.py
mbestavros: @stefanberger That's not correct - there **IS** a current configuration option to require signatures on runtime policies. The only difference from your proposal is that it's a verifier-side config option. The specific option is [`require_allow_list_signatures`](https://github.com/keylime/keylime/blob/master/keylime.conf#L383); it currently defaults to `false`, but if set, it'll check for a runtime policy's signature as it arrives on the verifier, and will return a `401` error code if the signature doesn't pass. This PR makes use of that option as well, and will check for DSSE signatures.

In any case, the current tenant-side logic in this pull request is to always verify the signature if it's detected as a DSSE envelope. The only way to not perform a signature check is to pass an unsigned policy.

Beyond that, another reason I'd err towards the CLI option is that the presence of a public key is not necessarily indicative of a signed policy. As mentioned in the description, it's possible to embed x509 certificates with an in-band public key into Keylime DSSE policies, which means you wouldn't need to pass a separate public key to verify the signature. In that case, you'd want to verify the envelope regardless, even without a path to a public key on disk.

stefanberger: Ah. It doesn't seem to make it into the verifier.conf file, though.

mbestavros:
I've finished my updates to the e2e tests - previously, they passed even though they were passing old options that wouldn't check the new code. They all seem to pass with the updates now!

I could use some help with the lint failures, though. The in-tree DSSE library is generating a number of complaints. @stefanberger or anyone else - do you have lint/formatting advice for any of the following new files?

- `keylime/dsse/dsse.py`
- `keylime/dsse/ecdsa.py`
- `keylime/dsse/x509.py`

stefanberger:
This solves a part of the issues. It's not clean and the pylint disablements can largely be removed.

```
diff --git a/keylime/dsse/dsse.py b/keylime/dsse/dsse.py
index 13eacb6..b15380d 100644
--- a/keylime/dsse/dsse.py
+++ b/keylime/dsse/dsse.py
@@ -8,26 +8,29 @@ SPDX-License-Identifier: Apache-2.0
 import base64
 import binascii
 import json
+from abc import abstractmethod

 # Protocol requires Python 3.8+.
 from typing import Iterable, List, Optional, Tuple


-class Signer:
-    def sign(self, message: bytes) -> bytes:  # pylint: disable=unused-argument,no-self-use,useless-option-value
+class Signer():
+    @abstractmethod
+    def sign(self, message: bytes) -> bytes:  # xpylint: disable=unused-argument,no-self-use,useless-option-value
         """Returns the signature of `message`."""

-    def keyid(self) -> Optional[str]:  # pylint: disable=no-self-use,useless-option-value
+    def keyid(self) -> Optional[str]:  # xpylint: disable=no-self-use,useless-option-value
         """Returns the ID of this key, or None if not supported."""


 class Verifier:
+    @abstractmethod
     def verify(
         self, message: bytes, signature: bytes
-    ) -> bool:  # pylint: disable=unused-argument,no-self-use,useless-option-value
+    ) -> bool:  # xpylint: disable=unused-argument,no-self-use,useless-option-value
         """Returns true if `message` was signed by `signature`."""

-    def keyid(self) -> Optional[str]:  # pylint: disable=no-self-use,useless-option-value
+    def keyid(self) -> Optional[str]:  # xpylint: disable=no-self-use,useless-option-value
         """Returns the ID of this key, or None if not supported."""


@@ -40,7 +43,7 @@ class VerifiedPayload:
     payload: bytes
     recognizedSigners: List[str]  # List of names of signers

-    def __init__(self, payloadType: str, payload: str, recognizedSigners: List[str]):
+    def __init__(self, payloadType: str, payload: bytes, recognizedSigners: List[str]):
         self.payloadType = payloadType
         self.payload = payload
         self.recognizedSigners = recognizedSigners
@@ -51,18 +54,18 @@ def b64enc(m: bytes) -> str:


 def b64dec(m: str) -> bytes:
-    m = m.encode("utf-8")
+    m_bytes = m.encode("utf-8")

-    decoded = None
+    decoded: Optional[bytes] = None
     try:
-        decoded = binascii.unhexlify(m)
+        decoded = binascii.unhexlify(m_bytes)
     except binascii.Error:
         pass
     if not decoded:
         try:
-            decoded = base64.b64decode(m, validate=True)
+            decoded = base64.b64decode(m_bytes, validate=True)
         except binascii.Error:
-            decoded = base64.b64decode(m, altchars="-_", validate=True)
+            decoded = base64.b64decode(m_bytes, altchars="-_", validate=True)  # type: ignore
     return decoded


diff --git a/keylime/dsse/ecdsa.py b/keylime/dsse/ecdsa.py
index a29310b..7bffb83 100644
--- a/keylime/dsse/ecdsa.py
+++ b/keylime/dsse/ecdsa.py
@@ -11,12 +11,12 @@ from cryptography.hazmat.primitives.asymmetric import ec


 class Signer:
-    def __init__(self, secret_key):
+    def __init__(self, secret_key: ec.EllipticCurvePrivateKey):
         self.secret_key = secret_key
         self.public_key = self.secret_key.public_key()

     @classmethod
-    def create(self, keypath="private_ecdsa.key"):  # pylint: disable=bad-classmethod-argument
+    def create(self, keypath:str ="private_ecdsa.key") -> "Signer":  ## pylint: disable=bad-classmethod-argument
         private_key = ec.generate_private_key(ec.SECP256R1(), default_backend())
         with open(keypath, "wb") as pem_out:
             pem_out.write(
@@ -39,7 +39,7 @@ class Signer:


 class Verifier:
-    def __init__(self, public_key):
+    def __init__(self, public_key: ec.EllipticCurvePublicKey) -> None:
         self.public_key = public_key

     def verify(self, message: bytes, signature: bytes) -> bool:
diff --git a/keylime/dsse/x509.py b/keylime/dsse/x509.py
index f35ef97..72c3dc7 100644
--- a/keylime/dsse/x509.py
+++ b/keylime/dsse/x509.py
@@ -11,15 +11,15 @@ from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import hashes, serialization
 from cryptography.hazmat.primitives.asymmetric import ec
 from cryptography.x509.oid import NameOID
-
+from cryptography.x509 import Certificate

 class Signer:
-    def __init__(self, secret_key):
+    def __init__(self, secret_key: ec.EllipticCurvePrivateKey):
         self.secret_key = secret_key
         self.certificate = self.construct(secret_key)

     @classmethod
-    def create(self, keypath="private_x509.key"):  # pylint: disable=bad-classmethod-argument
+    def create(cls, keypath:str="private_x509.key") -> "Signer":  ## xpylint: disable=bad-classmethod-argument
         private_key = ec.generate_private_key(ec.SECP256R1(), default_backend())
         with open(keypath, "wb") as pem_out:
             pem_out.write(
@@ -33,13 +33,13 @@ class Signer:

     @classmethod
     def construct(
-        self,
-        private_key,
-        subject_name="dsse_lib",
-        issuer_name="dsse_lib",
-        subject_alternative_name="dsse_lib",
-        expiration=30,
-    ):  # pylint: disable=bad-classmethod-argument
+        cls,
+        private_key: ec.EllipticCurvePrivateKey,
+        subject_name:str="dsse_lib",
+        issuer_name:str="dsse_lib",
+        subject_alternative_name:str="dsse_lib",
+        expiration:int=30,
+    ) -> Certificate:  # xpylint: disable=bad-classmethod-argument
         one_day = datetime.timedelta(1, 0, 0)
         public_key = private_key.public_key()
         builder = x509.CertificateBuilder()
@@ -88,13 +88,14 @@ class Signer:


 class Verifier:
-    def __init__(self, certificate):
+    def __init__(self, certificate: Certificate):
         self.certificate = certificate

     def verify(self, message: bytes, signature: bytes) -> bool:
         """Returns true if `message` was signed by `signature`."""
         try:
             public_key = self.certificate.public_key()
+            assert isinstance(public_key, (ec.EllipticCurvePublicKey))
             public_key.verify(signature, message, ec.ECDSA(hashes.SHA256()))
             return True
         except InvalidSignature as _:
```

keylime/dsse/dsse.py
```diff
@@ -8,26 +8,29 @@
...
+from abc import abstractmethod
 
 # Protocol requires Python 3.8+.
 from typing import Iterable, List, Optional, Tuple
 
 
 class Signer:
-    def sign(self, message: bytes) -> bytes:  # pylint: disable=unused-argument,no-self-use,useless-option-value
+    @abstractmethod
+    def sign(self, message: bytes) -> bytes:  # xpylint: disable=unused-argument,no-self-use,useless-option-value
```
stefanberger, line 19: this xpylint is just done for disablement and from what it looked like the whole thing pylint comment can be removed -- the other ones below as well

keylime/dsse/x509.py
```diff
@@ -10,16 +10,17 @@
...
 
 class Signer:
-    def __init__(self, secret_key):
+    def __init__(self, secret_key: ec.EllipticCurvePrivateKey):
         self.secret_key = secret_key
         self.certificate = self.construct(secret_key)
 
     @classmethod
-    def create(self, keypath="private_x509.key"):  # pylint: disable=bad-classmethod-argument
+    def create(self, keypath: str = "private_x509.key") -> "Signer":  #  xpylint: disable=bad-classmethod-argument
```
stefanberger, line 23: The `self` should become `cls`, but that's 'normal.'

keylime/dsse/dsse.py
```diff
@@ -8,26 +8,29 @@
...
         """Returns the signature of `message`."""
 
-    def keyid(self) -> Optional[str]:  # pylint: disable=no-self-use,useless-option-value
+    def keyid(self) -> Optional[str]:  # xpylint: disable=no-self-use,useless-option-value
```
stefanberger, line 22: odd: no @abstracmethod was needed here

keylime/dsse/dsse.py
```diff
@@ -50,19 +53,19 @@ def b64enc(m: bytes) -> str:
...
+        decoded = binascii.unhexlify(m_bytes)
     except binascii.Error:
         pass
     if not decoded:
         try:
-            decoded = base64.b64decode(m, validate=True)
+            decoded = base64.b64decode(m_bytes, validate=True)
         except binascii.Error:
-            decoded = base64.b64decode(m, altchars="-_", validate=True)
+            decoded = base64.b64decode(m_bytes, altchars="-_", validate=True)  # type: ignore
```
stefanberger, line 68: One of the tools didn't realize that altchars is allowed to be 2 ASCII characters (str), so what else can one do than have it ignore it...

stefanberger:
@mbestavros Please cherry-pick the top-most patch from here to complete the annotation: https://github.com/stefanberger/keylime/commits/stefanberger/mbestavros%2Bdsse-signing%2BANNOTATIONS

I am not sure about this modification of the code that is supposed to provide some sort of name in the first field of a tuple:

```
-            (hashlib.sha256(cert), dsse_x509.Verifier(load_pem_x509_certificate(cert))) for cert in certificates
+            (str(hashlib.sha256(cert).digest()), dsse_x509.Verifier(load_pem_x509_certificate(cert)))
+           for cert in certificates
```

mbestavros:
@stefanberger That's very helpful, thanks! The verifier name was my mistake, I think it should be `hexdigest()` instead of `digest()`. I've made that change.

keylime/signing.py
```diff
@@ -83,3 +92,49 @@
...
+            if not os.path.exists(pubkey_dir):
+                os.makedirs(pubkey_dir)
+            with open(pubkey_path, "wb") as f:
+                f.write(runtime_policy_key_bytes)
+        else:
+            runtime_policy_key_bytes = None
+            for sig in runtime_policy_json["signatures"]:
+                keyid = sig["keyid"]
+                pubkey_path = pubkey_dir + keyid + ".pub"
+                if os.path.isfile(pubkey_path):
```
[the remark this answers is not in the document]
mpeters: @mbestavros I assume these security warnings are fine to dismiss because we need to use the user input for the filename and we aren't just letting them reference a full file path or doing things like include wildcards or "../" by forcing the prefix and the suffix. But just wanted to confirm.

maugustosilva:
@mbestavros ok, we are almost there. Two things left: a) rebase and b) please squash some of the commits (not asking for a single commit, but just some consolidation). Thank you!

keylime/signing.py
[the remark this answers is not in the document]
mbestavros: @mpeters I'd say so, especially because it isn't even "user-provided" per se - this code writes keyfiles to disk, and it just names them with the SHA256 hash of the keyfile (to find it later). So the only "arbitrary input" here is the hash of the file. If an adversary is able to hijack SHA256, we have much bigger problems :)

keylime/dsse/x509.py
```diff
@@ -0,0 +1,109 @@
...
+        builder = builder.add_extension(
+            x509.BasicConstraints(ca=False, path_length=None),
+            critical=True,
+        )
+        certificate = builder.sign(
+            private_key=private_key,
+            algorithm=hashes.SHA256(),
+        )
+
+        with open("certificate.crt", "wb") as crt_out:
```
stefanberger, line 79: It writes this file and I think the help description for the x509 backend should describe it.

mbestavros: @stefanberger Where would you recommend I put documentation for that?

stefanberger: I think you should mention it as part of the x509 backend though the help for the "--backend' option seems also not such a great place. Another possibility would be to introduce an option that gives the user control over the name of the certificate file, just like the name of the key file.

mbestavros:
@maugustosilva I've rebased and split my commits into three: one for the signing tool, one for the signature support, and one for the updated tests. We can drop the last one as needed.

mbestavros:
@stefanberger I've added an option to the signing tool to configure the certificate output path.

@maugustosilva The associated [Packit test PR](https://github.com/RedHat-SP-Security/keylime-tests/pull/312) has been approved. Pending a final passing test run, is this ready to merge?

keylime/signing.py
```diff
@@ -83,3 +92,54 @@ def verify_signature(key: bytes, sig: bytes, body: bytes) -> bool:
...
+
+
+def get_runtime_policy_keys(runtime_policy: bytes, pubkey: Optional[str] = None) -> Optional[bytes]:
+    runtime_policy_json = json.loads(runtime_policy)
+    if runtime_policy_json.get("payload"):
+        pubkey_dir = "/var/lib/keylime/signing_keys/"
+        if pubkey:
+            runtime_policy_key_bytes = base64.b64decode(pubkey)
+            keyid = hashlib.sha256(runtime_policy_key_bytes).hexdigest()
+            pubkey_path = pubkey_dir + keyid + ".pub"
```
stefanberger, line 129: os.path.join() ?

keylime/signing.py
```diff
@@ -83,3 +92,54 @@ def verify_signature(key: bytes, sig: bytes, body: bytes) -> bool:
...
+            logger.info("Writing provided DSSE public key to disk at %s", pubkey_path)
+            if not os.path.exists(pubkey_dir):
+                os.makedirs(pubkey_dir)
+            with open(pubkey_path, "wb") as f:
+                f.write(runtime_policy_key_bytes)
+        else:
+            runtime_policy_key_bytes = None
+            for sig in runtime_policy_json["signatures"]:
+                keyid = sig["keyid"]
+                pubkey_path = pubkey_dir + keyid + ".pub"
```
stefanberger, line 139: Same here.

keylime/signing.py
mbestavros: Done in my latest push!

keylime/signing.py
mbestavros: Done!

mbestavros:
@kkaarreell Any idea why I'm seeing a single `500 Internal Server Error` in my test runs? I can't seem to track down any additional logging output to see what might be going wrong.

kkaarreell:
> @kkaarreell Any idea why I'm seeing a single `500 Internal Server Error` in my test runs? I can't seem to track down any additional logging output to see what might be going wrong.

Me neither. Is this error reported by the verifier? The only thing in the verifier log is
```
Mar 10 16:26:58 ip-172-31-29-126.us-east-2.compute.internal keylime_verifier[104393]: 2023-03-10 16:26:58.002 - keylime.verifier - INFO - POST returning 201
Mar 10 16:26:59 ip-172-31-29-126.us-east-2.compute.internal keylime_verifier[104391]: 2023-03-10 16:26:59.730 - keylime.signing - INFO - Writing provided DSSE public key to disk at /var/lib/keylime/signing_keys/5591d8b9b81380df046ddb8e734fb65c9163b7308674a252a733e0860d200348/.pub
Mar 10 16:27:01 ip-172-31-29-126.us-east-2.compute.internal keylime_verifier[104393]: 2023-03-10 16:27:01.483 - keylime.verifier - INFO - POST returning 201
Mar 10 16:27:04 ip-172-31-29-126.us-east-2.compute.internal keylime_verifier[104391]: 2023-03-10 16:27:04.927 - keylime.verifier - INFO - DELETE returning 204 response for allowlist: list1
```

mbestavros:
@kkaarreell Yes, the error should be on the verifier. I truly have no idea what's causing it, though the only time it happens is when we attempt to add a runtime policy from URL with a signature.

Maybe there's some bad code path on the verifier when verifying signed policies?

kkaarreell:
I think that the related code would deserve more debug output to make future troubleshooting easier.

mbestavros:
@kkaarreell Yep, it was a bug in the public key storage code. It's been fixed now.

I'm inclined to leave the code as is, since the error was because of a bug in code (that tests caught) and not a bug in behavior.

mbestavros:
@maugustosilva @mpeters My latest push should fix all tests. Is this ready to be merged?

mbestavros:
@maugustosilva Finally fixed the lint issue! I thought I'd gotten it before, but it was a tricky one. And all tests pass (besides CodeQL). Ready to merge?

kkaarreell:
> @kkaarreell I forget the process - does the [testing PR](https://github.com/RedHat-SP-Security/keylime-tests/pull/312) or this PR merge first?

Yes, please restore packit_ci.fmf file. And also redirect in main.fmf in tests. I will merge test PR and then we can merge this one right after that.

maugustosilva:
@mbestavros Pretty great, many thanks for you patience :-)

@kkaarreell Is there an outstanding PR on RedHat Tests that I should wait?

mbestavros:
@maugustosilva The testing PR has been merged, and I've removed the redirect commit - this is ready to merge!

maugustosilva:
@mbestavros I am pretty sure the error with FC36 was a fluke. Retesting...

maugustosilva:
As it appeared (from inspecting the test logs). Merging now.

PR #1318, opened 2023-02-20
Install create_policy as keylime_create_policy, annotate sources, and extend it

stefanberger:
This PR annotates the keylime/cmd/create_policy.py file (former scripts/create_policy.py) and makes it available as keylime_create_policy.

PR #1319, opened 2023-02-21
create_policy: Avoid duplicate entries in lists

stefanberger:
Every list in a map is basically a set of values (hashes) that shouldn't
have any duplicates. Implement add_to_list_in_map() to insert a value
into a list with only unique values (set) held by a key in a map.


aplanas:
IIUC there is only one commit extra from https://github.com/keylime/keylime/pull/1318. I do this too in some projects: create new PR that includes commits from non-(yet)-merged PR, and before merging rebase to make them disappear.

But because we changed recently the policy of squashing-before-merge, I think that makes sense to move the last commit into the old PR, and extend the reach of the original one.

stefanberger:
> But because we changed recently the policy of squashing-before-merge, I think that makes sense to move the last commit into the old PR, and extend the reach of the original one.

Ok, closing it.

keylime/cmd/create_policy.py
```diff
@@ -151,17 +157,11 @@ def process_ima_buf_in_measurement_list(
...
-                        keyrings_map[path] = [checksum_hash]
+                    keyrings_map[path] = add_to_list(keyrings_map.get(path), checksum_hash)
                     continue
 
                 if get_ima_buf:
-                    if path in ima_buf_map:
-                        ima_buf_map[path].append(checksum_hash)
-                    else:
-                        ima_buf_map[path] = [checksum_hash]
+                    ima_buf_map[path] = add_to_list(ima_buf_map.get(path), checksum_hash)
```
aplanas, line 164: This is only covering part of the places where this can goes wrong. I propose to add a new extra step around line 504, before`jsonpolicy = json.dumps(policy)`, that will normalize the policy removing duplicates. This can be done with something like: `policy["ima_buf"] = { k: sorted(list(set(v))) for k, v in policy["ima-buf"] }`, and do the same for `policy["digest"]`

stefanberger: > policy["ima_buf"] = { k: sorted(list(set(v))) for k, v in policy["ima-buf"] }

Sounds good. Add this to the other PR now.

stefanberger: @aplanas Thanks.

PR #1322, opened 2023-02-22
ima: do not validate against the allowlist if signature was already validated

THS-on:
Old behavior: If the path is also in the allowlist, it needs to also match the digest in there if the signature is valid.

New behavior: Entry is valid if the signature is valid, only when verification fails we check the allowlist.

### DIFF
--- a/keylime/ima/ima.py
+++ b/keylime/ima/ima.py
@@ -188,37 +188,26 @@ def _validate_ima_sig(
     signature: ast.Signature,
 ) -> Failure:
     failure = Failure(Component.IMA, ["validator", "ima-sig"])
-    valid_signature = False
     if ima_keyrings and signature:
         if exclude_regex is not None and exclude_regex.match(path.name):
             logger.debug("IMA: ignoring excluded path %s", path.name)
             return failure
 
-        if not ima_keyrings.integrity_digsig_verify(signature.data, digest.hash, digest.algorithm):
-            logger.warning("signature for file %s is not valid", path.name)
-            failure.add_event("invalid_signature", f"signature for file {path.name} is not valid", True)
+        if ima_keyrings.integrity_digsig_verify(signature.data, digest.hash, digest.algorithm):
+            logger.debug("signature for file %s is good", path)
             return failure
 
-        valid_signature = True
-        logger.debug("signature for file %s is good", path)
-
-    # If there is also an allowlist verify the file against that but only do this if:
-    # - we did not evaluate the signature (valid_siganture = False)
-    # - the signature is valid and the file is also in the allowlist
-    if (
-        allowlist is not None
-        and allowlist.get("digests") is not None
-        and ((allowlist["digests"].get(path.name, None) is not None and valid_signature) or not valid_signature)
-    ):
-        # We use the normal ima_ng validator to validate hash
+    # If signature validation failed check if the allowlist matches
+    if allowlist is not None:
+        logger.debug("signature for file %s could not be validated. Trying allowlist.", path.name)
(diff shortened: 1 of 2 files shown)


keylime/ima/ima.py
```diff
@@ -188,37 +188,26 @@ def _validate_ima_sig(
...
+        logger.debug("signature for file %s could not be validated. Trying allowlist.", path.name)
         return _validate_ima_ng(exclude_regex, allowlist, digest, path)
 
     # If we don't have a allowlist and don't have a keyring we just ignore the validation.
     if ima_keyrings is None:
         return failure
 
-    if not valid_signature:
-        failure.add_event("invalid_signature", f"signature for file {path.name} could not be validated", True)
+    logger.warning("signature for file %s could failed and no allowlist is available")
```
stefanberger, line 209: signature verification for file %s failed and no...

THS-on: fixed

stefanberger:
LGTM

keylime/ima/ima.py
```diff
@@ -188,37 +188,26 @@ def _validate_ima_sig(
...
-    # - we did not evaluate the signature (valid_siganture = False)
-    # - the signature is valid and the file is also in the allowlist
-    if (
-        allowlist is not None
-        and allowlist.get("digests") is not None
-        and ((allowlist["digests"].get(path.name, None) is not None and valid_signature) or not valid_signature)
-    ):
-        # We use the normal ima_ng validator to validate hash
+    # If signature validation failed check if the allowlist matches
+    if allowlist is not None:
```
aplanas, line 201: I am always confused by this double negation. I suggest: `if allowlist:`

aplanas (approved):
One nit, but I think that I get the reason of this change this change. LGTM

keylime/ima/ima.py
THS-on: I think it comes from that `{}` is also `False` and there is a difference between those two:

- `{}` empty allowlist everything will fail to validate against
- `None` no allowlist is supplied and therefore IMA validation is not enabled. Validation passes.

aplanas: Very subtle, but OK for me : )

aplanas:
seems that the testing-farm tests needs to be updated too:

```
:: [ 21:44:02 ] :: [   FAIL   ] :: File '/var/tmp/limeLib/verifier.log' should contain 'WARNING - signature for file /keylime-tests/basic-attestation-with-ima-signatures-17hor/echo is not valid' 

:: [ 21:44:02 ] :: [   PASS   ] :: File '/var/tmp/limeLib/verifier.log' should contain 'ERROR - IMA ERRORS: Some entries couldn't be validated' 
```
:

maugustosilva (approved):

maugustosilva:
LGTM, but the CI tests need to be fixed.

THS-on:
@kkaarreell the error now looks like this:
```
keylime.ima - WARNING - File not found in allowlist: /keylime-tests/basic-attestation-with-ima-signatures-gkdyq/echo
keylime.ima - ERROR - IMA ERRORS: Some entries couldn't be validated. Number of failures in modes: ImaSig 1.
```

Can you modify the check or should I do that?

kkaarreell:
> Can you modify the check or should I do that?

I will do it.

kkaarreell:
> > Can you modify the check or should I do that?
> I will do it.

Test update is merged now, you can reschedule tests.

THS-on:
@kkaarreell thank you!

maugustosilva:
@THS-on can you please rebase? I just merged the PR which disables Fedora Rawhide (due to a bug)

PR #1329, opened 2023-03-03
Update path mapping in codecov.yml

kkaarreell:


mpeters (approved):

kkaarreell:
Test failure on CS8 is not related to this change. Agent log states `ERROR - Agent did not return proper quote due to PCR race condition.` and most likely rerun would pass (not sure if it is worth it).

mpeters:
Might as well re-run it, no harm in doing so.

Issue #1283, opened 2023-01-26
verifier: agent event loops are not distributed on multiple processes after restart


THS-on:
# Environment
 * OS / version:  any
 * Processor architecture:  any
 * TPM Manufacturer: any
 * Keylime version:  master (6.5.3)

# Description 
When the verifier starts up it looks in the database for existing agents and restarts the event loop for them.
This is currently only done on the first verifier process: 

https://github.com/keylime/keylime/blob/7f9483af909500e8a5f90d07560b334ec573a9ce/keylime/cloud_verifier_tornado.py#L1472-L1474

Ideally we want to redistribute the agents evenly among the started processes to allow for better scalability.

cc: @stefanberger

stefanberger:
Is the attestation workload equally distributed among all the verifier processes when doing a `tenant ... -c add` many times? The way it must be working is that the processes are competing on the listening socket via the `accept()` and one of the processes gets the socket and processes the REST call, so theoretically this should lead to equal distribution...

THS-on:
> Is the attestation workload equally distributed among all the verifier processes when doing a tenant ... -c add many times? 

Should roughly be the case.

> The way it must be working is that the processes are competing on the listening socket via the accept() and one of the processes gets the socket and processes the REST call, so theoretically this should lead to equal distribution...

Yes that is how it is implemented, but I have not verified if that actually happens in practice.

stefanberger:
> Yes that is how it is implemented, but I have not verified if that actually happens in practice.

I have not verified it, either, but this is the only way it actually can work and on which the actual equal distribution of the load depends.

stefanberger:
I just ran a test with a single registered machine and it does end up on different processes on repeated `keylime_tenant -c add` & `keylime_tenant -c remove` cycles.

PR #1330, opened 2023-03-07
verifier: Activate every m-th agent starting at the n-th agent on a w…

stefanberger:
When restarting the verifier, equally distribute the agents over the verifier worker processes by activating every m-th agent (with m being the number of worker processes) starting at the n-th agent (with n being the id of a particular worker process; the ids start at 0).

Resolves issue #1283

### DIFF
--- a/keylime/cloud_verifier_tornado.py
+++ b/keylime/cloud_verifier_tornado.py
@@ -1569,9 +1569,8 @@ def main() -> None:
         loop.add_signal_handler(signal.SIGTERM, server_sig_handler)
 
         server.start()
-        if task_id == 0:
-            # Reactivate agents
-            asyncio.ensure_future(activate_agents(agents, verifier_host, int(verifier_port)))
+        # Reactivate agents
+        asyncio.ensure_future(activate_agents(agents, verifier_host, int(verifier_port)))
         tornado.ioloop.IOLoop.current().start()
         logger.debug("Server %s stopped.", task_id)
         sys.exit(0)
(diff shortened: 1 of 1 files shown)


keylime/cloud_verifier_tornado.py
```diff
@@ -1596,7 +1598,10 @@ def sig_handler(*_: Any) -> None:
     num_workers = config.getint("verifier", "num_workers")
     if num_workers <= 0:
         num_workers = tornado.process.cpu_count()
+
+    agents = get_agents_by_verifier_id(verifier_id)
     for task_id in range(0, num_workers):
-        process = Process(target=server_process, args=(task_id,))
+        active_agents = [agents[i] for i in range(task_id, len(agents), num_workers)]
```
aplanas, line 1604: I like this one. In 3.12 maybe we can use https://docs.python.org/3.12/library/itertools.html#itertools.batched

aplanas (approved):

keylime/cloud_verifier_tornado.py
stefanberger: I think what's there right now is better for equal distribution. I wouldn't want to neglect the last process started with a really short list versus the other ones having equally filled lists. :-) 

Until we all have 3.12 it's going to be a while...

Thanks.

THS-on (approved):
LGTM. The code coverage patch drop comes just from changing existing code lightly that was not covered before, so that is not an issue.

maugustosilva (approved):

PR #1331, opened 2023-03-08
Add JSON schema for runtime policy and validate user-provided runtime policies against it

stefanberger:
This PR adds a JSON schema for the runtime policy along with its package dependency on jsonschema. It then uses the JSON schema to validate the user-provided runtime policies before they are written to the Db.

### DIFF
--- a/keylime/ima/ima.py
+++ b/keylime/ima/ima.py
@@ -6,6 +6,8 @@ import hashlib
 import json
 from typing import Any, Dict, List, Optional, Pattern, Set, TextIO, Tuple, Type
 
+import jsonschema
+
 from keylime import config, keylime_logging, signing
 from keylime.agentstates import AgentAttestState
 from keylime.common import algorithms, validators
@@ -50,6 +52,55 @@ EMPTY_RUNTIME_POLICY: RuntimePolicyType = {
     "ima": {"ignored_keyrings": [], "log_hash_alg": "sha1", "dm_policy": None},
     "ima-buf": {},
     "verification-keys": "",
+}
+
+
+RUNTIME_POLICY_SCHEMA = {
+    "type": "object",
+    "required": ["meta", "release", "digests", "excludes", "keyrings", "ima", "ima-buf", "verification-keys"],
+    "properties": {
+        "meta": {
+            "type": "object",
+            "required": ["version", "generator"],
+            "properties": {
+                "version": {"type": "integer", "minimum": 0},
+                "generator": {
+                    "type": "integer",
+                    "minimum": min(list(RUNTIME_POLICY_GENERATOR)),
+                    "maximum": max(list(RUNTIME_POLICY_GENERATOR)),
+                },
+                "timestamp": {"type": "string"},
+            },
+        },
+        "release": {"type": "integer", "minimum": 0},
+        "digests": {"$ref": "#/definitions/digests-object"},
+        "excludes": {"type": "array", "items": {"type": "string"}},
+        "keyrings": {"$ref": "#/definitions/digests-object"},
+        "ima": {
+            "type": "object",
+            "required": ["ignored_keyrings", "log_hash_alg", "dm_policy"],
+            "properties": {
+                "ignored_keyrings": {"type": "array", "items": {"type": "string"}},
+                "log_hash_alg": {"type": "string", "enum": ["sha1", "sha256", "sha384", "sha512"]},
+                "dm_policy": {"type": "null"},
+            },
+        },
+        "ima-buf": {"$ref": "#/definitions/digests-object"},
+        "verification-keys": {"type": "string"},
+    },
+    "definitions": {
+        "digests-object": {
+            "type": "object",
+            "patternProperties": {
+                "^": {
+                    "type": "array",
+                    "items": {"type": "string", "pattern": "^[0-9a-f]{40,128}$"},
+                    "minItems": 1,
+                    "uniqueItems": False,
+                }
+            },
+        }
+    },
 }
 
 
@@ -481,6 +532,9 @@ def runtime_policy_db_contents(runtime_policy_name: str, runtime_policy: str, tp
         runtime_policy_db_format["ima_policy"] = runtime_policy
     runtime_policy_bytes = runtime_policy.encode()
     runtime_policy_dict = deserialize_runtime_policy(runtime_policy)
+
+    validate_runtime_policy(runtime_policy_dict)
+
     if "meta" in runtime_policy_dict:
         if "generator" in runtime_policy_dict["meta"]:
             runtime_policy_db_format["generator"] = runtime_policy_dict["meta"]["generator"]
@@ -554,3 +608,14 @@ def deserialize_runtime_policy(runtime_policy: str) -> RuntimePolicyType:
     # TODO: Extract IMA policy JSON from DSSE envelope, if applicable.
     runtime_policy_deserialized: RuntimePolicyType = json.loads(runtime_policy)
     return runtime_policy_deserialized
+
+
+def validate_runtime_policy(runtime_policy: RuntimePolicyType) -> None:
+    """
+    Validate a runtime policy against the schema.
+    """
+    try:
+        jsonschema.validate(instance=runtime_policy, schema=RUNTIME_POLICY_SCHEMA)
+    except Exception as error:
+        msg = str(error).split("\n", 1)[0]
+        raise ImaValidationError(message=f"{msg}", code=400) from error
--- a/keylime/ima/ima.py
+++ b/keylime/ima/ima.py
@@ -1,5 +1,6 @@
 import codecs
 import copy
+import enum
 import functools
 import hashlib
 import json
@@ -21,7 +22,7 @@ logger = keylime_logging.init_logging("ima")
 RUNTIME_POLICY_CURRENT_VERSION = 1
 
 
-class RUNTIME_POLICY_GENERATOR:
+class RUNTIME_POLICY_GENERATOR(enum.IntEnum):
     Unknown = 0
     EmptyAllowList = 1
     CompatibleAllowList = 2
--- a/keylime/tenant.py
+++ b/keylime/tenant.py
@@ -1239,7 +1239,10 @@ class Tenant:
         response = cv_client.post(
             f"/v{self.api_version}/allowlists/{self.runtime_policy_name}", data=body, timeout=self.request_timeout
         )
-        Tenant._jsonify_response(response)
+        response_json = Tenant._jsonify_response(response)
+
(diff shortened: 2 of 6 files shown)


stefanberger:
This only happened **only** on CentOS-9 : https://artifacts.dev.testing-farm.io/e02ae41e-fc30-41b8-b6d2-3ec5aa4ea407/
```
  File "/var/ARTIFACTS/work-e2e-with-revocationnti_c7o8/packit-ci/e2e-with-revocation/discover/default-0/tests/functional/basic-attestation-on-localhost/convert_runtime_policy.py", line 9, in <module>
    from keylime.ima import file_signatures, ima
  File "/usr/local/lib/python3.9/site-packages/keylime-6.6.0-py3.9.egg/keylime/ima/ima.py", line 8, in <module>
    import jsonschema
ModuleNotFoundError: No module named 'jsonschema'
```
@kkaarreell If we were to merge this you would have to update the installed modules? Though not for CentOS-8 and Fedora  ?

stefanberger:
This seems to only happen with older python versions: 
```
ImportError: Failed to import test module: ima.ima_test
  File "/usr/lib64/python3.11/unittest/loader.py", line 407, in _find_test_path
    module = self._get_module_from_name(name)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/unittest/loader.py", line 350, in _get_module_from_name
    __import__(name)
  File "/__w/keylime/keylime/keylime/ima/ima_test.py", line 4, in <module>
    from keylime.ima import ima
  File "/__w/keylime/keylime/keylime/ima/ima.py", line 15, in <module>
    from keylime.ima.ima_schema import validate_runtime_policy
  File "/__w/keylime/keylime/keylime/ima/ima_schema.py", line 3, in <module>
    import jsonschema
  File "/tmp/keylime_test/lib64/python3.11/site-packages/jsonschema/__init__.py", line 23, in <module>
    from jsonschema.validators import (
  File "/tmp/keylime_test/lib64/python3.11/site-packages/jsonschema/validators.py", line 432, in <module>
    meta_schema=_utils.load_schema("draft3"),
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/keylime_test/lib64/python3.11/site-packages/jsonschema/_utils.py", line 62, in load_schema
    return json.loads(data)
           ^^^^^^^^^^^^^^^^
  File "/__w/keylime/keylime/keylime/json.py", line 59, in loads
    return json_module.loads(s, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/__w/keylime/keylime/keylime/json.py", line 59, in loads
    return json_module.loads(s, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/__w/keylime/keylime/keylime/json.py", line 59, in loads
    return json_module.loads(s, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  [Previous line repeated 944 more times]
RecursionError: maximum recursion depth exceeded
```

stefanberger:
Above issue actually only appears when running tests like this: 

```
sudo python3 -m unittest discover -s keylime -p '*_test.py' -v
```

It works fine when running a test like this:
```
sudo python -m unittest keylime/ima/ima_test.py
```

sergio-correia:
> This only happened **only** on CentOS-9 : https://artifacts.dev.testing-farm.io/e02ae41e-fc30-41b8-b6d2-3ec5aa4ea407/
> 

I don't think we have this module available in c9s, which means we would have to add it there, if we were to merge this.

stefanberger:
> I don't think we have this module available in c9s, which means we would have to add it there, if we were to merge this.

Oh, right. You don't install from pip but from rpm repo. It's available in CentOS 8 ?

kkaarreell:
For upstream testing we eventually install also from pip but better avoid adding new dependencies unless really necessary as it would require us to add this package to RHEL with future rebase.

kkaarreell:
> Above issue actually only appears when running tests like this:
> 
> ```
> sudo python3 -m unittest discover -s keylime -p '*_test.py' -v
> ```
> 
> It works fine when running a test like this:
> 
> ```
> sudo python -m unittest keylime/ima/ima_test.py
> ```

I believe these two approaches end up running different set of tests.

stefanberger:
> For upstream testing we eventually install also from pip but better avoid adding new dependencies unless really necessary as it would require us to add this package to RHEL with future rebase.

It would be good to be able to validate input JSON with a schema rather than having to do this programmatically...

stefanberger:
@kkaarreell : python3-jsonschema-3.2.0-13.el9.noarch.rpm  seems to be available

kkaarreell:
> @kkaarreell : python3-jsonschema-3.2.0-13.el9.noarch.rpm seems to be available

Right, I have just checked it is present in AppStream repository.
I have created https://github.com/RedHat-SP-Security/keylime-tests/pull/327 to install this package.

stefanberger:
> Interesting, I am wondering what made such a difference.

If I knew...

kkaarreell:
> > @kkaarreell : python3-jsonschema-3.2.0-13.el9.noarch.rpm seems to be available
> 
> Right, I have just checked it is present in AppStream repository. I have created [RedHat-SP-Security/keylime-tests#327](https://github.com/RedHat-SP-Security/keylime-tests/pull/327) to install this package.

This is merged now, you can rerun failed Packit test.

requirements.txt
```diff
@@ -22,4 +22,5 @@ jinja2>=3.0.0 # BSD
 # Those bindings must match the local GPG version and therefore this package is not installed via PyPI.
 # typing-extensions is required for TypedDict and Literal on Python < 3.8
 # as well as NotRequired on Python < 3.11
-typing-extensions>=4.0.1 ; python_version < '3.11' #PSF-2.0
\ No newline at end of file
+typing-extensions>=4.0.1 ; python_version < '3.11' #PSF-2.0
+jsonschema # MIT
```
aplanas, line 26: any constrains about the version? In SLE we have 3.2.0, as it is supporting Python 3.6. In Tumbleweed / MicroOS we have 4.17.3 that drops 3.6

keylime/ima/ima.py
```diff
@@ -51,6 +53,50 @@ class RUNTIME_POLICY_GENERATOR:
...
+RUNTIME_POLICY_SCHEMA = {
+    "type": "object",
+    "required": ["meta", "release", "digests", "excludes", "keyrings", "ima", "ima-buf", "verification-keys"],
+    "properties": {
+        "meta": {
+            "type": "object",
+            "required": ["version", "generator"],
+            "properties": {
+                "version": {"type": "integer", "minimum": 0},
+                "generator": {"type": "integer", "minimum": 0, "maximum": 3},
```
aplanas, line 65: IIRC this 0 .. 3 comes from an enum-a-like, maybe the last value should be used, (RUNTIME_POLICY_GENERATOR.LegacyAllowList)

But the best approach is, if possible, use a real enum:

```python
import enum
class RUNTIME_POLICY_GENERATOR(enum.IntEnum):
    Unknown = 0
    EmptyAllowList = 1
    CompatibleAllowList = 2
    LegacyAllowList = 3
```

So now we can use the Enum API:

```python
...
"generator": {"type": "integer", "minimum": min(list(RUNTIME_POLICY_GENERATOR)), "maximum": max(list(RUNTIME_POLICY_GENERATOR))},
```

keylime/ima/ima.py
```diff
@@ -479,6 +525,11 @@ def runtime_policy_db_contents(runtime_policy_name: str, runtime_policy: str, tp
         runtime_policy_db_format["ima_policy"] = runtime_policy
     runtime_policy_bytes = runtime_policy.encode()
     runtime_policy_dict = deserialize_runtime_policy(runtime_policy)
+
+    err_msg = validate_runtime_policy(runtime_policy_dict)
+    if err_msg:
+        raise ImaValidationError(message=f"{err_msg}", code=400)
```
aplanas, line 531: Is not weird to wrap `jsonschema.validate` into a function that catch the exception, to here revert it and raise one if there is a returned string?

aplanas (commented):
Some comments, but LGTM

requirements.txt
stefanberger: From my testing it seems that 3.2.0 is sufficient. I'll add this to the constraint.

keylime/ima/ima.py
stefanberger: I also changed this now.

aplanas (approved):
Thanks for the changes.

stefanberger:
Does anyone have any more comments to this PR?

THS-on (approved):
LGTM

keylime/tenant.py
```diff
@@ -1239,7 +1239,10 @@ def do_add_runtime_policy(self, args: Dict[str, str]) -> None:
         response = cv_client.post(
             f"/v{self.api_version}/allowlists/{self.runtime_policy_name}", data=body, timeout=self.request_timeout
         )
-        Tenant._jsonify_response(response)
+        response_json = Tenant._jsonify_response(response)
+
+        if response.status_code >= 400 and response.status_code < 500:
+            raise UserError(response_json)
```
stefanberger, line 1245: I am going to have to drop this since it causes issues with existing tests that we probably should resolve separately.

```
2023-03-17 16:23:45.310 - keylime.tenant - ERROR - {'code': 409, 'status': 'Runtime policy with name list2 already exists', 'results': {}}

:: [ 16:23:45 ] :: [   FAIL   ] :: Command 'keylime_tenant -c addruntimepolicy --runtime-policy policy.json --runtime-policy-name list2' (Expected 0, got 1)
```

stefanberger:
/packit retest-failed

Hoping this goes away:
```
: [ 13:44:31 ] :: [  BEGIN   ] :: Running './patchcov.py /var/tmp/limeLib/coverage/patch.txt /var/tmp/limeLib/coverage/.coverage 0'

  File "/var/ARTIFACTS/work-e2e-with-revocationcv1ks0an/packit-ci/e2e-with-revocation/discover/default-0/tests/setup/generate_coverage_report/./patchcov.py", line 74

    if l == '}' || l == '},':

                 ^

SyntaxError: invalid syntax

:: [ 13:44:31 ] :: [   FAIL   ] :: Command './patchcov.py /var/tmp/limeLib/coverage/patch.txt /var/tmp/limeLib/coverage/.coverage 0' (Expected 0, got 1)

:: [ 13:44:31 ] :: [  BEGIN   ] :: Running 'rm -rf /tmp/tmp.RrOgtfS3CT'
```

kkaarreell:
I am sorry for that, should be fixed now.

PR #1333, opened 2023-03-10
codestyle: Define RuntimePolicyType and use it

stefanberger:


keylime/types.py
```diff
@@ -1,9 +1,48 @@
+import sys
 import typing
 
+from keylime.ima.ima_dm import Policies
```
stefanberger, line 4: The inclusion of this here could cause some issues down the road. Maybe the types would have to be moved into this file here or into keylime.ima.ima_dm_types.

THS-on: Maybe we can put all the IMA related types into `keylime.ima.types` and keep only the more general ones in `keylime.types`?

stefanberger: I moved it now.

THS-on (approved):
LGTM

PR #1339, opened 2023-03-29
Refactor tests

kamaxeon:
* Introduce unittest subtests
  * We can differentiate between test case definitions and testing code.
  * All subtests will be executed, even when the previous one failed.
  * When a test fails, we can see a message to identify the test case.
* Fix typo

### DIFF
--- a/test/test_api_version.py
+++ b/test/test_api_version.py
@@ -9,36 +9,57 @@ from keylime import api_version
 
 
 class APIVersion_Test(unittest.TestCase):
+    """Test to check API version."""
+
     def test_current_version(self):
+        """Test current_version."""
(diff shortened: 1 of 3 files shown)


aplanas (approved):
I do not mind the changes. IMHO tests should be trivial to read and write, like the old code was.

Smart code in tests is problematic, and this change goes one (small) step in this direction. Still easy to read and write, tho.

In fact, you can go one step further (IMHO, again, don't) and write the pattern of case iteration in a function that expect the cases and the function that needs to be called.

In any case, welcome : )

THS-on (approved):

maugustosilva (approved):

PR #1340, opened 2023-03-31
ca:CLI utilities (keylime_ca,keylime_tenant) read password from ca.conf

### LINKED ISSUE #1135: keystore password is required when adding compute into keylime_verifier with keylime_tenant

# Environment
 * OS / version: Ub22.04
 * Processor architecture: N/A
 * TPM Manufacturer: N/A
 * Keylime version: v6.5.1

# Description 
Keylime CA component requires a password set in its config (ca.conf)

# The keystore password
# This password is used to protect the generated CA private key.
password = some_password

if I don't set any password here, the keylime_verifier service will not start...and then if I set some password in this config, every time I am adding some compute/agent into verifier (via keylime_tenant) to be attestated, I am asked for this password from ca.conf:

This prompt shows up:
Please enter the password to decrypt your keystore:

would be great to have a way around this, so I can add compute/agent into verifier without being prompted for this password?

No keystore password is needed when adding compute/agent into keylime_verifier via keylime_tenant tool.

## Steps to reproduce problem
1. Start keylime server
2. Start keylime agent
3. Add compute/agent into verifier using keylime_tenant tool

# Relevant logs

maugustosilva:
Fixes #1135

This is a minimalistic fix for the issue of keystore certificate passords. Instead of prompting an user to enter a password during the exection of commands such as `keylime_ca --command init` or `keylime_tenant -c add ... --cert <cert dir>`, the new code now simply reads it from `/etc/keylime/ca.conf`.

Please note that an user can specify multiple different passwords (which can be required in case certificates to multiple nodes are created) by simply executing the commands prefaced by
`KEYLIME_CA_PASSWORD=<password>`


THS-on:
I think the test fail now because the expect script checking for that is no longer required.

How was the password option in the config used before, because if I remember correctly the created CAs for e.g. revocation messages just asked via user input?

keylime/ca_util.py
```diff
@@ -63,15 +63,22 @@ def load_cert_by_path(cert_path: str) -> Certificate:
     return cert
 
 
-def setpassword(pw: Optional[str]) -> None:
+def setpassword(key_store_pw: Optional[str]) -> None:
     global global_password
-    if not pw:
-        pw = getpass.getpass("Please enter the password to decrypt your keystore: ")
+    if not key_store_pw:
```
aplanas, line 68: Does make sense to have an empty password?, if so the condition should be `if key_store_pw is None:`

keylime/ca_util.py
```diff
@@ -63,15 +63,22 @@ def load_cert_by_path(cert_path: str) -> Certificate:
...
+        key_store_pw = config.get("ca", "password")
+        if key_store_pw:
+            if key_store_pw == "default":
+                if logger:
+                    logger.warning("Using 'default' password option from CA configuration file")
+            key_store_pw = str(key_store_pw)
```
aplanas, line 74: No need? The `config.get` already return a string (per the signature)

keylime/ca_util.py
```diff
@@ -63,15 +63,22 @@ def load_cert_by_path(cert_path: str) -> Certificate:
...
 
 
-def setpassword(pw: Optional[str]) -> None:
+def setpassword(key_store_pw: Optional[str]) -> None:
     global global_password
-    if not pw:
-        pw = getpass.getpass("Please enter the password to decrypt your keystore: ")
+    if not key_store_pw:
+        key_store_pw = config.get("ca", "password")
+        if key_store_pw:
```
aplanas, line 70: Should be noted that the default fallback is the empty string, so a missing password will signaled as the empty string.

keylime/ca_util.py
maugustosilva: No empty password possible. This code was basically transplanted from the `verifier` to here, in order to keep the code (and messages) as close as original as possible.

keylime/ca_util.py
maugustosilva: Again, the idea was the keep the messages emitted by the different components close to the "original behavior"

keylime/ca_util.py
maugustosilva: True, wil fix it.

maugustosilva:
> I think the test fail now because the expect script checking for that is no longer required.
> 
> How was the password option in the config used before, because if I remember correctly the created CAs for e.g. revocation messages just asked via user input?

I will open a PR on the tests. Or make the reading of the password from the file optional, still pondering.

maugustosilva:
> 

Yeah, I forgot to `run_local.sh` before pushing. Fixing it now.

keylime/ca_util.py
maugustosilva: Fixed

keylime/ca_util.py
```diff
@@ -63,23 +63,31 @@ def load_cert_by_path(cert_path: str) -> Certificate:
...
-def setpassword(pw: Optional[str]) -> None:
+def setpassword(key_store_pw: Optional[str], key_usage: Optional[str] = "server") -> None:
     global global_password
-    if not pw:
-        pw = getpass.getpass("Please enter the password to decrypt your keystore: ")
+    if not key_store_pw and not global_password:
+        key_store_pw = config.get("ca", "password", fallback="default")
+        if key_store_pw == "default":
+            if key_usage == "server":
+                logger.warning("Using 'default' password option from CA configuration file")
```
aplanas, line 72: Somehow `setpasword` is not anymore the correct function. Now is acting as "set password from the CLI", "ask password" and "read password from config file". I think that this should be splited in three and used when makes sense, and avoid the new `key_usage` pollution in the different signatures.

aplanas: Uhmm one option can be to have `set_password(key_store_pw: str)`, `read_password() -> str` and `ask_password() -> str`. The invariant now can be that `read_private(warn: bool = False) -> Tuple[Dict[str, Any], str]` expect that the `global_password` is always set (we can put an assert there and exit if this is not the case) but will never ask nor read the password as was done before.

Before using any `ca_util` function, the user will be sure that the password is set, calling always `set_password()` with the output of the read or ask version, depending if the entry point is from a cli or not.  That way we can remove `key_usage` from all the signatures.

maugustosilva: Hmmmm... I think I get it, but let me restate it for clarity: instead of letting the internal logic in `setpassword` to decide when it is acceptable to read the password from the configuration file, we simply move it to this decision to individual invocation of `setpassword`, replacing it with `read_password` and `ask_password` when needed?

maugustosilva: I think I can see the logic of what you're proposing, given that the server-side components (i.e., `verifier`) will never actually prompt a user for a password. Let me rework this a bit more then.

aplanas: Yes, thank you for using better wording!

aplanas (approved):

THS-on (approved):
LGTM

PR #1341, opened 2023-03-31
log: add a barebones log config in case configuration files not present

### LINKED ISSUE #1116: `keylime_tenant --help` backtraces if logging config exists but is not readable
# Environment
 * OS / version: Fedora 36
 * Keylime version: git

# Reproducer

```
$ KEYLIME_LOGGING_CONFIG=/etc/keylime.conf ./bin/keylime_tenant --help
Reading configuration from []
  File "/home/crobinso/.local/share/virtualenvs/keylime-x34CgAWE/./bin/keylime_tenant", line 5, in <module>
    from keylime.cmd.tenant import main
  File "/home/crobinso/.local/share/virtualenvs/keylime-x34CgAWE/lib/python3.10/site-packages/keylime/cmd/tenant.py", line 3, in <module>
  File "/home/crobinso/.local/share/virtualenvs/keylime-x34CgAWE/lib/python3.10/site-packages/keylime/keylime_logging.py", line 15, in <module>
    logging_config.fileConfig(config.get_config("logging"))
  File "/usr/lib64/python3.10/logging/config.py", line 72, in fileConfig
    formatters = _create_formatters(cp)
  File "/usr/lib64/python3.10/logging/config.py", line 105, in _create_formatters
    flist = cp["formatters"]["keys"]
  File "/usr/lib64/python3.10/configparser.py", line 964, in __getitem__
    raise KeyError(key)
KeyError: 'formatters'
```

This is kind of an artificial reporducer, so let me explain: keylime in fedora 36 is the non-latest 6.4.3 version. Run `keylime_tenant --help` as non-root with stock packages, and you get a similar backtrace. The issue is the config lookup just checks for essentially `os.path.exists("/etc/keylime.conf")` and not whether the user can read the config, and the error flows from there.

Upstream code config layout has changed, but the root problem still appears to be there, but whether it's busted out of the box or not on Fedora will depend on `/etc/keylime` directory permissions I think.

maugustosilva:
Fixes #1116

### DIFF
--- a/keylime/keylime_logging.py
+++ b/keylime/keylime_logging.py
@@ -5,7 +5,10 @@ from typing import Any, Callable, Dict
 
 from keylime import config
 
-logging_config.fileConfig(config.get_config("logging"))
+try:
+    logging_config.fileConfig(config.get_config("logging"))
+except KeyError:
+    logging.basicConfig(format="%(asctime)s %(name)-12s %(levelname)-8s %(message)s", level=logging.DEBUG)
 
 
 def set_log_func(loglevel: int, logger: Logger) -> Callable[..., None]:
--- a/keylime/config.py
+++ b/keylime/config.py
@@ -157,6 +157,10 @@ def get_config(component: str) -> RawConfigParser:
         # If no component was specified, return an empty configuration
         raise Exception("No component provided to get_config")
 
+    conf_create_msg = (
+        '\nPlease use "keylime_upgrade_config --defaults" to create a minimalistic set of configurations.\n'
+    )
+
     if component not in _config:  # pylint: disable=too-many-nested-blocks
         # Use RawConfigParser, so we can also use it as the logging config
         _config[component] = RawConfigParser()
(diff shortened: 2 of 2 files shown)


keylime/keylime_logging.py
```diff
@@ -5,7 +5,20 @@
...
+        version=1,
+        formatters={"f": {"format": "%(asctime)s %(name)-12s %(levelname)-8s %(message)s"}},
+        handlers={"h": {"class": "logging.StreamHandler", "formatter": "f", "level": logging.DEBUG}},
+        root={
+            "handlers": ["h"],
+            "level": logging.DEBUG,
+        },
+    )
+
+    logging_config.dictConfig(logging_barebones_config)
```
aplanas, line 21: Maybe a more direct approach can be:

```python
...
except KeyError:
    logging.basicConfig(format="%(asctime)s %(name)-12s %(levelname)-8s %(message)s", level=logging.DEBUG)
```

maugustosilva: Nice, you're right, it is more direct (and thus better).

maugustosilva:
> Looks like a good plan to fallback on debug logging if no config could be found, but maybe also output a warning that we are doing that.
> 
> Can you also address the failing pylint check:
> 
> ```
> ************* Module keylime.keylime_logging
> keylime/keylime_logging.py:11:47: R1735: Consider using '{"version": 1, "formatters": {'f': {'format': '%(asctime)s %(name)-12s %(levelname)-8s %(message)s'}}, ... }' instead of a call to 'dict'. (use-dict-literal)
> ```

Fixed it and, and added the appropriate warning.

THS-on (approved):

aplanas (approved):

PR #1344, opened 2023-04-06
da: non-zero exit code for attestation replay failures.

### LINKED ISSUE #1334: keylime_attest should return non-zero exit code on attestation failure

# Environment
 * OS / version: Fedora 37
 * Processor architecture:  x86_64
 * TPM Manufacturer: swtpm
 * Keylime version: current upstream [5278a84](https://github.com/keylime/keylime/commit/5278a84fcac927da49933a8396874909fc7dfa59)

# Description 
When running `keylime_attest` to do off-line attestation the script itself should return non-zero exit code (e.g. 1) when some agent fails the attestation.

Exit code is 0 both for successful and unsuccessful attestation.

## Steps to reproduce problem
1. fail the agent
2. run `keylime_attest` script
See e.g. https://github.com/RedHat-SP-Security/keylime-tests/pull/336 (durable-attestation test is passing because the test has not yet been updated to expect non-zero exit code)

# Relevant logs
```
               out: :: [ 10:39:24 ] :: [   PASS   ] :: Command 'keylime_attest' (Expected 0, got 0)
                out: :: [ 10:39:24 ] :: [   PASS   ] :: File '/var/tmp/rlRun_LOG.WlcJOaB1' should contain 'WARNING - File not found in allowlist: /bad-script.sh' 
                out: :: [ 10:39:24 ] :: [   PASS   ] :: File '/var/tmp/rlRun_LOG.WlcJOaB1' should contain 'ERROR - IMA ERRORS: Some entries couldn't be validated.' 
                out: 2023-03-13 10:39:24.014 - keylime.durable_attestation_fetch_and_replay - INFO - ---------- Agent d432fbb3-d2f1-4a97-9ef7-75bd81c00000 was NOT in "attested" state at 03/13/2023, 10:39:22 (TPM delta: 0): ima.validation.ima-ng.not_in_allowlist {"message": "File not found in allowlist: /bad-script.sh"}
                out: :: [ 10:39:24 ] :: [   PASS   ] :: File '/var/tmp/rlRun_LOG.WlcJOaB1' should contain 'keylime.durable_attestation_fetch_and_replay - INFO -.*Agent d432fbb3-d2f1-4a97-9ef7-75bd81c00000 was NOT in "attested" state at.*' 
```

maugustosilva:
`keylime_attest` CLI will build a list of nodes which failed at some point in the recorded history of attestation artifacts on the persistent store, and in case this list is non-empty, the exit code will be non-zero.

Fixes #1334

### DIFF
--- a/keylime/da/attest.py
+++ b/keylime/da/attest.py
@@ -1,6 +1,7 @@
 #!/usr/bin/python3
 import argparse
 import json
+import sys
 from datetime import datetime
 
 from keylime import cloud_verifier_common, config, keylime_logging
@@ -78,6 +79,8 @@ def main() -> None:
         logger.info("=> Focusing on the agent %s on the persistent store.", args.agent_uuid)
         agent_list = [args.agent_uuid]
 
+    failed_agents = []
+
     for agent_uuid in agent_list:
         logger.info("===> Getting all existing registration records for agent %s...", agent_uuid)
 
@@ -132,6 +135,8 @@ def main() -> None:
             if failure:
                 if failure.events:
                     quote_failed = True
+                    if agent_uuid not in failed_agents:
+                        failed_agents.append(agent_uuid)
 
             if quote_failed:
                 f_e_id = "NA"
@@ -148,7 +153,7 @@ def main() -> None:
                         f_e_id,
                         f_e_ctx,
                     )
-                    return
+                    break
             else:
                 logger.info(
                     '---------- Agent %s was in "attested" state at %s (TPM delta: %s)',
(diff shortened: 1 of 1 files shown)


keylime/da/attest.py
```diff
@@ -148,11 +153,20 @@ def main() -> None:
...
                     agent_uuid,
                     attestation_record["verifier" + "_timestamp"],
                     d_tpm_ts,
                 )
+
+    if failed_agents:
+        logger.info(
+            "=> The following agents failed attestation at some point in recorded history: %s", ",".join(failed_agents)
+        )
+        sys.exit(255)
```
aplanas, line 169: Usually an error is signed as `sys.exit(1)` in this codebase (there are some -1, that correspond to 255, and one 2).  Unless we want to provide meaning to the exit code (in that case I suggest to use os.EX_* family), we should normalize it to 1 here too.

keylime/da/attest.py
```diff
@@ -148,11 +153,20 @@ def main() -> None:
...
+    else:
+        logger.info("=> All agents were in attested state for the whole duration of recorded history")
+        sys.exit(0)
```
aplanas, line 172: not needed

keylime/da/attest.py
maugustosilva: Make sense.

aplanas (approved):

PR #1348, opened 2023-04-20
setup.cfg: use license_files instead of license_file

aplanas:
license_file is deprecated in favor of license_files in setup.cfg. This generate a DistDeprecationWarning exception, making some test to fail.


stefanberger:
LGTM

maugustosilva (approved):

PR #1349, opened 2023-04-20
Replace tpm2_print

stefanberger:
To simplify code and get rid of tpm2_print tool, implement a function that parses TPMS_ATTEST so far that it can extract TPMS_CLOCK_INFO.

### DIFF
--- a/keylime/tpm/tpm2_objects.py
+++ b/keylime/tpm/tpm2_objects.py
@@ -1,6 +1,6 @@
 import hashlib
 import struct
-from typing import Any, Tuple, Union, cast
+from typing import Any, Dict, Tuple, Union, cast
 
 import cryptography.hazmat.primitives.asymmetric.ec as crypto_ec
 from cryptography.hazmat.backends import default_backend
@@ -38,6 +38,7 @@ TPM_ALG_SHA512 = 0x000D
 TPM_ALG_AES = 0x0006
 TPM_ALG_CFB = 0x0043
 
+TPM_GENERATED_VALUE = 0xFF544347
 
 # These are the object attribute values important for EK certs
 OA_FIXEDTPM = 0x00000002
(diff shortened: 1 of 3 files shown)


keylime/tpm/tpm2_objects.py
```diff
@@ -440,3 +441,26 @@ def ek_low_tpm2b_public_from_pubkey(pubkey: pubkey_type) -> bytes:
...
+def parse_tpms_clock_info(clock_info: bytes) -> Dict[str, int]:
+    (clock, resetCount, restartCount, safe) = struct.unpack(">QIIB", clock_info)
+    return {"clock": clock, "resetCount": resetCount, "restartCount": restartCount, "safe": safe}
+
+
+def get_tpms_attest_clock_info(tpms_attest: bytes) -> Dict[str, Any]:
+    (
+        magic,
+        _,
+    ) = struct.unpack_from(">IH", tpms_attest, 0)
```
aplanas, line 455: nit: but I wonder why black is doing this, when `magic, _ =  struct.unpack_from(">IH", tpms_attest, 0)` seems shorter that other lines

aplanas (approved):
Very cool one.

keylime/tpm/tpm2_objects.py
stefanberger: don't try to overrule `black` ...

aplanas: yeah, it does not matter

stefanberger: I don't know why black did this...

aplanas: there is a chance that it is a bug somewhere, here it is working tho:

```
# pip list | grep black
black           23.3.0

# cat kk.py
import struct


def get_tpms_attest_clock_info(tpms_attest):
    magic, _ = struct.unpack_from(">IH", tpms_attest, 0)

# black --diff kk.py 
All done! ✨ 🍰 ✨1 file would be left unchanged.
```

THS-on (approved):

PR #1350, opened 2023-04-24
Enable Packit CI again on all Fedora releases, simplify packit plans

kkaarreell:
- Enable CI on all Fedora releases as bug 2171376 has been already fixed
- Remove duplicate content in packit CI test plans utilizing fmf inheritance

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -3,7 +3,6 @@ jobs:
   trigger: pull_request
   metadata:
     targets:
-    - fedora-37
-#    - fedora-all
+    - fedora-all
     - centos-stream-9-x86_64
     skip_build: true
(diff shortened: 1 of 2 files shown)


ansasaki:
I think we need to set the `TPM_BINARY_MEASUREMENTS` env var when installing from Copr to use the changes I introduced in https://github.com/keylime/rust-keylime/pull/554. 

Please check if https://github.com/RedHat-SP-Security/keylime-tests/pull/376 makes sense.

kkaarreell:
It does. Thank you! Also, it seems we may need to wait a day or two due to F38.

aplanas (approved):

Issue #1354, opened 2023-04-26
keylime has hardcoded check for cacert.crt


kkaarreell:
Is your issue a feature request? If so, please raise it as an [enhancement](https://github.com/keylime/enhancements)

# Environment
 * OS / version: Fedora 38
 * Processor architecture: x86_64
 * TPM Manufacturer: swtpm
 * Keylime version: current upstream 
fcc9ba9

# Description 
for `tls_dir = default` keylime has a hardcoded check for the presence of a file cacert.crt and fails if the file is not present.
https://github.com/keylime/keylime/blob/master/keylime/web_util.py#L125
This is confusing since one would expect that `default` here represents the path `/var/lib/keylime/ca_cv` and therefore using `tls_dir = default` along with custom CA certificate with a different file name seems legit.

Moreover, the error message presented by keylime is not very helpful to identify which certificate file posses an issue:
```
Exception: It appears that the verifier has not yet created a CA and certificates, please run the verifier first
```

It seems appropriate to simplify the meaning of setting default there and set `tls_dir = /var/lib/keylime/cv_ca`.

## Steps to reproduce problem
1. `tls_dir = default`
2. set `trusted_client_ca` and `trusted_server_ca` to point to a different file in `/var/lib/keylime/cv_ca`, like `cacert.pem`.

# Relevant logs

Attach any relevant log files that can help to debug your issue.

PR #1358, opened 2023-05-02
web_util: handle tls_dir default with cacerts correctly

THS-on:
Fixes: #1354


aplanas (approved):

sergio-correia (approved):

THS-on:
I missed the case when the `tls_dir` is set to `generate`. This should now also work in that instance.

maugustosilva (approved):

THS-on:
I've undid the change to `get_tls_options` because the handling there was already correct. Should be now ready to merge.

PR #1364, opened 2023-05-04
WIP: tpm: Replace tpm2_quote with python code

stefanberger:


stefanberger:
Something is fishy with the Nonce. I have the impression that the Nonce is **not** used on the Quote side. @ansasaki ?

stefanberger:
> Something is fishy with the Nonce. I have the impression that the Nonce is **not** used on the Quote side. @ansasaki ? But there's something mysterious about the TPMS_ATTEST structure that does hold the extraData (=Nonce) that can then also be passed to the signing API on the TPM 2 side. So it may be actually ok...

Actually, it is ok.

This is the code I was referring to. The Nonce is in certifyInfo and therefore in attest->t.attestationData and need not be available in qualifyingData again:
https://github.com/stefanberger/libtpms/blob/stable-0.9/src/tpm2/Attest_spt.c#L168-L179

THS-on (commented):
Really nice work! 

Some of the return types are quite complex. For example:
```
def __get_pcrs_from_blob(pcrblob: bytes) -> Tuple[int, Dict[int, int], List[bytes]]:
```
I think it makes sense to move them into either specific data classes, named tuple or typed dicts.

Can you add some unit tests for the parsing functions?

keylime/tpm/tpm_util.py
```diff
@@ -0,0 +1,156 @@
...
+    """Do signature verification with the given public key"""
+    if isinstance(pubkey, RSAPublicKey):
+        pubkey.verify(sig, digest, padding.PKCS1v15(), Prehashed(hashfunc))
+    elif isinstance(pubkey, EllipticCurvePublicKey):
+        pubkey.verify(sig, digest, ec.ECDSA(Prehashed(hashfunc)))
+
+
+def __get_pcrs_from_blob(pcrblob: bytes) -> Tuple[int, Dict[int, int], List[bytes]]:
+    """This function is specific to the Intel tools with data in little endian format.
+    Data structures were not marshalled but written as whole out of memory."""
```
aplanas, line 30: This can be a problem. We currently have the Rust agent in Intel, S390X and AArch64, that shares different endianness (little, big, bi)

I found issues in the tpm2-tools in the past (and present), so I am not sure if this was broken when using tpm2_quote, but this can comeback as a problem later.

aplanas (commented):
Moving out from tpm2_tools seems to be the right decision. This code is impressive but I am a bit worried about two points:

* We lost debug-ability. Capturing the blogs and using the CLI as in the previous code give me in the past valuable information on what was wrong. Translating the new code into de calls of the equivalent CLI commands is now more complex (I experienced that in the Rust side, but the chaining calls make me clues about how to build the CLI parameters)
* Re-implementing the parsing seems an step back. Has been https://github.com/tpm2-software/tpm2-pytss disused and discarded?

THS-on:
> We lost debug-ability. Capturing the blogs and using the CLI as in the previous code give me in the past valuable information on what was wrong. Translating the new code into de calls of the equivalent CLI commands is now more complex (I experienced that in the Rust side, but the chaining calls make me clues about how to build the CLI parameters)

Yeah, but I think we can mitigate this by documenting what the equivalent calls are to tpm2-tools and how to convert the raw data.

> Re-implementing the parsing seems an step back. Has been https://github.com/tpm2-software/tpm2-pytss disused and discarded?

It was discussed, but I don't think there was a final result. The main advantage of doing this way is, that it is in pure Python and are independent form the tpm2-tss bindings. Packaging is also a consideration, but have no insight on how hard it is to bring tpm2-pytss into distros.

keylime/tpm/tpm_util.py
THS-on: Yes, once we do major changes we should move to a more universal format instead of using the binary blob from tpm2-tools.

aplanas:
>>  Re-implementing the parsing seems an step back. Has been https://github.com/tpm2-software/tpm2-pytss disused and discarded?

> It was discussed, but I don't think there was a final result. The main advantage of doing this way is, that it is in pure Python and are independent form the tpm2-tss bindings. Packaging is also a consideration, but have no insight on how hard it is to bring tpm2-pytss into distros.

IIRC it was easy: https://build.opensuse.org/package/show/openSUSE:Factory/python-tpm2-pytss

It is not much a matter of how hard is to package or introduce into a distribution, but how may other packages demand it.

keylime/tpm/tpm_util.py
stefanberger: First step is to replace the tpm2-tools code with python code while maintaining compatibility. 2nd step is to **maybe** extend the format but probably still maintain compatibility between agent and verifier.

stefanberger:
> Moving out from tpm2_tools seems to be the right decision. This code is impressive but I am a bit worried about two points:
> 
>     * We lost debug-ability. Capturing the blogs and using the CLI as in the previous code give me in the past valuable information on what was wrong. Translating the new code into de calls of the equivalent CLI commands is now more complex (I experienced that in the Rust side, but the chaining calls make me clues about how to build the CLI parameters)

When I was fixing things with the tpm2-tools in regards to being able to use them also on big endian machines, I was doing what you described also plus added some printf's to the tpm2-tools code but also was running the tools with gdb . That was during development time. Luckily I didn't have to go back to that. Now I would anticipate that this code is like no other and would again, as I just did during development, go back and add logger.error()'s into it and run `keylime_tenant -c add...`.

> 
>     * Re-implementing the parsing seems an step back. Has been https://github.com/tpm2-software/tpm2-pytss disused and discarded?

The only function we would need for this code is to be able to unmarshal TPMS_ATTEST...

keylime/tpm/tpm2_objects.py
```diff
@@ -61,6 +70,15 @@ def _pack_in_tpm2b(val: bytes) -> bytes:
 )
 
 
+# The hash functions used by TPM
+HASH_FUNCS = {
+    TPM_ALG_SHA1: hashes.__dict__.get("SHA1"),
```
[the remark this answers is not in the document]
stefanberger: Fixed.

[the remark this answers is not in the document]
stefanberger: mypy is now inhappy:

```
keylime/tpm/tpm_util.py:87: error: Cannot instantiate abstract class "HashAlgorithm" with abstract attributes "block_size", "digest_size" and "name"  [abstract]
keylime/tpm/tpm_util.py:125: error: Cannot instantiate abstract class "HashAlgorithm" with abstract attributes "block_size", "digest_size" and "name"  [abstract]
keylime/tpm/tpm_util.py:126: error: Cannot instantiate abstract class "HashAlgorithm" with abstract attributes "block_size", "digest_size" and "name"  [abstract]
keylime/tpm/tpm_util.py:128: error: Cannot instantiate abstract class "HashAlgorithm" with abstract attributes "block_size", "digest_size" and "name"  [abstract]
keylime/tpm/tpm_util.py:133: error: Cannot instantiate abstract class "HashAlgorithm" with abstract attributes "block_size", "digest_size" and "name"  [abstract]
keylime/tpm/tpm_util.py:149: error: Cannot instantiate abstract class "HashAlgorithm" with abstract attributes "block_size", "digest_size" and "name"  [abstract]
```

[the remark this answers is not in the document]
aplanas: That is weird ... as should be exactly the same thing ...

```Python
>>> hashes.SHA1 is hashes.__dict__.get("SHA1")
True
```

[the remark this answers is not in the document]
aplanas: `SHA1` is indeed abstract, as inherit from `HashAlgorithm` that it is abstract, and the class is not implementing the methods: https://github.com/pyca/cryptography/blob/main/src/cryptography/hazmat/primitives/hashes.py#L98

But the mypy complain seems ... not totally correct as you can instantiate it (done in tpm_utils, at the indicated lines). IIUC the magic happens in `digest = hashes.Hash(hashfunc(), backend=backends.default_backend())` that build the complete class in a dynamic way.

So this is another limitation of conflating a static type system in a dynamic language.

I guess that there is no complain when accessing dict as the type can be considered `Any` for the inference.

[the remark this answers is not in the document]
stefanberger: > That is weird ... as should be exactly the same thing ...

Maybe mypy isn't runtime checking things but doing some sort of parsing only ?

stefanberger:
> It was discussed, but I don't think there was a final result. The main advantage of doing this way is, that it is in pure Python and are independent form the tpm2-tss bindings. Packaging is also a consideration, but have no insight on how hard it is to bring tpm2-pytss into distros.

It's not widely packaged so far: https://pkgs.org/download/python3-tpm2-pytss

keylime/tpm/tpm_util.py
maugustosilva: I second @stefanberger here. This is an incremental, yet important, step in the right direction

keylime/tpm/tpm2_objects.py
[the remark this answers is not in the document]
aplanas: mypy is static, not runtime analysis, so it is limited. As commented, I think that mypy gets confused when accessing the object via `__dict__`, thinking that the object that return `.get("XXX")` is `Any`, and that can be instantiated.

But when referring the real object (via `.XXX`) it sees the type and get confused later when the code instantiate it, as believe that instantiating an ABC is a mistake (but in this case it is OK)

The fix seems easy: use the real object here and ignore the complain in the lines when it is instantiated.

keylime/tpm/tpm_util.py
aplanas: > I second @stefanberger here. This is an incremental, yet important, step in the right direction

I half disagree.

This change is introducing a backward compatibility issue in the code. Big endian architectures gets now unsupported because of the new parser.

I am not super familiar with the Intel tss code, but when I worked on it I saw bugs in the [de]marshaling code (honest bugs, miss interpretations on the standard, retro compatibility, etc). That for me is an indication that this is not a trivial task.

@stefanberger code is remarkable, and moving out of tpm2_tools CLI can be a good step, but re-implementing the parsers is not. If possible, I would ask pytss to do the parsing, as it will delegate in into the tss stack.

stefanberger: > This change is introducing a backward compatibility issue in the code. Big endian architectures gets now unsupported because of the new parser.

That is not true. 
The pcrblob was always written in little native format and per the changes I made back then they are now written in little endian format on all platforms. The function `__get_pcrs_from_blob()` explicitly parses the data in little endian format using the '<' specifier.
The aikblob is the PEM encoded public key. No issues with endianess here.
The sigblob is written in big endian and parsed as such.
The quoteblob is written in big endian and parsed as such.

> @stefanberger code is remarkable, and moving out of tpm2_tools CLI can be a good step, but re-implementing the parsers is not. If possible, I would ask pytss to do the parsing, as it will delegate in into the tss stack.

If we rely on packages available across all the distros we want to run this on, we will have to wait. If we can rely on pip I **may** have a look.

stefanberger: I was referring to these PR here for tpm2-tools that did some of the changes converting from native to little endian format for at least the quote aspects (read and write) of the tools. Using little endian, where native indian was used before, was the least disruptive for interoperability with existing x86/ppc64/arm systems and resolved it for s390. The rust agent TPM 2 library for sure inherited that format (as we know due to interoperability with the tpm2 tools we are about to get rid of) and probably also took over explicit writing in little endian :
- https://github.com/tpm2-software/tpm2-tools/pull/2228
- https://github.com/tpm2-software/tpm2-tools/pull/2224
- also: https://github.com/tpm2-software/tpm2-tools/pull/2230

stefanberger:
> The only function we would need for this code is to be able to unmarshal TPMS_ATTEST...

pytss will pull in the tpm2 tss 'C' stack for parsing this data structure. It has an implementation of makeidentity but also that can be addressed.

keylime/tpm/tpm_util.py
aplanas: > That is not true.
> The pcrblob was always written in little (native on x86) format and per the changes I made back then they are now written in little endian format on all platforms.

My apologies then.

AFAICT the tools are still not fully endian agnostics:

https://github.com/tpm2-software/tpm2-tools/issues/3060
https://github.com/tpm2-software/tpm2-tools/pull/3184

so my worries was in that direction, as some fails were related with the checkquote test code.

stefanberger: Makes me think we shouldn't rely on it...

stefanberger: IOW, with this PR we control more of the few TPM related pieces we need.

maugustosilva: so, we all agree that merging this code will not bring a backward compatibility issue?

aplanas: > so, we all agree that merging this code will not bring a backward compatibility issue?

I think we can agree on this!

I still think that parsing the binary blob in here is doing that in the wrong level of abstraction.  But if I am the only one thinking that, I will not oppose to the merge, of course.

stefanberger: > I still think that parsing the binary blob in here is doing that in the wrong level of abstraction.

I did not see code in the pytss project that would parse that pcrblob (which we now also write in Rust).

aplanas: > I did not see code in the pytss project that would parse that pcrblob (which we now also write in Rust).

... that is also true, the format from tpm2_tools seems to be very adhoc ...

I consider myself defeated! : )

aplanas (approved):

maugustosilva:
@THS-on I am ready to merge this, do you have any additional changes required?

stefanberger:
Rebased and added an assert.

PR #1365, opened 2023-05-05
installer.sh: support Alibaba Cloud Linux OS whose ID is alinux

hustliyilin:
This patch fix the following problem:

image

and the `/etc/os-release` of Alibaba Cloud Linux OS is like

image

### DIFF
--- a/installer.sh
+++ b/installer.sh
@@ -148,6 +148,26 @@ case "$ID" in
         fi
     ;;
 
+    # Alibaba Cloud Linux
+    alinux)
+        case "${VERSION_ID}" in
+            3*)
+                echo "${ID} selected."
+                PACKAGE_MGR=$(command -v dnf)
+                PYTHON_PREIN="python3 python3-devel python3-setuptools git wget patch rust cargo"
+                PYTHON_DEPS="python3-pip gcc gcc-c++ openssl-devel swig python3-pyyaml python3-zmq python3-cryptography python3-tornado python3-requests python3-gpg yaml-cpp-devel procps-ng python3-psutil python3-lark-parser python3-alembic python3-setuptools"
+                if [ "$(uname -m)" = "x86_64" ]; then
+                     PYTHON_DEPS+=" efivar-libs"
+                fi
+                pip3 install setuptools_rust
+                TPM2_TOOLS_PKGS="tpm2-tools tpm2-tss"
+            ;;
+            *)
+                echo "Version ${VERSION_ID} of ${ID} not supported"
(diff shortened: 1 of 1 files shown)


installer.sh
```diff
@@ -148,6 +148,28 @@ case "$ID" in
...
+                PACKAGE_MGR=$(command -v dnf)
+                PYTHON_PREIN="python3 python3-devel python3-setuptools git wget patch rust cargo"
+                PYTHON_DEPS="python3-pip gcc gcc-c++ openssl-devel swig python3-pyyaml python3-zmq python3-cryptography python3-tornado python3-requests python3-gpg yaml-cpp-devel procps-ng python3-psutil python3-lark-parser python3-alembic python3-setuptools"
+                if [ "$(uname -m)" = "x86_64" ]; then
+                     PYTHON_DEPS+=" efivar-libs"
+                fi
+                BUILD_TOOLS="openssl-devel libtool make automake pkg-config m4 libgcrypt-devel autoconf autoconf-archive libcurl-devel libstdc++-devel uriparser-devel dbus-devel gnulib-devel doxygen libuuid-devel json-c-devel"
+                pip3 install setuptools_rust
+                TPM2_TOOLS_PKGS="tpm2-tools tpm2-tss"
+                NEED_BUILD_TOOLS=0
```
aplanas, line 165: If the build tools are not needed, maybe you can skip setting the "BUILD_TOOLS" list. OTOH IIUC alibaba should be similar to RH/Centos, and this have "NEED_BUILD_TOOLS" as 1 for 8 and 0 for 9

hustliyilin: > If the build tools are not needed, maybe you can skip setting the "BUILD_TOOLS" list. OTOH IIUC alibaba should be similar to RH/Centos, and this have "NEED_BUILD_TOOLS" as 1 for 8 and 0 for 9

Thanks, I fix it.

aplanas: Ah maybe I was not clear : D

If `NEED_BUILD_TOOLS` is "0", it will not install "BUILD_TOOLS". So if you are sure that you do not need them to build a new version of the TPM tools, you can remove the "BUILD_TOOLS" line too.

hustliyilin: Thanks @aplanas  .  We can use the `dnf install` command to install the `TPM tools` in the Alibaba Cloud Linux OS, so I removed the "BUILD_TOOLS" line too.

aplanas (approved):
LGTM

stefanberger (approved):

PR #1367, opened 2023-05-09
README.md: remove the invalid URL about IMA stub service.

hustliyilin:
IMA stubbing support is removed already and the URL in README.md is invalid. so we need to remove the invalid URL correspondingly.

Fixes: #1368 


THS-on (approved):
Note that the stub service now has moved to the rust package: https://github.com/keylime/rust-keylime/

Maybe we should include a reference to that.

hustliyilin:
Do you mean we should change the invalid URL here to `https://github.com/keylime/rust-keylime/tree/master/keylime-ima-emulator` ?

THS-on:
Yes I think that is better, but good catch I missed that when I updated the README

hustliyilin:
OK. I fixed it and fork-pushed it.

aplanas (approved):

PR #1369, opened 2023-05-13
First of several PRs to clean up measured boot attestation

galmasi:
This PR is the first of several (hopefully nondisruptive) changes to clean up the code base we use for measured boot attestation.

Higher level goals
---

The *state of the art* of measured boot attestation in Keylime is that ... well, it works ... but the code base is disorganized, without a proper API and poorly documented.

The *ultimate goal* we have in sight is to 
(a) move all the files related to measured boot attestation into a single directory
(b) create a common API for calling MBA functionality from other parts of keylime
(c) make said API *pluggable* so that MBA implementation can be replaced at will

This PR
---

Needless to say, this PR does not do that. What does is 
* Propose a first tentative cut of the API (`keylime/mba/mba.py`), identifying MBA functions into three broad categories -- initialization, event log parsing and event log evaluation.
* Move all the files related to event log evaluation (aka `elchecking`) into a subdirectory; restrict interaction between `elchecking` and the rest of Keylime to go through the API (`keylime/mba/mba.py`)
* Prepare the ground  for the next PR, which will move event log parsing altogether into `keylime/mba/elparsing`

@THS-on since you are an active user of MBA, please think and comment about the proposed new MBA API. It is purposefully vague at this point, but the initial thought is to minimally formalize the types of binary event logs, parsed event logs, policy files, and write parsing and evaluation in terms of these types. 

Next steps
---

The next PR will move event log parsing completely into `mba` under `elparsing` and subsequently make it pluggable. This would remove a lot of less-than-perfect code from `tpm_main.py`, and continue to collect and sequester code that is strewn around the keylime directories. 

We are also considering creating an event log authentication function, which would remove another hefty chunk of not very transparent code from `tpm_main.py`. We have not [yet] added this proposed function to `mba`.


keylime/tpm/tpm_main.py
```diff
@@ -670,7 +669,7 @@ def _parse_mb_bootlog(self, log_b64: str) -> typing.Tuple[Failure, typing.Option
 
     def parse_mb_bootlog(
         self, mb_measurement_list: Optional[str], hash_alg: algorithms.Hash
-    ) -> typing.Tuple[Dict[str, int], typing.Optional[Dict[str, List[str]]], RefState, Failure]:
+    ) -> typing.Tuple[Dict[str, int], typing.Optional[Dict[str, List[str]]], Dict[str, Any], Failure]:
```
aplanas, line 672: Why RefState is replaced here? Seems to be defined as `typing.Mapping[str, tests.Data]`

keylime/mba/mba.py
```diff
@@ -0,0 +1,100 @@
...
+
+# ###########
+# when something goes wrong with MBA policies
+# ###########
+
+
+class PolicyException(Exception):
+    """
+    This exception is raised if policies cannot be loaded.
+    TODO final format including a reason code.
```
aplanas, line 16: This will come later in this PR, right? Would make sense to move into Draft until all the code is there?

aplanas (commented):
(will continue later)

keylime/tpm/tpm_main.py
galmasi: I made a conscious decision to remove all types defined in `elchecking` from the rest of the code.

* Ultimately `refstate` should be redefined as "mba policy", because refstate is not even a term anyone in TCG would recognize; we made it up. 
* The type of what is now "refstate" should be eventually defined in `mba.py` and become part of the API.

I just didn't want to move *everything* at once, it would have made the PR even more unwieldy and complicated. My focus is on preserving functionality while we are slowly cleaning up.

keylime/mba/mba.py
maugustosilva: @aplanas My main argument in favor of merging this particular PR, the first of "many" (at least 3) is that we have a good precedent in the project of effecting these significant changes in an incremental manner. It was so when we first introduced "measured boot" (3 PRs) and "durable attestation" (4 PRs, with probably more needed).

keylime/mba/elchecking/mbpolicy.py
```diff
@@ -0,0 +1,78 @@
...
+
+    # no evaluation if the refstate does not load as JSON
+    if not mb_refstate_data:
+        return failure
+
+    # load policy name
+    mb_policy_name = config.get("verifier", "measured_boot_policy_name", fallback="accept-all")
+
+    # pylint: disable=import-outside-toplevel
+    from keylime.mba.elchecking import policies as eventlog_policies
```
stefanberger, line 33: Does this need to be here?

keylime/mba/mba.py
THS-on: Maybe we can change it to a ValueError for now and introduce the specific exception when needed.

keylime/mba/mba.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def parse_bootlog() -> None:
+    """
+    This is a place holder for an external entity parsing the binary boot log and returning a JSON
+    of the parsed bootlog including the predicted PCR values and boot aggregates.
+    Inputs: the binary file
+    Output: a single file, JSON format to be formalized. It is a bit of ad-hoc at present.
+    """
+    return None
```
THS-on, line 100: Move this into the commit where it is then going to be used, because currently we still rely on the tpm abstraction class.

keylime/cloud_verifier_tornado.py
```diff
@@ -1436,10 +1435,11 @@ def main() -> None:
...
     # Check if measured boot was configured correctly
-    mb_policy_name = config.get("verifier", "measured_boot_policy_name", fallback="accept-all")
-    if mb_policy_name and policies.get_policy(mb_policy_name) is None:
-        logger.error('Measured boot policy "%s" could not be found!', mb_policy_name)
-        raise Exception(f'Measued boot policy "{mb_policy_name}" could not be found!')
+    # TODO remove this completely. It's redundant.
+    # mb_policy_name = config.get("verifier", "measured_boot_policy_name", fallback="accept-all")
+    # if mb_policy_name and policies.get_policy(mb_policy_name) is None:
+    #    logger.error('Measured boot policy "%s" could not be found!', mb_policy_name)
+    #    raise Exception(f'Measued boot policy "{mb_policy_name}" could not be found!')
```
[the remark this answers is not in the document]
galmasi: will do. Was being tentative about it :) but you are right, it should go.

maugustosilva:
@kkaarreell @Koncpa @ansasaki scratching my head looking at the CI errors here. Please help me out. 

It looks like the (rust) `agent` is sending "zero" as the values for PCRs 0-9 and 14. 

That was not the case with the python `agent` and @galmasi's new code "does the right thing", attempts to check it based on the values from the file `binary_bios_measurements` and fails.

kkaarreell:
Hi @maugustosilva , in the failing scenarios we are using empty mb policy so there aren't reference PCR values. Previously, the verifier logged
```
keylime.tpm - WARNING - PCR #0 in quote (from agent d432fbb3-d2f1-4a97-9ef7-75bd81c00000) not found in tpm_policy, skipping.
```
while now it fails with
```
keylime.tpm - ERROR - For PCR 0 and hash sha256 the boot event log has value 'fcb620568efe4ac4e15f6dcbc6431cad79bc85c7f2f592e08dde0bf37da6df39' but the agent d432fbb3-d2f1-4a97-9ef7-75bd81c00000 returned '0000000000000000000000000000000000000000000000000000000000000000'
```
I belief it should not fail as before.

maugustosilva:
So, the way it was originally designed (@mpeters @THS-on please keep me honest here) is that the addition of **any** MB policy (even an empty one) would automatically modify the "TPM PCR mask" to add PCRs 0-9,14 to the quote sent by the `agent`, which would result in the `verifier` comparing the **expected** values from these PCRs with the values from `binary_bios_measurements`.

Looking at the `tenant` code, I see the "TPM PCR Mask" is still "automatically updated". The fact this is **overridden** by an empty `TPM policy` during the verification later is, IMO, a change in behavior. I do apologize for not having caught it when it change, and I am even ok with modifying it now.

However, I must ask: is there a way for the agent to either extend the correct PCRs on the tested VM?

ansasaki:
> So, the way it was originally designed (@mpeters @THS-on please keep me honest here) is that the addition of **any** MB policy (even an empty one) would automatically modify the "TPM PCR mask" to add PCRs 0-9,14 to the quote sent by the `agent`, which would result in the `verifier` comparing the **expected** values from these PCRs with the values from `binary_bios_measurements`.
> 
> Looking at the `tenant` code, I see the "TPM PCR Mask" is still "automatically updated". The fact this is **overridden** by an empty `TPM policy` during the verification later is, IMO, a change in behavior. I do apologize for not having caught it when it change, and I am even ok with modifying it now.
> 
> However, I must ask: is there a way for the agent to either extend the correct PCRs on the tested VM?

I'm not sure what you are proposing here. There is the possibility to use fake measured boot logs for the keylime agent by compiling it with `testing` feature enabled (which is done in the testing rpms we create on Copr) and setting the `TPM_BINARY_MEASUREMENTS` env var. But the TPM quote will be based on the SW TPM state. Is the request to create a measured boot emulator, similar to the IMA emulator?

maugustosilva:
Thanks for the short and productive conf call @ansasaki! 

FYI @kkaarreell @Koncpa. I have asked @galmasi to restore the old (and IMO, wrong) behavior - do not validate the PCR 0-9,14 values that `verifier` will get from `TPM_BINARY_MEASUREMENTS` if `TPM Policy` is empty - and later on we will open a PR on the RedHat CI testing to have a (very simple) "PCR 0-9,14 extender" which will deal with these PCRs even before the `agent` starts (@galmasi has some code which can do this).

kkaarreell:
> FYI @kkaarreell @Koncpa. I have asked @galmasi to restore the old (and IMO, wrong) behavior - do not validate the PCR 0-9,14 values that `verifier` will get from `TPM_BINARY_MEASUREMENTS` if `TPM Policy` is empty - and later on we will open a PR on the RedHat CI testing to have a (very simple) "PCR 0-9,14 extender" which will deal with these PCRs even before the `agent` starts (@galmasi has some code which can do this).

Sounds good, thank you. 

Like Anderson said, we have an option to "redirect" agent to a custom tpm_binary_measurement which allows us to test MB policy also on a system with runtime TPM emulation. We can definitely adopt MB emulator once it becomes available in keylime if that would improve the range of testable scenarios.

galmasi:
Thank you @kkaarreell . I have offered to provide the MBA emulator based on the recent work I did with the python-only event log parser. At the moment the event log parser can only produce simulated PCR values, but with fairly trivial changes I could make it do what @ansasaki 's IMA emulator does -- to make it talk to a TPM through the TPM interfaceand produce a set of PCRs the emulator would need. Then we could amend the test infrastructure to provide a meaningful set of MBA tests, and I think everyone will be happy[er].

And to confirm what @maugustosilva said just above, I _will_ revert this PR to the (technically incorrect) behavior of not authenticating event logs when the policy is an empty string. That should avert the current crisis.

stefanberger:
FYI: When you rebase on upstream with a command like 'git rebase upstream master' my IMA patches will disappear from your PR.

keylime/mba/mba.py
galmasi: done with next commit.

keylime/mba/elchecking/mbpolicy.py
```diff
@@ -0,0 +1,78 @@
...
+from keylime.failure import Component, Failure
+from keylime.mba.elchecking import tests
+
+logger = keylime_logging.init_logging("measured_boot")
+
+
+def evaluate_policy(
+    mb_refstate_str: Optional[str],
+    mb_measurement_data: tests.Data,
+    mb_pcrs_inquote: Set[int],
```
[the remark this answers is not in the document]
galmasi: will do.

keylime/mba/mba.py
```diff
@@ -0,0 +1,100 @@
...
+    """
+    MBA API front-end for importing the policy engine.
+    Inputs: <none>, but may read from keylime configuration files.
+    Outputs: <none>, but the keylime policy engine is now loaded.
+    Errors: Possible exceptions raised if the policy engine cannot be loaded.
+    """
+    try:
+        policies.load_policies()
+    except Exception as e:
+        raise PolicyException from e
```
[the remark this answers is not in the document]
galmasi: Changed to ValueError. There is no exception handling in the tpm check_quote code that I know of. Will re-evaluate when we introduce PolicyException again: once we have a published exception, there is _something_ intelligent that the calling code will do.

keylime/mba/elchecking/mbpolicy.py
```diff
@@ -0,0 +1,78 @@
...
+        mb_policy_name = "reject-all"
+        mb_policy = eventlog_policies.RejectAll()
+
+    # figure out whether the quote contains all quotes to evaluate the policy
+    # if there are any PCRs in the policy that are not in the quote, we canot evaluate.
+    mb_pcrs_policy = mb_policy.get_relevant_pcrs()
+    missing_pcrs = list(mb_pcrs_policy.difference(mb_pcrs_inquote))
+    if len(missing_pcrs) > 0:
+        logger.error("PCRs specified for policy %s not in quote: %s", mb_policy_name, str(missing_pcrs))
+        failure.add_event("missing_pcrs", {"context": "PCRs are missing in quote", "data": missing_pcrs}, True)
```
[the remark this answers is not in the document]
galmasi: Not originally my code. Let's optimize _later_ please. This PR is large enough already even with minimal changes made in Mike's code from 2yrs ago. 

But _will_ open a separate PR with your observations, which are relevant.

keylime/mba/elchecking/mbpolicy.py
```diff
@@ -0,0 +1,78 @@
...
+from keylime import config, keylime_logging
+from keylime.failure import Component, Failure
+from keylime.mba.elchecking import tests
+
+logger = keylime_logging.init_logging("measured_boot")
+
+
+def evaluate_policy(
+    mb_refstate_str: Optional[str],
+    mb_measurement_data: tests.Data,
```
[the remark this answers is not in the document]
galmasi: Agreed. Will fix.

keylime/mba/mba.py
galmasi: done (ValueError)

galmasi:
A bit of fumbling with the rebase, but complete now. I think mergeable.

galmasi:
MBA integration test passed. If code coverage can indeed be ignored, this is ready to merge.

maugustosilva:
I see an approval from @stefanberger, a "LGTM" from @THS-on and resolved conversations from @aplanas ... it is time to merge it.

maugustosilva (approved):

kkaarreell:
> MBA integration test passed. If code coverage can indeed be ignored, this is ready to merge.

I am sorry for that, it turns out that finding a reliable and free file hosting that works with `curl` is not that simple. I am doing some changes in our scripts that would hopefully improve the situation.

PR #1372, opened 2023-05-16
keylime/cloud_verifier_tornado.py: set ssl_ctx None when the enable_agent_mtls config of verifier is set false

hustliyilin:
Fixes: #1373

### DIFF
--- a/keylime/cloud_verifier_tornado.py
+++ b/keylime/cloud_verifier_tornado.py
@@ -1532,7 +1532,11 @@ def main() -> None:
     keylime_api_version.log_api_versions(logger)
 
     # Get the server TLS context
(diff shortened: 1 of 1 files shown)


ansasaki:
This is an incorrect interpretation of the meaning of the `enable_agent_mtls` option. This option is meant to configure whether the verifier should access the agent via HTTP or HTTPS. It is not about the verifier Rest API.

If you want to test the verifier disabling the client certificate verification, you can set the `trusted_client_ca` with the keyword `all`, meaning that any client certificate will be accepted. In the same way, you can disable the server certificate verification by setting the `trusted_server_ca` with the keyword `all`.

hustliyilin:
Thank you very much for the explanation. I understand now, I close this PR now,

PR #1375, opened 2023-05-16
Replace tpm2_makecredential with a python implementation

stefanberger:
This PR implements the algorithms necessary for running TPM2_MakeCredential outside of a TPM 2 but follows the TPM 2 reference implementation. It then replaces the tpm2_makecredential command line tool invocation with a call to the python function that returns its result in the same format as tpm2_makecredential did.

### DIFF
--- a/keylime/tpm/ec_crypto_helper.py
+++ b/keylime/tpm/ec_crypto_helper.py
@@ -1,0 +1,101 @@
+# pylint: disable=protected-access
+
+from typing import Optional
+
+from cryptography.hazmat import backends
+from cryptography.hazmat.backends.openssl.backend import Backend as OpenSSLBackend
+from cryptography.hazmat.bindings.openssl import binding
+from cryptography.hazmat.primitives.asymmetric.ec import EllipticCurvePrivateKey, EllipticCurvePublicKey
+
+
+class EcCryptoHelper:
+    instance: Optional["EcCryptoHelper"] = None
+
+    @staticmethod
+    def get_instance() -> "EcCryptoHelper":
+        """Create and return a singleton EcCryptoHelper"""
+        if not EcCryptoHelper.instance:
+            EcCryptoHelper.instance = EcCryptoHelper()
+        return EcCryptoHelper.instance
+
+    def __init__(self) -> None:
+        self.backend = backends.default_backend()
+        if not isinstance(self.backend, OpenSSLBackend):
+            raise Exception("The python cryptography default backend must be using OpenSSL")
+
+        self.binding = binding.Binding()
+        lib = self.binding.lib
+        assert lib
+
+        for func in [
+            "BN_CTX_get",
+            "BN_free",
+            "EC_KEY_get0_group",
+            "EC_POINT_new",
+            "EC_POINT_free",
+            "EC_POINT_mul",
+        ]:
+            if not hasattr(lib, func):
+                raise Exception(f"{func} not available in lib {str(dir(lib))}")
+
+        if hasattr(lib, "EC_POINT_set_affine_coordinates_GFp"):
+            self.set_affine_coordinates = lib.EC_POINT_set_affine_coordinates_GFp  # pyright: ignore
+        elif hasattr(lib, "EC_POINT_set_affine_coordinates"):
+            self.set_affine_coordinates = lib.EC_POINT_set_affine_coordinates  # pyright: ignore
+        else:
+            raise Exception(f"EC_POINT_set_affine_coordinates/GFp not available in lib {str(dir(lib))}")
+
+        if hasattr(lib, "EC_POINT_get_affine_coordinates_GFp"):
+            self.get_affine_coordinates = lib.EC_POINT_get_affine_coordinates_GFp  # pyright: ignore
+        elif hasattr(lib, "EC_POINT_get_affine_coordinates"):
+            self.get_affine_coordinates = lib.EC_POINT_get_affine_coordinates  # pyright: ignore
+        else:
+            raise Exception(f"EC_POINT_get_affine_coordinates/GFp not available in _lib [{str(dir(lib))}]")
+
+    def point_multiply_x(self, public_key: EllipticCurvePublicKey, private_key: EllipticCurvePrivateKey) -> bytes:
+        """Perform a point multiplication of the given public key with the private_key (scalar).
+        Return the x component of the result."""
+        ec_cdata = private_key._ec_key  # type: ignore
+        group = self.binding.lib.EC_KEY_get0_group(ec_cdata)  # pyright: ignore
+
+        point = self.binding.lib.EC_POINT_new(group)  # pyright: ignore
+        if point == self.binding.ffi.NULL:
+            raise Exception("EC_POINT_new(group) returned NULL")
+        result_point = self.binding.ffi.gc(point, self.binding.lib.EC_POINT_free)  # pyright: ignore
+
+        # convert the public_key into a point
+        pn = public_key.public_numbers()
+        bn_x = self.binding.ffi.gc(self.backend._int_to_bn(pn.x), self.binding.lib.BN_free)  # pyright: ignore
+        bn_y = self.binding.ffi.gc(self.backend._int_to_bn(pn.y), self.binding.lib.BN_free)  # pyright: ignore
+
+        point = self.binding.lib.EC_POINT_new(group)  # pyright: ignore
+        if point == self.binding.ffi.NULL:
+            raise Exception("EC_POINT_new(group) returned NULL")
+        pubkey_point = self.binding.ffi.gc(point, self.binding.lib.EC_POINT_free)  # pyright: ignore
+
+        # convert private key to scalar
+        privkey_scalar = self.binding.ffi.gc(
+            self.backend._int_to_bn(private_key.private_numbers().private_value),
+            self.binding.lib.BN_free,  # pyright: ignore
+        )
+
+        with self.backend._tmp_bn_ctx() as bn_ctx:
+            res = self.set_affine_coordinates(group, pubkey_point, bn_x, bn_y, bn_ctx)
+            if res != 1:
+                raise Exception("set_affine_coordinates failed")
+
+            res = self.binding.lib.EC_POINT_mul(  # pyright: ignore
+                group, result_point, self.binding.ffi.NULL, pubkey_point, privkey_scalar, bn_ctx
+            )
+            if res != 1:
+                raise Exception("EC_POINT_mul failed")
+
+            # return the x component of result_point
+            bn_x2 = self.binding.lib.BN_CTX_get(bn_ctx)  # pyright: ignore
+            res = self.get_affine_coordinates(group, result_point, bn_x2, self.binding.ffi.NULL, bn_ctx)
+            if res != 1:
+                raise Exception("get_affine_coordinates failed")
+
+            x: int = self.backend._bn_to_int(bn_x2)
+
+        return x.to_bytes((x.bit_length() + 7) >> 3, "big")
--- a/keylime/tpm/ec_crypto_helper.py
+++ b/keylime/tpm/ec_crypto_helper.py
@@ -1,6 +1,6 @@
 # pylint: disable=protected-access
 
-from typing import Optional
+from typing import Any, Optional
 
 from cryptography.hazmat import backends
 from cryptography.hazmat.backends.openssl.backend import Backend as OpenSSLBackend
@@ -52,6 +52,21 @@ class EcCryptoHelper:
         else:
             raise Exception(f"EC_POINT_get_affine_coordinates/GFp not available in _lib [{str(dir(lib))}]")
 
+    def int2bn(self, num: int) -> Any:
+        binary = num.to_bytes((num.bit_length() + 7) >> 3, "big")
+        bn = self.binding.lib.BN_bin2bn(binary, len(binary), self.binding.ffi.NULL)  # pyright: ignore
+        if bn == self.binding.ffi.NULL:
+            raise Exception("BN_bin2bn returned NULL")
+        return bn
+
+    def bn2int(self, bn: Any) -> int:
+        num_bytes = self.binding.lib.BN_num_bytes(bn)  # pyright: ignore
+        binary = self.binding.ffi.new("unsigned char[]", num_bytes)
+        binary_len = self.binding.lib.BN_bn2bin(bn, binary)  # pyright: ignore
+        if binary_len < 0:
+            raise Exception("BN_bn2int failed")
+        return int.from_bytes(self.binding.ffi.buffer(binary)[:binary_len], "big")
+
     def point_multiply_x(self, public_key: EllipticCurvePublicKey, private_key: EllipticCurvePrivateKey) -> bytes:
         """Perform a point multiplication of the given public key with the private_key (scalar).
         Return the x component of the result."""
@@ -65,8 +80,8 @@ class EcCryptoHelper:
 
         # convert the public_key into a point
         pn = public_key.public_numbers()
-        bn_x = self.binding.ffi.gc(self.backend._int_to_bn(pn.x), self.binding.lib.BN_free)  # pyright: ignore
-        bn_y = self.binding.ffi.gc(self.backend._int_to_bn(pn.y), self.binding.lib.BN_free)  # pyright: ignore
+        bn_x = self.binding.ffi.gc(self.int2bn(pn.x), self.binding.lib.BN_free)  # pyright: ignore
+        bn_y = self.binding.ffi.gc(self.int2bn(pn.y), self.binding.lib.BN_free)  # pyright: ignore
 
         point = self.binding.lib.EC_POINT_new(group)  # pyright: ignore
         if point == self.binding.ffi.NULL:
@@ -75,7 +90,7 @@ class EcCryptoHelper:
 
         # convert private key to scalar
         privkey_scalar = self.binding.ffi.gc(
-            self.backend._int_to_bn(private_key.private_numbers().private_value),
+            self.int2bn(private_key.private_numbers().private_value),
             self.binding.lib.BN_free,  # pyright: ignore
         )
 
@@ -96,6 +111,6 @@ class EcCryptoHelper:
             if res != 1:
                 raise Exception("get_affine_coordinates failed")
 
-            x: int = self.backend._bn_to_int(bn_x2)
+            x: int = self.bn2int(bn_x2)
 
         return x.to_bytes((x.bit_length() + 7) >> 3, "big")
--- a/keylime/tpm/ec_crypto_helper.py
+++ b/keylime/tpm/ec_crypto_helper.py
@@ -2,8 +2,6 @@
 
 from typing import Any, Optional
 
-from cryptography.hazmat import backends
-from cryptography.hazmat.backends.openssl.backend import Backend as OpenSSLBackend
 from cryptography.hazmat.bindings.openssl import binding
 from cryptography.hazmat.primitives.asymmetric.ec import EllipticCurvePrivateKey, EllipticCurvePublicKey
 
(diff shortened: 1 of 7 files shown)


keylime/tpm/tpm2_objects.py
```diff
@@ -266,21 +271,23 @@ def _extract_tpm2b(vals: bytes) -> Tuple[bytes, bytes]:
...
+    # Extract type, nameAlg, and [objectAttributes] (we don't care about the
+    #  latter)
+    (alg_type, name_alg, _) = struct.unpack(">HHI", public[0:8])
     # Ignore the authPolicy
     (_, sym_parms) = _extract_tpm2b(public[8:])
     # Ignore the non-asym-alg parameters
-    (sym_mode,) = struct.unpack(">H", sym_parms[0:2])
-    # Ignore the sym_mode and keybits (4 bytes), possibly symmetric (2) and sign
+    (sym_alg, sym_key_bits) = struct.unpack(">HH", sym_parms[0:4])
+    # Ignore the sym_mode (2 bytes), possibly symmetric (2) and sign
```
stefanberger, line 287: I have to restore the comment here ...

keylime/tpm/tpm_util.py
```diff
@@ -176,3 +179,145 @@ def checkquote(
...
+
+
+def compute_outer_integrity(
+    name: bytes,
+    hashfunc: abc.ABCMeta,
+    seed: bytes,
+    sensitive_data_2b_size: int,
+    ciphertext: bytes,
+) -> bytes:
+    assert sensitive_data_2b_size == len(ciphertext)
```
stefanberger, line 268: remove parameter `sensitive_data_2b_size` from function

aplanas (commented):
I did not review the deep part of the code (byte mangling)

keylime/tpm/tpm2_objects.py
```diff
@@ -78,6 +79,10 @@ def _pack_in_tpm2b(val: bytes) -> bytes:
     TPM_ALG_SHA512: hashes.__dict__.get("SHA512"),
 }
 
+SYMCIPHER_FUNCS = {
+    TPM_ALG_AES: algorithms.__dict__.get("AES"),
```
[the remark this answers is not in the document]
stefanberger: same as a few lines above ...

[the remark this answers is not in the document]
stefanberger: @aplanas See discussion here: https://github.com/keylime/keylime/pull/1364#discussion_r1185842102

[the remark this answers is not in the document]
aplanas: yes, we talked about that in the other PR: this is somehow hiding the type and bypassing the type checker. The complain of the type checker was some limitation of the static analysis and maybe the only option is to add those comments that skip the error

keylime/tpm/tpm_util.py
```diff
@@ -176,3 +179,140 @@ def checkquote(
         raise Exception("The digest of the quoteblob differs from the quote's digest")
 
     return pcrs_dict
+
+
+def makecredential(ek_tpm: bytes, challenge: bytes, aik_name: bytes) -> bytes:
+    public_key, hash_alg, _, _ = tpm2_objects.pubkey_parms_from_tpm2b_public(ek_tpm)
+    # python crypto only supports encryption with RSA public key
+    if not isinstance(public_key, (RSAPublicKey)):
```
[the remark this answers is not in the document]
stefanberger: Removed...

keylime/tpm/tpm2_objects.py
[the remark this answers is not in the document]
aplanas: > @aplanas See discussion here: [#1364 (comment)](https://github.com/keylime/keylime/pull/1364#discussion_r1185842102)

exactly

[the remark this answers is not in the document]
stefanberger: What are 'those comments that skip the error'?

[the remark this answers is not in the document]
aplanas: (not tested)

```python
    TPM_ALG_AES: algorithms.AES,  # type: ignore[abstract]
```

keylime/tpm/tpm_util.py
```diff
@@ -176,3 +179,140 @@ def checkquote(
...
+    credentialblob = secret_to_credential(challenge, aik_name, random, ek_tpm, hashfunc)
+
+    # Create tpm2-tools specific format
+    hdr = struct.pack(">II", 0xBADCC0DE, 1)
+    tail = struct.pack(f">H{len(secret)}s", len(secret), secret)
+
+    return hdr + credentialblob + tail
+
+
+def secret_to_credential(secret: bytes, name: bytes, seed: bytes, ek_tpm: bytes, hashfunc: abc.ABCMeta) -> bytes:
```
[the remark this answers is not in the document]
stefanberger: If we can solve the issue further above we can easily solve it here...

keylime/tpm/tpm2_objects.py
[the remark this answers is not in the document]
stefanberger: For the hashes SHA1, SHA256, etc. it works with abc.ABCMeta only -- at least for me. I have tried it now with either hashes.HashContext or hashes.HashAlgorithm, and for both the type checker complains:

```
/home/stefanb/keylime/keylime/tpm/tpm_util.py
  /home/stefanb/keylime/keylime/tpm/tpm_util.py:268:19 - error: Object of type "HashContext" is not callable (reportGeneralTypeIssues)
  /home/stefanb/keylime/keylime/tpm/tpm_util.py:272:29 - error: Object of type "HashContext" is not callable (reportGeneralTypeIssues)
  /home/stefanb/keylime/keylime/tpm/tpm_util.py:307:28 - error: Object of type "HashContext" is not callable (reportGeneralTypeIssues)
  /home/stefanb/keylime/keylime/tpm/tpm_util.py:316:17 - error: Object of type "HashContext" is not callable (reportGeneralTypeIssues)
4 errors, 0 warnings, 0 informations
```


```
/home/stefanb/keylime/keylime/tpm/tpm_util.py
  /home/stefanb/keylime/keylime/tpm/tpm_util.py:268:19 - error: Object of type "HashAlgorithm" is not callable (reportGeneralTypeIssues)
  /home/stefanb/keylime/keylime/tpm/tpm_util.py:272:29 - error: Object of type "HashAlgorithm" is not callable (reportGeneralTypeIssues)
  /home/stefanb/keylime/keylime/tpm/tpm_util.py:307:28 - error: Object of type "HashAlgorithm" is not callable (reportGeneralTypeIssues)
  /home/stefanb/keylime/keylime/tpm/tpm_util.py:316:17 - error: Object of type "HashAlgorithm" is not callable (reportGeneralTypeIssues)
4 errors, 0 warnings, 0 informations
```

See: https://github.com/pyca/cryptography/blob/main/src/cryptography/hazmat/primitives/hashes.py#L98

keylime/tpm/tpm_util.py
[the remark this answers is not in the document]
stefanberger: FYI: Due to python crypto we cannot easily use EC EK keys since makecredential doesn't work for EC keys. For EC key support we would probably have to resort to openssl command line tool or another crypto library...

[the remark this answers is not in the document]
sergio-correia: > FYI: Due to python crypto we cannot easily use EC EK keys since makecredential doesn't work for EC keys. For EC key support we would probably have to resort to openssl command line tool or another crypto library...

What is the issue with python crypto and EC keys?

keylime/tpm/tpm2_objects.py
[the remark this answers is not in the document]
stefanberger: Actually the AES one is solvable, the hashes I do not know how to solve.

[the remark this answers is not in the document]
aplanas: This did the trick:

```patch
diff --git a/keylime/tpm/tpm2_objects.py b/keylime/tpm/tpm2_objects.py
index 937be5b..bcb9b21 100644
--- a/keylime/tpm/tpm2_objects.py
+++ b/keylime/tpm/tpm2_objects.py
@@ -71,11 +71,11 @@ AK_EXPECTED_ATTRS = (
 
 
 # The hash functions used by TPM
-HASH_FUNCS = {
-    TPM_ALG_SHA1: hashes.__dict__.get("SHA1"),
-    TPM_ALG_SHA256: hashes.__dict__.get("SHA256"),
-    TPM_ALG_SHA384: hashes.__dict__.get("SHA384"),
-    TPM_ALG_SHA512: hashes.__dict__.get("SHA512"),
+HASH_FUNCS: dict[int, hashes.HashAlgorithm] = {
+    TPM_ALG_SHA1: hashes.SHA1(),
+    TPM_ALG_SHA256: hashes.SHA256(),
+    TPM_ALG_SHA384: hashes.SHA384(),
+    TPM_ALG_SHA512: hashes.SHA512(),
 }
 
 
diff --git a/keylime/tpm/tpm_util.py b/keylime/tpm/tpm_util.py
index 7d63070..8122d37 100644
--- a/keylime/tpm/tpm_util.py
+++ b/keylime/tpm/tpm_util.py
@@ -87,7 +87,7 @@ def __hash_pcr_banks(
     if not hashfunc:
         raise ValueError(f"Unsupported hash with id {hash_alg:#x} in signature blob")
 
-    digest = hashes.Hash(hashfunc(), backend=backends.default_backend())
+    digest = hashes.Hash(hashfunc, backend=backends.default_backend())
 
     idx = 0
     pcrs_dict: Dict[int, str] = {}
@@ -145,15 +145,15 @@ def checkquote(
     hashfunc = tpm2_objects.HASH_FUNCS.get(hash_alg)
     if not hashfunc:
         raise ValueError(f"Unsupported hash with id {hash_alg:#x} in signature blob")
-    if hashfunc().name != exp_hash_alg:
-        raise ValueError(f"Quote was expected to use {exp_hash_alg} but used {hashfunc().name} instead")
+    if hashfunc.name != exp_hash_alg:
+        raise ValueError(f"Quote was expected to use {exp_hash_alg} but used {hashfunc.name} instead")
 
-    digest = hashes.Hash(hashfunc(), backend=backends.default_backend())
+    digest = hashes.Hash(hashfunc, backend=backends.default_backend())
     digest.update(quoteblob)
     quote_digest = digest.finalize()
 
     try:
-        verify(pubkey, signature, quote_digest, hashfunc())
+        verify(pubkey, signature, quote_digest, hashfunc)
     except InvalidSignature:
         logger.error("Invalid quote signature!")
 
@@ -168,7 +168,7 @@ def checkquote(
     if retDict["attested.quote.pcrDigest"] != compare_digest:
         raise Exception("The digest used for quoting is different than the one that was calculated")
 
-    digest = hashes.Hash(hashfunc(), backend=backends.default_backend())
+    digest = hashes.Hash(hashfunc, backend=backends.default_backend())
     digest.update(quoteblob)
     quoteblob_digest = digest.finalize()

```

keylime/tpm/tpm_util.py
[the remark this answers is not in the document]
stefanberger: > > FYI: Due to python crypto we cannot easily use EC EK keys since makecredential doesn't work for EC keys. For EC key support we would probably have to resort to openssl command line tool or another crypto library...
> 
> What is the issue with python crypto and EC keys?

```# python crypto only supports encryption with RSA public key```

File: keylime/tpm/tpm2_objects.py (hunk 3 above)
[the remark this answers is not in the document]
Comment (aplanas) [reply keylime/tpm/tpm2_objects.py:83]: > his solution provides a single instance of a SHA1, SHA256 etc. hasher. 

It provides a single instance of a SHA!, SHA256 etc. algorithm.  The hasher is created later (`digest`), that is the one that contains `.update()` and `.finalize()`

File: keylime/tpm/tpm_util.py (hunk 4 above)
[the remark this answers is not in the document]
Comment (stefanberger) [reply keylime/tpm/tpm_util.py]: @ansasaki Encryption with EC keys does exist: https://github.com/stefanberger/libtpms/blob/master/src/tpm2/crypto/openssl/CryptEccCrypt.c#L101   

> If makecredential is dependent on encryption using an RSA key, then there is no way to workaround this by using openssl.

When EC keys are used we would need to implement something like this here, which seems to be a different method: https://github.com/stefanberger/libtpms/blob/master/src/tpm2/CryptUtil.c#L504-L558

File: keylime/tpm/tpm2_objects.py [hunk 6]
```diff
@@ -71,11 +72,15 @@ def _pack_in_tpm2b(val: bytes) -> bytes:
...
-    TPM_ALG_SHA512: hashes.__dict__.get("SHA512"),
+HASH_FUNCS: Dict[int, hashes.HashAlgorithm] = {
+    TPM_ALG_SHA1: hashes.SHA1(),
+    TPM_ALG_SHA256: hashes.SHA256(),
+    TPM_ALG_SHA384: hashes.SHA384(),
+    TPM_ALG_SHA512: hashes.SHA512(),
+}
+
+SYMCIPHER_FUNCS = {
+    TPM_ALG_AES: algorithms.AES,
```
stefanberger, line 83: The user will have to instantiate this one with the actual key material, so we cannot instantiate it here..

keylime/tpm/tpm_util.py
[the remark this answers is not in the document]
stefanberger: If the EK is an EC key and to be used for MakeCredential then the biggest concerns for implementing CryptSecretEncrypt are the functions 
1) CryptEccIsPointOnCurve and 
2) CryptEccPointMultiply. 

For 1) this here may work: https://github.com/pyca/cryptography/blob/main/src/cryptography/hazmat/primitives/asymmetric/ec.py#L169

For 2) We may have to do something like this here: **https://github.com/pyca/cryptography/blob/c7146f9ed3830a9c3120722bc99dda81ce77037d/src/cryptography/hazmat/backends/openssl/backend.py#L1228-L1230**

stefanberger:
I think all discussions have been resolved and we should be able to merge this.

THS-on:
@stefanberger this currently does not support ECC keys, right? Do we want to add support for this later?

stefanberger:
> @stefanberger this currently does not support ECC keys, right? Do we want to add support for this later?

The tpm2 tools do not support it, either.
Later, maybe. I am not sure what advantage ECC EK keys provide to users or keylime over RSA keys... For as long as TPM 2's have RSA keys I think we are fine using RSA keys.

THS-on:
Ah, thanks for the clarification.

When we currently want to use an ECC AK, then we also generate an ECC EK, but we should probably split that up into two options.

stefanberger:
> Ah, thanks for the clarification.
> 
> When we currently want to use an ECC AK, then we also generate an ECC EK, but we should probably split that up into two options.

Oh, you are right about the ECC support. I hadn't see it before. This here does it: https://github.com/tpm2-software/tpm2-tools/blob/00848b281c5065ec5da0373704e02f02fc8a5425/tools/tpm2_makecredential.c#LL105C46-L105C46

How does one trigger the EC EK being picked up or it is just not picking up the NIST P384 key of swtpm?

keylime/tpm/ec_crypto_helper.py
```diff
@@ -0,0 +1,101 @@
...
+
+class EcCryptoHelper:
+    instance: Optional["EcCryptoHelper"] = None
+
+    @staticmethod
+    def get_instance() -> "EcCryptoHelper":
+        """Create and return a singleton EcCryptoHelper"""
+        if not EcCryptoHelper.instance:
+            EcCryptoHelper.instance = EcCryptoHelper()
+        return EcCryptoHelper.instance
```
THS-on, line 18: I'm wondering if we actually need a class and singleton pattern here. The backend could be just declared in the module and the rest are then normal functions. 

@aplanas what is the pythonic way to do this?

stefanberger: There are other ways of doing this of course but the singleton is there to check the validity of the backend class only once...

maugustosilva:
LGTM

keylime/tpm/ec_crypto_helper.py
aplanas: Not sure to understand why a single instance is needed. Is because the __init__ checks are too costly?

keylime/tpm/ec_crypto_helper.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def __init__(self) -> None:
+        self.backend = backends.default_backend()
+        if not isinstance(self.backend, OpenSSLBackend):
+            raise Exception("The python cryptography default backend must be using OpenSSL")
+
+        # sanity check the backend
+        if "_lib" not in dir(self.backend):
```
aplanas, line 26: Also this should be done with `hasattr()`

keylime/tpm/ec_crypto_helper.py
```diff
@@ -0,0 +1,103 @@
...
+            raise Exception("self.backend does not have _lib")
+
+        if "_ffi" not in dir(self.backend):
```
aplanas, line 29: Same ...

keylime/tpm/ec_crypto_helper.py
```diff
@@ -0,0 +1,103 @@
...
+
+        lib_dir = dir(self.backend._lib)
+        for func in [
+            "BN_CTX_get",
+            "BN_free",
+            "EC_KEY_get0_group",
+            "EC_POINT_new",
+            "EC_POINT_free",
+            "EC_POINT_mul",
+        ]:
```
aplanas, line 40: Ah ... I see now why need to access _lib.

There is a consideration here. `_lib` name is an indication that it is private, and should not be used because it can change in later releases (or disappear)

Seems to me that you want to access openssl directly, via the default backend. In this case I would suggest to use pyOpenSSL, as maybe it is exporting the symbols in a more stable way

keylime/tpm/ec_crypto_helper.py
```diff
@@ -0,0 +1,103 @@
...
+            raise Exception("The python cryptography default backend must be using OpenSSL")
+
+        # sanity check the backend
+        if "_lib" not in dir(self.backend):
+            raise Exception("self.backend does not have _lib")
+
+        if "_ffi" not in dir(self.backend):
+            raise Exception("self.backend does not have _ffi")
+
+        lib_dir = dir(self.backend._lib)
```
aplanas, line 32: `dir` should not be user here neither

keylime/tpm/ec_crypto_helper.py
stefanberger: I cannot respond to your comment above...

> Not sure to understand why a single instance is needed. Is because the init checks are too costly?

There's no need to check the interface more than once.

aplanas: > I cannot respond to your comment above...

You mean the `hasattr()`? If so is that `dir()` is not really expected to be used in code but in REPL. The natural code would be something like `if not hasattr(self.backend, "_lib")`

keylime/tpm/ec_crypto_helper.py
stefanberger: To avoid another dependency (pyOpenSSL)  just for point multiplication I think I will do the same as this here and use Binding directly: https://github.com/pyca/cryptography/blob/main/src/cryptography/hazmat/backends/openssl/backend.py#L147
Binding is documented and there's a good chance it will not change: https://cryptography.io/en/3.0/hazmat/bindings/openssl/

aplanas: A cool idea. Yes. Using public interfaces is more correct.

keylime/tpm/ec_crypto_helper.py
stefanberger: Switched to hasattr now.

keylime/tpm/ec_crypto_helper.py
stefanberger: Switch to hasattr() now. Would squash the top 3 patches at some point.

keylime/tpm/ec_crypto_helper.py
stefanberger: Done.

keylime/tpm/ec_crypto_helper.py
```diff
@@ -0,0 +1,127 @@
...
+
+    @staticmethod
+    def get_instance() -> "EcCryptoHelper":
+        """Create and return a singleton EcCryptoHelper"""
+        if not EcCryptoHelper.instance:
+            EcCryptoHelper.instance = EcCryptoHelper()
+        return EcCryptoHelper.instance
+
+    def __init__(self) -> None:
+        self.backend = backends.default_backend()
```
aplanas, line 22: Why this is needed? Can be because some side effects, like some object creation, loading of the *.so ... ? If so, can you add a comment? Maybe should not be stored as a field in the class?

aplanas (approved):
Just a comment about self.backed, but LGTM

keylime/tpm/ec_crypto_helper.py
stefanberger: It's not need anymore... I removed it.

PR #1388, opened 2023-05-22
tpm: Get rid of AbstractTPM class

stefanberger:
AbstractTPM may have served as an interface-defining class while TPM 1.2 support was still there but now it's obsolete, so remove it.


keylime/tpm/tpm_main.py
```diff
@@ -29,8 +29,9 @@
 logger = keylime_logging.init_logging("tpm")
 
 
-class tpm(tpm_abstract.AbstractTPM):
+class tpm:
```
aplanas, line 32: Maybe a chance to rename it as "Tpm". Usually lowercase names for classes are for primitive types (list, dict) but not for user defined classes

keylime/tpm/tpm_main.py
```diff
@@ -29,8 +29,9 @@
...
     tools_version: str = ""
+    EXIT_SUCCESS: int = 0
```
aplanas, line 34: As a constant, I would put it outside the class scope

keylime/tpm/tpm_main.py
```diff
@@ -90,7 +91,7 @@ def run(self, cmd: Sequence[str]) -> cmd_exec.RetDictType:
     def __run(
         self,
         cmd: Sequence[str],
-        expectedcode: int = tpm_abstract.AbstractTPM.EXIT_SUCESS,
+        expectedcode: int = EXIT_SUCCESS,
```
aplanas, line 94: that should fail, shouldn't?

THS-on (commented):
Once the style and linter checks are fixed this LGTM. 

Note that https://github.com/keylime/keylime/pull/1369 also modifies the TPM code so it defintetly requires then a rebase, but I think merging this first makes sense

keylime/tpm/tpm_main.py
```diff
@@ -399,6 +629,18 @@ def check_quote_timing(
...
 
+    @staticmethod
+    def hashdigest(payload: bytes, algorithm: Hash) -> Optional[str]:
+        digest = algorithm.hash(payload)
+        if digest is None:
+            return None
+        return codecs.encode(digest, "hex").decode("utf-8")
+
+    @staticmethod
+    def START_HASH(algorithm: Hash) -> str:
```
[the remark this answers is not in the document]
stefanberger: Fixed.

keylime/tpm/tpm_main.py
stefanberger: Done.

THS-on:
I checked the CodeQL warnings and they are false positives, because they assume agent_id can leak the mtls certificate.

stefanberger:
> Once the style and linter checks are fixed this LGTM.

Fixed.

THS-on (approved):

PR #1389, opened 2023-05-22
IMA: cleanups and simplifications

stefanberger:
This PR does some cleanups and simplifications on IMA code.

### DIFF
--- a/keylime/ima/ima.py
+++ b/keylime/ima/ima.py
@@ -573,8 +573,13 @@ def verify_runtime_policy(
             code=400,
         )
 
+    # validate that the runtime_policy is proper JSON
+    try:
+        runtime_policy_json = json.loads(runtime_policy)
+    except Exception as error:
+        raise ImaValidationError(message="Runtime policy is not valid JSON!", code=400) from error
+
     # detect if runtime policy is DSSE
-    runtime_policy_json = json.loads(runtime_policy)
     if runtime_policy_json.get("payload"):
         if verify_sig:
             if not signing.verify_dsse_envelope(runtime_policy, runtime_policy_key):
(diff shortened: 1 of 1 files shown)


keylime/ima/ima.py
```diff
@@ -212,18 +212,19 @@ def _validate_ima_ng(
             failure.add_event("not_in_allowlist", f"File not found in allowlist: {path.name}", True)
             return failure
 
-        if codecs.encode(digest.hash, "hex").decode("utf-8") not in accept_list:
+        hex_hash = digest.hash.hex()
```
stefanberger, line 214: My guess is we didn't know about the hex() method on bytes and therefore used the more complicated code...

keylime/ima/ima.py
```diff
@@ -322,7 +321,7 @@ def _process_measurement_list(
     errors: Dict[Type[ast.Mode], int] = {}
     pcrval_bytes = b""
     if pcrval is not None:
-        pcrval_bytes = codecs.decode(pcrval.encode("utf-8"), "hex")
+        pcrval_bytes = bytes.fromhex(pcrval)
```
stefanberger, line 324: My guess here is we didn't know about the fromhex() method on bytes and therefore used the more complicated code...

stefanberger:
```
>>> bytes.fromhex("123456abcd")
b'\x124V\xab\xcd'
>>> codecs.decode("123456abcd".encode("utf-8"), "hex")
b'\x124V\xab\xcd'
```

Even number of letters required:
```
>>> bytes.fromhex("123456abc")
  File "<stdin>", line 1, in <module>
ValueError: non-hexadecimal number found in fromhex() arg at position 9

>>> codecs.decode("123456abc".encode("utf-8"), "hex")
  File "/usr/lib64/python3.10/encodings/hex_codec.py", line 19, in hex_decode
    return (binascii.a2b_hex(input), len(input))
binascii.Error: Odd-length string

The above exception was the direct cause of the following exception:

  File "<stdin>", line 1, in <module>
binascii.Error: decoding with 'hex' codec failed (Error: Odd-length string)
```

non-hex letters make both fail:
```
>>> bytes.fromhex("123456gg")
  File "<stdin>", line 1, in <module>
ValueError: non-hexadecimal number found in fromhex() arg at position 6

>>> codecs.decode("123456gg".encode("utf-8"), "hex")
  File "/usr/lib64/python3.10/encodings/hex_codec.py", line 19, in hex_decode
    return (binascii.a2b_hex(input), len(input))
binascii.Error: Non-hexadecimal digit found

The above exception was the direct cause of the following exception:

  File "<stdin>", line 1, in <module>
binascii.Error: decoding with 'hex' codec failed (Error: Non-hexadecimal digit found)
```

stefanberger: Note: We should probably do this throughout the whole code base where codec is used for writing hex numbers or parsing hex numbers

THS-on: Yes that was the case, we just need to make sure that the behavior is the same and now catch a  `ValueError` instead of a `binascii.Error`.

THS-on (approved):

maugustosilva (approved):

PR #1391, opened 2023-05-23
pypright: Remove ignored files that do not exist anymore

stefanberger:


THS-on (approved):
`test/test_ima_dm.py` still exists, but we can remove it from the ignore list.

stefanberger:
> `test/test_ima_dm.py` still exists, but we can remove it from the ignore list.

Right, restored that one in the list.

stefanberger:
@THS-on Iirc it was an class-upcast that was causing the pyright failures in test_ima_dm.py.

This here resolves it:
```
-            ima_buf: ast.ImaBuf = table_load_verity.mode
+            ima_buf: ast.ImaBuf = cast(ast.ImaBuf, table_load_verity.mode)
```

Wdyt?

THS-on:
A cast fine for the test.

stefanberger:
> A cast fine for the test.
Fixed up this test and added to this PR now.

pyproject.toml
```diff
@@ -27,9 +27,6 @@ ignore = [
     "keylime/da/examples/redis.py",
     "keylime/da/examples/file.py",
     "keylime/da/record.py",
-    "test/oldtest.py",
-    "test/test_ima_dm.py",
-    "test/test_restful.py",
```
aplanas, line 32: Didn't we propose to ignore all the tests? IMHO those should not be type checked ...

aplanas (approved):

pyproject.toml
stefanberger: Now it's done and having the test cases type-checked as well caused me to fix a missing field in the Policies type, which does have some benefits already.

PR #1394, opened 2023-05-30
More TPM related cleanups

stefanberger:
This PR provides more TPM related cleanups.


keylime/tpm/tpm_util.py
```diff
@@ -434,3 +435,47 @@ def crypt_kdfe(
         size -= hashfunc.digest_size
 
     return result[:size_in_bytes]
+
+
+class TPM_Utilities:
```
THS-on, line 440: I think we can drop the class wrapper all together.

stefanberger: Yeah, it doesn't buy much. Dropping it.

THS-on (approved):

keylime/tpm/tpm_util.py
```diff
@@ -434,3 +435,45 @@ def crypt_kdfe(
...
+        password += chars[(rand[i]) % len(chars)]
+    return password
+
+
+def readPolicy(configval: str) -> Dict[str, Any]:
+    policy: Dict[str, Any] = json.loads(configval)
+
+    # compute PCR mask from tpm_policy
+    mask = 0
+    for key in list(policy.keys()):
```
aplanas, line 460: `for key in policy:`

aplanas (approved):
Just a comment, but LGTM

keylime/tpm/tpm_util.py
stefanberger: Fixed. Thanks.

aplanas (approved):

kkaarreell:
Could you please rerun Packit tests (/packit test)? I am wondering why there were so many failures.

ansasaki:
> Could you please rerun Packit tests (/packit test)? I am wondering why there were so many failures.
 
@kkaarreell For most of the failures, I see that the installation of the RPM from Copr failed due to a `404` error (probably caused by an outage), and then the old agent version 1.0.0 is installed instead.

On Rawhide, I updated the agent to version 2.1.0. The results there are closer to what they should be. I see failures on durable attestation and measured boot tests, apparently due to missing PCRs in the quote. I'll investigate.

kkaarreell:
I believe we were discussing those mb failures in one of the former PRs. I assume these would required test updates. Regarding other failures, yes, I wanted to double check that they would not appear again.

kkaarreell:
The COPR issue should be resolved now. You can rerun tests.

kkaarreell:
@stefanberger Could you please also manually rerun the failing coverage check? It Is not handled by Packit and therefore wasn't rescheduled.

stefanberger:
> @stefanberger Could you please also manually rerun the failing coverage check? It Is not handled by Packit and therefore wasn't rescheduled.

I don't think I have the rights to do this. @maugustosilva ?

PR #1395, opened 2023-05-30
Yet more cleanup

stefanberger:
Cleanup continues...


aplanas (approved):

THS-on (approved):
Most of that looks like it was added when a VerifierMain object was passed to that function. Because no caller is doing  that anymore the changes LGTM.

We just need to make sure that we don't copy anything from the database by reference accidentally.

Issue #1396, opened 2023-06-01
The version of tpm2-tools in installer.sh is too low：Parsing of binary boot measurements failed with: ['{"message": "running tpm2_eventlog failed"}'] 


hustliyilin:
# Environment
 * Keylime version:  master

# Description 

In tpm2-tools, https://github.com/tpm2-software/tpm2-tools/commit/b3dac7d32f9cdc6154720ecb01d0220cca6d2f79
fix the error `Cannot allocate memory when using --eventlog-version=2`.

![image](https://github.com/keylime/keylime/assets/6824093/948d907b-3e94-4065-84f9-5fb6e2517ca7)

In keylime, https://github.com/keylime/keylime/blob/master/keylime/tpm/tpm_main.py#L711
also uses the tpm2_eventlog command to pasre MokListTrusted, which gets
a similar error ` - root - ERROR - Parsing of binary boot measurements failed with: ['{"message": "running tpm2_eventlog failed"}']`

![image](https://github.com/keylime/keylime/assets/6824093/928772d2-db00-4c7c-89dd-b0f7da8f3b5d)

So we need to 5.4 to fix the bug of parsing for MokListTrusted using the tpm2_eventlog command.

PR #1397, opened 2023-06-01
installer.sh: update TPM2TOOLS_VER to 5.5 and cherry-pick patches to fix the bug of parsing for most newer logs with the tpm2_eventlog command.

hustliyilin:
In tpm2-tools, https://github.com/tpm2-software/tpm2-tools/commit/b3dac7d32f9cdc6154720ecb01d0220cca6d2f79 fix the error `Cannot allocate memory when using --eventlog-version=2`. In keylime, https://github.com/keylime/keylime/blob/master/keylime/tpm/tpm_main.py#L711 also uses the tpm2_eventlog command to pasre MokListTrusted, which gets a similar error ` - root - ERROR - Parsing of binary boot measurements failed with: ['{"message": "running tpm2_eventlog failed"}']`

So we need to 5.4 to fix the bug of parsing for MokListTrusted using the tpm2_eventlog command.

Fixes: #1396 

### DIFF
--- a/installer.sh
+++ b/installer.sh
@@ -12,7 +12,7 @@ TPM2SIM_SRC=http://sourceforge.net/projects/ibmswtpm2/files/ibmtpm1119.tar.gz/do
 KEYLIME_VER="master"
 TPM4720_VER="master"
 TPM2TSS_VER="3.2.x"
-TPM2TOOLS_VER="5.1.X"
+TPM2TOOLS_VER="5.5"
 
 # Minimum version requirements
 MIN_PYTHON_VERSION="3.6.7"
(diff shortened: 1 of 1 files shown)


THS-on:
We can also update to 5.5 and cherry-pick https://github.com/tpm2-software/tpm2-tools/commit/9735dc332b782ce94371aff8746ce6a68094b038 and https://github.com/tpm2-software/tpm2-tools/commit/576a31bcc910da517067b29667f45fbe78e812e0. This would cover parsing for most newer logs.


@mpeters for packagers: if they want to use the measured boot feature on a modern distro they would need to update also tpm2-tools to a recent version.

installer.sh
```diff
@@ -12,7 +12,7 @@ TPM2SIM_SRC=http://sourceforge.net/projects/ibmswtpm2/files/ibmtpm1119.tar.gz/do
 KEYLIME_VER="master"
 TPM4720_VER="master"
 TPM2TSS_VER="3.2.x"
-TPM2TOOLS_VER="5.1.X"
+TPM2TOOLS_VER="5.4"
```
aplanas, line 15: Why not to update to 5.5?

THS-on (approved):

aplanas (approved):

PR #1399, opened 2023-06-02
bump version to 7.2.5

maugustosilva:


THS-on (approved):
Unconventional version jump, but inline with what we already tagged.

PR #1402, opened 2023-06-06
ima: remove code used for reading the IMA log from securityfs

THS-on:


stefanberger (approved):
This was agent code, so it can go.

aplanas (approved):

Issue #1404, opened 2023-06-07
The default value of  `tpm_ownerpassword` causes authentication failure


ansasaki:
Is your issue a feature request? If so, please raise it as an [enhancement](https://github.com/keylime/enhancements)

# Environment
 * OS / version: N/A
 * Processor architecture: N/A
 * TPM Manufacturer: N/A
 * Keylime version: current master (6e58ee5)

# Description 
The default configuration value for the agent option `tpm_ownerpassword` is the string `"keylime"`. Differently from the old python agent, the rust agent does not change the TPM password, but will try to use the password set in the configuration file to authenticate. This results in authentication failure in the default case when the user never changed the TPM password.

To avoid such problem, the default value of the `tpm_ownerpassword` should be the empty password.

It is true that if the user uses the `keylime-agent.conf` configuration from the keylime/rust-keylime repository, then it will work as expected since the default value for the `tpm_ownerpassword` option there is the empty string. But it is not uncommon that the user installs the components from keylime/keylime repo, and generates the configuration files for all components. If the user uses the configuration file generated by the `keylime_upgrade_config --defaults`, then the rust agent will fail authentication due to the value set in `tpm_ownerpassword` option.

## Expected behavior vs. actual behavior 
Expected behavior: the agent using the generated default configuration files should work in the default installation

Actual behavior: the agent using the generated default configuration files does not work due to authentication failure caused by the value set in `tpm_ownerpassword` option

## Steps to reproduce problem
1. In a machine where the TPM owner password is the default empty string, try to run the Keylime agent using the generated configuration file with the default values
2. The agent will fail authentication due to the `"keylime"` string set in `tpm_ownerpassword` option

# Relevant logs

An error similar to the following will occur:
```
WARNING:esys:src/tss2-esys/api/Esys_CreatePrimary.c:400:Esys_CreatePrimary_Finish() Received TPM Error
ERROR:esys:src/tss2-esys/api/Esys_CreatePrimary.c:135:Esys_CreatePrimary() Esys Finish ErrorCode (0x000009a2)
Error: Tpm(Tss2 { err: Tss2Error(FormatOne(FormatOneResponseCode { .0: 2466, error_number: 34, parameter: false, format_selector: true, number: 9 })), kind: Some(BadAuth), message: "authorization failure without DA implications (associated with session number 1)" })
```

PR #1405, opened 2023-06-07
templates: Set empty string as default value for tpm_ownerpassword

ansasaki:
The default value for the TPM owner password should be the empty string to avoid authentication failure on machines with default TPM owner password configuration.

Resolves: #1404

### DIFF
--- a/templates/2.0/mapping.json
+++ b/templates/2.0/mapping.json
@@ -100,7 +100,7 @@
             "tpm_ownerpassword": {
                 "section": "cloud_agent",
                 "option": "tpm_ownerpassword",
-                "default": "keylime"
+                "default": ""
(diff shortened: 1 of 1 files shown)


aplanas (approved):

stefanberger (approved):

ansasaki:
> I suppose the 'machines with default TPM owner password configuration' is an anticipated password that users typically set when they set a TPM owner password, which then happens to be the **typical** empty string?

It is more the password that comes set when you get a new machine with a new TPM hardware: it is the empty string by default (I believe even the TPM specification recommends the empty string to be the initial owner password string). Unless the user set the TPM password for some reason, it will be the empty password.

PR #1406, opened 2023-06-07
mba: moving the boot event log parsing to the MBA subdirectory

galmasi:
This is another preparatory PR for making measured boot attestation pluggable.

In this PR we remove all helper functions from tpm_main into a separate directory.
* tpm_main is now officially free of any involvement with `tpm2-tools` -- thanks to previous PRs by @stefanberger .
* the only access to parsing boot event logs is through `keylime/mba`
* changing `keylime/mba` to allow pluggable function implementations for parsing and policy checking will be the subject of the next (3rd) PR in this sequence. That PR will allow replacement of `tpm2-tools` based parsing with python-only parsing; we are also working on using OPA for policy checking.

Further todos to be tackled in the 3rd PR is more consistent use of exception handling for all these putatively pluggable functions.


stefanberger:
> This is another preparatory PR for making measured boot attestation pluggable.

What does plugable actually mean in this context? When the tenant tool provides the necessary parameters to cause measured boot and now the boot log and quote are passed to the server then how does the keylime server react when measured boot has become plugable ? I am wondering about this in the context of IMA as well...

galmasi:
@stefanberger very good and topical question, because we'd like to make the IMA policy engine pluggable as well; we are just concentrating on MBA right now because we are on more familiar ground.

To me, "pluggable" specifically means that there is a documented API for MBA execution that any Python library can implement. The keylime verifier can load this python library by reading a (yet to be defined) keylime configuration option. 

The third PR will actually implement the dynamic load mechanism I'm describing here, while keeping all the defaults. I want to have a tpm2-tool-less event log parser to be available as an option -- who knows, the default eventually. 

I will also have said documentation ready for the PR.

In the longer term we also want to institutionalize this plugin approach with IMA, with registration and eventually with the make/activatecredential loop.

stefanberger:
> @stefanberger very good and topical question, because we'd like to make the IMA policy engine pluggable as well; we are just concentrating on MBA right now because we are on more familiar ground.

The topic 'plugability' seems to be worth a discussion to have and at least I would need some clarity since I am not sure where making MB and IMA truly plugable would even start. Parameters concerning measured boot and IMA are all over the place, starting in `cloud_verifier_tornado.py` where we are handling things like `mb_refstate` or `tpm_policy`  etc. Iirc Keylime was originally only supporting IMA measurements and therefore in its core it has support for IMA measurements all over the code since this is what it only supported. So, what is the expected gain when making IMA and MB plugable? If we want to support multiple different types of attestation etc. then the problem IMHO lies somewhere else. We would need a different format to carry over the REST calls or different REST calls just to begin with so that we can detect that the message that was sent to the server contains IMA attestation and it can then be routed to the IMA attestation handler and if it contains XYZ stuff it is handled by the XYZ hander. But we're nowhere close to this.

galmasi:
Pluginizing MBA is not going to be very hard in principle. 
* Take a boot event log, parse it, produce a JSON formatted event log, a set of PCRs to authenticate, and a boot aggregate.
* Take the JSON formatted event log, a "policy file" aka "reference state" and an arbitrary policy engine, and return a (possibly empty) set of policy match failures.

Note that we are carefully not saying what the format of the policy file should be, or how the policy engine should do its job. We are just documenting the format of the parsed event log and the format of the event log policy engine's output. As long as the plugin satisfies these requirements, it's usable.

The current "elchecking" event log policy engine _does_ specify the JSON format of the reference state. We are going to keep it around as the "default" implementation of the plug. But we have ideas to replace it with OPA in the longer term, since OPA has a larger customer base and is generally more tested and proven fit for the purpose.

stefanberger:
> * we are also working on using OPA for policy checking.

Along the lines of OPA I am wondering whether Rego can handle hashing (replay of PCRs/PCR 10) and IMA signature verification with multiple verification keys? How well does it handle dynamic policies? Do you have a Rego policy that can read allowlists from a 'dynamic part' (system specific hashes) that may be different per system?

maugustosilva:
@stefanberger The picture is not totally clear in my head, I freely admit, but based on separate discussions with both @galmasi and @THS-on the very first step towards "pluginization" of IMA in Keylime would similar to what we do today: first  go through the IMA log, extending the (virtual) PCR 10, and the comparing the value with what we have on the quote, and then, only then, separately, processing the entries looking for checksum matches (and taking into account the exclude list). 

I am aware that extending the (virtual) PCR 10 is not as simple as it is in the measured boot case, given that we need to search for a match on IMA log, which could be several lines before the tip of the log (due to mismatch between PCR 10 on the quote and the values on the log). IMO this is yet another reason to have these two pieces of code: a) matching of PCR 10 and b) checking of checksums against allowlist/excludelist clearly separated. These are basically co-located currently on the code. 

Of course, the "validade each individual entry on IMA log", as it is today, would be our "default IMA plugin".

stefanberger:
> @stefanberger The picture is not totally clear in my head, I freely admit, but based on separate discussions with both @galmasi and @THS-on the very first step towards "pluginization" of IMA in Keylime would similar to what we do today: first go through the IMA log, extending the (virtual) PCR 10, and the comparing the value with what we have on the quote, and then, only then, separately, processing the entries looking for checksum matches (and taking into account the exclude list).
> 
> I am aware that extending the (virtual) PCR 10 is not as simple as it is in the measured boot case, given that we need to search for a match on IMA log, which could be several lines before the tip of the log (due to mismatch between PCR 10 on the quote and the values on the log). IMO this is yet another reason to have these two pieces of code: a) matching of PCR 10 and b) checking of checksums against allowlist/excludelist clearly separated. These are basically co-located currently on the code.

We have to loop through the measurement log twice then. Other than it being less efficient, why would it be 'better'? I also would not see the point of just sending a quote without a log or a log without a quote so that it would make sense to separate them.

maugustosilva:
I do concede that it would be less efficient in the "vanilla" case (i.e., "default" plugin, whenever we get changes on the IMA log), but very, much like the measure boot pluginization, we are taking a (small, but shall be properly measured) performance hit to allow for more flexibility. 

The tradeoff here is that now one can replace the "default" IMA log checksum evaluator "plugin" with different ones, including some which might process IMA logs taking into account multiple allow and exclude lists, including ones generated from containers (a necessity in our production environment, where Kubernetes/Openshift is deployed on bare-metal). 

Given that this processing could be done in parallel (and by an external entity) for multiple `agents`, there are scenarios - mostly at scale - where having these two loops separated might be actually more efficient from a `verifier` stand point.

stefanberger:
> I do concede that it would be less efficient in the "vanilla" case (i.e., "default" plugin, whenever we get changes on the IMA log), but very, much like the measure boot pluginization, we are taking a (small, but shall be properly measured) performance hit to allow for more flexibility.
> 
> The tradeoff here is that now one can replace the "default" IMA log checksum evaluator "plugin" with different ones, including some which might process IMA logs taking into account multiple allow and exclude lists, including ones generated from containers (a necessity in our production environment, where Kubernetes/Openshift is deployed on bare-metal).

Is the tenant tool and the required database entries/format that need to somehow support and reflecting this kind of desired process of multiple lists per system (which is obviously also not possible today) going to become plugin's - or part of this IMA Attestation plugin - as well?
I would just generate one plugin that can take multiple of these lists instead of one that can only handle one allowlist and one exclude list. The newer one will immediately supersede the older one.

> 
> Given that this processing could be done in parallel (and by an external entity) for multiple `agents`, there are scenarios - mostly at scale - where having these two loops separated might be actually more efficient from a `verifier` stand point.

And what is the external entity in this case? OPA?

Anyway, I think it's better to wait to see how this works out for MBA.

galmasi:
@stefanberger re: OPA efficiency that is a question in my mind as well. OPA is adequate for MBA, very much unsure about how it will handle a 200k entry IMA allow list. We have not yet measured. ... that is why we are looking at other policy agents as well, say something like Red Hat's `seedwing`. Preliminary stages only. 

FWIW -- and this is probably offtopic for this PR -- keylime plugins are coming for us whether we want it or not. Handling IMA namespaces ... as you _well_ know ... does not have a simple or single answer. Attestation of confidential VMs and container, ditto.

Shorter term -- this PR and the next -- I don't think anyone will object to the laudable goal of further reducing the keylime dependencies on `tpm2-tools`. We are tantalizingly close to independence now.

maugustosilva:
> > I do concede that it would be less efficient in the "vanilla" case (i.e., "default" plugin, whenever we get changes on the IMA log), but very, much like the measure boot pluginization, we are taking a (small, but shall be properly measured) performance hit to allow for more flexibility.
> > The tradeoff here is that now one can replace the "default" IMA log checksum evaluator "plugin" with different ones, including some which might process IMA logs taking into account multiple allow and exclude lists, including ones generated from containers (a necessity in our production environment, where Kubernetes/Openshift is deployed on bare-metal).
> 
> Is the tenant tool and the required database entries/format that need to somehow support and reflecting this kind of desired process of multiple lists per system (which is obviously also not possible today) going to become plugin's - or part of this IMA Attestation plugin - as well? I would just generate one plugin that can take multiple of these lists instead of one that can only handle one allowlist and one exclude list. The newer one will immediately supersede the older one.
> 
> > Given that this processing could be done in parallel (and by an external entity) for multiple `agents`, there are scenarios - mostly at scale - where having these two loops separated might be actually more efficient from a `verifier` stand point.
> 
> And what is the external entity in this case? OPA?
> 
> Anyway, I think it's better to wait to see how this works out for MBA.

The key takeaway is, yes, we should wait and use MBA pluginization as the "case study before proceeding. I just wanted to communicate (and get feedback from you) on the rationale behind this "full pluginization" for whole of Keylime.

stefanberger:
> @stefanberger re: OPA efficiency that is a question in my mind as well. OPA is adequate for MBA, very much unsure about how it will handle a 200k entry IMA allow list. We have not yet measured. ... that is why we are looking at other policy agents as well, say something like Red Hat's `seedwing`. Preliminary stages only.

My 1st stage concern is 'suitability' of OPA. It may be used industry-wide but maybe it just isn't right for us. IMA Attestation has large dynamic (updateable) lists of hashes and exclude lists with regex's, we have to replay PCRs extensions, validate quotes and we have keyrings and signature verification of executables. Now which part goes into OPA?

> 
> FWIW -- and this is probably offtopic for this PR -- keylime plugins are coming for us whether we want it or not. Handling IMA namespaces ... as you _well_ know ... does not have a simple or single answer. Attestation of confidential VMs and container, ditto.

Correct me if I am wrong, but IMA namespaces would primarily need **different** runtime policies, very likely much shorter than those of physical machines or VMs, but otherwise the same format as we have it today. Since IMA namespaces from the perspective of attestation today works the same way in containers as it works inside VMs or on physical machines, there should not be any difference in how they are used. Real problem here more in the direction of pull versus push model.

Also, I don't see plugins as the 1st level problem.The first level problem IMHO is having a  format to transport the messages along with extension to runtime policy, database entries, tenant tool etc. and then one can invoke handlers with the data that was sent. Problem is just that MBA and IMA Attestation specific stuff is all over the code right now because that's what Keylime was designed for. And that's where the problem is: How do you draw a line in the code that defines a plugin so that you can define an interface for a IMA Attestation or MBA handler? From what I can see we have specifics for IMA Attestation in the database tables, the runtime policy, REST calls that support the runtime policy, tenant tool and server code.

galmasi:
> Problem is just that MBA and IMA Attestation specific stuff is all over the code right now because that's what Keylime was designed for.

That is specifically what this PR -- and its predecessor -- is supposed to fix; for MBA. `keylime/mba/mba.py` is supposed to end up as the abstract class describing interactions with the measured boot attestation subsystem, and the implementation is hidden behind it. Don't nail me to the wall, the API is far from perfect. Small steps.

Niteesh Dubey has found places in the verifier code that make implicit assumptions about the policy being JSON. Separate PR in the works to get that fixed.

keylime/mba/elparsing/tpm2_tools_elparser.py
```diff
@@ -0,0 +1,264 @@
+import base64
```
aplanas, line 1: Why moving TPM related stuff from keylime.tpm to keylime.mba.elparsing?

aplanas: OK. So the goal is to remove the event log parser from the TPM class, that happens to use tpm2-tools for doing that for this implementation.

Maybe we should not reference tpm2_tools in the module name, then? Seems that the use of tpm2_eventlog is in this case a technical detail. Replacing the implementation to not use the tpm2_tools should not imply a renaming of the module and a change in the code that import it.

maugustosilva: If I understood your argument correctly @aplanas, I believe the question here is whether or not we are going to keep the invocation of the `tpm2-tools` eventlog parser as a possible "plugin" or not. This is a separate and interesting question, one that I do not believe we can collectively answer today.

Be as it may, even if we decide to keep it around, we should see something like both `tpm2_tools_elparser.py` and `native_python_elparser.py` (just invented the name for the sake of argument) on the aforementioned directory in a **future** PR (with the appropriate `importlib` code, of course!) and thus I favor that we merge "as is" now (there is third PR coming right after this one, FYI).

stefanberger: My $0.02: we should only have one and not two since if we had two we would have to test two.

maugustosilva: I tend to agree with you. I believe our whole effort of avoiding installing `tpm2-tools` is a worthy one, but even if the majority of admins decide otherwise, we would be covered by having a "modular" implementation (in subsequent PRs, of course)

maugustosilva:
LGTM

keylime/mba/elparsing/tpm2_tools_elparser.py
galmasi: I agree with @stefanberger also. The native event log parser is coming to keylime, and I think the incremental price in upkeep to for the parser is worth paying to be free of this: https://github.com/keylime/keylime/blob/be94e4f2ab2002155908f045ef99de86492236e9/keylime/tpm/tpm_main.py#L64-L84

PR #1410, opened 2023-06-21
mba: making MBA policy parser and checker pluggable

galmasi:
## Pluggable MBA theory of operations

The module `keylime/mba/mba` provides an interface for measured boot attestation verification that consists of four functions.

* `bootlog_parse()`
   * converts a boot log from binary to JSON.
   * authenticates the log by checking the consistency of its digests.
   * provides a set of PCRs that can be verified against the TPM quote.
   * provides a boot aggregate for use by Keylime IMA verification.
* `policy_load()`
   * load a policy from a file into a Python string
   * optionally validate that the policy is syntactically correct (this is TBD)
*  `policy_is_valid()`
   *  test whether a string is valid policy
* `bootlog_evaluate()`
   * check the parsed boot log against the loaded policy, and return a list of policy failures.

A default implementation for all four functions is predefined in Keylime, and is backwards compatible (i.e. this is *not* a breaking change). The functions are implemented in `keylime/mba/elparsing` and `keylime/mba/elchecking` respectively.

However, these functions can be overridden by specifying a list of modules to load in the configuration option `verifier.measured_boot_imports`. Upon initialization MBA loads the modules as specified by the verifier option. These modules can contain implementations of the four functions above.

At runtime keylime chooses the active implementation for each of the functions described above by enumerating the ordered list of imported modules and picking the first implementation of the function it can find.

### DIFF
--- a/keylime/cloud_verifier_common.py
+++ b/keylime/cloud_verifier_common.py
@@ -10,6 +10,7 @@ from keylime.db.verifier_db import VerfierMain
 from keylime.failure import Component, Event, Failure
 from keylime.ima import file_signatures
 from keylime.ima.types import RuntimePolicyType
+from keylime.mba import mba
 from keylime.tpm import tpm_util
 from keylime.tpm.tpm_main import Tpm
 
@@ -259,8 +260,7 @@ def prepare_get_quote(agent: Dict[str, Any]) -> Dict[str, Union[str, int]]:
 def process_get_status(agent: VerfierMain) -> Dict[str, Any]:
     has_mb_refstate = 0
     try:
-        mb_refstate = json.loads(cast(str, agent.mb_refstate))
-        if mb_refstate and mb_refstate.keys():
+        if mba.policy_is_valid(cast(str, agent.mb_refstate)):
             has_mb_refstate = 1
     except Exception as e:
         logger.warning(
--- a/keylime/mba/elchecking/elchecker.py
+++ b/keylime/mba/elchecking/elchecker.py
@@ -3,17 +3,63 @@ from typing import Optional, Set
 
 from keylime import config, keylime_logging
 from keylime.failure import Component, Failure
-from keylime.mba.elchecking import tests
+from keylime.mba.elchecking import policies, tests
 
 logger = keylime_logging.init_logging("measured_boot")
 
 
-def evaluate_bootlog(
+def policy_load(policy_path: Optional[str] = None) -> str:
+    """
+    Load (and validates) an actual policy file.
+    :param policy_path: <optional> name of policy file to load
+    :returns: a string defining the policy.
+    Errors: if the policy file cannot be read, or contains errors, this function may
+    cause exceptions. Validation in this case is to confirm that the file
+    is formatted as proper JSON; nothing more.
+
+    TODO: default policy should probably not be defined, because it depends on the policy engine itself?
+    """
+    try:
+        if policy_path is None:
+            policy_path = config.get("tenant", "mb_refstate")
+        with open(policy_path, encoding="utf-8") as f:
+            mb_policy_data = json.load(f)
+            return json.dumps(mb_policy_data)
+    except Exception as e:
+        raise ValueError from e
+
+
+def policy_is_valid(mb_refstate: Optional[str]) -> bool:
+    """
+    Returns true if the policy argument is a valid nonempty policy
+    """
+    if not mb_refstate:
+        return False
+    try:
+        mb_refstate_obj = json.loads(mb_refstate)
+    except Exception as _:
+        return False
+    if not mb_refstate_obj:
+        return False
+    if len(mb_refstate_obj) == 0:
+        return False
+    return True
+
+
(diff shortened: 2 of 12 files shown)


keylime/mba/elchecking/elchecker.py
```diff
@@ -76,3 +119,6 @@ def evaluate_bootlog(
             True,
         )
     return failure
+
+
+logger.info("mba.elchecking.elchecker: policy names = %s", str(policies.get_policy_names()))
```
mpeters, line 124: What's the purpose of this log line? It seems to be outside the scope of any methods.

galmasi: The log announcement runs during initialization (during the import of the `elchecker` module). 

The actual registration of policies happens during import in the elchecker design. The actual line of code is [here](https://github.com/galmasi/keylime/blob/03fb3ad3df3a52dbce37fb4b82d9e960420ec160/keylime/mba/elchecking/example.py#L404
). I won't claim to like it, but my goal throughout was to mess with the working code as little as possible. I moved things around, but I didn't change what I didn't have to.

mpeters: Yeah, I got that it happened during module init, I was more curious about why it was needed. I now see the more dynamic nature of the module being loaded and seeing the order would be useful when multiple modules are loaded. Maybe it should be a debug instead?

galmasi: Yes, I agree, "debug" is better. Will make the change.

keylime/mba/mba.py
```diff
@@ -21,46 +46,98 @@
...
+    Side effects: Reads from keylime configuration files.
+      After execution all imports required by MBA are loaded and initialized.
+    Exceptions: If any policy engine in the list cannot be loaded.
+    """
+    try:
+        imports = config.getlist("verifier", "measured_boot_imports")
+        # these are the defaults
+        imports.append("keylime.mba.elchecking.elchecker")
+        imports.append("keylime.mba.elchecking.example")
+        imports.append("keylime.mba.elparsing.tpm2_tools_elparser")
```
aplanas, line 73: Those that are always in the list, should unconditionally imported with "import". Why to do that dynamically?

keylime/mba/mba.py
```diff
@@ -21,46 +46,98 @@
...
-        policies.load_policies()
+        m = _find_implementation("bootlog_parse")
+        return m.bootlog_parse(bootlog_b64, hash_alg)  # type: ignore[no-any-return]
     except Exception as e:
         raise ValueError from e
 
 
 # ##########
 # read a policy definition (aka "refstate") from a file.
+# TODO: do we need a policy file validator?
```
aplanas, line 110: is not the task of policy_is_valid defined later?

keylime/mba/elchecking/elchecker.py
```diff
@@ -3,17 +3,63 @@
...
 from keylime import config, keylime_logging
 from keylime.failure import Component, Failure
-from keylime.mba.elchecking import tests
+from keylime.mba.elchecking import policies, tests
 
 logger = keylime_logging.init_logging("measured_boot")
 
 
-def evaluate_bootlog(
+def policy_load(policy_path: Optional[str] = None) -> str:
```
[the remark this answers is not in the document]
galmasi: This was the initial behavior of policy_load. Load from an argument or else load from location specified in [tenant]/mb_refstate. I didn't change this because of backward compatibility.

keylime/mba/mba.py
galmasi: I want these modules in the list because `_find_implementation` will scan that list to find functions to call. There might be nicer ways to implement this, but for now having an explicit list works.

keylime/mba/mba.py
galmasi: Yes. Documentation fail. Fixing.

maugustosilva:
@aplanas can we talk briefly on Slack CNCF? Thanks!

keylime/cloud_verifier_common.py
```diff
@@ -259,8 +260,7 @@ def prepare_get_quote(agent: Dict[str, Any]) -> Dict[str, Union[str, int]]:
 def process_get_status(agent: VerfierMain) -> Dict[str, Any]:
     has_mb_refstate = 0
     try:
-        mb_refstate = json.loads(cast(str, agent.mb_refstate))
-        if mb_refstate and mb_refstate.keys():
+        if mba.policy_is_valid(cast(str, agent.mb_refstate)):
```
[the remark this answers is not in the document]
galmasi: Bah the cast was needed -- I removed it and caused a `pyright` violation -- sqlalchemy strings are not strings. 
So I put the cast back.

keylime/mba/mba.py
aplanas: oh bummer :+1:

aplanas:
@galmasi I think that if you rebase, the failing test should disappear

PR #1412, opened 2023-06-22
Adding kubectl to tenant docker image for Kubernetes init jobs in helm charts

mheese:
This installs kubectl into the tenant docker image. This is a requirement for the Kubernetes init jobs in the helm charts.

See the related PR for the helm charts here: https://github.com/keylime/attestation-operator/pull/13

### DIFF
--- a/docker/release/tenant/Dockerfile.in
+++ b/docker/release/tenant/Dockerfile.in
@@ -1,4 +1,15 @@
 FROM _source_keylime_base:_version_ AS keylime_tenant
+
+# install latest stable kubectl version - required for Kubernetes init job in the helm charts
+RUN export GOARCH="$( uname -m | sed -e 's/x86_64/amd64/' -e 's/aarch64/arm64/' )" && \
+    export KUBERNETES_RELEASE="$( curl -L -s https://dl.k8s.io/release/stable.txt )" && \
+    curl -LO "https://dl.k8s.io/release/${KUBERNETES_RELEASE}/bin/linux/${GOARCH}/kubectl" && \
+    curl -LO "https://dl.k8s.io/${KUBERNETES_RELEASE}/bin/linux/${GOARCH}/kubectl.sha256" && \
+    echo "$(cat kubectl.sha256)  kubectl" | sha256sum --check && \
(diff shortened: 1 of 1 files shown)


maugustosilva:
In principle, I agree that having `kubectl` binary on the image is not a big ask. By the price of basically two `cur` commands we basically enable the use of `init` containers. All the alternatives to this requirement (`agents` require TLS certs after all) sound significantly more complex (including installing `kubectl` on-the-fly, which is a problem for air gapped deployments). 

The only thing I ask here is to have the architecture as a parameter, auto-discovered (something like `uname -m`) instead of hard-coded (I know we don't have everything architecture-independent currently, but we should strive to be in this position)

@ansasaki do you have any issues with this addition?

mpeters:
kubectl is tied to particular versions of k8s correct? So does this mean that the images will only run on certain versions?

mheese:
@mpeters there could be remote possibilities where this might not work in certain clusters, but usually kubectl compatibility between cluster versions is not an issue. I have not experienced any issues for this in a very very long time. If there are issues, we would run into them the same way if we would call the APIs programmatically. And then, the kubectl calls are to get and create Kubernetes secrets, which is part of the core API which is stable for a long time.

@maugustosilva @ansasaki good point about the architecture. I can make use of the `ARG TARGETARCH` here which gets set by docker automatically? Requires BuildKit though. Otherwise it's a couple of lines more code (`uname -m` would return `x86_64` but we need `amd64`, and `arm64` instead of `aarch64`).

Also, again as I mentioned in the meeting this morning: we _can_ make this simply another docker image altogether which could be based on the tenant image?

maugustosilva:
@mheese I am perfectly fine with `uname -m | sed -e 's/x86_64/amd64/' -e 's/aarch64/arm64/'` tbh :-)

I believe we should keep the number of images to a minimum for the time being, if @mpeters and/or @ansasaki approve it as is, I will merge

mpeters (approved):

mheese:
@maugustosilva made the changes as discussed and force-pushed the branch! :+1:

maugustosilva (approved):

PR #1416, opened 2023-06-29
Initial PR to add support for IDevID and IAK

Isaac-Matthews:
This is the first of a series of PRs that will enable the use of IDevIDs and IAKs as proposed in [enhancement 81.](https://github.com/keylime/enhancements/blob/master/81-IDevID_and_IAK_support.md)
The modifications to the Keylime repositories are:

**Keylime**
Columns for the IDevID and IAK are added to the database
The IAK and IDevID sent by the agent are processed
The AK is verified by the IAK

**Rust-Keylime**
Config options to enable IDevID and IAK generation and use are added, with choices on how to generate the IDevID and IAK
The IAK and IDevID are generated according to config options
The IAK is used to certify the AK in the registration process

The purpose of this PR is to begin enabling the use of IAK and IDevID in keylime, starting with their generation. Options for the server config are not yet added so the inclusion of IAK and IDevID are left as optional for the agent.
The changes are all backwards compatible. The use of the IAK to certify and verify the AK is to act as a simple early use case in the process. Down the road this will be changed to encompass all the options listed in the enhancement, and those options will be included in the Keylime config.

N.B.
The certify in Rust-keylime currently just uses the UUID as qualifying data. This should perhaps be changed to include some temporal data at some point.


keylime/tpm/tpm_main.py
```diff
@@ -40,11 +41,89 @@ def encryptAIK(uuid: str, ek_tpm: bytes, aik_tpm: bytes) -> Optional[Tuple[bytes
...
+        attestd, attestpath = tempfile.mkstemp()
+        with open(attestpath, 'wb') as f:
+            f.write(iak_attest)
+        digestd, digestpath = tempfile.mkstemp()
+        command = [
+            "tpm2_hash",
+            "-C",
+            "e",
+            "-g",
+            "sha256",
```
stefanberger, line 74: This shouldn't have to be hardcoded. I would think you should be able to find the TPM identifier for the hash that was used for creating the signature in some of the data structures involved here.

keylime/tpm/tpm_main.py
```diff
@@ -40,11 +41,89 @@ def encryptAIK(uuid: str, ek_tpm: bytes, aik_tpm: bytes) -> Optional[Tuple[bytes
...
+            "-o",
+            digestpath,
+            attestpath
+        ]
+        EXIT_SUCCESS: int=0
+        retDict = cmd_exec.run(cmd=command, expectedcode=EXIT_SUCCESS, raiseOnError=False)
+
+        # process iak pub key and import into tpm, get pub key context from tpm
+        iakpub = tpm2_objects.pubkey_from_tpm2b_public(iak_tpm)
```
stefanberger, line 83: You should call this function here for signature verification:

https://github.com/keylime/keylime/blob/master/keylime/tpm/tpm_util.py#L23

like called from here:

https://github.com/keylime/keylime/blob/master/keylime/tpm/tpm_util.py#L159

The iakpub is the public key that can be used as 1st parameter to verify().

The signature (2nd) parameter is probably what you have as `iak_sign`. It should be 256bytes long for an RSA2048 key.

The you need to calculate the hash over what you want to verify and pass the hash (bytes) as 3rd parameter.

The forth parameter is you hash_func that you can get like this:

https://github.com/keylime/keylime/blob/master/keylime/tpm/tpm_util.py#L148

The hash_alg is the TPM identifier of the hash that you should be able to extract from one of these data  structures here. 0xb represents SHA256.

stefanberger:
How far does the certificate chain of the IAK go so we can identify the device as a TPM that we trust for quoting valid PCR states rather than a coffee machine that can fake quotes and PCR states?

Isaac-Matthews:
The IAK goes back to the OEM of the platform.

stefanberger:
> The IAK goes back to the OEM of the platform.

I suppose these certs are different from TPM manufacturers' certs, so we would have to have them as well in a directory so we can check the chain. However, we still have to be sure that it's a real TPM that we can trust and not the crypto-enabled coffee machine.

Isaac-Matthews:
Yes there will need to be certs for the TPM manufacturer and the platform manufacturer for the EK and IAK respectively. The IAK should have been ensured to be colocated in the same tpm as the EK by the platform manufacturer so the guarantee that it is a genuine tpm is still achieved using the IAK but the proposal uses EK as well as IAK.

stefanberger:
> The changes are all backwards compatible. The use of the IAK to certify and verify the AK is to act as a simple early use case in the process.

And the IAK will be verified as well to be originating in a TPM?

keylime/registrar_common.py
```diff
@@ -266,6 +266,24 @@ def do_POST(self) -> None:
...
+            # Check if IDevID and IAK are received
+            # If not it is reported with logger info, but this behaviour will change when configuration requirements are added
+            idevid_received = False
+            try:
+                iak_tpm = json_body["iak_tpm"]
+                idevid_tpm = json_body["idevid_tpm"]
+                iak_attest = base64.b64decode(json_body["iak_attest"])
+                iak_sign_raw = json_body["iak_sign"]
+                iak_sign = base64.b64decode(iak_sign_raw)
+                if idevid_tpm == "AA==":
```
THS-on, line 280: Why is the separate check required? If there is no idevid this field just should be None

keylime/tpm/tpm_main.py
Isaac-Matthews: Thanks for the advice, I have removed the hard coded algorithm and removed all use of cli tpm2_tools from the pr in the most recent commit.

keylime/registrar_common.py
Isaac-Matthews: The difference is if an old agent is being used this field doesn't exist, but if a new version is being used but idevid is disabled this field is '0'. I can change this from the agent side and then remove this check if that would be preferable.

Isaac-Matthews: I have just committed a change to the agent to remove sending the empty fields so I have removed the checks in the newest commit.

keylime/tpm/tpm_util.py
```diff
@@ -20,13 +21,49 @@
...
 
-def verify(pubkey: SupportedKeyTypes, sig: bytes, digest: bytes, hashfunc: hashes.HashAlgorithm) -> None:
+def verify(pubkey: SupportedKeyTypes, sig: bytes, digest: bytes, hashfunc: hashes.HashAlgorithm, sigalg=tpm2_objects.TPM_ALG_RSASSA, saltlen=0) -> None:
     """Do signature verification with the given public key"""
     if isinstance(pubkey, RSAPublicKey):
-        pubkey.verify(sig, digest, padding.PKCS1v15(), Prehashed(hashfunc))
+        if sigalg == tpm2_objects.TPM_ALG_RSAPSS:
+            pubkey.verify(sig, digest, padding.PSS(mgf=padding.MGF1(hashfunc), salt_length=saltlen), Prehashed(hashfunc))
+        if sigalg == tpm2_objects.TPM_ALG_RSASSA:
+            pubkey.verify(sig, digest, padding.PKCS1v15(), Prehashed(hashfunc))
```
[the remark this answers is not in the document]
stefanberger: Looks better now but before we invest more time in this ... I am still wondering whether we can use IDevId/IAK to verify that the AIK is indeed from a TPM, which depends on the certificate chain and for what the involved CA vouches. If we cannot verify that our trust is rooted in a TPM then we may open up Keylime to untrusted devices that just make up PCR states, quotes, and logs.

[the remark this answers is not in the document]
Isaac-Matthews: From my understanding we are able to do this. From the TPM keys for device identity doc the procedure for the creation of the IAK and IDevID provides assurances that:
1.  The TPM is authentic and represented by the valid EK cert
2. The IAK is co-resident in the TPM identified by the EK
3.  The IDevID is in the same TPM as the IAK
With those assurances, I think we can demonstrate that the AK resides in the same TPM as the IAK/IDevID using certify/verify.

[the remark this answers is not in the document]
stefanberger: > From my understanding we are able to do this. From the TPM keys for device identity doc the procedure for the creation of the IAK and IDevID provides assurances that:
> 
>     1. The TPM is authentic and represented by the valid EK cert
> 
>     2. The IAK is co-resident in the TPM identified by the EK
> 
>     3. The IDevID is in the same TPM as the IAK
> 
> 
> With those assurances, I think we can demonstrate that the AK resides in the same TPM as the IAK/IDevID using certify/verify.

As long as some well-known certificate assures that these steps were taken as described ...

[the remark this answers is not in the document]
stefanberger: For  reference the procedure with IDevID and IAK is described in this document here: https://trustedcomputinggroup.org/wp-content/uploads/TCG_IWG_DevID_v1r2_02dec2020.pdf

The CA mentioned in section 4 should be publicly available and we want to see its certificate and this CA must vouch for the procedure and key properties you mentioned above. This is similar to the CA certificates of TPM manufacturers we have collected in this project that vouch for the EK certificates they minted are in fact related to EK keys in the TPM. [ We just added an Alibaba CA cert that presumably is used to certify vTPMs EK certs. ]

Another procedure would be to have an LDevID and LAK (sect 5.5) with their certificates created by your local IT department for example and their CA vouches for the procedure and key properties you mentioned. Again, we also want to see this certificate and someone would add the IT department's (LDdevID/LAK issuing) CA locally to the collection of certificates,  though maybe in a different directory than where the TPM manufacturers' CA certs are.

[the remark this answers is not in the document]
stefanberger: I think you are right. Since you are still pulling the EK certificate, this is all good.

stefanberger:
@Isaac-Matthews I think a fair question is whether your future PRs going to go all the way and

  1. provide IAK and EK certs (for various vendors via some sort of plug-in arch on the agent) from the rust agent side
  2. provide mechanisms to verify the CAs for IAKs and IDevIDs
  3. get rid of the transfer of EKs and EKcerts since they are then not needed anymore
  4. validate the certificate for the IDevID and IAKs to ensure that they are from a TPM (SAN)
  5. use the IDevID key in the TLS endpoint connection?
  6. ....

So far this PR only shows a small part of the story.
I think particularly 5 moves the needle for assurance that we talk to the proper endpoint.

Isaac-Matthews:
@stefanberger  yes we still plan to fully implement [enhancement 81.](https://github.com/keylime/enhancements/blob/master/81-IDevID_and_IAK_support.md)

IDevID and IAK certs will be collected by the agent and then sent to and verified by the registrar. This PR shows the creation and use of the keys, the next will add trust to the keys by incorporating their certificates enabling their use in place of the EK for registration. 

5 is out of scope for this enhancement but could be put into an enhancement request.

stefanberger:
> @stefanberger yes we still plan to fully implement [enhancement 81.](https://github.com/keylime/enhancements/blob/master/81-IDevID_and_IAK_support.md)
> 
> IDevID and IAK certs will be collected by the agent and then sent to and verified by the registrar. This PR shows the creation and use of the keys, the next will add trust to the keys by incorporating their certificates enabling their use in place of the EK for registration.
> 
> 5 is out of scope for this enhancement but could be put into an enhancement request.

Once we end up adding lots of code for IAK/IDevID/LAK/LDevID etc. support this could still turn out to be the most important part actually that ties TLS endpoint to the same platform with the same  TPM where the logs and quotes are coming from.

"Using IDevID, provides important security guarantees to users. It assures them the device is genuine and one that they own (via the serial number in the IDevID). In many cases the devices will be geographically distributed in remote or branch offices, for example. So securely identifying the specific device is important."

Is 'availability of a serial number' going to play a role in the implementation or how are existing procedures going to change because of IDevID? Similar statements could be made about the Ek and EK cert and platform cert as well just that it's not a serial number but the EK that identifies the TPM.

maugustosilva:
@Isaac-Matthews Please don't forget to fix the `mypy` errors. Thanks!

maugustosilva (approved):

maugustosilva:
@Isaac-Matthews I am ready to merge, a very welcome functional addition, and awaiting with anticipation the subsequent PRs! I do have a couple of final requests before the merge: 1) fix the (very simple) `isort` error and please squash your commits. Thanks for your patience :-).

maugustosilva:
@Isaac-Matthews something wrong with the tests, it looks like there is a conflict. Could please rebase and push again? As soon as the tests pass, I will merge, thanks.

PR #1419, opened 2023-07-07
Several improvements for the "create_runtime_policy.sh" script

maugustosilva:
In addition to introducing parametrization for all paths, there is now the possibility for the user to supply an exclude list (-e/--exclude_list), which will be automatically incorporated into the policy

Usage: /KL/keylime/scripts/create_runtime_policy.sh -o/output_file FILENAME [-a/--algo ALGO] [-x/--ramdisk-location PATH] [-y/--boot_aggregate-location PATH] [-z/--rootfs-location PATH] [-e/--exclude_list FILENAME] [-s/--skip-path PATH]" optional arguments:
    -a/--algo (default sha1sum)
    -x/--ramdisk-location (default /boot/, set to "none" to skip)
    -y/--boot_aggregate-location (default
/sys/kernel/security/ima/ascii_runtime_measurements, set to "none" to skip)
    -z/--rootfs-location (default /, cannot be skipped)
    -e/--exclude_list (default none)
    -s/--skip-path (default none)

Not extensively tested (and AFAIK, not part of the testing CI)

### DIFF
--- a/scripts/create_runtime_policy.sh
+++ b/scripts/create_runtime_policy.sh
@@ -1,8 +1,23 @@
-#!/bin/bash
+#!/usr/bin/env bash
 ################################################################################
 # SPDX-License-Identifier: Apache-2.0
 # Copyright 2017 Massachusetts Institute of Technology.
 ################################################################################
+
+function announce {
+    # 1 - MESSAGE
+
+	MESSAGE=$(echo "${1}" | tr '\n' ' ')
+	MESSAGE=$(echo $MESSAGE | sed "s/\t\t*/ /g")
+
+	echo "==> $(date) - ${0} - $MESSAGE"
+}
+
+function valid_algo {
+        local algo=$1
+
+        [[ " ${ALGO_LIST[@]} " =~ " ${algo} " ]]
+}
 
 # Configure the installer here
 INITRAMFS_TOOLS_GIT=https://salsa.debian.org/kernel-team/initramfs-tools.git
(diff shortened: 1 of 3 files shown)


maugustosilva:
@Koncpa @kkaarreell I believe there is a problem with fedora-rawhide. This is the second PR where this particular test keeps failing consistently with ` Failure during Ansible playbook execution`

aplanas:
I was under the impression that this should now be consolidated as part of [create_policy.py](https://github.com/keylime/keylime/blob/master/keylime/cmd/create_policy.py).  In fact, it is there where I added the code to create the policy based on the remote repo information (and avoid the analysis of the local system)

maugustosilva:
@aplanas This is a very good question. The documentation still refers to `scripts/create_runtime_policy.sh` and as it was the issue which motivated this PR, first-time users, researchers and students (among others) still derive some value from the ability to create an allowlist from the analysis of the local filesystem. What I did here was to make sure, given that the script exists and is documented, it operates correctly.

I am not opposed to remove it altogether, although in my experience, having a tool that produces a list of checksums by analyzing the filesystem has been proven useful (including in production environments).

THS-on:
@aplanas  yes, ideally we should re-implement the features of `create_runtime_policy.sh` in `create_policy.py`, so that we only have one tool. For now the fixes I think are fine, but we should not start extending the shell script.

aplanas:
> I am not opposed to remove it altogether, although in my experience, having a tool that produces a list of checksums by analyzing the filesystem has been proven useful (including in production environments).

I agree with you that the local analysis is very important to bootstrap the tool, so the feature should not removed.

scripts/create_runtime_policy.sh
```diff
@@ -1,138 +1,254 @@
...
+        shift
 done
 
-if [ ! "$OUTPUT" ]
+if valid_algo $ALGO
 then
-    echo "Missing argument for -o" >&2;
-    echo "Usage: $0 -o [filename] -h [hash-algo]" >&2;
+    true
+else
```
aplanas, line 152: You can skip the else with `if ! valid_algo $ALGO; then`

scripts/create_runtime_policy.sh
```diff
@@ -1,138 +1,254 @@
...
 # Convert to runtime policy
-echo "Converting created allowlist to Keylime runtime policy"
-python3 $WORKING_DIR/../keylime/cmd/convert_runtime_policy.py -a $OUTPUT -o $OUTPUT
+announce "Converting created allowlist to Keylime runtime policy"
+CONVERT_CMD_OPTS="--allowlist ${OUTPUT}_allowlist --output_file $OUTPUT"
+ls $EXCLUDE_LIST  > /dev/null 2>&1
+if [[ $? -eq 0 ]]
+then
+    CONVERT_CMD_OPTS=$CONVERT_CMD_OPTS" --excludelist $EXCLUDE_LIST"
+fi
```
aplanas, line 249: `[ -f $EXCLUDE_LIST ] && CONVERT_CMD_OPTS="$CONVERT_CMD_OPTS --excludelist $EXCLUDE_LIST"`

aplanas (approved):
Once the algorithm type is fixed, LGTM

PR #1421, opened 2023-07-11
Non-blocking Registrar SSL socket

flozilla:
This PR fixes an issue in the Keylime Registrar that allows performing a DoS by opening a single connection to its HTTPS interface. As long as the connection is open, the Register is blocked and cannot serve any further clients, which prevents the Register from working as expected.

The issue happens as Python SSL sockets block by default, unlike regular sockets. To fix this, a non-blocking SSL handshake is implemented. The fix is inspired by Python's documentation on non-blocking SSL sockets (https://docs.python.org/3/library/ssl.html#notes-on-non-blocking-sockets).

A cleaner solution would be using the tornado framework in the Registrar (like in the Verifier). However, this requires several changes to the Registrar code.

### DIFF
--- a/keylime/registrar_common.py
+++ b/keylime/registrar_common.py
@@ -4,6 +4,7 @@ import ipaddress
 import os
 import signal
 import socket
+import ssl
 import sys
 import threading
 from http.server import BaseHTTPRequestHandler, HTTPServer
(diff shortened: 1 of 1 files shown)


aplanas (approved):
That is a good one!

Can you sing the commit (to fix the DCO test case)?

stefanberger (approved):

flozilla:
Added sign-off.

Thanks for making me aware of the failed DCO test case @aplanas!

maugustosilva:
This PR fixes CVE-2023-38200

ansasaki:
@aplanas Hi, did you test this? Apparently it is missing some imports (`select` and `sock` are undefined)

PR #1422, opened 2023-07-11
tpm_util: Remove useless comparison of always identical hashes

stefanberger:


mheese (approved):
@stefanberger that's how I'm seeing this too. However, the raised exception gives you the feeling that somebody truly wanted to compare something of importance here. So I'm wondering if the 2nd digest which looks redundant was supposed to be calculating over something else? I truly don't see though what else this could possibly be.

Approved IMHO, but I would like to see somebody else with more knowledge of that part of the code base to comment/approve.

stefanberger:
> @stefanberger that's how I'm seeing this too. However, the raised exception gives you the feeling that somebody truly wanted to compare something of importance here. So I'm

That somebody was 'me'. From what I can see there's nothing else to compare. The nonce has been validated that it was used for the quote (it's inside the quoteblob) and the hash over the quoted PCRs' values has been compared against what was quoted (it's also inside the quoteblob; so now the PCRs in pcrblob should not be fictitious) and the signature is over the quoteblob's hash.

A real problem is solved in the next PR...

mheese:
> > @stefanberger that's how I'm seeing this too. However, the raised exception gives you the feeling that somebody truly wanted to compare something of importance here. So I'm
> 
> That somebody was 'me'. From what I can see there's nothing else to compare. The nonce has been validated that it was used for the quote (it's inside the quoteblob) and the hash over the quoted PCRs' values has been compared against what was quoted (it's also inside the quoteblob; so now the PCRs in pcrblob should not be fictitious) and the signature is over the quoteblob's hash.
> 
> A real problem is solved in the next PR...

@stefanberger link me on that please, as I'm implementing the `checkquote()` functionality in Golang right now for the Kubernetes controller, I want to know what needs fixing ... I essentially ported the python code

aplanas (approved):

PR #1423, opened 2023-07-11
tpm_util: Replace a logger.error with an Exception in case of invalid…

stefanberger:
… signature

This fixes a possibly severe issue in 7.2.5 & 7.3.0.

### DIFF
--- a/keylime/tpm/tpm_util.py
+++ b/keylime/tpm/tpm_util.py
@@ -3,7 +3,6 @@ import string
 import struct
 from typing import Any, Dict, List, Optional, Tuple, Union
 
-from cryptography.exceptions import InvalidSignature
 from cryptography.hazmat import backends
 from cryptography.hazmat.primitives import hashes, hmac, serialization
 from cryptography.hazmat.primitives.asymmetric import ec, padding
@@ -155,10 +154,7 @@ def checkquote(
     digest.update(quoteblob)
     quote_digest = digest.finalize()
 
-    try:
-        verify(pubkey, signature, quote_digest, hashfunc)
-    except InvalidSignature:
-        logger.error("Invalid quote signature!")
+    verify(pubkey, signature, quote_digest, hashfunc)
 
     # Check that reported nonce is expected one
     retDict = tpm2_objects.unmarshal_tpms_attest(quoteblob)
--- a/keylime/tpm/tpm_util_test.py
+++ b/keylime/tpm/tpm_util_test.py
@@ -2,6 +2,7 @@ import base64
 import unittest
 from unittest import mock
 
+from cryptography.exceptions import InvalidSignature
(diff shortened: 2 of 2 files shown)


stefanberger:
@mheese This is the more severe issue.

maugustosilva (approved):

keylime/tpm/tpm_util.py
```diff
@@ -157,8 +157,8 @@ def checkquote(
 
     try:
         verify(pubkey, signature, quote_digest, hashfunc)
-    except InvalidSignature:
-        logger.error("Invalid quote signature!")
+    except InvalidSignature as e:
+        raise Exception("Invalid quote signature!") from e
```
aplanas, line 161: At this point I wonder why to catch an specific exception to throw a generic one. Maybe is better to drop the try all together and bubble up the original exception

keylime/tpm/tpm_util_test.py
```diff
@@ -60,6 +60,33 @@ def test_checkquote(self) -> None:
         except Exception as e:
             self.fail(f"checkquote failed with {e}")
 
+        try:
+            # flip bit
+            quoteblob[5] ^= 0x1
+            checkquote(aikblob, nonce, sigblob, quoteblob, pcrblob, "sha256")
+            self.fail(f"checkquote did not fail signature verification as expected")
```
aplanas, line 67: This assert is more clear: https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises

keylime/tpm/tpm_util.py
stefanberger: Fixed.

keylime/tpm/tpm_util_test.py
stefanberger: Thanks. Fixed.

aplanas (approved):
The lint complain that now the import for InvalidSignature should be removed, but beside that LGTM

stefanberger:
Do we want to open branches for 7.3.x and 7.2.5.x to apply this patch to ?

stefanberger:
This patch fixes CVE-2023-3674.

aplanas:
> CVE-2023-3674

Is this ID correct? Is the CVE public or is under embargo? In the last case I think should not be published after the release?

stefanberger:
> > CVE-2023-3674
> 
> Is this ID correct? Is the CVE public or is under embargo? In the last case I think should not be published after the release?

This is the ID I received in the email. I know, compared to the 'other' one it is a much lower number.

ansasaki:
> > CVE-2023-3674
> 
> Is this ID correct? Is the CVE public or is under embargo? In the last case I think should not be published after the release?

Yes, let's please stop disclosing security issues.

stefanberger:
Delete my messages.

PR #1431, opened 2023-07-20
installer.sh: support Anolis OS whose ID is anolis

Jingshui1037:
This patch fix the problem that anolis is not currently supported when using the ./installer.sh, so add the installer scripts about anolis.

### DIFF
--- a/installer.sh
+++ b/installer.sh
@@ -128,17 +128,25 @@ case "$ID" in
         esac
     ;;
 
-    fedora)
+    fedora | anolis)
         echo "${ID} selected."
         PACKAGE_MGR=$(command -v dnf)
         PYTHON_PREIN="python3 python3-devel python3-setuptools git wget patch"
-        PYTHON_DEPS="python3-pip gcc gcc-c++ openssl-devel swig python3-pyyaml python3-zmq python3-cryptography python3-tornado python3-requests python3-gpg yaml-cpp-devel procps-ng python3-psutil python3-lark-parser python3-alembic"
+        PYTHON_DEPS="python3-pip gcc gcc-c++ openssl-devel python3-pyyaml python3-zmq python3-cryptography python3-tornado python3-requests python3-gpg yaml-cpp-devel procps-ng python3-psutil python3-lark-parser python3-alembic"
         if [ "$(uname -m)" = "x86_64" ]; then
             PYTHON_DEPS+=" efivar-devel"
         fi
         BUILD_TOOLS="openssl-devel libtool make automake pkg-config m4 libgcrypt-devel autoconf autoconf-archive libcurl-devel libstdc++-devel uriparser-devel dbus-devel gnulib-devel doxygen libuuid-devel json-c-devel"
-        if [[ ${VERSION_ID} -ge 30 ]] ; then
-        # if fedora 30 or greater, then using TPM2 tool packages
+        if [ "${ID}" = "anolis" ]; then
+            pip3 install setuptools_rust
+            BUILD_TOOLS+=" rust cargo"
+        elif [ "${ID}" = "fedora" ]; then
+            PYTHON_DEPS+=" swig"
+        fi
+
+        MAJOR_VERSION_ID=$(echo ${VERSION_ID} | awk -F '.' '{print$1}')
+	if ([[ "${ID}" = "fedora" ]] && [[ "${MAJOR_VERSION_ID}" -ge 30 ]]) || ([[ "${ID}" = "anolis" ]] && [[ "${MAJOR_VERSION_ID}" -ge 23 ]]) ; then
+        # if fedora 30/anolis 23 or greater, then using TPM2 tool packages
             TPM2_TOOLS_PKGS="tpm2-tools tpm2-tss tss2"
             NEED_BUILD_TOOLS=0
         else


aplanas:
Can you describe what is anolis? Google says that Anolis OS is a distribution, but I am not able to find more information : )

Jingshui1037:
OK @aplanas . OpenAnolis community (https://openanolis.cn/anolisos?lang=en and https://github.com/openanolis ) is a non-profit open-source community established by enterprises and institutions, colleges, scientific research units, non-profit organizations and individuals in China.

 Anolis OS 8 ( https://openanolis.cn/?lang=en ) is an OS product about OpenAnolis and it is compatible with Redhat 8.

aplanas:
> it is compatible with Redhat 8.

Ah thanks. I see that the code is also similar to Fedora. Do you think that can be integrated in the Fedora case?

Jingshui1037:
> > it is compatible with Redhat 8.
> 
> Ah thanks. I see that the code is also similar to Fedora. Do you think that can be integrated in the Fedora case?

OK，I'll have a try.

Jingshui1037:
Hi @aplanas ，an anolis OS has a version like 8.8. When it is integrated in the Fedora case, it is an error like this 

image

image

so I think may be it is better to put anolis in a  alone case.

aplanas:
> so I think may be it is better to put anolis in a alone case.

The showed error suggest that there is indeed a bug in the code. Most probably a missing quote here https://github.com/keylime/keylime/blob/master/installer.sh#L140 and in some other place that introduced your change (I do not see the code that you used, so is hard to diagnose, maybe you can show the code)

IMHO those errors are not a reason to add anolis as a independent case, but maybe other reviewers have different opinions

Jingshui1037:
Hi @aplanas, 
I finished the modification according to your suggestion. Could you review it again? Thanks.

aplanas (approved):

aplanas:
LGTM!

THS-on:
If you are mainly based on RHEL 8 and use their numbering scheme it probably makes more sense to merge it with the "rhel | centos" section, than the fedora one.

Jingshui1037:
> If you are mainly based on RHEL 8 and use their numbering scheme it probably makes more sense to merge it with the "rhel | centos" section, than the fedora one.

Hi @THS-on , 
      Anolis is also compatible with Fedora. I already incorporated anolis into the Fedora case according to the suggestion of @aplanas,  I think it is enough for anolis. In addition, the next version of anolis (After anolis 8) is no longer anolis 9 but anolis 23, so I think it is better for us to put anolis into the fedora case.

installer.sh
```diff
@@ -128,16 +128,24 @@ case "$ID" in
...
-        if [[ ${VERSION_ID} -ge 30 ]] ; then
+        if [ "${ID}" = "anolis" ]; then
+            pip3 install setuptools_rust
+            BUILD_TOOLS+=" rust cargo"
+        elif [ "${ID}" = "fedora" ]; then
+            PYTHON_DEPS+=" swig"
+        fi
+
+        MAJOR_VERSION_ID=$(echo ${VERSION_ID} | awk -F '.' '{print$1}')
+        if [[ ${MAJOR_VERSION_ID} -ge 30 ]] ; then
```
THS-on, line 148: this also is probably fedora specific

Jingshui1037: At present, this judgment condition can also be met by anolis (because anolis currently doesn't have a version above 30), if this condition is not met in the future, a new judgment condition will be added.

ansasaki: Better fix now than when it breaks and another PR is necessary.

Jingshui1037:
Hi @THS-on @ansasaki  ,
I finished the modification according to your suggestion. Could you review it again? Thanks.

THS-on (approved):

installer.sh
```diff
@@ -128,17 +128,25 @@ case "$ID" in
...
-        # if fedora 30 or greater, then using TPM2 tool packages
+        if [ "${ID}" = "anolis" ]; then
+            pip3 install setuptools_rust
+            BUILD_TOOLS+=" rust cargo"
+        elif [ "${ID}" = "fedora" ]; then
+            PYTHON_DEPS+=" swig"
+        fi
+
+        MAJOR_VERSION_ID=$(echo ${VERSION_ID} | awk -F '.' '{print$1}')
+        if [[ "${ID}" = "fedora" ]] && [[ "${MAJOR_VERSION_ID}" -ge 30 ]] || [[ "${ID}" = "anolis" ]] && [[ "${MAJOR_VERSION_ID}" -ge 23 ]] ; then
```
stefanberger, line 148:
```
$ if [[ "foo" = "foo" ]] && [[ 30 -ge 30 ]] || [[ "bar" = "bar" ]] && [[ 20 -ge 23 ]] ; then echo "yes"; fi
$ if ([[ "foo" = "foo" ]] && [[ 30 -ge 30 ]]) || ([[ "bar" = "bar" ]] && [[ 20 -ge 23 ]]) ; then echo "yes"; fi
yes
```
==> add `()` to the tests!

stefanberger (changes requested):

installer.sh
Jingshui1037: Thanks @stefanberger , I fixed it.

Jingshui1037: OK， Thanks @stefanberger ，I fixed it.

stefanberger (approved):

PR #1433, opened 2023-07-20
mba: Manage the number of measured boot attestation

niteeshkd:
It allows to control the number of measured boot policy evaluation.

### DIFF
--- a/keylime/tpm/tpm_main.py
+++ b/keylime/tpm/tpm_main.py
@@ -163,6 +163,7 @@ class Tpm:
         mb_measurement_list: Optional[str],
         mb_refstate: Optional[str],
         hash_alg: Hash,
+        count: int,
     ) -> Failure:
         failure = Failure(Component.PCR_VALIDATION)
 
@@ -335,13 +336,32 @@ class Tpm:
             failure.add_event("missing_pcrs", {"context": "PCRs are missing in quote", "data": list(missing)}, True)
 
         if not mb_failure and mba.policy_is_valid(mb_refstate):
-            mb_policy_failure = mba.bootlog_evaluate(
-                mb_refstate,
-                mb_measurement_data,
-                pcrs_in_quote,
-                agentAttestState.get_agent_id(),
-            )
-            failure.merge(mb_policy_failure)
+            mb_evaluate = config.get("verifier", "measured_boot_evaluate", fallback="once")
+
+            # Value of measured_boot_evaluate can be only 'once' or 'always'
+            if mb_evaluate not in ("once", "always"):
+                logger.error("Invalid value %s of measured_boot_evaluate", mb_evaluate)
+                failure.add_event(
+                    "invalid_measured_boot_evaluate", "correct value of measured_boot_evaluate is required", True
+                )
+
+            if mb_evaluate == "always":
+                mb_policy_failure = mba.bootlog_evaluate(
+                    mb_refstate,
+                    mb_measurement_data,
+                    pcrs_in_quote,
+                    agentAttestState.get_agent_id(),
+                )
+                failure.merge(mb_policy_failure)
+
+            elif mb_evaluate == "once" and count == 0:
+                mb_policy_failure = mba.bootlog_evaluate(
+                    mb_refstate,
+                    mb_measurement_data,
+                    pcrs_in_quote,
+                    agentAttestState.get_agent_id(),
+                )
+                failure.merge(mb_policy_failure)
 
         return failure
 
@@ -360,6 +380,7 @@ class Tpm:
         mb_measurement_list: Optional[str] = None,
         mb_refstate: Optional[str] = None,
         compressed: bool = False,
+        count: int = -1,
     ) -> Failure:
         if tpm_policy is None:
             tpm_policy = {}
@@ -408,6 +429,7 @@ class Tpm:
             mb_measurement_list,
             mb_refstate,
             hash_alg,
+            count,
         )
 
     @staticmethod
--- a/templates/2.0/verifier.j2
+++ b/templates/2.0/verifier.j2
@@ -158,6 +158,15 @@ measured_boot_policy_name = {{ verifier.measured_boot_policy_name }}
 # The default value for this config item is the empty list.
 measured_boot_imports = {{ verifier.measured_boot_imports }}
 
+# This is used to manage the number of times measure boot attestation
+# is done. In other words, it controls the number of times the call
+# to the measure boot policy engine is made to evaluate the boot log
+# against the policy specified.
+# Here are its possible values and number of bootlog evaluations.
+# once (default)  : Bootlog evaluation will be done for only one time.
+# always          : Bootlog evaluation will always be done (i.e. for unlimited times).
+measured_boot_evaluate = {{ verifier.measured_boot_evaluate }}
+
 # Severity labels for revocation events strictly ordered from least severity to
 # highest severtiy.
 severity_labels = {{ verifier.severity_labels }}
--- a/keylime.conf
+++ b/keylime.conf
@@ -321,6 +321,15 @@ measured_boot_policy_name = accept-all
 # The default value for this config item is the empty string.
 # measured_boot_imports = keylime.mba.elchecking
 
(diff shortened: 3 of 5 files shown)


templates/2.0/verifier.j2
```diff
@@ -158,6 +158,9 @@ measured_boot_policy_name = {{ verifier.measured_boot_policy_name }}
 # The default value for this config item is the empty list.
 measured_boot_imports = {{ verifier.measured_boot_imports }}
 
+# This is used to control the number of measured boot policy evaluation.
```
aplanas, line 161: The description here in the template is different from the one in the keylime.conf

keylime/tpm/tpm_main.py
```diff
@@ -335,13 +335,30 @@ def check_pcrs(
...
+            if mb_eval_count < 0:
+                mb_policy_failure = mba.bootlog_evaluate(
+                    mb_refstate,
+                    mb_measurement_data,
+                    pcrs_in_quote,
+                    agentAttestState.get_agent_id(),
+                )
+                failure.merge(mb_policy_failure)
+
+            # Call bootlog_evalaute() as long as the current value of mb_evaluate_count is less
```
aplanas, line 351: typo: evalaute

keylime/tpm/tpm_main.py
```diff
@@ -335,13 +335,30 @@ def check_pcrs(
...
+            #  than the configured value of measured_boot_evalaute_count.
+            elif mb_eval_count > agentAttestState.get_mb_evaluate_count():
```
aplanas, line 353: same typo

keylime/agentstates.py
```diff
@@ -194,6 +197,14 @@ def get_ima_dm_state(self) -> Optional[bytes]:
     def set_ima_dm_state(self, state: bytes) -> None:
         self.ima_dm_state = state
 
+    def get_mb_evaluate_count(self) -> int:
+        """Get measured boot evaulation count"""
+        return self.mb_evaluate_count
+
+    def update_mb_evaluate_count(self) -> None:
```
aplanas, line 204: nit: maybe a better name is `increment_mb_evaluate_count`

aplanas (commented):
Something that I do not understand: why would want to attest the measure boot 5 times only, for example?

THS-on:
I agree with @aplanas here that it probably only makes sense to evaluate the log once or every time. What we could do is to evaluate the log once, save the computed PCR values and only assert that they don't change between quotes. This way the policy engine only runs once, but we still fully validate all the PCRs in the quote.

stefanberger:
> Something that I do not understand: why would want to attest the measure boot 5 times only, for example?

You may want to do this because it doesn't change anymore. At least I haven't seen a system where the measured boot log still changes after Linux has taken over.

aplanas:
> You may want to do this because it doesn't change anymore.

That is clear, but this invariant is not proved when testing 5 times, or 8 times, or 114 times. The idea is that only three values makes sense to me: testing 0 time, testing 1 time or always testing.

maugustosilva:
> I agree with @aplanas here that it probably only makes sense to evaluate the log once or every time. What we could do is to evaluate the log once, save the computed PCR values and only assert that they don't change between quotes. This way the policy engine only runs once, but we still fully validate all the PCRs in the quote.

This is the design philosophy that I would like to see in Keylime as whole. Keep the PCRs in the quote, ship it to `verifier`, validate it (via log replay) every time, but let the user decide when (how many times) will the policy engine be called.

maugustosilva:
> > You may want to do this because it doesn't change anymore.
> 
> That is clear, but this invariant is not proved when testing 5 times, or 8 times, or 114 times. The idea is that only three values makes sense to me: testing 0 time, testing 1 time or always testing.

@aplanas I generally agree that evaluating the MB log "N" time does not make much sense (I cannot think of a case where it would). 

However, my practical consideration is, given the cases which do "make sense" (0, 1, "infinity") are controlled via an attribute mapped to an integer, why not simply allow an user to specify "nonsensical" (but harmless) values if he chooses so? We could certainly have extra code to enforce the set of values, but I do wonder if we couldn't simply skip it.

stefanberger:
> I agree with @aplanas here that it probably only makes sense to evaluate the log once or every time. What we could do is to evaluate the log once, save the computed PCR values and only assert that they don't change between quotes. This way the policy engine only runs once, but we still fully validate all the PCRs in the quote.

It's the same then as incremental attestation that we have for IMA except that the next entry/entries for MBA never show up... Yes, we should make sure that the PCRs don't change.

niteeshkd:
@stefanberger , @THS-on , @aplanas & @maugustosilva , thanks a lot for your valuable suggestions. I have modified it slightly but kept the mechanism same to include the call to the policy engine (for evaluation of boot log against the policy specified) '0' time, '1' time or unlimited times.

aplanas:
> '0' time, '1' time or unlimited times.

I do not see that in the code nor in the documentation. Am I misreading?

niteeshkd:
> > '0' time, '1' time or unlimited times.
> 
> I do not see that in the code nor in the documentation. Am I misreading?

The bootlog will be evaluated '0', '1' and 'unlimited' times if the configurable item measured_boot_evaluate_count (defined in keylime.conf) is '0', '1'(default) and '<0' respectively.

aplanas:
> > I do not see that in the code nor in the documentation. Am I misreading?

> The bootlog will be evaluated '0', '1' and 'unlimited' times if the configurable item measured_boot_evaluate_count (defined in keylime.conf) is '0', '1'(default) and '<0' respectively.

Sorry, maybe I am blind. This is not what I see in the current code:

1) The documentation of the feature (in the template and in keylime.conf) explain the old behavior
  https://github.com/keylime/keylime-vagrant-ansible-tpm-emulator/issues/74
  https://github.com/keylime/keylime/pull/1433/files#diff-3ada17b9664265cd2f3c4760d9e7f0869ec9bf020a1f7839be2ad378710ba016R164

2) The code still calls  `increment_mb_evaluate_count` and compares with `get_mb_evaluate_count`
  https://github.com/keylime/keylime/pull/1433/files#diff-a5eab463887b7b666a404889fb335a379949c1e4912b164fecff8c77e7fb8fbaR360

Maybe you forgot to push the new changes?

maugustosilva:
@aplanas I just had a quick chat with @niteeshkd, and just wanted to confirm that maybe we have a potential misunderstanding here: he did **not** add any code to do "parameter value critical evaluation", limiting to values in the set [0, 1, -1]. He just mentioned that the code, as is, does accept these three values in the way we discussed, but it does not have any mechanism to detect values outside of the range and prevent the setting of a "nonsensical" value, like another integer higher than 1.

If you firmly believe that an error or warning message is in order, I am not opposed to it, but I do offer that it will be 3-4 extra lines of code which are not strictly necessary from a functional standpoint.

niteeshkd:
Added/modified the comments for clarity.

aplanas:
@niteeshkd thanks updating the comments in the template, this will help.

@maugustosilva I do not have a hard opinion, but this feature is presented to the user in an ambiguous way.

Users will asks the same questions: why would I want to test measured boot 5 times? If testing more that 1 time makes sense, why are tested all sequentially? How it is that -1 and -8 represent the same condition, but 1 and 8 not? The feature is useful and should be merged, but IMHO should be presented in a different way, maybe using an enum to represent the closed set different options (internally can be an int as it is presented here). That will make some states not represented, and the user will have a more clear picture of what is this is about.

But in any case, this is only my opinion and if any other review approves it I will be totally OK seeing this merged as it is.

niteeshkd:
With the suggestions of @maugustosilva and @galmasi , I modified the code to allow user to choose either 'once'  or 'always' for controlling the number of times of bootlog evaluation against the measured boot policy as a part of measured boot attestation. Also, it now uses the existing variable attestation_count of the verifier database to handle the case when the bootlog evaluation is to be done once.

THS-on (approved):

maugustosilva:
Ok, I see an **entirely unrelated** `pyright` error in `create_policy.py` (which was not modified by this PR) ...  I see no reason to hold the merge because of it.

PR #1434, opened 2023-07-21
tpm_util: Add the BSD license to the file due to functions from TPM 2 code

stefanberger:
The functions to implement TPM2's MakeCredential were taken from the TPM 2 reference implementation that requires that the license and notices be kept but allows for distribution and creation of detrivative works.

### DIFF
--- a/keylime/tpm/tpm_util.py
+++ b/keylime/tpm/tpm_util.py
@@ -1,3 +1,41 @@
+# Some of the functions in this file were adapted from Microsoft's TPM 2
+# reference implementation that states the following:
+#
+# Microsoft Reference Implementation for TPM 2.0
+#
+#  The copyright in this software is being made available under the BSD License,
+#  included below. This software may be subject to other third party and
+#  contributor rights, including patent rights, and no such rights are granted
+#  under this license.
+#
+#  Copyright (c) Microsoft Corporation
+#
+#  All rights reserved.
+#
+#  BSD License
+#
+#  Redistribution and use in source and binary forms, with or without modification,
+#  are permitted provided that the following conditions are met:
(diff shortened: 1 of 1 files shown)


mpeters:
@ansasaki @sergio-correia or @aplanas Do you know of any problems including a copyright like this might have on packaging/distribution?

aplanas:
> @ansasaki @sergio-correia or @aplanas Do you know of any problems including a copyright like this might have on packaging/distribution?

Ah thanks a lot for pinging.

I am not sure to recognize the license, is this "BSD-3-Clause"? Can this coexist with the current "Apache-2.0" that is the one in Keylime?

ansasaki:
> @ansasaki @sergio-correia or @aplanas Do you know of any problems including a copyright like this might have on packaging/distribution?

As @aplanas mentioned above, the question is if this license can coexist with the current used licensing.

If this is not based on a known good license, it will also requires some extra work and special analysis from legal to check if we can include it as part of products and distributions. For Fedora, we would have to also work to include it in https://spdx.dev/ licenses list.

In any case, if we took part of the code from somewhere that has that licensing requirements, we have to abide.

stefanberger:
Microsoft reference implementation uses BSD, so changed it to that as well.

aplanas:
LGTM, IIUC the new package license will be "Apache-2.0 AND BSD-3-Clause", and that should be OK (IANAL)?

stefanberger:
> LGTM, IIUC the new package license will be "Apache-2.0 AND BSD-3-Clause", and that should be OK (IANAL)?

I think so.

aplanas (approved):

PR #1436, opened 2023-07-26
codestyle: Have pyright check mba/elchecking/ except for example.py

stefanberger:


THS-on (approved):

keylime/mba/elchecking/elchecker.py
```diff
@@ -92,7 +92,7 @@ def bootlog_evaluate(
     if len(missing_pcrs) > 0:
         logger.error("PCRs specified for policy %s not in quote: %s", mb_policy_name, str(missing_pcrs))
         failure.add_event("missing_pcrs", {"context": "PCRs are missing in quote", "data": missing_pcrs}, True)
-
+        reason = None
```
aplanas, line 95: Good catch. Maybe I suggest to move it just before the `if`? If in the future this piece get refactored, would be easy to stop that both branches define the variable

aplanas (approved):
Just a nit, but LGTM

aplanas (approved):

PR #1441, opened 2023-08-02
templates/2.0/mapping.json: fix the default registrar_port error in the verifier config

Jingshui1037:
Fixes: #1442 


aplanas (approved):
Indeed

PR #1444, opened 2023-08-03
installer.sh: use the -i parameter variable to set the default binding and listening IP about the agent, verifier, and registrar server is 127.0.0.1 or 0.0.0.0

Jingshui1037:
Keylime agent, verifier, and registrar can be deployed on different machines, so it is better to use the -i parameter to set the default binding and listening IP about the agent, verifier, and registrar server is 127.0.0.1 or 0.0.0.0

Fixes: #1445 

### DIFF
--- a/installer.sh
+++ b/installer.sh
@@ -182,8 +182,11 @@ esac
 # Command line params
 STUB=0
 KEYLIME_DIR=
-while getopts ":hkmp:" opt; do
+# LISTEN_ALL_IPS=1 means than will set the binding and listening IP about the agent, verifier, and registrar server as 0.0.0.0
+LISTEN_ALL_IPS=0
+while getopts ":hikmp:" opt; do
     case $opt in
+        i) LISTEN_ALL_IPS=1 ;;
         k) STUB=1 ;;
         p)
             KEYLIME_DIR=$OPTARG
@@ -196,6 +199,7 @@ while getopts ":hkmp:" opt; do
         h)
             echo "Usage: $0 [option...]"
             echo "Options:"
+            echo $'-i \t\t\t\t Set the default listen IP as 0.0.0.0 rather than 127.0.0.1 for all components(agent, verifier, and registrar)'
             echo $'-k \t\t\t\t Download Keylime (stub installer mode)'
             echo $'-m \t\t\t\t Use modern TPM 2.0 libraries; this is the default'
             echo $'-p PATH \t\t\t Use PATH as Keylime path'
@@ -418,6 +422,13 @@ echo "==========================================================================
 mkdir -p /etc/keylime
 mkdir -p config
 python3 -m keylime.cmd.convert_config --defaults --out config --templates templates
+
+# set the default binding and listening IP about the agent, verifier, and registrar server as 0.0.0.0 when LISTEN_ALL_IPS is set 1
+if [[ "$LISTEN_ALL_IPS" -eq "1" ]] ; then
+    for comp in "agent" "verifier" "registrar"; do
+        sed -i 's/^ip = "127.0.0.1"/ip = "0.0.0.0"/' config/$comp.conf
+    done
+fi
 
 for comp in "agent" "verifier" "tenant" "registrar" "ca" "logging"; do
     mkdir -p /etc/keylime/$comp.conf.d
--- a/installer.sh
+++ b/installer.sh
@@ -182,7 +182,7 @@ esac
 # Command line params
 STUB=0
 KEYLIME_DIR=
-while getopts ":shckmp:" opt; do
+while getopts ":hkmp:" opt; do
     case $opt in
         k) STUB=1 ;;
         p)
--- a/templates/2.0/registrar.j2
+++ b/templates/2.0/registrar.j2
@@ -4,8 +4,8 @@
 # The configuration file version number
 version = {{ registrar.version }}
 
-# The registrar server IP address and port
-ip = {{ registrar.ip }}
+# The binding address and port for the registrar server
+ip = "{{ registrar.ip }}"
 port = {{ registrar.port }}
 tls_port = {{ registrar.tls_port }}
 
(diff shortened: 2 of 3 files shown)


templates/2.0/mapping.json
```diff
@@ -25,7 +25,7 @@
             "ip": {
                 "section": "cloud_agent",
                 "option": "cloudagent_ip",
-                "default": "127.0.0.1"
+                "default": "0.0.0.0"
```
aplanas, line 28: IMHO that is not correct. The default, 127.0.0.1, is more secure as there is control of the interface that will be attached.  Using 0.0.0.0 by default will attach the service into all interfaces by default, a much less secure one.

As a disclosure, I am using 0.0.0.0 in the RPM package so the user can star the service soon, but it is expected that the user will change it to a more secure option.

maugustosilva: I second @aplanas evaluation here. Thanks to both our `keylime_upgrade_config` tool as well as the ability to override values on configuration files from environment variables, I simply cannot see the need for making the **less** secure option (i.e., listen on all interfaces) the default on the template.

maugustosilva:
@Jingshui1037 by the way, I am open to have it (i.e., all components listen to `0.0.0.0`) as a **non-default** option on `installer.sh` (which no distribution uses directly anyway)

Jingshui1037:
> @Jingshui1037 by the way, I am open to have it (i.e., all components listen to `0.0.0.0`) as a **non-default** option on `installer.sh` (which no distribution uses directly anyway)

Thanks @maugustosilva @aplanas , I understand it. Now I change to modify it (i.e., all components listen to 0.0.0.0) in the `installer.sh`. Could you review it again?

Jingshui1037:
> @Jingshui1037 by the way, I am open to have it (i.e., all components listen to `0.0.0.0`) as a **non-default** option on `installer.sh` (which no distribution uses directly anyway)

Thanks @maugustosilva， I understand it.  Do you mean that I can choose to set the default configuration by adding the corresponding variable in `installer.sh`? I updated and force-pushed the codes. Could you review it again?

templates/2.0/agent.j2
```diff
@@ -19,7 +19,7 @@ version = "{{ agent.version }}"
 uuid = "{{ agent.uuid }}"
 
 # The binding address and port for the agent server
-ip = "{{ agent.ip }}"
+ip = {{ agent.ip }}
```
aplanas, line 22: No needed change. The correct one is with quotes, but there is code to add them if they are missing. But I see why is there, to match the rest of the templates.

installer.sh
```diff
@@ -419,6 +421,20 @@ mkdir -p /etc/keylime
 mkdir -p config
 python3 -m keylime.cmd.convert_config --defaults --out config --templates templates
 
+# set the default binding and listening IP about the agent, verifier, and registrar server as 0.0.0.0 when LISTEN_ALL_IPS is set 1
+for arg in "$@"
+do
+    if [[ $arg == LISTEN_ALL_IPS=* ]]; then
```
aplanas, line 427: IIUC the parameter will be used like this?

```
./install.sh LISTEN_ALL_IPS=1
```

If so that is not how it is expected. There is code to parse the parameters here: https://github.com/keylime/keylime/blob/a19bac371a8d38858882fd4e288aa61aa1f0a141/installer.sh#L187

installer.sh
```diff
@@ -182,8 +182,11 @@ esac
 # Command line params
 STUB=0
 KEYLIME_DIR=
-while getopts ":shckmp:" opt; do
+# LISTEN_ALL_IPS=1 means than will set the binding and listening IP about the agent, verifier, and registrar server as 0.0.0.0
+LISTEN_ALL_IPS=0
+while getopts ":shcikmp:" opt; do
     case $opt in
+	i) LISTEN_ALL_IPS=1 ;;
```
aplanas, line 189: nit: check the space / tabs

installer.sh
```diff
@@ -196,6 +199,7 @@ while getopts ":shckmp:" opt; do
         h)
             echo "Usage: $0 [option...]"
             echo "Options:"
+	    echo $'-i \t\t\t\t Set the default listen IP as 0.0.0.0 rather than 127.0.0.1 for all components(agent, verifier, and registrar)'
```
aplanas, line 202: (same, I think that there is a tab at the beginning that should be spaces)

aplanas (approved):
Some nits, but LGTM

installer.sh
```diff
@@ -182,8 +182,11 @@ esac
 # Command line params
 STUB=0
 KEYLIME_DIR=
-while getopts ":shckmp:" opt; do
+# LISTEN_ALL_IPS=1 means than will set the binding and listening IP about the agent, verifier, and registrar server as 0.0.0.0
+LISTEN_ALL_IPS=0
+while getopts ":shcikmp:" opt; do
```
stefanberger, line 187: `s` and `c` are not use (anymore), so they could be removed.

Jingshui1037: Thanks @stefanberger , I fixed it.

installer.sh
```diff
@@ -419,6 +423,13 @@ mkdir -p /etc/keylime
 mkdir -p config
 python3 -m keylime.cmd.convert_config --defaults --out config --templates templates
 
+# set the default binding and listening IP about the agent, verifier, and registrar server as 0.0.0.0 when LISTEN_ALL_IPS is set 1
+if [[ "$LISTEN_ALL_IPS" -eq "1" ]] ; then
+    for comp in "agent" "verifier" "registrar"; do
+        sed -i 's/^ip = \"127.0.0.1\"/ip = \"0.0.0.0\"/' config/$comp.conf
```
stefanberger, line 429: Since you are using single quotes you wouldn't need to escape the double quotes.

Jingshui1037: @stefanberger  OK，I fixed it.

stefanberger (approved):

Issue #1455, opened 2023-08-14
webhook worker doesn't close opened session correctly


kkaarreell:
Is your issue a feature request? If so, please raise it as an [enhancement](https://github.com/keylime/enhancements)

# Environment
 * OS / version:  Fedora, CentOS Stream 9
 * Processor architecture: x86_64
 * TPM Manufacturer: swtpm
 * Keylime version: 7.3.0

# Description 
With `openssl s_server` running as a webhook listener I am getting the following failure
```
804056A6FF7F0000:error:0A000126:SSL routines:ssl3_read_n:unexpected eof while reading:ssl/record/rec_layer_s3.c:320:
```
It seems that `session` is not properly closed in `worker_webhook`
https://github.com/keylime/keylime/blob/fc5cc69c0c1e1b5694240881dfd483ff8ce7c9e9/keylime/revocation_notifier.py#L132

## Expected behavior vs. actual behavior 
session is closed

## Steps to reproduce problem
1. configure webhook URL, use `openssl s_server` as a listener.

# Relevant logs
```
Using default temp DH parameters

ACCEPT

-----BEGIN SSL SESSION PARAMETERS-----

MIGCAgEBAgIDBAQCEwIEIG2z1skUm2k+Tz1y0jQcTrF3lA6+RAdlXotc047p9Wz0

BDC0JjW8R1s2vuHFiC5tTRHP3kfqoAz24rii0a86vAY+wAWB6LuULnzAlobLx1pq

iBihBgIEZMukA6IEAgIcIKQGBAQBAAAArgYCBBFXKcezAwIBFw==

-----END SSL SESSION PARAMETERS-----

Shared ciphers:TLS_AES_256_GCM_SHA384:TLS_AES_128_GCM_SHA256:TLS_AES_128_CCM_SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384:DHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-CCM:ECDHE-ECDSA-AES128-CCM:DHE-RSA-AES256-CCM:DHE-RSA-AES128-CCM

[...]

PR #1456, opened 2023-08-15
Close session in worker_webhook function

kkaarreell:
Resolves https://github.com/keylime/keylime/issues/1455


kkaarreell:
There is a Packit service outage and therefore Packit tests have not been started yet.

stefanberger (approved):

sergio-correia (approved):

keylime/revocation_notifier.py
```diff
@@ -159,6 +159,8 @@ def worker_webhook(tosend: Dict[str, Any], url: str) -> None:
 
             time.sleep(next_retry)
 
+        session.close()
```
aplanas, line 162: I think that a bit more correct fix is to use the context manager: https://requests.readthedocs.io/en/latest/user/advanced/#session-objects

kkaarreell: Updated.

stefanberger (approved):

sergio-correia (approved):

PR #1459, opened 2023-08-23
Implement automatic agent API version bump

### LINKED ISSUE #1457: Updating agent with API bump breaks ongoing attestation

# Environment
 * OS / version: N/A
 * Processor architecture: N/A
 * TPM Manufacturer: N/A
 * Keylime version: latest master (22f0656)

# Description 
If the agent is updated and the supported API version changes, the ongoing attestation fails as the verifier will try to reach out the agent using the old API version.

Expected: the ongoing attestation does not fail as the verifier uses the new API version
Actual: the ongoing attestation fails as the verifier uses the old API version to reach out the agent

## Steps to reproduce problem
 
1. Add an old agent (<v0.2.2) to a verifier
2. Without deleting the agent, update the agent to a newer version (v0.2.2+)

# Relevant logs
 
 The keylime verifier will log something like: 
```
keylime.verifier - CRITICAL - Unexpected Get Quote response error for cloud agent d432fbb3-d2f1-4a97-9>
```
The agent will log something like:
```
WARN  keylime_agent::errors_handler  > GET returning 400 response. API version not supported: v2.0
```

Originally reported as https://github.com/keylime/rust-keylime/issues/642

ansasaki:
Automatically update the agent supported API version in the database if the agent is updated and its API version is bumped.

Previously, if an agent was added to a verifier while it used an old API version, and then it is updated with an API version bump, the attestation would fail as the verifier would try to reach the agent using the old API version.

Fixes #1457

### DIFF
--- a/keylime/cloud_verifier_tornado.py
+++ b/keylime/cloud_verifier_tornado.py
@@ -32,6 +32,7 @@ from keylime import (
 )
 from keylime.agentstates import AgentAttestState, AgentAttestStates
 from keylime.common import retry, states, validators
+from keylime.common.version import str_to_version
 from keylime.da import record
 from keylime.db.keylime_db import DBEngineManager, SessionManager
 from keylime.db.verifier_db import VerfierMain, VerifierAllowlist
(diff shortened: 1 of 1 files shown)


keylime/cloud_verifier_tornado.py
```diff
@@ -998,6 +998,76 @@ def data_received(self, chunk: Any) -> None:
         raise NotImplementedError()
 
 
+async def update_agent_api_version(agent: Dict[str, Any], timeout: float = 60.0) -> (Dict[str, Any] | None):
+    agent_id = agent["agent_id"]
+
+    logger.warning("Agent %s API version bump detected, trying to update stored API version", agent_id)
```
aplanas, line 1004: INFO?

keylime/cloud_verifier_tornado.py
```diff
@@ -998,6 +998,76 @@ def data_received(self, chunk: Any) -> None:
...
+    kwargs = {}
+    if agent["ssl_context"]:
+        kwargs["context"] = agent["ssl_context"]
```
aplanas, line 1007: `kwargs = {"context": agent["ssl_context"] if agent["ssl_context"]}`, but I am not sure that `"ssl_context" in agent` is always true ...

keylime/cloud_verifier_tornado.py
```diff
@@ -998,6 +998,76 @@ def data_received(self, chunk: Any) -> None:
...
+                    agent_db = dict(agent)
+                    for key in exclude_db:
+                        if key in agent_db:
+                            del agent_db[key]
+
+                    session.query(VerfierMain).filter_by(agent_id=agent_id).update(agent_db)  # pyright: ignore
+                    session.commit()
+                except SQLAlchemyError as e:
+                    logger.error("SQLAlchemy Error updating API version for agent %s: %s", agent_id, e)
+                    return None
```
aplanas, line 1055: All this inner `try`/`except` can be dropped (or the external one)

keylime/cloud_verifier_tornado.py
```diff
@@ -998,6 +998,76 @@ def data_received(self, chunk: Any) -> None:
...
+            except SQLAlchemyError as e:
+                logger.error("SQLAlchemy Error updateing API version for agent %s: %s", agent_id, e)
```
aplanas, line 1057: typo in updating

keylime/cloud_verifier_tornado.py
ansasaki: OK!

keylime/cloud_verifier_tornado.py
maugustosilva: `ssl_context` indeed can be `None` if `agent_mtls_cer` is disabled

THS-on:
If I see correctly that the agent could theoretically downgrade it's version. This is not a good idea from a security perspective.

ansasaki:
> If I see correctly that the agent could theoretically downgrade it's version. This is not a good idea from a security perspective.

That is true, I'll add a check if the new API version is greater than the stored one.

ansasaki:
Switched temporarily to a different test branch which includes a new specific test for this PR

maugustosilva (approved):

ansasaki:
Dropped the temporary commit as the tests were merged in the tests repository.

ansasaki:
@maugustosilva Can we merge this?

Issue #1462, opened 2023-08-29
Keylime silently discards signatures it could not parse (wrong signature length indicator) 


stefanberger:
Is your issue a feature request? If so, please raise it as an [enhancement](https://github.com/keylime/enhancements)

No

# Environment
 * OS / version: CentOS 9 stream
 * Processor architecture: x86_64
 * TPM Manufacturer: swtpm
 * Keylime version: 7.5.0

# Description 

*Some* of the signature length indicators in CentOS 9 stream (of today at least) seem to be in little endian while they should be in big endian. Once Keylime cannot parse a signature header, it silently discards the signature but then gets the pcr template hash wrong (missing appended signature) and in the end one ends up with an error message like this one:

```
2023-08-29 11:39:46.041 - keylime.ima - ERROR - IMA measurement list does not match TPM PCR f5e0272e484d74460b3cb2e4adb7ba2930e097b98795df27157cdfdc781b243a
```

## Expected behavior vs. actual behavior 

Keylime should at least emit an error in the log so that the user knows what may be happening.

## Steps to reproduce problem
1. Use CentOS with IMA and ima-sig template

# Relevant logs

```
2023-08-29 11:39:46.041 - keylime.ima - ERROR - IMA measurement list does not match TPM PCR f5e0272e484d74460b3cb2e4adb7ba2930e097b98795df27157cdfdc781b243a
```

PR #1463, opened 2023-08-29
ima: Emit a warning when a file signature could not be parsed

stefanberger:
CentOS currently has file signatures where the signature length indicator seems to be written in little endian. A failure to parse the length indicator causes Keylime to silently discard the signature and the result then is that the pcr template hash is not calculated correctly for the file and the running hash becomes wrong and one ends up with this error message in the log:

IMA measurement list does not match TPM PCR f5e0272e4....

Log the failure if a non zero-length signature was attempted to be parsed.

Resolves: https://github.com/keylime/keylime/issues/1462


THS-on:
@stefanberger can you add one example of those signatures also to the tests?

aplanas (approved):

PR #1467, opened 2023-09-15
Fix an inappropriate test expression to remove a logical short circuit

munahaf:
In file: web_util.py, the comparison of Collection length creates a logical short circuit. I suggested that the Collection length comparison should be done without creating a logical short circuit. 

Sponsorship and Support:

This work is done by the security researchers from OpenRefactory and is supported by the [Open Source Security Foundation (OpenSSF)](https://openssf.org/): [Project Alpha-Omega](https://alpha-omega.dev/). Alpha-Omega is a project partnering with open source software project maintainers to systematically find new, as-yet-undiscovered vulnerabilities in open source code - and get them fixed – to improve global software supply chain security.

The bug is found by running the Intelligent Code Repair (iCR) tool by OpenRefactory and then manually triaging the results.

### DIFF
--- a/keylime/web_util.py
+++ b/keylime/web_util.py
@@ -413,7 +413,7 @@ def get_restful_params(urlstring: str) -> Dict[str, Union[str, None]]:
 
     # If first token looks like an API version, validate it and make sure it's supported
     api_version = "0"
-    if path_tokens[0] and len(path_tokens[0]) >= 0 and re.match(r"^v?[0-9]+(\.[0-9]+)?", path_tokens[0]):
+    if path_tokens[0] and len(path_tokens[0]) > 0 and re.match(r"^v?[0-9]+(\.[0-9]+)?", path_tokens[0]):
         version = keylime_api_version.normalize_version(path_tokens[0])
 
         if keylime_api_version.is_supported_version(version):


ansasaki:
@munahaf Thank you for the patch! Could you please add a sign-off to the commit so that the DCO check passes?

keylime/ima/ast.py
```diff
@@ -95,7 +95,7 @@ def __init__(self, data: str):
         _, _, _, _, sig_size = struct.unpack(fmt, self.data[:hdrlen])
 
         if hdrlen + sig_size != len(self.data):
-            raise ParserError("Invalid signature: malformed header")
+            raise ParserError(f"Invalid signature: malformed header ({hdrlen} + {sig_size} != {len(self.data)})")
```
aplanas, line 98: nit: not sure that showing "132 != 128" will be helpful

keylime/ima/ast.py
```diff
@@ -269,8 +269,9 @@ def create_signature(hexstring: str) -> Optional[Signature]:
         """Create the Signature object if the hexstring is a valid signature"""
         try:
             return Signature(hexstring)
-        except ParserError:
-            pass
+        except ParserError as e:
+            if hexstring:
```
aplanas, line 273: I agree on not hiding the exception under the pass, but in that case I am not sure why to hide it if the signature is empty or None

keylime/web_util.py
```diff
@@ -413,7 +413,7 @@ def get_restful_params(urlstring: str) -> Dict[str, Union[str, None]]:
 
     # If first token looks like an API version, validate it and make sure it's supported
     api_version = "0"
-    if path_tokens[0] and len(path_tokens[0]) >= 0 and re.match(r"^v?[0-9]+(\.[0-9]+)?", path_tokens[0]):
+    if path_tokens[0] and len(path_tokens[0]) > 0 and re.match(r"^v?[0-9]+(\.[0-9]+)?", path_tokens[0]):
```
aplanas, line 416: Indeed the `if` statement there is super weird. If the first element is `True` imply that the second one is `True` ... always. So why not to drop it?

keylime/ima/ast.py
stefanberger: This shouldn't happen very often at all. What it was showing in case of the endianess mixup were values like 32768 != 128.

keylime/ima/ast.py
stefanberger: This commit has already been applied and shouldn't be part of this PR.

Issue #1475, opened 2023-09-25
Add new IDevID and IAK configuration options to templates


ansasaki:
Is your issue a feature request? If so, please raise it as an [enhancement](https://github.com/keylime/enhancements)

# Environment
 * OS / version: N/A
 * Processor architecture: N/A
 * TPM Manufacturer: N/A
 * Keylime version: current master

# Description 
The IDevID and IAK related changes introduced new configuration options: https://github.com/keylime/rust-keylime/pull/608

This issue is to track the addition of the new configuration options to the configuration templates. Ideally this should be done with a minor configuration version bump.

## Expected behavior vs. actual behavior 
Expected: The IDevID and IAK related changes are present in the configuration templates
Actual: The IDevID and IAK related changes are not present in the configuration templates

## Steps to reproduce problem
1. Check configuration templates in `templates`

Isaac-Matthews:
I will be adding this along with the finalised idevid/iak related config options in the next PR. Thanks.

PR #1477, opened 2023-10-03
Add certificates and certificate checking for IDevID and IAK keys

Isaac-Matthews:
This is a draft PR for feedback.

This is the next set of PRs to enable the use of IDevIDs and IAKs as proposed in [enhancement 81.](https://github.com/keylime/enhancements/blob/master/81-IDevID_and_IAK_support.md)
The previous PRs added IDevID and IAK creation, sending, storing and using to verify: [keylime PR](https://github.com/keylime/keylime/pull/1416), [rust-keylime PR](https://github.com/keylime/rust-keylime/pull/608)
The modifications to the Keylime repositories are:

**Keylime**
Columns for IDevID and IAK certificates
Config to require IDevID and IAK
IDevID and IAK certificate checks

**Rust-Keylime**
IDevID and IAK certificate config options
Loading and sending IDevID and IAK certificates from paths

In this PR the certificates must be loaded from path. In future, the ability to load from TPM nvram will need to be added.

I am adding a second commit that includes the template changes requested in [1475](https://github.com/keylime/keylime/issues/1475) to make it easier to review the template changes, as I am not sure if they are correct.


keylime/cert_utils.py
```diff
@@ -71,46 +71,60 @@ def verify_ek(ekcert: bytes, tpm_cert_store: str) -> bool:
...
+            logger.debug("Cert to verify matched cert: %s", cert_file)
             return True
     except Exception as e:
         # Log the exception so we don't lose the raw message
         logger.exception(e)
-        raise Exception("Error processing ek/ekcert. Does this TPM have a valid EK?").with_traceback(sys.exc_info()[2])
+        raise Exception("Error processing certificate.").with_traceback(sys.exc_info()[2])
 
-    logger.error("No Root CA matched EK Certificate")
+    logger.error("No Root CA matched Certificate")
```
stefanberger, line 109: It may be good to pass in the type of certificate as a string parameter so that EK can re-appear here and above.

keylime/registrar_common.py
```diff
@@ -291,20 +292,73 @@ def do_POST(self) -> None:
...
+                    web_util.echo_json_response(self, 400, "Error: IDevID certificate could not be verified")
+                    return
+                logger.info("IDevID cert verified.")
+                if not cert_utils.verify_cert(iak_cert_509, config.get("tenant", "tpm_cert_store")):
+                    logger.warning("Invalid IAK certificate for %s", agent_id)
+                    web_util.echo_json_response(self, 400, "Error: IAK certificate could not be verified")
+                    return
+                logger.info("IAK cert verified.")
+                # NEEDS CRYPTO >= 38.0.0
+                # Check that the agent-generated keys match the certificates
```
stefanberger, line 332: Could we not just send the IAK and IDevID certs from the agent? These seem like cross checks that also the agent could do if we don't trust the key-generation algorithm...

Isaac-Matthews: Thanks for the comments on including cert types, I have pushed a commit with that included.

For this version we could only send the certs and take the keys from the certs. I did this here as I think there is a possibility we would want to allow sending the keys without certs from the agent and performing validation out-of-band and doing this here would reduce the amount of logic and config on the agent. 

Not especially attached to this though as that will require other changes in future as well, so happy to move these checks to the agent if you would prefer that. (This section will still require crypto 38.0.0 though)

keylime/registrar_common.py
```diff
@@ -291,20 +292,73 @@ def do_POST(self) -> None:
...
-            if "iak_tpm" in json_body:
-                iak_tpm = json_body["iak_tpm"]
-                idevid_tpm = json_body["idevid_tpm"]
-                iak_attest = base64.b64decode(json_body["iak_attest"])
-                iak_sign = base64.b64decode(json_body["iak_sign"])
-                logger.info("IDevID and IAK received")
-                idevid_received = True
-            else:
-                logger.info("No IDevID/IAK received")
+            idevid_required = config.getboolean("registrar", "require_idevid_iak")
```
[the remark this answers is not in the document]
Isaac-Matthews: Sorry I am not sure I understand this. Could you rephrase or elaborate?

keylime/registrar_common.py
stefanberger: Also, you should check that the certificates we receive are indeed those of a TPM and not some other type of device that happens to have two certificates. The certificates have to validate against the certificate chain of the issuer and should distinguishable by the content of the SAN field where it indicates that it is a TPM (2.23.133.1.2).

keylime/registrar_common.py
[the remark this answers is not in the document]
stefanberger: If you require IDevID and IAK at this place then that may work for the few newer devices you have in your data center. What about all the other older ones that only have an EK cert? Are you going to be able to still use them or will the not get past this check here?

keylime/registrar_common.py
stefanberger: > Not especially attached to this though as that will require other changes in future as well, so happy to move these checks to the agent if you would prefer that. (This section will still require crypto 38.0.0 though)

I think the agent should report an error if it cannot properly recreate the key and bail out.

keylime/registrar_common.py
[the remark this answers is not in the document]
Isaac-Matthews: If you want to be able to use an EK to register then you can leave require_idevid_iak as false (the default) and continue to just use EK. Is there much to be gained by enabling IDevID and IAK use but still allowing devices to register without them?

keylime/registrar_common.py
Isaac-Matthews: > Also, you should check that the certificates we receive are indeed those of a TPM and not some other type of device that happens to have two certificates. The certificates have to validate against the certificate chain of the issuer and should distinguishable by the content of the SAN field where it indicates that it is a TPM (2.23.133.1.2).

I have added a check for this in [this commit](https://github.com/keylime/keylime/pull/1477/commits/69b08931c6b5631f38ffae0a2c766cd8ad81e88c)

stefanberger: > I have added a check for this in [this commit](https://github.com/keylime/keylime/pull/1477/commits/69b08931c6b5631f38ffae0a2c766cd8ad81e88c)

Great!

keylime/registrar_common.py
```diff
@@ -326,6 +326,16 @@ def do_POST(self) -> None:
...
                 logger.debug("IAK cert verified.")
+
+                #TPM cert checks
+                if not cert_utils.check_tpm_origin(idevid_cert_509, "IDevID"):
+                    web_util.echo_json_response(self, 400, "Error: IDevID certificate might not be from a TPM")
+                    return
+                if not cert_utils.check_tpm_origin(iak_cert_509, "IAK"):
+                    web_util.echo_json_response(self, 400, "Error: IAK certificate might not be from a TPM")
+                    return
+                logger.info("IDevID and IAK seem to come from a TPM")
```
stefanberger, line 337: To avoid the noisy log make this .debug ?

keylime/registrar_common.py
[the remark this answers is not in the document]
Isaac-Matthews: Would adding a second config option work as a fix for this? I.e. enable_idevid and fallback_to_ek? Reg can fail if falback_to_ek is false and idevid fails, or can continue if it is true.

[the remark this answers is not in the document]
stefanberger: Maybe a single configuration option that either requires ( ` = required` ) the idevid & iak or makes them optional ( `= optional` , or `allow_ek` )?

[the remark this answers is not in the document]
stefanberger: Maybe like this?

```
device_identity = ek_cert | ek_cert_or_iak_idevid | iak_idevid
```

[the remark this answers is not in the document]
stefanberger: Or a bit more TPM specific (who knows what the future brings in terms of devices...)

```
tpm_identity = ek_cert | ek_cert_or_iak_idevid | iak_idevid
```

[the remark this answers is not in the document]
mpeters: I like this approach the best I think.

keylime/registrar_common.py
```diff
@@ -291,20 +292,84 @@ def do_POST(self) -> None:
...
+                ) == idevidpub_reg.public_bytes(
+                    serialization.Encoding.PEM, serialization.PublicFormat.SubjectPublicKeyInfo
+                ):
+                    logger.info("IDevID cert and key match")
+                else:
+                    logger.warning("Generated IDevID does not match certificate %s", agent_id)
+                    web_util.echo_json_response(
+                        self, 400, "Error: Generated IDevID does not match certificate. Check selected template."
+                    )
+                    return
```
stefanberger, line 368: This do_POST is already so long. If you could put all these IAK/IDevID checks (starting at `# IDevID and IAK cert checks`) into a function returning bool that would be helpful. If it returns FALSE, just do `return` then.

keylime/cert_utils.py
```diff
@@ -24,6 +25,9 @@
 # when parsing them with python-cryptography fails, and in this case, we
 # try to read the parsed certificate again into python-cryptograhy.
 
+OID_HW_MODULE_NAME = "1.3.6.1.5.5.7.8.4"
+OID_HWTYPE_TPM = "2.23.133.1.2"
```
aplanas, line 29: From where this comes from?

keylime/cert_utils.py
```diff
@@ -71,46 +76,78 @@ def verify_ek(ekcert: bytes, tpm_cert_store: str) -> bool:
...
                 continue
 
-            logger.debug("EK cert matched cert: %s", cert_file)
+            logger.debug("Cert to verify matched cert: %s", cert_file)
             return True
     except Exception as e:
         # Log the exception so we don't lose the raw message
         logger.exception(e)
-        raise Exception("Error processing ek/ekcert. Does this TPM have a valid EK?").with_traceback(sys.exc_info()[2])
+        raise Exception("Error processing %s certificate.", cert_type).with_traceback(sys.exc_info()[2])
```
aplanas, line 112: Exceptions require that you interpolate the string (are different from the log entries, where this way will work): `raise Exception(f"Error processing {cert_type} certificate").with...`

keylime/cert_utils.py
```diff
@@ -71,46 +76,78 @@ def verify_ek(ekcert: bytes, tpm_cert_store: str) -> bool:
...
+
+def verify_ek(ekcert: bytes, tpm_cert_store: str) -> bool:
+    """Verify that the provided EK certificate is signed by a trusted root
+    :param ekcert: The Endorsement Key certificate in DER format
+    :param tpm_cert_store: The path for the TPM certificate store
+    :returns: True if the certificate can be verified, False otherwise
+    """
+    try:
+        ek509 = x509_der_cert(ekcert)
+    except Exception as e:
```
aplanas, line 144: (seems that this `e` will raise a complain, so `err`)

keylime/cert_utils.py
Isaac-Matthews: This is from the certificate fields section of the [TPM 2.0 Keys for device ID spec](https://trustedcomputinggroup.org/wp-content/uploads/TPM-2p0-Keys-for-Device-Identity-and-Attestation_v1_r12_pub10082021.pdf).

You can find it in section 8.1, the hardwareModuleName part of the SubjectAltName on page 57.

keylime/cert_utils.py
Isaac-Matthews: Thanks good spot, I will fix this.

keylime/cert_utils.py
Isaac-Matthews: should it be this:
```
except Exception as e:
    logger.error("Error processing the EK certificate: %s", e)
    return False
```

or do you mean something else?

keylime/registrar_common.py
stefanberger: Better: have it return an error string and if there is one send it via web_util.echo_json_reponse()...

stefanberger (changes requested):

keylime/registrar_common.py
Isaac-Matthews: I have moved the cert checks into a new function in cert_utils, and also rearranged so all idevid stuff is within one block

keylime/cert_utils.py
ansasaki: Please add comments pointing to the origin of these OIDs.

Isaac-Matthews: I have added this in the newest commit

keylime/registrar_common.py
```diff
@@ -289,22 +290,73 @@ def do_POST(self) -> None:
...
                 iak_sign = base64.b64decode(json_body["iak_sign"])
                 logger.info("IDevID and IAK received")
+                idevid_cert = json_body["idevid_cert"]
+                iak_cert = json_body["iak_cert"]
                 idevid_received = True
-            else:
-                logger.info("No IDevID/IAK received")
+
+                # IDevID and IAK cert checks
+                cert_checks, iak_tpm_pub, idevid_tpm_pub = cert_utils.iak_idevid_cert_checks(
```
stefanberger, line 320: I would rename cert_checks to error

keylime/registrar_common.py
```diff
@@ -289,22 +290,73 @@ def do_POST(self) -> None:
...
+                    base64.b64decode(iak_cert), base64.b64decode(idevid_cert), config.get("tenant", "tpm_cert_store")
+                )
+
+                if iak_tpm_pub is None or idevid_tpm_pub is None:
```
stefanberger, line 324: ```if error:```

stefanberger (changes requested):

keylime/registrar_common.py
Isaac-Matthews: OK I will rename

stefanberger:
It would be good to squash the fix patches into the first one.

stefanberger:
The linter problems are fixed now in master. Please rebase.

requirements.txt
```diff
@@ -3,7 +3,7 @@
 # Once an enhancement has been merged, the package needs to be added to the installer.sh, the CI Docker images and issues should be raised on
 # keylime/keylime-rpm keylime/keylime-deb and the keylime ansible installer
 
-cryptography>=3.3.2 # BSD/Apache-2.0
+cryptography>=38.0.0 # BSD/Apache-2.0
```
stefanberger, line 6: With this we would loose CentOS9 and RHEL 9, which are on v36.0: https://rpmfind.net/linux/rpm2html/search.php?query=python3-cryptography&submit=Search+...

Which functions require v38.0?

Isaac-Matthews: The functions that require it have been moved to cert_utils:
```   
    # NEEDS CRYPTO >= 38.0.0
    idevid_pub = idevid_cert_509.public_key()
    iak_pub = iak_cert_509.public_key()
```
I think CentOS9 and RHEL should both be able to install using the installer as pip will provide a wheel.

Isaac-Matthews: You can use this to reproduce:
https://gist.github.com/Isaac-Matthews/1ec57e5536b53ecc2349115193c7416c

stefanberger: > I think CentOS9 and RHEL should both be able to install using the installer as pip will provide a wheel.

SuSE is even only on 3.3.2

The issue is that keylime from rpm packages is supposed to work right away with what is available.

@mpeters , @aplanas : Any chance that at least SuSE and RHEL can update to a more recent python3-cryptography?

ansasaki: @stefanberger Probably not soon, I'll ask the responsible team.

Maybe that is not a problem for now as in CentOS/RHEL keylime is still based on version 7.3.0 (and will be at least for the next 6 months).

> I think CentOS9 and RHEL should both be able to install using the installer as pip will provide a wheel.

Probably installing directly from pip is vetoed in most enterprises. For RHEL, all the dependencies must also be packaged as rpms, meaning that we cannot include these changes in keylime unless the python-cryptography package is updated.

aplanas: SLE / Leap is in 3.3.2 and will not change in the 15.X live cycle. But in MicroOS / Tumbleweed we have 41.0.4. For SUSE we have already a Keylime release, so in that regard is in maintenance mode, so this change is OK.

Isaac-Matthews: Is it possible to use the `%pyproject_wheel` macro to build and install keylime as a wheel, with the dependencies collected via `%pyproject_buildrequires` as described [here?](https://src.fedoraproject.org/rpms/pyproject-rpm-macros/)

I do not have a great deal of understanding for rpms/packaging so that may not make sense.

ansasaki: I'll try to explain without going into RPM details: we cannot use dependencies from pip in CentOS/RHEL for security reasons, not because of technical limitations.

Which means that if we include the changes as they are now and python cryptography is not updated in CentOS/RHEL, we cannot build keylime either. We would be stuck in the current release until python-cryptography is updated, which might mean forever for RHEL 9.

I'd suggest to make the feature enabled conditionally, depending on the available python-cryptography version instead of changing the keylime python-cryptography requirement unconditionally.

Isaac-Matthews: OK understood. In that case would we be happy with:
- Checking crypto version before idevid chunk and setting tpm_identity as ek_cert so the idevid chunk is never used, maybe putting out a warning if the config option gets changed by this event
- Including as a note in the config options the crypto requirement

as a solution here?

stefanberger: > I'd suggest to make the feature enabled conditionally, depending on the available python-cryptography version instead of changing the keylime python-cryptography requirement unconditionally.

I guess it's good to have the fallback mechanism to support IAK + IDevID and also EK cert, of which all the agent should provide. With older versions of python we will then always have to fall back to EK cert.

stefanberger: > * Including as a note in the config options the crypto requirement

Yes, write it to the description of `tpm_identity = default`.

Isaac-Matthews: Am I right in saying the requirements.txt should keep my change? Since it is only used by pip installs afaik, the packagers ignore it and just add python3-cryptography as a build requirement, which if I add the version check will not cause an issue (they just wont have access to this feature).

Isaac-Matthews: Check added in [this commit.](https://github.com/keylime/keylime/pull/1477/commits/3e925f4aa2d49c49e16b74d734357112ccdf4de3)

ansasaki: > Am I right in saying the requirements.txt should keep my change? Since it is only used by pip installs afaik, the packagers ignore it and just add python3-cryptography as a build requirement, which if I add the version check will not cause an issue (they just wont have access to this feature).

Better revert the change in the `requirements.txt` as it would be misleading. Having the version `38.0.0` is not a hard requirement.

Isaac-Matthews: This would mean that anyone installing keylime with an old version of cryptography already installed wouldn't be able to use this. I suppose the problem is that requirements.txt is being used both as a pip installer file and as description of the project. In which case maybe the minimum requirements should be added to documentation instead and the file could be renamed to something pip related. 

At the moment these changes will work for users installing via installer.sh, and will work for rpm install with access to both new and old cryptography with the caveat that old crypto users will only have access to EK registration. Reverting will stop this working for a subset of instller.sh users. If we are OK with that I can just revert, but I just wanted to make sure as this will reduce the quantity of people that will have access to this.

keylime/cert_utils.py
stefanberger: He mean to replace `e` with `err` since some linters (that are not active yet) will complain about variables that are too short.

keylime/cert_utils.py
```diff
@@ -146,3 +190,33 @@ def verify_ek_script(script: Optional[str], env: Optional[Dict[str, str]], cwd:
...
         return False
     return True
+
+
+def iak_idevid_cert_checks(
+    idevid_cert: bytes, iak_cert: bytes, tpm_cert_store: str
+) -> Tuple[str, Optional[tpm2_objects.pubkey_type], Optional[tpm2_objects.pubkey_type]]:
+    idevid_cert_509 = x509_der_cert(idevid_cert)
+    iak_cert_509 = x509_der_cert(iak_cert)
+    # NEEDS CRYPTO >= 38.0.0
```
stefanberger, line 200: Some sort of test case would be helpful, like a test case in `test/test_cert_utils.py`

requirements.txt
stefanberger: I would also revert it for now since probably not a whole lot of TPMs have IDevID and IAK support, yet, except for those used in some clouds of course.

stefanberger (changes requested):

requirements.txt
Isaac-Matthews: @ansasaki @stefanberger this has been reverted now so if you are happy with the check this can be resolved.

keylime/cert_utils.py
ansasaki: @Isaac-Matthews Wouldn't it be better to simply skip the test? This way, the test won't fail if the support is backported to an older version and the parsing start to work.

https://docs.python.org/3/library/unittest.html#skipping-tests-and-expected-failures

Isaac-Matthews: Thanks I was not aware of this option. [Changed here](https://github.com/keylime/keylime/pull/1477/commits/16b0be9542b7c04111aa2d86a8d9a5b2285c0563)

maugustosilva:
Looks like all the tests are passing now (we are not holding merges due to code coverage measurement errors). Is there anything else required before merging this @stefanberger @ansasaki ?

keylime.conf
```diff
@@ -665,6 +665,15 @@ transparency_log_sign_algo = sha256
...
 signed_attributes = ek_tpm,aik_tpm,ekcert
 
+# What TPM-based identity is allowed to be used to register agents.
+# The options "default" and "iak_idevid" will only allow registration with IAK and IDevID if python cryptography is version 38.0.0 or higher.
+# The following options are accepted:
+# "default": either an EK or IAK and IDevID may be used. In the case that cryptography version is <38.0.0 only EK will be used
+# "ek_cert_or_iak_idevid": this is equivalent to default
+# "ek_cert": only allow agents to use an EK to register
+# "iak_idevid": only allow agents with an IAK and IDevID to register
+tpm_identity = default
```
stefanberger, line 675: Should this be in a j2 file as well?

ansasaki: Nice catch! Yes, this should be in the `registrar.j2`. I see it is already in the `mapping.json`, but if it is not in the template, it will not be added to the `registrar.conf` in the output

stefanberger: Who is now using the j2 in 2.1 directory?

ansasaki: This PR is introducing the 2.1 version for the configuration templates.

maugustosilva: @Isaac-Matthews if you fix this last problem, I can include it on this month's release

stefanberger: > This PR is introducing the 2.1 version for the configuration templates.

So convert_config will now pick up the 2.1 templates rather than the old 2.0 ones?

templates/2.1/registrar.j2
```diff
@@ -0,0 +1,116 @@
...
+# artifacts, such as quotes and logs) to be signed by the registrar. The use of "all"
+# will result in the whole "package" (agent + artifacts) being signed and leaving it empty
+# will mean no signing should be done.
+signed_attributes = {{ registrar.signed_attributes }}
+
+# Require an IDevID and IAK for registration. If this option is 'True',
+# any registration that does not provide IDevID and IAK keys and certificates
+# will fail. The certificates are checked in the registration process and the 
+# AK is verified by the IAK
+require_idevid_iak = {{ registrar.require_idevid_iak }}
```
stefanberger, line 116: I believe this here is old.

keylime.conf
Isaac-Matthews: Ah sorry looks like I only updated the mappings when we changed config option. This is now updated with the old option removed.

templates/2.1/registrar.j2
Isaac-Matthews: Yes thanks, I have replaced this with tpm_identity

keylime.conf
stefanberger: Now that we have 2.1 templates is anyone still going to use the 2.0 templates?

ansasaki: The upgrade script applies all mappings in sequence and then outputs the configuration files using the latest templates, but can also provide `--target` to stop at an intermediary version. So, the version `2.0` can still be used.

mpeters (approved):
LGTM

stefanberger (approved):

templates/2.1/mapping.json
```diff
@@ -0,0 +1,838 @@
+{
+    "version": "2.1",
+    "components": {
+        "agent": {
+            "version": {
+                "section": "agent",
+                "option": "version",
+                "default": "2.1"
+            },
+            "revocation_notification_ip": {
```
stefanberger, line 10: I think someone should still look at the diff between the 2.0 mapping.json and this one. There seem to be lots of changes unrelated to IDevID and IAK support and I don't know as much about these files.

ansasaki: I'll check them

ansasaki: The changes LGTM. The part that may be confusing is that the `mapping.json` is now mapping the sections and options from version `2.0` to the options in version `2.1`, and not from the legacy `1.0` (when we had a single `keylime.conf` file).

With the new feature I'm adding in #1486 this kind of change will be much smaller in future as only the differences will show up in the mappings updates.

PR #1480, opened 2023-10-06
create_policy: Set the generator to LegacyAllowList

stefanberger:


maugustosilva (approved):

maugustosilva:
LGTM. It fixes a bug (shouldn't compare to "1" and properly populates policy metadata on generation) while keeping the intended behavior, status-wise (i.e., `has_runtime_policy` has to be equal 1 if `--runtime-policy` was added to the `keylime_tenant` invocation)

keylime/cmd/create_policy.py
```diff
@@ -584,6 +585,9 @@ def main() -> None:
     policy["excludes"] = sorted(list(set(policy["excludes"])))
     policy["ima"]["ignored_keyrings"] = sorted(list(set(policy["ima"]["ignored_keyrings"])))
 
+    policy["meta"]["generator"] = ima.RUNTIME_POLICY_GENERATOR.LegacyAllowList
```
ansasaki, line 588: What is the meaning of `LegacyAllowList`?

All generated runtime policies will be unconditionally be marked as `LegacyAllowList`, is this correct?

stefanberger: Yes. I don't know what `LegacyAllowList` means, nor the meaning of the alternative `CompatibelAllowList`.

ansasaki: @maugustosilva You introduced this on c6d3d0b

What should be the value set here?

maugustosilva: The main goal on the introduction of a `generator` attribute was to allow the `verifier` to be able to differentiate between a scenario where an `agent` was added (via `keylime_tenant`) with vs without `--allowlist`. 

Since this option is no longer valid, and now all policies are automatically converted to the new format, we could switch to `CompatibleAllowList`, although I don't think it has to be done on this PR. The `generator` field only has effect on deciding the value of `has_runtime_policy` attribute, and not on the actual processing of the policy and `agent` attestation.

Issue #1464, opened 2023-09-05
`require_allow_list_signatures` is missing in templates


ansasaki:
Is your issue a feature request? If so, please raise it as an [enhancement](https://github.com/keylime/enhancements)

# Environment
 * OS / version: N/A
 * Processor architecture: N/A
 * TPM Manufacturer: N/A
 * Keylime version: Current master (e9c0b49)

# Description 
The `require_allow_list_signatures` configuration option is included in the mappings for configuration upgrade, but it is not referred in the templates, making it to not be used at all in the configuration files generated by the configuration upgrade tool.

## Expected behavior vs. actual behavior 
Expected: The `required_allow_list_signature` is used in the `verifier.j2` configuration template, including description and accepted values.
Actual: The `required_allow_list_signature` is not used in the `verifier.j2` configuration template

## Steps to reproduce problem
1. Check the `verifier.j2` configuration template

# Relevant logs

Attach any relevant log files that can help to debug your issue.

PR #1482, opened 2023-10-16
Add require_allow_list_signature to the verfier config file

stefanberger:
This PR adds the require_allow_list_signature config option to the verifier config file but sets it to 'False' on default. Also, when signatures are required, raise an error if a plain runtime policy is being used that doesn't have the 'payload' key in its map (The payload key in the signed policy contains the b64 encoded actual runtime policy).

Fixes: #1314
Resolves: #1464

### DIFF
--- a/templates/2.0/mapping.json
+++ b/templates/2.0/mapping.json
@@ -342,7 +342,7 @@
             "require_allow_list_signatures": {
                 "section": "cloud_verifier",
                 "option": "require_allow_list_signatures",
-                "default": "True"
+                "default": "False"
             },
             "durable_attestation_import" : {
                 "section": "cloud_verifier",
--- a/keylime/dsse/dsse.py
+++ b/keylime/dsse/dsse.py
@@ -63,7 +63,7 @@ def b64dec(m_str: str) -> bytes:
         try:
             decoded = base64.b64decode(m, validate=True)
         except binascii.Error:
-            decoded = base64.b64decode(m, altchars="-_", validate=True)  # type: ignore
+            decoded = base64.b64decode(m, altchars="-_", validate=True)
     return decoded
 
 
--- a/keylime/ima/ima.py
+++ b/keylime/ima/ima.py
@@ -501,6 +501,11 @@ def verify_runtime_policy(
     except Exception as error:
         raise ImaValidationError(message="Runtime policy is not valid JSON!", code=400) from error
 
+    if verify_sig and not runtime_policy_key:
+        raise ImaValidationError(
+            message="Runtime policy signature verification required but no key was given!", code=401
+        )
+
     # detect if runtime policy is DSSE
     if runtime_policy_json.get("payload"):
         if verify_sig:
--- a/keylime/ima/ima.py
+++ b/keylime/ima/ima.py
@@ -509,6 +509,9 @@ def verify_runtime_policy(
             logger.info("Runtime policy passed DSSE signature verification")
         else:
             runtime_policy = dsse.b64dec(runtime_policy_json["payload"])
+    else:
+        if verify_sig:
+            raise ImaValidationError(message="Runtime policy is not signed!", code=400)
 
     # Validate exclude list contains valid regular expressions
     _, excl_err_msg = validators.valid_exclude_list(runtime_policy_json.get("exclude"))
--- a/templates/2.1/mapping.json
+++ b/templates/2.1/mapping.json
@@ -372,7 +372,7 @@
             "require_allow_list_signatures": {
                 "section": "verifier",
                 "option": "require_allow_list_signatures",
-                "default": "True"
+                "default": "False"
             },
             "durable_attestation_import": {
                 "section": "verifier",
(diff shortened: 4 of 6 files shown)


mbestavros (approved):
This makes sense to me!

maugustosilva:
LGTM, but wanted to get @ansasaki 's input on the question of **introducing** this new `verifier`-specific parameter (`require_allow_list_signatures`). "Technically", shouldn't it be placed on the `2.1` version of templates?

stefanberger:
For me the biggest issue is the self-verification of an allowlist by a key that the allowlist itself carries...

maugustosilva (approved):

stefanberger:
> @stefanberger Given that we are about to cut a new release, once you move the template from `2.0` to `2.1`, we could merge this.

I think we should discuss whether the self-verification of the allowlist is a good thing before we enable this.

maugustosilva:
What do you see as an alternative in this case? I understand and actually agree with your point DSSE spec resulting in an arrangement which is "a bit odd" (more than a bit actually), but what are the other possible courses of action here?

stefanberger:
> What do you see as an alternative in this case? I understand and actually agree with your point DSSE spec resulting in an arrangement which is "a bit odd" (more than a bit actually), but what are the other possible courses of action here?

Adding support for key management and disabling the self-verification feature.

maugustosilva:
@ansasaki shouldn't it (trusted public keys) be implemented in the broader context of "key management" which @stefanberger proposes then?

ansasaki:
> @ansasaki shouldn't it (trusted public keys) be implemented in the broader context of "key management" which @stefanberger proposes then?

Sure, I think we meant the same thing

stefanberger:
The patches now enforce that when signature verification is required that the user pass in a key. This avoids that the policy passes signature verification due to the embedded self-signed certificate on which the verification is performed.

```
openssl ecparam -name prime256v1 -genkey -noout -out ec-private-key.pem
openssl ec -in ec-private-key.pem -pubout -out ec-pub-key.pem

keylime_create_policy -o mypolicy.json -i .ima  -A /etc/keys/ima/fedora-39-ima.der -A /etc/keys/ima/fedora-38-ima.der
keylime_sign_runtime_policy -r mypolicy.json -o mypolicy-signed.json  -b x509 -k ec-private-key.pem -c /tmp/crt
keylime_tenant -f ./myfile --uuid d432fbb3-d2f1-4a97-9ef7-75bd81c00000 --runtime-policy ./mypolicy-signed.json -c update --runtime-policy-sig-key ./ec-pub-key.pem
```

I suppose we need to copy all the templates to a version 2.3?

stefanberger:
If these patches are now providing a way forward (never pass policy signature verification by verifying the self-sign cert carried in the signed policy) then we will have to adjust the test cases.

stefanberger:
I'd merge it if someone gives me a thumbs-up on the template changes.

templates/2.0/mapping.json
```diff
@@ -342,7 +342,7 @@
             "require_allow_list_signatures": {
                 "section": "cloud_verifier",
                 "option": "require_allow_list_signatures",
-                "default": "True"
+                "default": "False"
```
aplanas, line 345: do we need to worry about changing the old templates default?

stefanberger: I don't want a True to ever appear in `require_allow_list_signatures` because it would change the behavior of keylime_tenant tool requiring policies to be signed.

aplanas (approved):

ansasaki:
I'm sorry that I did not review this before it was merged. There are some mistakes here:

-  The previous already released template mappings should not be modified
   - This includes the template version `2.2` which was already released on version `7.9.0`
- The `replace` action should be used instead of the `add` action. Please refer to the documentation: 
   - https://keylime.readthedocs.io/en/latest/developers/config_update.html#update-mapping-components-format

The correct way of handling this would be adding a new template version `2.3` where the mapping replaces the old default value using the `replace` action:

```
{
    "version": "2.3",
    "type": "update",
    "components": {
        "verifier": {
            "replace": {
                "require_allow_list_signature": {
                    "section": "verifier",
                    "option": "require_allow_list_signature",
                    "default": "False"
                },
            }
        }
    }
}
```

The template for the verifier, `templates/2.3/verifier.j2` should contain the changes that were made here in `templates/2.2/verifier.j2`

I understand the will to avoid the old default value `True` to appear in the configuration files, but it is not possible to modify the past, what is released is set in stone. Changing old templates will not fix deployments using older releases as they have the default values that were in that release. The only way to fix past mistakes in the configuration is to create configuration updates so that deployments using older releases can upgrade their configuration (if the configurations are compatible, which is the case here).

ansasaki:
I'll take back part of what I wrote: actually this option was never used in the old templates, even though the option is set in the mappings (with the incorrect default value).

Then I think it is worth changing the unused default value in the old mappings, so that it does not influence the newly generated configuration with default templates. The part about creating a new version still stands, though.

I'll propose a PR moving the changes to the `2.2` template to the `2.3` version

PR #1483, opened 2023-10-20
Fix some issues detected by pyright

stefanberger:
This PR fixes some issues detected by pyright.


THS-on (approved):

aplanas (approved):
IMHO we should exclude test for the type checking. But LGTM

PR #1486, opened 2023-10-26
configuration upgrade: Add support for update mapping

ansasaki:
Support upgrading the configuration by making changes on top of the previous version.

This introduces the concept of mapping types. The current supported types are "full" and "update". If a mapping is missing the "type" field, it is treated as a full mapping for compatibility.

On full mappings, all the options need to be defined. The update mappings uses the previous version mapping as the base and make changes on top of it.

Supported update operations are:

* Add a new option
* Replace an option with another (e.g. when renaming options)
* Remove an option


stefanberger (approved):

docs/developers/contributing.rst
```diff
@@ -0,0 +1,129 @@
+Contributing
+============
+
+When contributing any keylime repository, please first discuss the change you wish
+to make via an issue in the relevant repository for your change or email to the
+`keylime mailing list <https://groups.io/g/keylime>`_
```
stefanberger, line 6: I didn't even know we had a mailing list...

THS-on: The mailing list is pretty inactive, but exists. Maybe point them to the Slack channel instead?

ansasaki: I'll change the documentation to point to the Slack channel

PR #1489, opened 2023-11-03
ima: Report an error if no quote forward-progress was made

stefanberger:
Report an error if the TPM's quote was not making forward-progress, which means that the PCR that was previously quoted was lagging behind the end of the log (= was not quote of PCR state corresponding to last entry) AND that the current quote is still quoting the same entry as the last time (which is the 1st log entry in the current log) and thus has not caught up with the log even though it should have.

Add a test case testing the check_quote_progress() function.


stefanberger:
/pack-it retest-failed


Interesting error condition but only on CentOS 9:
```
Nov 04 00:33:43 ip-172-31-19-118.us-east-2.compute.internal keylime_verifier[51201]: 2023-11-04 00:33:43.207 - keylime.tpm - INFO - Checking IMA measurement list on agent: d432fbb3-d2f1-4a97-9ef7-75bd81c00000

Nov 04 00:33:43 ip-172-31-19-118.us-east-2.compute.internal keylime_verifier[51201]: 2023-11-04 00:33:43.207 - keylime.ima - ERROR - PCR quote did not make progress to catch up with log

Nov 04 00:33:43 ip-172-31-19-118.us-east-2.compute.internal keylime_verifier[51201]: 2023-11-04 00:33:43.232 - keylime.verifier - WARNING - Agent d432fbb3-d2f1-4a97-9ef7-75bd81c00000 failed, stopping polling

Nov 04 00:34:01 ip-172-31-19-118.us-east-2.compute.internal keylime_verifier[51201]: 2023-11-04 00:34:01.765 - keylime.verifier - INFO - GET returning 200 response for agent_id list
```

THS-on:
Looks fine to me. Just that I understand it correctly the following case it not an error:

1. PCR matched at line 10 but log has 12 entries
2. Next quote PCR matched at line 11, but log has still 12 or more entires

stefanberger:
> Looks fine to me. Just that I understand it correctly the following case it not an error:
> 
>     1. PCR matched at line 10 but log has 12 entries
> 
>     2. Next quote PCR matched at line 11, but log has still 12 or more entires

That would be considered (slow) forward-progress, so it would be ok.

THS-on (approved):

stefanberger:
> > Looks fine to me. Just that I understand it correctly the following case it not an error:
> > ```
> > 1. PCR matched at line 10 but log has 12 entries
> > 
> > 2. Next quote PCR matched at line 11, but log has still 12 or more entires
> > ```
> 
> That would be considered (slow) forward-progress, so it would be ok.

I have changed this now so it has to entirely catch up with the lag, so the above scenario would have been an error.

stefanberger:
I would like to defer this one until after #1495 has been merged.

PR #1490, opened 2023-11-09
Incorrect boolean expression causing ECs to be disallowed

Isaac-Matthews:
The cert_utils certificate checking function incorrectly disallows using EC keys in certs due to wrong expression.


stefanberger (approved):
This fixes an issue I had seen with EC keys.

sergio-correia (approved):

PR #1495, opened 2023-11-13
Fix some codestyle issue

stefanberger:
Fix some codestyle issue reported by mypy and pyright.


keylime/config.py
```diff
@@ -226,7 +226,9 @@ def get_config(component: str) -> RawConfigParser:
                         raise Exception(f"Invalid component {component}")
 
                     for d in (x for x in CONFIG_SNIPPETS_DIRS[component] if os.path.exists(x)):
-                        snippets = sorted(filter(os.path.isfile, (os.path.join(d, f) for f in os.listdir(d) if f)))
+                        snippets: List[str] = sorted(
+                            filter(os.path.isfile, (os.path.join(d, f) for f in os.listdir(d) if f))
+                        )
```
aplanas, line 231: I wonder if with this we can avoid the type annotation:
```
[os.path.join(d, f) for f in os.listdir(d) if f and os.path.isfile(os.path.join(d, f))]
```

The type of this one should be `List[str]`, and not other type of Iterator

stefanberger: Yes, this also work.s

stefanberger: @aplanas I updated the patch with your suggestion.

aplanas (approved):

PR #1497, opened 2023-11-15
added import function for private keys

koh-osug:
I needed to import an existing CA . Since the private key format seems to be a custom format I have added an import tool integrated into keylime_ca.

I added unit tests and fixed the test execution by adding the gpg requirement. The test suite is not working on a fresh system otherwise, is this part maintained?


test/run_tests.sh
```diff
@@ -98,19 +98,19 @@ fi
...
     PYTHON_DEPS="python36-pip python36-dbus"
 # Ubuntu / Debian
 elif [ $PACKAGE_MGR = "apt-get" ]; then
-    PYTHON_PREIN="python3 openssl"
+    PYTHON_PREIN="python3 openssl libgpgme-dev"
     PYTHON_DEPS="python3-pip python3-dbus"
 # SUSE
 elif [ $PACKAGE_MGR = "zypper" ]; then
-    PYTHON_PREIN="python3 openssl"
+    PYTHON_PREIN="python3 openssl gpgme"
```
stefanberger, line 113: Is this a dependency you  added for test/test_signing.py? Could you split this off into a separate patch that fixes just this issue?

requirements.txt
```diff
@@ -24,3 +24,4 @@ jinja2>=3.0.0 # BSD
 # as well as NotRequired on Python < 3.11
 typing-extensions>=4.0.1 ; python_version < '3.11' #PSF-2.0
 jsonschema>=2.6.0 # MIT
+gpg>=1.10.0
```
stefanberger, line 27: The license here is missing. We haven't need to add gpg here so far or anywhere else actually.

Due to this line I get the following issue:
```
$ tox -e mypy
[...]  Installing build dependencies ... done
  Getting requirements to build wheel ... error

  error: subprocess-exited-with-error

  × Getting requirements to build wheel did not run successfully.
  │ exit code: 1
  ╰─> [2 lines of output]
      <string>:163: SyntaxWarning: invalid escape sequence '\s'
      Could not find gpgme-config.  Please install the libgpgme development package.
      [end of output]

  note: This error originates from a subprocess, and is likely not a problem with pip.
error: subprocess-exited-with-error

× Getting requirements to build wheel did not run successfully.
│ exit code: 1
╰─> See above for output.

note: This error originates from a subprocess, and is likely not a problem with pip.
mypy: exit 1 (6.11 seconds) /home/stefanb/keylime> python -I -m pip install mypy -r /home/stefanb/keylime/requirements.txt -r /home/stefanb/keylime/test-requirements.txt pid=48250
  mypy: FAIL code 1 (6.25 seconds)
  evaluation failed :( (6.37 seconds)
```

koh-osug: `run_tests.sh` fails when `gpg`is not available. Also the github action is failing without it. It is referenced in https://github.com/keylime/keylime/blob/c26708fd3e55e2eccf98c7fe532b4a16a4426be5/keylime/signing.py#L8 How does it come that it is not needed?

stefanberger: > How does it come that it is not needed?

Maybe it's being installed as a dependency for some other package but I don't know . So far we never needed it here.

koh-osug: OK, let me recheck, where my issue are then coming from. Indeed the workflow does not need it, but `run_tests.sh` fails without it. I will create another version of the PR.

koh-osug: I have updated the PR not manipulating the github workflows.

stefanberger: I still think the gpg stuff in the installer files should be separated from this PR that tries to import a private key.

koh-osug: trying to separate it

THS-on: The reason that gpg is missing from the requirements.txt is, that it needs the same gpg version installed and is normally provided by the gpg package for the distribution.
That's why there is also a comment about it in there:
https://github.com/keylime/keylime/blob/c26708fd3e55e2eccf98c7fe532b4a16a4426be5/requirements.txt#L21-L22

koh-osug: I have now removed the test fixes and this PR is just for the import function.

stefanberger:
I think the patch description should have the title 'ca_util: add import functions for private keys' and then in the body it should state what it does because that's what people will read when they look through the patches. The description above in this PR is good but we also need something like this in the patch itself.

koh-osug:
I modified the git commit message

keylime/ca_util.py
```diff
@@ -151,6 +152,20 @@ def cmd_mkcert(workingdir: str, name: str, password: Optional[str] = None) -> No
         os.chdir(cwd)
 
 
+def cmd_import_priv(priv_pem_file: str, lastserial: int) -> None:
```
stefanberger, line 155: This function seems to need workingdir as a parameter and then you have to ch_dir to this directory otherwise the key goes into an unexpected place.

koh-osug: added

keylime/ca_util.py
```diff
@@ -151,6 +152,25 @@ def cmd_mkcert(workingdir: str, name: str, password: Optional[str] = None) -> No
         os.chdir(cwd)
 
 
+def cmd_import_priv(workingdir: str, priv_pem_file: str, lastserial: int) -> None:
+    cwd = os.getcwd()
+    try:
+        fs_util.ch_dir(workingdir)
+        with open(priv_pem_file, "rb") as priv_pem:
```
stefanberger, line 159: If you were to read this file before changing the directory one could pass `-f privkey.pem`, otherwise one has to provide the full path.

koh-osug: looking into this ...

koh-osug: fixed

keylime/ca_util.py
```diff
@@ -151,6 +152,25 @@ def cmd_mkcert(workingdir: str, name: str, password: Optional[str] = None) -> No
...
+            pem = priv_pem.read()
+        priv = read_private(False)
+        private_key = load_pem_private_key(pem, None, default_backend())
+        priv[0]["ca"] = private_key.private_bytes(
+            encoding=serialization.Encoding.PEM,
+            format=serialization.PrivateFormat.PKCS8,
+            encryption_algorithm=serialization.NoEncryption(),
+        )
+        priv[0]["lastserial"] = int(lastserial) + 1
+        fs_util.ch_dir(workingdir)
```
stefanberger, line 168: For consistency reasons I would put the `ch_dir` before the `read_private()` because all other functions have it also in this order to read the key from the `workingdir`. Maybe there's also a benefit of not loosing possibly already revoked keys once importing/changing a private key?

stefanberger (changes requested):

keylime/ca_util.py
koh-osug: Ok. Fixed. That's it?

stefanberger: LGTM now.

stefanberger (approved):

PR #1498, opened 2023-11-17
support for user mode in run-test.sh

koh-osug:
The test script supports a user mode, but fails when it is used.

### DIFF
--- a/test/run_tests.sh
+++ b/test/run_tests.sh
@@ -119,7 +119,6 @@ fi
 
 if [[ "$USER_MODE" -ne "0" ]] ; then
     echo -e "Packages '$PYTHON_DEPS' are required for this script.  Please manually install them, or run this script as root." 1>&2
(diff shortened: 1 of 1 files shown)


test/run_tests.sh
```diff
@@ -80,7 +80,6 @@ esac
 # Permissions-related sanity checking
 if [[ "$USER_MODE" -eq "0" && $EUID -ne 0 ]]; then
    echo "This script must be run as root in order to install dependencies.  Use -u for user mode." 1>&2
-   exit 1
```
stefanberger, line 83: Why not exit after this error message? Same comment below.

koh-osug: Oh, right, this one was too much. Fixed. The other one is need, otherwise the script is never accepting `-u`.

maugustosilva (approved):

PR #1502, opened 2023-11-23
Add documentation for IAK and IDevID

Isaac-Matthews:
Some documentation for the recent IDevID and IAK patches. Feedback and typo-spotting is very welcome.


docs/user_guide/idevid_and_iak.rst
```diff
@@ -0,0 +1,56 @@
...
+
+==========  ===============  ==========
+ Template    asymmetricAlg    nameAlg
+==========  ===============  ==========
+H-1         RSA 2048         SHA256
+H-2         ECC NIST P256    SHA256
+H-3         ECC NIST P384    SHA384
+H-4         ECC NIST P521    SHA512
+H-5         ECC SM2 P256     SM3_256
+==========  ===============  ==========
```
stefanberger, line 51: I think users should not have to deal with these. Instead, the agent should read a certificate from a file or later on from an NVRAM location, determine the type (EC vs. RSA) and size of the key and deduce template and nameAlg from it.

Isaac-Matthews: OK I can work on a patch for that. Would it be worth including the documentation as is for now so people can use the feature as it exists today, then changing when that is added?

maugustosilva:
Just a couple of minor comments:

1) Make sure to mention (just for convenience and completeness, I know most of it is already mentioned on config file's comments)  that `tpm_identity` parameter is part of `registrar.conf` (doesn't hurt to mention that it can be overridden by setting the environment variable `KEYLIME_REGISTRAR_TPM_IDENTITY`

2) Same could be said for `agent.conf`

3) I know it should be pretty clear, but again, I am just trying to eliminate reasons for people to go back to the slack channel (or email list) and ask for clarification later. The `TPM_CERT_STORE` can hold certificates for multiple authorities simultaneously (right?).

docs/user_guide/idevid_and_iak.rst
maugustosilva: I propose we merge as is, and a future patch (for the `agent`) is followed by a new PR just to update the documentation. Let's make sure the code capabilities and documentation are on sync.

Isaac-Matthews:
@maugustosilva @stefanberger  hopefully [this addresses all comments](https://github.com/keylime/keylime/pull/1502/commits/f9fc66199a31a09bce51c3052b586a18164078be), let me know if more is needed.

maugustosilva (approved):

PR #1505, opened 2023-12-05
docs: add additional reading section

THS-on:
This adds an initial additional reading section based on the list from @maugustosilva and @galmasi plus some additional papers and talks that I found.


maugustosilva (approved):

galmasi (approved):
Ditto. LGTM, and a very nice contribution. Thank you.

PR #1512, opened 2024-01-16
codestyle: Remove a 'type: ignore' comment (mypy)

stefanberger:
Remove the 'type: ignore' comment from migration.py to allow mypy to pass without errors.


stefanberger:
This PR has to go in before pyright PR since the lint jobs stops on mypy issues before pyright runs.

aplanas:
Ah sorry. I missed this.

aplanas (approved):

PR #1526, opened 2024-02-13
mba: Add a separate table for measured boot policies.

niteeshkd:
In the next PR, similar to named runtime policies, this table will be used to provide support for named measured boot policies and thier management.

### DIFF
--- a/keylime/db/verifier_db.py
+++ b/keylime/db/verifier_db.py
@@ -28,7 +28,8 @@ class VerfierMain(Base):
     ima_policy = relationship("VerifierAllowlist", back_populates="agent", uselist=False)
     ima_policy_id = Column(Integer, ForeignKey("allowlists.id"))
     ima_sign_verification_keys = Column(Text().with_variant(Text(429400000), "mysql"))
-    mb_refstate = Column(Text().with_variant(Text(429400000), "mysql"))
+    mb_policy = relationship("VerifierMbpolicy", back_populates="agent", uselist=False)
+    mb_policy_id = Column(Integer, ForeignKey("mbpolicies.id"))
     revocation_key = Column(String(2800))
     accept_tpm_hash_algs = Column(JSONPickleType(pickler=JSONPickler))
     accept_tpm_encryption_algs = Column(JSONPickleType(pickler=JSONPickler))
@@ -62,3 +63,12 @@ class VerifierAllowlist(Base):
     generator = Column(Integer)
     tpm_policy = Column(Text())
     ima_policy = Column(Text().with_variant(Text(429400000), "mysql"))
+
+
+class VerifierMbpolicy(Base):
+    __tablename__ = "mbpolicies"
+    __table_args__ = (schema.UniqueConstraint("name", name="uniq_mbpolicies_name"),)
+    id = Column(Integer, primary_key=True)
+    agent = relationship("VerfierMain", back_populates="mb_policy")
+    name = Column(String(255), nullable=False)
+    mb_policy = Column(Text().with_variant(Text(429400000), "mysql"))
--- a/keylime/migrations/versions/f4196d13fe45_migrate_mb_refstates_to_mbpolicies.py
+++ b/keylime/migrations/versions/f4196d13fe45_migrate_mb_refstates_to_mbpolicies.py
@@ -1,0 +1,83 @@
+"""migrate mb_refstates to mbpolicies
+
+Revision ID: f4196d13fe45
+Revises: 32902c0a8d90
+Create Date: 2024-02-08 04:16:21.554062
+
+"""
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "f4196d13fe45"
+down_revision = "32902c0a8d90"
+branch_labels = None
+depends_on = None
+
+
+def upgrade(engine_name):
+    globals()[f"upgrade_{engine_name}"]()
+
+
+def downgrade(engine_name):
+    globals()[f"downgrade_{engine_name}"]()
+
+
+def upgrade_registrar():
+    pass
+
+
+def downgrade_registrar():
+    pass
+
+
+def upgrade_cloud_verifier():
+    # Migrate existing agent info to the mbpolicies table.
+    conn = op.get_bind()
+
+    res = conn.execute(sa.text("SELECT agent_id, mb_refstate FROM verifiermain"))
+    results = res.fetchall()
+    old_policy = [{"name": r[0], "mb_policy": r[1]} for r in results]
+
+    # get existing table metadata
+    meta = sa.MetaData()
+    meta.reflect(bind=conn, only=("mbpolicies",))
+    mbpolicies = meta.tables["mbpolicies"]
+
+    # Insert the existing MB policies to the mbpolicies table
+    op.bulk_insert(mbpolicies, old_policy)
+
+    # Modify verifiermain table (i.e. add mb_policy_id and remove mb_refstate)
+    with op.batch_alter_table("verifiermain") as batch_op:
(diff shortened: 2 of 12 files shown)


niteeshkd:
@THS-on @aplanas ,  can I request you to please take a look into this PR  and provide any comment? Thanks!

keylime/migrations/versions/f4196d13fe45_migrate_mb_refstates_to_mbpolicies.py
```diff
@@ -0,0 +1,77 @@
...
+
+    # Insert the existing MB policies to the mbpolicies table
+    op.bulk_insert(mbpolicies, old_policy)
+
+    # Modify verifiermain table (i.e. add mb_policy_id and remove mb_refstate)
+    with op.batch_alter_table("verifiermain") as batch_op:
+        batch_op.add_column(
+            sa.Column("mb_policy_id", sa.Integer(), sa.ForeignKey("mbpolicies.id", name="fk_verifiermain_mbpolicies"))
+        )
+        batch_op.drop_column("mb_refstate")
```
stefanberger, line 55: Do you also need to add this column back when doing the downgrade?

niteeshkd: Yes, i need to add mb_refstate back to verifiermain. I am adding it back while downgrading through its next upgrade file (i.e. *associate*.py ) but i should have added it here. I will do that.

niteeshkd: > Yes, i need to add mb_refstate back to verifiermain. I am adding it back while downgrading through its next upgrade file (i.e. 9d2f6fab52b1_associate_moved_mb_refstates_to_agents.py ) but i should have added it here. I will do that.

Done!

keylime/mba/mba.py
```diff
@@ -185,3 +185,18 @@ def _find_implementation(functionname: str) -> Any:
...
+
+
+def mb_policy_db_contents(mb_policy_name: str, mb_policy: Optional[str]) -> Dict[str, Any]:
+    """Assembles a mb_policy dictionary to be written on the database"""
+    mb_policy_db_format: Dict[str, Any] = {}
+
+    mb_policy_db_format["name"] = mb_policy_name
+    mb_policy_db_format["mb_policy"] = mb_policy
+
+    return mb_policy_db_format
```
THS-on, line 202: This function seems to not really necessary.

keylime/db/verifier_db.py
```diff
@@ -62,3 +63,12 @@ class VerifierAllowlist(Base):
     generator = Column(Integer)
     tpm_policy = Column(Text())
     ima_policy = Column(Text().with_variant(Text(429400000), "mysql"))
+
+
+class VerifierMbpolicy(Base):
+    __tablename__ = "mbpolicies"
+    __table_args__ = (schema.UniqueConstraint("name", name="uniq_mbpolicies0name"),)
```
THS-on, line 70: Any reason why the name contains a "0" instead of "_"?

THS-on (commented):
Besides the small comments, looks fine.

How is deletion of the entry implemented once the agent is removed. If the table was added implicitly, it should also be removed automatically.

niteeshkd:
> Besides the small comments, looks fine.
> 
> How is deletion of the entry implemented once the agent is removed. If the table was added implicitly, it should also be removed automatically.

Th entry is being deleted through the function cloud_verifier_tornado.py:verifier_db_delete_agent() which is called when the agent is deleted.

keylime/mba/mba.py
niteeshkd: You are right. Somehow, i wrote this thinking about a generic fuction to return a db format for a given name and policy and policy could be verified in the same function before returing the format. But, we don't need that  at this time.

keylime/db/verifier_db.py
THS-on: tbh this looks like a typo we didn't catch. I would use an underscore

niteeshkd: Sure, I will use underscore. Should we also correct the underscore of `allowlists` with this PR?

niteeshkd: > Sure, I will use underscore. Should we also correct the underscore of `allowlists` with this PR?

I think we should handle the underscore of `allowlists` separately.

THS-on: > Should we also correct the underscore of allowlists with this PR?

No because it is mostly cosmetic and yet still requires DB migration

keylime/migrations/versions/f4196d13fe45_migrate_mb_refstates_to_mbpolicies.py
```diff
@@ -0,0 +1,83 @@
...
+    meta.reflect(bind=conn, only=("mbpolicies",))
+    mbpolicies = meta.tables["mbpolicies"]
+
+    # Insert the existing MB policies to the mbpolicies table
+    op.bulk_insert(mbpolicies, old_policy)
+
+    # Modify verifiermain table (i.e. add mb_policy_id and remove mb_refstate)
+    with op.batch_alter_table("verifiermain") as batch_op:
+        batch_op.add_column(
+            sa.Column("mb_policy_id", sa.Integer(), sa.ForeignKey("mbpolicies.id", name="fk_verifiermain_mbpolicies"))
```
stefanberger, line 53: When I tried downgrading I found that the CONSTRAINT still remained in `verifiermain`:
```
CONSTRAINT fk_verifiermain_allowlists FOREIGN KEY(ima_policy_id) REFERENCES allowlists (id)
```

stefanberger: I used this command here to downgarde:
```
sudo bash -c "PYTHONPATH=. alembic -c keylime/migrations/alembic.ini downgrade 160f932fde5b"
```

niteeshkd: > When I tried downgrading I found that the CONSTRAINT still remained in `verifiermain`:
> 
> ```
> CONSTRAINT fk_verifiermain_allowlists FOREIGN KEY(ima_policy_id) REFERENCES allowlists (id)
> ```
> 
> Can you drop this CONSTRAINT during a downgrade?

That constraint is beacuse of the existing `allowlists` table and it should not be removed when we downgrade till 160f932fde5b .

stefanberger: > That constraint is beacuse of the existing `allowlists` table and it should not be removed when we downgrade till 160f932fde5b .

I downgraded to `160f932fde5b` with this command:

```
sudo bash -c "PYTHONPATH=. alembic -c keylime/migrations/alembic.ini downgrade 160f932fde5b
```

stefanberger: @niteeshkd You're right. The constraint is gone.

keylime/mba/mba.py
niteeshkd: This function is removed with the last commit.

ansasaki (approved):
Sorry for taking so long to review. LGTM!

Please, consider also contributing a end-to-end test to the https://github.com/RedHat-SP-Security/keylime-tests

niteeshkd:
> Sorry for taking so long to review. LGTM!
> 
> Please, consider also contributing a end-to-end test to the https://github.com/RedHat-SP-Security/keylime-tests

Thank you!

PR #1531, opened 2024-02-28
verifier, tenant: make payload for agent completely optional

THS-on:
This now allows an agent to be added without using the payload feature.
Note that then also no U/V key is sent to the agent on successful attestation.

### DIFF
--- a/keylime/tenant.py
+++ b/keylime/tenant.py
@@ -307,11 +307,9 @@ class Tenant:
             self.runtime_policy_key,
         ) = policies.process_policy(cast(policies.ArgsType, args))
 
-        # if none
-        if args["file"] is None and args["keyfile"] is None and args["ca_dir"] is None:
-            raise UserError(
-                "You must specify one of -k, -f, or --cert to specify the key/contents to be securely delivered to the agent"
-            )
+        # Check if verify flag is not set when no payload is added
+        if args["verify"] and not (args["file"] or args["ca_dir"] or args["keyfile"]):
+            raise UserError("--verify only works when a payload is provided to the agent via -k, -f, or --cert")
 
         if args["keyfile"] is not None:
             if args["file"] is not None or args["ca_dir"] is not None:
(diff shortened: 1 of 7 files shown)


maugustosilva:
Given we are "changing" the column `v` on `verifier`'s database , don't we need an `alembic` migration?

THS-on:
We don't because this is the default. I just wanted to make it explicit that we are intending to put null in there.

keylime/tenant.py
```diff
@@ -570,7 +564,9 @@ def validate_tpm_quote(self, public_key: str, quote: str, hash_alg: algorithms.H
 
     def do_cvadd(self) -> None:
         """Initiate v, agent_id and ip and initiate the cloudinit sequence"""
-        b64_v = base64.b64encode(self.V).decode("utf-8")
+        b64_v = None
+        if self.V:
+            b64_v = base64.b64encode(self.V).decode("utf-8")
```
aplanas, line 569: nit: `b64_v = base64.b64encode(self.V).decode("utf-8") if self.V else None`

aplanas:
> Note that then also no U/V key is sent to the agent on successful attestation.

What are the consequences of this?

keylime/tenant.py
THS-on: Fixed

THS-on:
> > Note that then also no U/V key is sent to the agent on successful attestation.
> 
> What are the consequences of this?

The consequence of this is that the agent has no idea if it passed initial attestation at some point and that the `/v2.1/keys/verify` interface wont work.  So the `--verify` in the tenant won't work.

stefanberger:
> > > Note that then also no U/V key is sent to the agent on successful attestation.
> > 
> > 
> > What are the consequences of this?
> 
> The consequence of this is that the agent has no idea if it passed initial attestation at some point and that the `/v2.1/keys/verify` interface wont work. So the `--verify` in the tenant won't work.

I  tested this. It works for me. 
I hadn't used `--verify` before. When not passing `-f <file>` and passing `--verify` on `-c add` then we get an error. I think we should be able to catch this case and error out telling the user that `--verify` requires `-f <file>`. `-c update' also supports `--verify` but there I am not sure how to handle this other than having `-f <file>` described as a prerequisite for `--verify` in the help screen.

THS-on:
@stefanberger yes will add better error handling for this case.

THS-on:
@stefanberger @kkaarreell now updated the tenant to include a better error message when no payload is specified, but verify is set.

stefanberger (approved):

maugustosilva:
LGTM

PR #1536, opened 2024-03-19
sign_runtime_policy: Display error message if non-EC key is provided

stefanberger:
Instead of showing an assert and traceback display a user-friendly
message when the user provides anything else than an EC private key.


THS-on (commented):
LGTM

THS-on (approved):

PR #1539, opened 2024-03-22
templates: Add version 2.3

ansasaki:
This contains the changes made in #1482 to the templates and mappings for the version `2.2` which was already released on version `7.9.0`

This also fixes a bug with the default usage when generating the configuration files with the `--default` option where the first value set as default for an option is kept. This changes this logic and uses the default value set for the option in the latest template version.

EDIT: Some of the previously used default values came from template version 2.0 and needed adjustment using the `adjust.py` script. These were the default values used before the change in the `--default` option.

To not require an `adjust.py` script to be added to the template version 2.3, the default values were replaced to the adjusted format, keeping the same default values previously used.

### DIFF
--- a/keylime/cmd/convert_config.py
+++ b/keylime/cmd/convert_config.py
@@ -357,6 +357,7 @@ def strip_quotes(config: RawConfigParser) -> None:
 def process_full_mapping(
     mapping: Dict[str, Any],
     old_config: RawConfigParser,
+    use_defaults: bool,
     target: Optional[Tuple[int, int]] = None,
     logger: Logger = logging.getLogger(__name__),
 ) -> RawConfigParser:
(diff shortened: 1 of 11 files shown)


stefanberger (approved):

keylime/cmd/convert_config.py
```diff
@@ -357,6 +357,7 @@ def strip_quotes(config: RawConfigParser) -> None:
 def process_full_mapping(
     mapping: Dict[str, Any],
     old_config: RawConfigParser,
+    use_defaults: bool,
```
aplanas, line 360: seems that the tests needs to be updated to use this new parameter?

aplanas (changes requested):

keylime/cmd/convert_config.py
ansasaki: Yes. I'll make this a WIP and fix the issues in tests

ansasaki: This should be solved now

maugustosilva (approved):

maugustosilva:
LGTM

maugustosilva:
@aplanas ok with merging it after @ansasaki addressed the issues you pointed out?

aplanas (approved):

PR #1542, opened 2024-04-08
WIP: IMA: Implement support for an internal text-based IMA policy based on rules

stefanberger:
This PR implements support for an internal text-based IMA policy based on rules rather than the hardcoded evaluation that we currently have. The idea is that this will provide more flexibility for evaluating IMA log entries in the future where the user gets control over adding such rules to the IMA runtime policy. For now we need an internal policy that provides (ideally) 100% compatibility to the hard coded evaluation. The current internal policy looks like this:

LEARN-KEYS: ignore-keyrings-ref=ignored_keyrings allowed-hashes-ref=keyrings
DEVICE-MAPPER-CHECK
FILE-NAMES: regex-ref=excludes
IMA-SIGNATURE-CHECK
FILE-HASHES: map-ref=digests target=ACCEPT


THS-on:
Before we try to invent our own policy language, I think it makes sense to look if we then take it a step further and implement something like Rego (https://www.openpolicyagent.org/docs/latest/policy-language/) fully.

I know that @galmasi has some experience with this and ran some experiments regarding measured boot.

stefanberger:
> Before we try to invent our own policy language, I think it makes sense to look if we then take it a step further and implement something like Rego (https://www.openpolicyagent.org/docs/latest/policy-language/) fully.

I had at least looked at it before but it wasn't clear to me how to use this for what we have so far, such as list with regular expressions and maps with filenames and list of hashes and verification of signature and handling of device mapper and the learning of keys. I would be curious to see how these things would be dealt with in this language.

stefanberger:
Other format of rules introduced in this PR look like these here:

```
FILE-HASHES: target=ACCEPT filename=/usr/bin/dd hash=sha256:d33d5d13792292e202dbf69a6f1b07bc8a02f01424db8489ba7bb7d43c0290ef
FILE-HASHES: target=ACCEPT filename=/usr/bin/zmore hash=sha256:b8ae0b8dd04a5935cd8165aa2260cd11b658bd71629bdb52256a675a1f73907b

FILE-NAMES: regex=ACCEPT:/usr/bin/dd
FILE-HASHES: target=ACCEPT filename=/usr/bin/zmore hash=sha256:0000000000000000000000000000000000000000000000000000000000000000 hash=sha256:b8ae0b8dd04a5935cd8165aa2260cd11b658bd71629bdb52256a675a1f73907b

FILE-NAMES: regex=ACCEPT:/usr/bin/dd regex=ACCEPT:/usr/bin/zmore
```

ansasaki:
> Other format of rules introduced in this PR look like these here:
> 
> ```
> FILE-HASHES: target=ACCEPT filename=/usr/bin/dd hash=sha256:d33d5d13792292e202dbf69a6f1b07bc8a02f01424db8489ba7bb7d43c0290ef
> FILE-HASHES: target=ACCEPT filename=/usr/bin/zmore hash=sha256:b8ae0b8dd04a5935cd8165aa2260cd11b658bd71629bdb52256a675a1f73907b
> 
> FILE-NAMES: regex=ACCEPT:/usr/bin/dd
> FILE-HASHES: target=ACCEPT filename=/usr/bin/zmore hash=sha256:0000000000000000000000000000000000000000000000000000000000000000 hash=sha256:b8ae0b8dd04a5935cd8165aa2260cd11b658bd71629bdb52256a675a1f73907b
> 
> FILE-NAMES: regex=ACCEPT:/usr/bin/dd regex=ACCEPT:/usr/bin/zmore
> ```

I don't know if I'm the only one with this feeling, but the proposed language doesn't look very intuitive, meaning that is not clear for someone to read it and immediately understand what is happening. I understand the focus is for machines to read and generate the policies, but it would be nice to keep the language sufficiently simple that allows manual adjustment.

I think we should discuss the format a bit further with the community before implementing it.

Are we considering how policies will be combined, or extended? For example, a policy that uses a base policy and adds more rules on top of it.

Also we should consider the case where a system is updated as the main use case as that is probably the most common case when the policies need to be adjusted.

stefanberger:
> I don't know if I'm the only one with this feeling, but the proposed language doesn't look very intuitive, meaning that is not clear for someone to read it and immediately understand what is happening. I understand the focus is for machines to read and generate the policies, but it would be nice to keep the language sufficiently simple that allows manual adjustment.

> I think we should discuss the format a bit further with the community before implementing it.

Sure, make a proposal for a language.

> Are we considering how policies will be combined, or extended? For example, a policy that uses a base policy and adds more rules on top of it.

Rules need a processing order so just concatenating them from two different policies is certainly not possible. They would need to be merged in some 'right' way. It may not always be possible for a machine to do it.

> Also we should consider the case where a system is updated as the main use case as that is probably the most common case when the policies need to be adjusted.

Right, we need tools to be able to get a handle on the policy.

kkaarreell:
> Are we considering how policies will be combined, or extended? For example, a policy that uses a base policy and adds more rules on top of it.

My gut feeling is that something like that may become handy in the future. The advantage of separate excludelist, allowlist, rejectlist etc. processing is that each of these lists can be extended and the result still makes sense. Once we start combining allow/exclude/reject rules within a single file then making such extensions won't be "guaranteed", although such a format would be more "rich". ATM I can't decide which benefits out-weights. Would the ability to freely combine rules really provide benefits for real deployments? It would be nice to have those backed up by some "user stories".

stefanberger:
> My gut feeling is that something like that may become handy in the future. The advantage of separate excludelist, allowlist, rejectlist etc. processing is that each of these lists can be extended and the result still makes sense.

For as long as you have good tools to manage these lists and don't want to do fancy things. Moving all those packages with recent CVEs from an allowlist into a rejectlist obviously needs a lot more involvement than what one would like and basically means starting over with a new policy, though I wouldn't know where to get the hashes for the packages with CVEs from. Or adding those packages I just installed from EPEL/pypy/etc. to the allowlist while the rest is covered by signature verification is another interesting challenge -- or removing those old packages' hashes from the allowlist that I just upgraded. Policy management is going to be the ultimate challenge.

maugustosilva:
Having worked on the measured boot policy evaluation language, I share @ansasaki concern about ease of use and applicability (I do not think what we have done on mb is particularly usable and maintainable).

While I do not have a proposal for a language ready at hand (I know for a fact that @galmasi and @niteeshkd have been looking into this), I do believe that a crucial property is the ability to combine different policies - as in "layers" - in order to form "meta-policies". 

Our (at least myself @galmasi @niteeshkd) grand vision has always been to provide keylime with the ability of offloading the analysis of a "fully checked" (i.e., replayed against a virtual PCR and compared with the quote) IMA log to an "custom policy engine" capable of evaluating the contents (which could be even an external engine). Following this vision, I believe if we could keep the old "built-in" evaluation and have this one as the first example implementation of this "pluggable" policy, the barriers for acceptance would be lowered.

stefanberger:
> While I do not have a proposal for a language ready at hand (I know for a fact that @galmasi and @niteeshkd have been looking into this), I do believe that a crucial property is the ability to combine different policies - as in "layers" - in order to form "meta-policies".

Then let's see what others have.

stefanberger:
> Following this vision, I believe if we could keep the old "built-in" evaluation and have this one as the first example implementation of this "pluggable" policy, the barriers for acceptance would be lowered.

I would try to avoid enabling multiple pluggable policies if possible. The policies need to be reflected not just in keylime verifier but also in all the tools and the more different policies we have the more we need to maintain... My, $0.02.

ansasaki:
> Rules need a processing order so just concatenating them from two different policies is certainly not possible. They would need to be merged in some 'right' way. It may not always be possible for a machine to do it.

TL;DR: I was thinking if the processing order is fixed and the order of the entries in the "new language" doesn't matter.

Considering the expected order of processing:

1. Process exclude list
2. Process reject list
3. Process signature verification
4. Process allow list
5. Reject everything else

Is there a reason, for example, to accept some entry before checking if it should be ignored?

I was thinking that maybe it doesn't make sense to allow the policy to change the order of operations, which means that the entries of the policy should not be executed in the order they are declared.

This would remove the importance of the order in which the entries appear in the new policy format, as all excluded entries would be processed first, followed by rejected entries, and so on. If this is the case, it would be easy to combine 2 policies by simply concatenating them.

stefanberger:
> > Rules need a processing order so just concatenating them from two different policies is certainly not possible. They would need to be merged in some 'right' way. It may not always be possible for a machine to do it.
> 
> TL;DR: I was thinking if the processing order is fixed and the order of the entries in the "new language" doesn't matter.

If we give users control over the language they can put the policy 'commands' in any order they want to. If we don't give them control over it we don't have to deal with the issue of potentially having to merge such rules. Maybe we should just hard code a reject list into the current code and be done with it.

> 
> Considering the expected order of processing:
> 
>     1. Process exclude list
> 
>     2. Process reject list
> 

I would actually revert the order of the first two now so that an entry that is rejected will never be just accepted by the exclude list.

>     3. Process signature verification
> 
>     4. Process allow list
> 
>     5. Reject everything else
> 
> 
> Is there a reason, for example, to accept some entry before checking if it should be ignored?

Not that I would know a reason atm.

> 
> I was thinking that maybe it doesn't make sense to allow the policy to change the order of operations, which means that the entries of the policy should not be executed in the order they are declared.

If we give users control over the policy language then they should also have control over the processing order. Otherwise let's not give them control over it at all. As mentioned, a valid choice is to continue with the current hard coded processing and not even go near a policy language. Unless we want to give users control over the language there's probably no need to persue it.

> 
> This would remove the importance of the order in which the entries appear in the new policy format, as all excluded entries would be processed first, followed by rejected entries, and so on. If this is the case, it would be easy to combine 2 policies by simply concatenating them.

When I look at a list of rules I look at the ordering as well. I don't think it's very intuitive that the rules wouldn't be processed in the order they are shown.

stefanberger:
Let's aim for a simpler approach.

PR #1543, opened 2024-04-09
PSS padding fix - salt length set according to wrong length

Isaac-Matthews:
The salt length for PSS should be the byte length of the digest getting signed. This has only been working when it was the same size as the signature.

For example, if the item is hashed into 32 bytes (sha256) and signed with that amount, the verification would pass as the salt would receive the correct length (256/8 = 32). However if the digest is 48 bytes, and the signature is 512 then it will fail (512/8 = 64) as a salt length of 64 is longer than 48.

### DIFF
--- a/keylime/tpm/tpm_main.py
+++ b/keylime/tpm/tpm_main.py
@@ -79,7 +79,7 @@ class Tpm:
             if sig_alg in [tpm2_objects.TPM_ALG_RSASSA, tpm2_objects.TPM_ALG_RSAPSS]:
                 try:
                     (signature,) = struct.unpack_from(f"{sig_size}s", iak_sign, 6)
-                    tpm_util.verify(iak_pub, signature, digest, hashfunc, sig_alg, int(sig_size / 8))
+                    tpm_util.verify(iak_pub, signature, digest, hashfunc, sig_alg, hashfunc.digest_size)
                     logger.info("Agent %s AIK verified with IAK", uuid)
                     return True
                 except InvalidSignature:


stefanberger:
You should also write in the  commit what has been working and what has not been working so that, if someone looks for a fix, (s)he finds it.

stefanberger:
> Just to check, do you mean renaming the commit to be more descriptive, i.e. "PSS padding fix" to "PSS padding fix - salt length changed to byte length of digest from length of signature" ?

I think you should describe the problem as to when the verification worked and when it didn't. +1 for a more descriptive title of the patch.

stefanberger:
I think you should put what you wrote above into the patch description:
"
The salt length for PSS should be the byte length of the digest getting signed. This has only been working when it was the same size as the signature.

For example, if the item is hashed into 32 bytes (sha256) and signed with that amount, the verification would pass as the salt would receive the correct length (256/8 = 32). However if the digest is 48 bytes, and the signature is 512 then it will fail (512/8 = 64) as a salt length of 64 is longer than 48.
"

Isaac-Matthews:
> I think you should put what you wrote above into the patch description: " The salt length for PSS should be the byte length of the digest getting signed. This has only been working when it was the same size as the signature.
> 
> For example, if the item is hashed into 32 bytes (sha256) and signed with that amount, the verification would pass as the salt would receive the correct length (256/8 = 32). However if the digest is 48 bytes, and the signature is 512 then it will fail (512/8 = 64) as a salt length of 64 is longer than 48. "

Ah sorry I thought you were referring to the PR description. Is patch description == commit message or should I be looking somewhere else?

stefanberger (approved):

PR #1545, opened 2024-05-06
Add support for a reject list in runtime policy

stefanberger:
This PR adds support for a reject list in the runtime policy. This list is actually a map with filenames and hashes.
The user can add entries to the reject list using the keylime_create_policy tool from either a (trimmed) IMA measurement list or from local RPMs. 

The assumption is that a user who wants to have keylime verify the signature on files may also want to prevent certain packages from being used, such as those that contain applications with CVEs. The number of packages added to the reject list should therefore be kept manageable/small, which may be in turn be a challenge if certain packages for example have had CVEs for a while.


keylime/ima/ima.py
```diff
@@ -42,7 +42,7 @@ class RUNTIME_POLICY_GENERATOR(enum.IntEnum):
 # the kernel maintainers decide to use a different algorithm for template-hash.
 EMPTY_RUNTIME_POLICY: RuntimePolicyType = {
     "meta": {
-        "version": RUNTIME_POLICY_CURRENT_VERSION,
+        "version": 1,  # Does not contain fields added after v1
```
aplanas, line 45: Should still be `RUNTIME_POLICY_CURRENT_VERSION`, and `rejects` should be added, shouldn't?  

Why we want to generate both v1 and v2 policy depending on the fields? This will make the schema evolution hard to track

stefanberger: > Why we want to generate both v1 and v2 policy depending on the fields? This will make the schema evolution hard to track

The primary reason for this was that someone creating a v2 policy without a rejects entry or an empty reject map should still be able to use the policy with a previous version of keylime.

aplanas: I understand, yes. But how it is done now will make the future changes of the schema hard to maintain.

The json-schema has this "required" field to list the required elements, and "rejects" is not there in your change, so is optional and the parser will be OK with moving the version to 2 unconditionally

There is a cmd/convert_runtime_policy.py that will do the conversion for us, in direction 1->2 if required.

stefanberger: I changed the version now in the EMPTY_RUNTIME_POLICY and added an empty rejects map. Wherever this EMPTY_RUNTIME_POLICY is now being used as a base policy and where the policy is written out later on (file or DB) I am now also calling ima.trim_runtime_policy() to get the version to the minimum that's required. If that's also undesirable I can drop it but you cannot use the policy on older versions of keylime anymore, even if there's no reason it is marked as a version 2 (because it uses no features of v2).

keylime/cloud_verifier_tornado.py
```diff
@@ -631,6 +631,7 @@ def post(self) -> None:
                     if not runtime_policy_name and not runtime_policy:
                         logger.info("IMA policy data not provided with request! Using default empty IMA policy.")
                         runtime_policy = json.dumps(cast(Dict[str, Any], ima.EMPTY_RUNTIME_POLICY))
+                        ima.trim_runtime_policy(runtime_policy, False)
```
aplanas, line 634: The code the convert the policy should not be out of the creator (create_policy.py, ima.py) or converter (convert_runtime_policy.py). 

I mean, maybe the next change of policy is not about trimming, but about extending, or translating. IMHO this piece should take the policy as such and not do any conversion.

keylime/cmd/convert_runtime_policy.py
```diff
@@ -123,9 +123,6 @@ def update_runtime_policy(
     policy: RuntimePolicyType, excludelist_path: Optional[str] = None, verification_keys: Optional[List[str]] = None
 ) -> RuntimePolicyType:
     if policy["meta"]["version"] < ima.RUNTIME_POLICY_CURRENT_VERSION:
-        print(
-            f"Provided policy has version {policy['meta']['version']}; latest policy has version {ima.RUNTIME_POLICY_CURRENT_VERSION}. Updating to latest version."
-        )
```
aplanas, line 128: That is a fair notice, why not to keep it?

keylime/cmd/convert_runtime_policy.py
```diff
@@ -214,6 +211,7 @@ def main() -> None:
     policy_out = update_runtime_policy(
         policy, excludelist_path=args.excludelist, verification_keys=args.verification_keys
     )
+    ima.trim_runtime_policy(policy_out, False)
```
aplanas, line 214: Somehow I think that the "trimming" code should be part of the conversion, and should live here and not in the IMA module.

keylime/cmd/create_policy.py
```diff
@@ -531,6 +543,8 @@ def main() -> None:
             ignored_keyrings = base_policy.get("ima", {}).get("ignored_keyrings", [])
             policy["ima"]["ignored_keyrings"] = ignored_keyrings
             policy["verification-keys"] = base_policy.get("verification-keys", "")
+            if base_policy.get("rejects", None) is not None:
+                policy["rejects"] = base_policy.get("rejects", {})
```
aplanas, line 547: Why not to create v2 policies here? Seems that `policy["rejects"] = base_policy.get("rejects", {})` will do the right thing is "rejects" is missing from the base_policy

keylime/cmd/create_policy.py
```diff
@@ -589,7 +611,7 @@ def main() -> None:
     policy["meta"]["timestamp"] = str(datetime.datetime.now())
 
     try:
-        ima.validate_runtime_policy(policy)
+        ima.trim_runtime_policy(policy, True)
```
aplanas, line 614: IMHO should be a call to the validator, the trim code seems to makes sense only in the converter side

keylime/migrations/versions/8c0f8ded1f90_convert_allowlists_to_ima_policies.py
```diff
@@ -81,6 +81,7 @@ def upgrade_cloud_verifier():
                     logger.info("Runtime policy field '%s' not found in existing allowlist; using default value", key)
                 else:
                     new_ima_policy[key] = alist_json[key]
+        ima.trim_runtime_policy(new_ima_policy, False)
```
aplanas, line 84: I do not think that migration code can be changed

keylime/ima/ima.py
aplanas: Ah, I read this comment after writing mines, so I lacked the context provided here.

I think your goal should be kept, but I am not sure that the current approach is easy to maintain the the future. For me it is OK to move to v2 policy, even if is just a new optional field. Moving all to v2 and converting v1 to v2 should be OK too.

Let me check some bits in the codebase and I will try to comeback with a better comment.

keylime/cmd/convert_runtime_policy.py
stefanberger: Because we may not upgrade it to v2 after trimming.

keylime/cmd/convert_runtime_policy.py
stefanberger: It's called from many places.

keylime/cmd/create_policy.py
stefanberger: Ok, will fix this.

keylime/migrations/versions/8c0f8ded1f90_convert_allowlists_to_ima_policies.py
stefanberger: This is to prevent writing a policy with v2 and then failing to run with the policy after downgrade to code that only supports v1 where we may check for a version number and not support a v2 policy.

keylime/cmd/create_policy.py
stefanberger: Ok, so you are saying if we generate a policy here that is v2 and this policy has no features of a v2 policy then this is fine. Problems with this policy may occur if we were to downgrade keylime to pre-v2  or  want to use the policy on another keylime that only understands v1.

keylime/migrations/versions/8c0f8ded1f90_convert_allowlists_to_ima_policies.py
stefanberger: I have to say I haven't tested this and it's not so easy to test anyway but running into trouble with a v2 policy in the DB while keylime only understands v1 (after a downgrade) would be bad. Only a true v2 policy, one that uses the new features of the policy, in the DB should cause issues. Keylime should refuse to use it since keylime won't behave correctly: does not reject files as expected.

keylime/cmd/create_policy.py
aplanas: Right. My main issue can be summarized as: What to do with policy v3, v4, ..., vN?

The current proposal does not work in the general case, it is an ad-hoc instance for v1 <-> v2 (trim, conflating validation, keeping the version number, etc). Honestly the pre-patched code also does not provide a nice solution to this, but seems to suggest one: work always with the last version.

> to downgrade keylime to pre-v2 or want to use the policy on another keylime that only understands v1.

OK. I accept that using the minimal version can be a valid approach. I will review based on this.

keylime/cloud_verifier_tornado.py
stefanberger: The problem in this case is that the ima.EMPTY_RUNTIME_POLICY unnecessarily is version 2 even though it only has an empty reject list. And then it may get written into the DB and prevent a should prevent a downgrade of the DB since we now have a v2 policy in the DB that no prior version of keylime knows how to deal with (keylime should reject using it since it wouldn't behave properly since it doesn't know what was added in v2).

PR #1546, opened 2024-05-07
CI: Enable Packit testing on centos-stream-10

kkaarreell:


kkaarreell:
I think it is too early to do this change. Let's postpone it.

PR #1551, opened 2024-06-07
add config options for the persisted idevid and iak handles, passwords

### LINKED ISSUE #1549: Add IDevID/IAK password options to configuration upgrade template

# Environment
 * OS / version: N/A
 * Processor architecture: N/A
 * TPM Manufacturer: N/A
 * Keylime version: current master (829169e)

# Description 
New configuration options were added to the agent configuration as part of the PR https://github.com/keylime/rust-keylime/pull/785:

* `idevid_password`
* `idevid_handle`
* `iak_password`
* `iak_handle`

Expected: The new options are part of the configuration upgrade templates
Actual: The new options are not included in any configuration upgrade template

## Steps to reproduce problem
1.  search for the missing options in all configuration versions in `templates` directories

# Relevant logs

Isaac-Matthews:
Fixes #1549 

Adds config options to the templates for the IDevID and IAK handles if the keys have been persisted, and their passwords if they are password protected.

These are the template changes for the Rust agent changes here [PR 785](https://github.com/keylime/rust-keylime/pull/785).

### DIFF
--- a/templates/2.3/agent.j2
+++ b/templates/2.3/agent.j2
@@ -169,7 +169,10 @@ tpm_signing_alg = "{{ agent.tpm_signing_alg }}"
 # create a new EK upon startup, and neither will it flush the EK upon exit.
 ek_handle = "{{ agent.ek_handle }}"
 
-# Enable IDevID and IAK usage and set their algorithms.
+# Enable IDevID and IAK usage 
+enable_iak_idevid = {{ agent.enable_iak_idevid }}
+
+# Select IDevID and IAK templates or algorithms for regenerating the keys.
 # By default the template will be detected automatically from the certificates. This will happen in iak_idevid_template is left empty or set as "default" or "detect".
 # Choosing a template will override the name and asymmetric algorithm choices. To use these choices, set iak_idevid_template to "manual"
 # Templates are specified in the TCG document found here, section 7.3.4: 
@@ -179,11 +182,19 @@ ek_handle = "{{ agent.ek_handle }}"
 # iak_idevid_template:        default, detect, H-1, H-2, H-3, H-4, H-5, manual
 # iak_idevid_asymmetric_alg:   rsa, ecc
 # iak_idevid_name_alg:        sha256, sm3_256, sha384, sha512
-enable_iak_idevid = {{ agent.enable_iak_idevid }}
 iak_idevid_template = "{{ agent.iak_idevid_template }}"
 # In order for these values to be used, set the iak_idevid_template option to manual
 iak_idevid_asymmetric_alg = "{{ agent.iak_idevid_asymmetric_alg }}"
 iak_idevid_name_alg = "{{ agent.iak_idevid_name_alg }}"
+
+# Alternatively if the keys are persisted, provide the handles for their location below, and optionally their passwords.
+# If handles are provided, they will take priority over templates/algorithms selected above.
+# To use a hex password, use the prefix "hex:" at the start of the password.
+idevid_password = "{{ agent.idevid_password }}"
+idevid_handle = "{{ agent.idevid_handle }}"
+
+iak_password = "{{ agent.iak_password }}"
+iak_handle = "{{ agent.iak_handle }}"
 
 # The name of the file containing the X509 IAK certificate.
 # If set as "default", the "iak-cert.crt" value is used
--- a/templates/2.3/mapping.json
+++ b/templates/2.3/mapping.json
@@ -38,6 +38,14 @@
                     "default": "['ecschnorr', 'rsassa']"
                 }
             }
+        },
+        "agent": {
+            "add" : {
+                "idevid_password": "",
+                "idevid_handle": "",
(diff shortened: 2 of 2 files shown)


maugustosilva:
Given that https://github.com/keylime/rust-keylime/pull/785 was already merged, I see no reason to delay the merging of this PR.

However, I do see an error on `fedora-rawhide` which is affecting all PRs. @kkaarreell I had a brief discussion with @ansasaki and he mentioned that perhaps we should no longer block on `rawhide` failures.

maugustosilva (approved):

kkaarreell:
> However, I do see an error on `fedora-rawhide` which is affecting all PRs. @kkaarreell I had a brief discussion with @ansasaki and he mentioned that perhaps we should no longer block on `rawhide` failures.

We can either keep Rawhide tests running but ignore known failures or disable Rawhide tests completely. I would prefer to do the earlier since those Rawhide failures may give us early warning, the issue would most likely appear in Fedora stable later.

However, would you know why tests are failing in this particular case? In the 2nd test I see
```
:: [ 10:42:13 ] :: [  BEGIN   ] :: Running 'rpmsign --addsign --signfiles --fskpath=/etc/keys/privkey_evm.pem /root/rpmbuild/RPMS/noarch/rpm-ima-sign-test-1-1.noarch.rpm'
hash(sha384): 562eb1111ef3e3c2ec9c60e53b27af259ff984f31a804d97bfd49df13e843128
sign_hash_v2: signing failed: (invalid digest length) in EVP_PKEY_sign
openssl: error:1C8000A6:Provider routines::invalid digest length
error: sign_hash failed
error: signFile failed
/root/rpmbuild/RPMS/noarch/rpm-ima-sign-test-1-1.noarch.rpm:
```
which might be a but in the package or in the test, I will look into it.
However, for  /compatibility/basic-attestation-on-localhost-api-version-bump I do not see the issue there. The tests expects the verifier to log 'API version updated' but this is not happening. The agent is using v2.1, the verifier seems to use 2.2. Also, when starting agent with v2.1 again the failure is expected but it is not happening. It looks like the verifier uses also 2.1 but logs says:
```
keylime.verifier - INFO - Current API version 2.2
```

@ansasaki Would you know what's happening here?

ansasaki:
> However, for /compatibility/basic-attestation-on-localhost-api-version-bump I do not see the issue there. The tests expects the verifier to log 'API version updated' but this is not happening. The agent is using v2.1, the verifier seems to use 2.2. Also, when starting agent with v2.1 again the failure is expected but it is not happening. It behaves like if the verifier would be using also 2.1 but the log says:
> 
> ```
> keylime.verifier - INFO - Current API version 2.2
> ```
> 
> @ansasaki Would you know what's happening here?

This is weird. What the test does is:

1. Backup the installed agent binary (from the log `keylime-agent-rust-0.2.5-1.20240508132448021528.master.4.gc91fba3.fc41.x86_64` was installed)
2. Compile the "old" version and replace the binary
3. Run the old version to register the old API
4. Stop the agent, restore the original binary from backup, and run to trigger the bump detection
5. Stop the agent, replace the binary again with the old version, and run to check that it does not allow rollback

Looking into the logs, the "old" version was compiled and run correctly with the old API version. Then, the new binary should be restored from the backup and run with then new API version to trigger the bump detection, but it is still using the API version `2.1`.

Since the preinstalled version is correct, it should be running with the new API version after restoring the backup. My suspicion is that maybe the preinstalled binary, for some reason, is in another place in Rawhide (not in `/usr/bin/keylime_agent`). I'll investigate.

kkaarreell:
I have tracked all these issue down to different packages. I believe it is safe to merge this PR.
The rust agent API related issue is being addressed in https://github.com/RedHat-SP-Security/keylime-tests/pull/588,
there are more issues to fix in Rawhide though.

ansasaki:
Just a note that the changes here are only valid while the version `7.11.0` is not released since it would release the config version `2.3`.

If this PR is not included in `7.11.0`, the correct approach is to introduce a new config version `2.4`

@maugustosilva FYI

maugustosilva:
@kkaarreell I have no fundamental opposition on keep rawhide running in an "advisory" capacity (i.e., a failure there does not block anything). I will just have to formalize it at our next monthly meeting. Thanks.

kkaarreell:
> @kkaarreell I have no fundamental opposition on keep rawhide running in an "advisory" capacity (i.e., a failure there does not block anything). I will just have to formalize it at our next monthly meeting. Thanks.

OK. This is a "risk" of running e2e tests, they are dependencies outside of the keylime scope and any of them can introduce a regression. This time rpm (btw, https://bugzilla.redhat.com/show_bug.cgi?id=2291183 has a fix already), usually it is selinux-policy (we ignore AVCs on Rawhide) or some tpm2-* package. Rawhide is just a bit more exposed but it may happen on other Fedoras or CS as well. Anyway, I think things are on a good track and we should see Rawhide tests green again in a few days.

PR #1560, opened 2024-06-20
Change Docker and Action Tags to Digests

lukehinds:
Currently the Keylime's dockerfiles and action workflows are using tags over digests.

Tags are risky. Pinning to digests for Actions and Dockerfiles ensures you are using a specific, verified version of a dependency, otherwise its possible someone could replace the code for an existing tag, which could lead to certain attacks being triggered by the workflow runs.

Check out [GitHubs documentation](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions) and [the following](https://candrews.integralblue.com/2023/09/always-use-docker-image-digests/) for Dockerfiles

I will make two follow up PR's. One to introduce Dependabot to keep the actions updated. The 2nd will be for Frizbee-action, an OSS tool that will automatically flip tags for any new actions added.

If you like we can add 'pip' as an ecosystem for dependabot to keep the Python deps updated.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -27,11 +27,11 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v3
+      uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@v2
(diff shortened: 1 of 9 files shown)


THS-on:
Nice to see you back contributing to Keylime! Changes LGTM.

> If you like we can add 'pip' as an ecosystem for dependabot to keep the Python deps updated.

Not sure if we can easily do this, as the lowest version is based on the oldest distribution we support

THS-on (approved):

maugustosilva (approved):

PR #1566, opened 2024-06-28
revocation_notifier: Take into account CA certificates added via configuration

ansasaki:
Previously, the revocation notifier would use only the system installed CA certificates, ignoring those added via `trusted_ca_certificates` configuration. Also, if the webhook didn't offer a certificate, the revocation notifier would ignore the missing certificate and continue the communication in plaintext, even when TLS is enabled via configuration. This can be dangerous if sensitive data is sent to the webhook.

This change makes the revocation notification to use the configured TLS options.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -40,9 +40,9 @@ jobs:
     - name: Install build dependencies
       run: sudo dnf -y install swig
     - name: Install Python dependencies
-      run: sudo dnf -y install python3.6 tox python3-pip
+      run: sudo dnf -y install python3.10 tox python3-pip
     - name: Run lints
-      run: tox -vv -e 'pylint'
+      run: tox -vv -e 'pylint,pylint310'
     - name: Run mypy
       run: tox -vv -e 'mypy'
     - name: Run pyright
--- a/keylime/requests_client.py
+++ b/keylime/requests_client.py
@@ -1,3 +1,4 @@
+import re
 import ssl
 from typing import Any, Dict, Optional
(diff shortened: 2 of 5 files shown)


kkaarreell:
What about simply passing the default trust store location to post()?
```
 response = session.post(url, json=tosend, timeout=5, verify=requests.cert.where())
```

ansasaki:
> What about simply passing the default trust store location to post()?
> 
> ```
>  response = session.post(url, json=tosend, timeout=5, verify=requests.cert.where())
> ```

This would fix the current issue, indeed. It would keep current functionality.

What it does not do is to account for additional CA certificates that are added through `trusted_server_ca` option, it will only consider the system installed certificates.

ansasaki:
@kkaarreell I'll move this PR to draft and create a specific issue for what I'm trying to fix here:

- Take into account CA certificates configured on `trusted_server_ca` for the webhook notification connection
- Do not allow plaintext communication with the webhook server when TLS is enabled

I'll open a separate PR with the workaround for the requests CA certificate issue

galmasi (approved):
IMHO this PR should be merged. A revocation that will not be delivered because of a missing CA is not worth having -- it makes revocation unreliable.

maugustosilva (approved):

ansasaki:
Sorry, this shouldn't be merged before dropping the temporary commit. I'll fill another PR reverting that.

ansasaki:
Opened https://github.com/keylime/keylime/pull/1619

PR #1568, opened 2024-07-01
Enhancement 110: add tool to handle keylime policies 

sergio-correia:
Initial steps of enhancement #110, which is to provide a single
tool that provides the functionality of the existing tools we
have.

In this commit, we add the `keylime_policy` tool, that is currently
able to handle the creation of runtime policies.

To see the available options:

`keylime_policy --help`

To see the options available related to creating runtime policies:

`keylime_policy create runtime --help`

Some of the options available include e.g. using an IMA measurement
list, legacy allowlist/excludelist files, local/remote RPM repositories,
specifying a rootfs filesystem, initrds, a base policy, etc.

PR #1573, opened 2024-07-09
docker: Update images to use Fedora 40

ansasaki:
The Fedora 37 is no longer supported. This updates the images to be based on Fedora 40.

Also add some small changes to fix the `build_locally.sh` script by disabling SELinux relabeling during the build.

EDIT: Added test dependencies needed for #1568

### DIFF
--- a/docker/ci/Dockerfile
+++ b/docker/ci/Dockerfile
@@ -5,9 +5,9 @@
 # It is not recommended for use beyond testing scenarios.
 ##############################################################################
 
-FROM quay.io/fedora/fedora:37-x86_64@sha256:eab075950eeef382b80b6c2ffb693381e6a7c507d3337949197c37ad244842e9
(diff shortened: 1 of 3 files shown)


mpeters:
I definitely approve of this change and will merge it after the test failures are figured out.

mpeters (approved):

maugustosilva:
@ansasaki We can ignore the Rawhide failure in this case, can't we?

maugustosilva (approved):

ansasaki:
> @ansasaki We can ignore the Rawhide failure in this case, can't we?

I think we can. Unfortunately, Rawhide changed something that broke the tests again. This is expected as Rawhide is not stable. The goal of keeping the tests running is to aware of future changes to the stable releases that may affect Keylime.

I'll rebase and then I think it would be ready to merge regardless of failures in Rawhide

maugustosilva:
`Rawhide` is the only remaining failure, LGTM

PR #1580, opened 2024-07-15
ci: use CODECOV_TOKEN for coverage file upload

kkaarreell:


kkaarreell:
FYI, I have asked @maugustosilva to modify CODEDOV_TOKEN, once this is done I will rerun the upload to confirm it is working properly.

kkaarreell:
I just realized that secrets won't work for PRs created from a fork. I have tried to define that secret in my fork as well so maybe that will help, let's wait for the results. If not, we may consider merging this without "verification".

kkaarreell:
Workflow log still says: `[2024-07-15T16:25:38.995Z] ['info'] -> No token specified or token is empty` but I believe that I have configured it properly and the problem is that I am running the workflow from a fork. Can we try to merge it to see if it truly works?

ansasaki:
The tests on Rawhide are still failing. I'll ignore them for now and merge this.

PR #1583, opened 2024-07-15
Verify that KDFa = KBKDF, KDFe = ConcatKDF, and point multiply = ECDH

ansasaki:
Add tests to show that:
 - `KDFa` is equivalent to `KBKDFHMAC` implemented by python-cryptography
 - `KDFe` is equivalent to `ConcatKDFHash` implemented by python-cryptography
 - The point multiplication implemented locally is equivalent to ECDH implemented by python-cryptography

This shows that the local implementation of `KDFa` and `KDFe` can be replaced by the KDFs implemented in python-cryptography

EDIT: added a test case to show that the point multiplication implemented locally is equivalent to ECDH implemented in python cryptography

### DIFF
--- a/keylime/tpm/tpm_util_test.py
+++ b/keylime/tpm/tpm_util_test.py
@@ -5,6 +5,7 @@ from unittest import mock
 
 from cryptography.exceptions import InvalidSignature
 from cryptography.hazmat.primitives import hashes
+from cryptography.hazmat.primitives.asymmetric import ec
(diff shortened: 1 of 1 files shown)


stefanberger (approved):

ansasaki:
My goal with this PR was to provide a quick proof that we can replace the locally implemented crypto operations with the implementation from python-cryptography. There is no need to merge these tests.

PR #1585, opened 2024-07-15
tpm: Replace KDFs and ECDH implementations with python-cryptography

ansasaki:
Replace the local implementations of the KDFa, KDFe, and point multiplication with the respective implementations of KBKDFHMAC, ConcatKDFHash, and ECDH from python-cryptography.


stefanberger (approved):

THS-on:
@ansasaki the linter errors are in files not touched by this PR, so can be ignored.

@stefanberger @ansasaki this test covers that the generation of makecredential that uses those functions is correct, right?

https://github.com/keylime/keylime/blob/b81832fd3c629b10b29c8130b24b03c2ae99809f/keylime/tpm/tpm_util_test.py#L126-L151

stefanberger:
> @stefanberger @ansasaki this test covers that the generation of makecredential that uses those functions is correct, right?

I would say that it is correct. Any concerns that it is not correct?

THS-on:
> I would say that it is correct. Any concerns that it is not correct?

No just was just checking that we cover all the changes with a test, as it is fundamental to TPM attestation.

THS-on (approved):

PR #1593, opened 2024-07-25
ima: list names of the runtime policies

niteeshkd:
Add listruntimepolicy option with 'keylime_tenant -c' to list names of the runtime policies.

### DIFF
--- a/keylime/cloud_verifier_tornado.py
+++ b/keylime/cloud_verifier_tornado.py
@@ -856,53 +856,67 @@ class AllowlistHandler(BaseHandler):
     def head(self) -> None:
         web_util.echo_json_response(self, 400, "Allowlist handler: HEAD Not Implemented")
 
-    def __validate_input(self, method: str) -> Optional[str]:
+    def __validate_input(self, method: str) -> Tuple[bool, Optional[str]]:
         """Validate the input"""
         if self.request.uri is None:
             web_util.echo_json_response(self, 400, "Invalid URL")
-            return None
+            return False, None
         rest_params = web_util.get_restful_params(self.request.uri)
         if rest_params is None or "allowlists" not in rest_params:
             web_util.echo_json_response(self, 400, "Invalid URL")
-            return None
+            return False, None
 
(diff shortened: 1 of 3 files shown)


THS-on:
@niteeshkd can you also include an update to the documentation to reflect this new API endpoint?

niteeshkd:
> @niteeshkd can you also include an update to the documentation to reflect this new API endpoint?

@THS-on which document should I update? I can't find any document related to named runtime policies under docs/ .

niteeshkd:
> @niteeshkd REST APIs are documented here: https://github.com/keylime/keylime/blob/master/docs/rest_apis.rst

Thanks! I just updated that doc.

keylime/tenant.py
```diff
@@ -1360,6 +1360,15 @@ def do_show_runtime_policy(self, name: Optional[str]) -> None:  # pylint: disabl
         if response.status_code >= 400:
             raise UserError(response_json)
 
+    def do_list_runtime_policy(self) -> None:  # pylint: disable=unused-argument
```
THS-on, line 1363: Instead of the pylint ignore, you can declare this a static method.

keylime/cloud_verifier_tornado.py
```diff
@@ -856,62 +856,76 @@ class AllowlistHandler(BaseHandler):
     def head(self) -> None:
         web_util.echo_json_response(self, 400, "Allowlist handler: HEAD Not Implemented")
 
-    def __validate_input(self, method: str) -> Optional[str]:
+    def __validate_input(self, method: str) -> Tuple[bool, Optional[str]]:
```
THS-on, line 859: The change here is a bit confusing, but I also don't see a simpler way to make it nicer.

keylime/tenant.py
niteeshkd: Pylint ignore was not needed.

maugustosilva:
The `lint` errors here, very much like #1607, are unrelated to the code changes. We should not hold off this merge because of it.

maugustosilva (approved):

PR #1601, opened 2024-07-31
workflow/base-image: Drop duplicated job ID

ansasaki:
The jobs to build the base image and CI image had the same id. Since the jobs are not referred anywhere, we can drop the id from both jobs.


sergio-correia (approved):
LGTM, thanks

Issue #1541, opened 2024-04-05
keylime_tenant seems to read verifier.conf


kkaarreell:
Is your issue a feature request? If so, please raise it as an [enhancement](https://github.com/keylime/enhancements)

# Environment
 * OS / version: Fedora 40
 * Processor architecture: x86_64
 * TPM Manufacturer: swtpm
 * Keylime version: current upstream [ea0cc9f](https://github.com/keylime/keylime/commit/ea0cc9f9522c1a0c753e10272fba90df75dd3a42)

# Description 
in keylime_tenant outout one can see that verifier.conf is read.
For example:
```
$ keylime_tenant -v 127.0.0.1 -t 127.0.0.1 -u d432fbb3-d2f1-4a97-9ef7-75bd81c00000 --verify --runtime-policy policy.json --cert default -c add

INFO:keylime.config:Reading configuration from ['/etc/keylime/logging.conf']
2024-03-20 13:05:58.186 - keylime.config - INFO - Reading configuration from ['/etc/keylime/tenant.conf']
2024-03-20 13:05:58.187 - keylime.tenant - INFO - Setting up client TLS...
2024-03-20 13:05:58.187 - keylime.tenant - INFO - Using default client_cert option for tenant
2024-03-20 13:05:58.187 - keylime.tenant - INFO - Using default client_key option for tenant
2024-03-20 13:05:58.187 - keylime.tenant - INFO - No value provided in client_key_password option for tenant, assuming the key is unencrypted
2024-03-20 13:05:58.194 - keylime.tenant - INFO - TLS is enabled.
2024-03-20 13:05:58.194 - keylime.config - INFO - Reading configuration from ['/etc/keylime/verifier.conf']
2024-03-20 13:05:58.240 - keylime.cli.policies - INFO - TPM PCR Mask from policy is 0x0
2024-03-20 13:05:58.240 - keylime.config - INFO - Reading configuration from ['/etc/keylime/ca.conf']
2024-03-20 13:05:58.240 - keylime.ca-util - WARNING - The 'default' password option from CA configuration file cannot be used with keylime CLI (keylime_tenant or keylime_ca)
...
```
Previously, I have reported similar issue https://github.com/keylime/keylime/issues/1446, I don't know if this is related though.

## Expected behavior vs. actual behavior 
keylime_tenant should not access verifier.conf, probably even ca.conf

## Steps to reproduce problem
1.  run keylime_tenant command

kkaarreell:
[...]

PR #1607, opened 2024-08-07
mba: Add a skip custom policies option when loading mba.

marcostork:
Allowing the skipping of custom policies, known only to the verifier, will enable the tenant to call MBA without consulting the verifier config file.

Resolves: #1541

### DIFF
--- a/keylime/mba/mba.py
+++ b/keylime/mba/mba.py
@@ -56,17 +56,20 @@ _mba_imports = []
 # ###########
 
 
-def load_imports() -> None:
+def load_imports(skip_custom_policies: Optional[bool] = False) -> None:
     """
     MBA API front-end for importing any modules needed by measured boot attestation.
-    No inputs.
+    :param skip_custom_policies: may be used to prevent the loading of custom policies.
     No outputs.
     Side effects: Reads from keylime configuration files.
       After execution all imports required by MBA are loaded and initialized.
     Exceptions: If any policy engine in the list cannot be loaded.
     """
     try:
-        imports = config.getlist("verifier", "measured_boot_imports")
+        imports: List[Any] = []
+        if not skip_custom_policies:
+            # import custom policies
+            imports = config.getlist("verifier", "measured_boot_imports")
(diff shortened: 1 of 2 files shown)


ansasaki (changes requested):
Could you please update the `load_imports()` documentation comments to reflect the changes?

And you need to add you sign-off to all commits to pass the DCO test

ansasaki:
The failint linting issues are unrelated with the changes.

@kkaarreell Would it be possible to confirm this makes the tenant to not load the `verifier.conf anymore`

kkaarreell:
> The failint linting issues are unrelated with the changes.
> 
> @kkaarreell Would it be possible to confirm this makes the tenant to not load the `verifier.conf ` anymore?

Hi, according to tenant logs it no longer reads verifier.conf, it still accesses ca.conf though.

ansasaki:
> > The failint linting issues are unrelated with the changes.
> > @kkaarreell Would it be possible to confirm this makes the tenant to not load the `verifier.conf ` anymore?
> 
> Hi, according to tenant logs it no longer reads verifier.conf, it still accesses ca.conf though.

This is expected, the `ca.conf` and `logging.conf` were never separated per component. Maybe this is an improvement we should aim to fix. In my opinion, it makes sense to be able to configure them separately per component.

ansasaki (approved):
I think with the proposed changes we make a compromise: the quote is still verified by the tenant, but will not support special modules to be used by the measured boot policy, only the default modules.

And this fixes the regression of the tenant requiring the `verifier.conf`.

@maugustosilva Would this be ok for you?

maugustosilva:
This is exactly what we wanted as the compromise (i.e., quote has to be checked, PCR#16 and all that), and if the `lint` issues are not related to this change LGTM.

maugustosilva (approved):

maugustosilva:
Ok, the `lint` errors are not related to this PR. I will merge and carve out some time to open a separated PR to deal with it.

PR #1618, opened 2024-08-15
Lint: ignore reportArgumentType and reportInvalidTypeForm errors

marcostork:
Ignoring pyright reportArgumentType and reportInvalidTypeForm errors until a better solution can be found.


maugustosilva:
@ansasaki @sergio-correia I believe this temporarily fixes the problem, as we previously discussed. In a new major release, when we have the opportunity to increase the minimum version of python, we can go back and remove these pragmas.

maugustosilva (approved):

ansasaki (approved):

Issue #1564, opened 2024-06-26
The `verification-keys` field in runtime policy is a string, but the specification expected an array


ansasaki:
Is your issue a feature request? If so, please raise it as an [enhancement](https://github.com/keylime/enhancements)

# Environment
 * OS / version: N/A
 * Processor architecture: N/A
 * TPM Manufacturer: N/A
 * Keylime version: current master (691f713)

# Description 
The enhancement that introduced runtime policies specified the `verification-keys` as an array:

https://github.com/keylime/enhancements/blob/2a462b60a1f40f257abf728af3269c1301752adc/70_ima_policy.md?plain=1#L241-L246

The code implemented the field as a single string:

https://github.com/keylime/keylime/blob/0b57e6c97cbfde83d2ab4a04952e2cf33b550e6f/keylime/ima/ima.py#L89

## Expected behavior vs. actual behavior 
Expected: The implemented schema should follow the specified schema
Actual: The implemented schema differs from the specified schema 

## Steps to reproduce problem
1. Compare the [specification](https://github.com/keylime/enhancements/blob/2a462b60a1f40f257abf728af3269c1301752adc/70_ima_policy.md?plain=1#L241-L246) with the [implementation](https://github.com/keylime/keylime/blob/0b57e6c97cbfde83d2ab4a04952e2cf33b550e6f/keylime/ima/ima.py#L89)

# Relevant logs

Attach any relevant log files that can help to debug your issue.

ansasaki:
@mbestavros @THS-on Could you please comment about this?

ansasaki:
The changes required to fix this are extensive, with no benefits: in the end, using a JSON string is actually more flexible than replacing it with a list of  the public keys.

I propose we modify the specification documentation to align it with what was actually implemented. Also, the changes should include a schema validation of the JSON used in the `verification-keys`

THS-on:
@ansasaki yes agree, then lets go this route.

ansasaki:
The JSON schema validation for the `verification-keys` content was added as part of https://github.com/keylime/keylime/pull/1638, via commit c724a9d51d527bdbeba0840d398d017325b46fdf

The missing part is fixing the documentation to reflect the actual implemented schema

PR #1626, opened 2024-08-28
keylime_policy: setting `log_hash_alg` to `sha1` (template-hash algo)

### LINKED ISSUE #1609: `keylime_policy` claim to support `sm3_256` digest algorithm, but does not

# Environment
 * OS / version: N/A
 * Processor architecture: N/A 
 * TPM Manufacturer: N/A
 * Keylime version: current master (dc118c780a06cb72725f72bc8100665a07a5a12f)

# Description 
`keylime_policy create runtime --help` shows that the `sm3_256` algorithm is supported, but it fails when it is provided.
The unsupported algorithm should be dropped.

Actual:

`keylime_policy create runtime --help` shows in the `--algo` option the choices `{sha1,sha256,sha384,sha512,sm3_256}`

When running the policy generation with the `sm3_256` algorithm, an error is generated, similar to:

`2024-08-08 08:11:23.399 - policy.create_runtime_policy - ERROR - Base policy is not a valid runtime policy: 'sm3_256' is not one of ['sha1', 'sha256', 'sha384', 'sha512']`

Expected:

`keylime_policy create runtime --help` shows in the `--algo` option only the supported choices `{sha1,sha256,sha384,sha512}`

## Steps to reproduce problem
1. Run '`keylime_policy create runtime --rootfs some_dir --algo sm3_256`'

# Relevant logs
An error similar to the following is returned

`2024-08-08 08:11:23.399 - policy.create_runtime_policy - ERROR - Base policy is not a valid runtime policy: 'sm3_256' is not one of ['sha1', 'sha256', 'sha384', 'sha512']`

### LINKED ISSUE #1623: keylime_policy create runtime requires ascii_runtime_measurements even when processing allowlist

# Environment
 * OS / version: CentOS Stream 10
 * Processor architecture: c86_64
 * TPM Manufacturer:  swtpm
 * Keylime version: upstream ad79ae8
 
# Description 
When building runtime policy from allowlist it should not be necessary to access ascii_runtime_measurements file due to boot_aggregation since this value is already in the allowlist.

```
# cat allowlist.txt 
ccd4fd4876d9f5651259d3cd0c1e1887497ea978588693e50e9553cf717dd688  /keylime-tests/basic-attestation-on-localhost-efrM8/good-script2.sh
670f5af88685baeb32096ba9b49ed3759f9b6bfb6f80dc0f46bc8278cfdc92a8  /keylime-tests/basic-attestation-on-localhost-efrM8/good-script1.sh
0000000000000000000000000000000000000000000000000000000000000000  boot_aggregate
# keylime_policy create runtime -a allowlist.txt
  File "/usr/local/bin/keylime_policy", line 33, in <module>
    sys.exit(load_entry_point('keylime==7.11.0', 'console_scripts', 'keylime_policy')())
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/keylime-7.11.0-py3.11.egg/keylime/cmd/keylime_policy.py", line 53, in main
    args.func(args)
  File "/usr/local/lib/python3.11/site-packages/keylime-7.11.0-py3.11.egg/keylime/policy/create_runtime_policy.py", line 750, in create_runtime_policy
    algo, _ = boot_aggregate_from_file(args.ima_measurement_list)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/keylime-7.11.0-py3.11.egg/keylime/policy/create_runtime_policy.py", line 320, in boot_aggregate_from_file
    with open(ascii_runtime_file, "r", encoding="UTF-8") as f:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/sys/kernel/security/ima/ascii_runtime_measurements'

## Steps to reproduce problem
1. try to build a policy from allowlist on a system that doesn't have kernel IMA configured

marcostork:
Set the algorithm for the template-hash (`log_hash_alg`) to `sha1`; the kernel currently hardcodes it to sha1. Now, the specified algorithm is only used for hashing the ramdisk and rootfs. If one is not provided, it defaults to sha256.
This change fixes #1609 because the policy schema prevents the use of sm3 for `log_hash_alg`. It also fixes #1623 since it removes the lookup for the boot_aggregate algorithm by setting `log_hash_alg` to sha1 instead.

### DIFF
--- a/keylime/policy/create_runtime_policy.py
+++ b/keylime/policy/create_runtime_policy.py
@@ -745,14 +745,13 @@ def create_runtime_policy(args: argparse.Namespace) -> Optional[RuntimePolicyTyp
         )
 
     algo = args.algo
-    if args.algo == "":
-        # Need to find the algo from the boot_aggregate.
-        algo, _ = boot_aggregate_from_file(args.ima_measurement_list)
+    if algo == "":
+        algo = FALLBACK_HASH_ALGO
 
     policy = ima.empty_policy()
 
-    # Set hash algo.
-    policy["ima"]["log_hash_alg"] = algo
+    # Set the algorithm for the template-hash; the kernel currently hardcodes it to sha1.
+    policy["ima"]["log_hash_alg"] = "sha1"
 
     if args.base_policy:
         merged_policy = merge_base_policy(policy, cast(str, args.base_policy))


maugustosilva:
@ansasaki  While I do understand hard-coding give us pause, it is important to note the code, as of today is basically "broken". Changing  `log_hash_alg` to anything other than `sha1` basically causes runtime attestation to fail, in a way which is poorly explained and therefore confusing to the user.

@marcostork Here is my proposal. We merge this PR as is, just to fix the pressing issue, and have a subsequent PR, more complete and complex, when we actually introduce **two** hash algorithms, one for  `template-hash` and one for `filedata-hash`. At this point, if the user chooses to change it (i.e., any of the algos), the attestation failures which he will see will be properly identified and explained.

ansasaki:
I finished my investigation and indeed that field from the runtime policy is used only for the IMA template hash, which on kernel can be only `sha1` or `md5`. Effectively, it can be considered hardcoded as `sha1`.

I agree with your proposal @maugustosilva, but beware the avalanche of changes that will propagate from that change to the policy format (I started similar changes to try to fix #1564, but decided that it was not worth for that single change).

To list some of the required changes for that:
- Support both versions of runtime policies
- Update all tools related to creation/modification of runtime policies
- Create the database migrations for the existing policies

ansasaki:
A possible intermediate solution would be to hardcode `SHA-1` internally for the template hash algorithm, and repurpose the field from the policy to be used as the `filedata-hash` algorithm. This would lead to no functional changes and allow the implementation of the validation of the `filedata-hash` algorithm in near future.

marcostork:
It would be beneficial to include the file content algorithm in the policy; this way, the tenant could validate the policy before sending it to the verifier, avoiding failures with valid agents due to a malformed policy (using the wrong algorithm). The issue with repurposing `log_hash_alg` for this is that we could only show warnings, as policies created by the old tool have this set to sha1.

ansasaki:
> It would be beneficial to include the file content algorithm in the policy; this way, the tenant could validate the policy before sending it to the verifier, avoiding failures with valid agents due to a malformed policy (using the wrong algorithm). The issue with repurposing `log_hash_alg` for this is that we could only show warnings, as policies created by the old tool have this set to sha1.

Sure, that is why I said is an intermediary solution. There is no doubt that adding separate options for the `template_hash` and `file_data_hash` is the best approach

ansasaki (approved):
Let's merge this as it is to be able to move forward

sergio-correia (approved):

PR #1629, opened 2024-08-29
tests: apply workarounds to known bugs

kkaarreell:
Before running tests apply workarounds to known
distribution bugs so we reduce the amount of
test failures.


kkaarreell:
F41 test failures is due to a known bug https://bugzilla.redhat.com/show_bug.cgi?id=2297927

ansasaki (approved):

Issue #1612, opened 2024-08-08
IDevIDs/IAKs that fail the check_tpm_origin check


Isaac-Matthews:
# Description 
When registering a device using an IDevID and IAK, the Keylime registrar performs a check on the certificates that the non-critical `subjectAltName` field specifies a TPM 2.0.

This check is performed by the `check_tpm_origin` function [here](https://github.com/keylime/keylime/blob/master/keylime/cert_utils.py#L144). The result of this check is used by `iak_idevid_cert_checks` [here](https://github.com/keylime/keylime/blob/master/keylime/cert_utils.py#L218), and if the check fails this function returns an error causing registration to fail and giving a 400 response. 

If a legitimate certificate does not have the `subjectAltName` field it will not be possible to register with Keylime (we have some certificates that do not have this field). 

# Reproducing The Problem
To reproduce this, attempt to register with an IDevID that does not contain the `subjectAltName` field. A quick way to try this out would be to run the functional test [iak-idevid-register-with-certificates](https://github.com/RedHat-SP-Security/keylime-tests/tree/main/functional/iak-idevid-register-with-certificates), but remove the `subjectAltName` fields from the `.conf` files in the tests directory.

# Proposed Fix
To remedy this I propose one of two fixes:

1.  Add a configuration option in the registrar to allow certificates that do not have this field (with a warning being produced). The default will be to reject the certificates with a descriptive error.
2.  Change this from a failing error into a warning.

If we can agree on one of these I will implement and put up a PR.

## Other Information
Some extra context on this issue: while this field seems to be required by the [TCG document](https://trustedcomputinggroup.org/wp-content/uploads/TPM-2p0-Keys-for-Device-Identity-and-Attestation_v1_r12_pub10082021.pdf) in section 8, the original spec for IDevIDs, [IEEE Std 802.1AR](https://standards.ieee.org/ieee/802.1AR/6995/) has the `subjectAltName` as optional. This possibly has caused confusion in provisioning device identities.

Isaac-Matthews:
Consensus from the 28/08/24 meeting was to relegate this to a warning and see if any issues arise in future. I will implement this change.

PR #1637, opened 2024-09-11
Change check_tpm_origin check to a warning that does not prevent registration

Isaac-Matthews:
Fixes [1612](https://github.com/keylime/keylime/issues/1612)

This changes the `check_tpm_origin` check in `cert_utils.py` to emit warnings if the SAN or hardware filed in the SAN are missing but still return true, and return false an incorrect HW identifier is provided.

The possibilities are:
* SAN missing: warn and pass
* SAN present but relevant hw field missing: warn and pass
* SAN present with HW info: pass or fail with error based on hw ID, no warning either way

### DIFF
--- a/keylime/cert_utils.py
+++ b/keylime/cert_utils.py
@@ -148,21 +148,26 @@ def check_tpm_origin(cert: Certificate, cert_type: str = "") -> bool:
     """Verify that the provided certificate is from a TPM according to the SAN
     :param cert: The certificate as a cryptography.x509.Certificate
     :param cert_type: Type of certificate as string for logging
-    :returns: True if the certificate came from a TPM
+    :returns: True if the certificate came from a TPM, or we are not sure. False if it did not come from a TPM
     """
     san_ext = cert.extensions.get_extension_for_oid(ExtensionOID.SUBJECT_ALTERNATIVE_NAME)
-    assert isinstance(san_ext.value, x509.SubjectAlternativeName)
+    if not isinstance(san_ext.value, x509.SubjectAlternativeName):
+        logger.warning("%s Certificate did not contain a SubjectAltName. This may not have come from a TPM.", cert_type)
+        return True
     othername = san_ext.value.get_values_for_type(x509.OtherName)[0]
     if othername.type_id.dotted_string != OID_HW_MODULE_NAME:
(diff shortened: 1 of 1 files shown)


Isaac-Matthews:
> The issue was that the SAN might be fully missing in some cases, right?
> 
> I still would fail on the case that their is a SAN, but not the TPM OID. I would expect the following behavior:
> 
> * SAN missing: issue warning but passes
> * SAN there: validate as usual

I think the SAN is general enough that it is very possible it will have other entries, so I'm not sure we should fail just if the SAN is there but doesn't contain the fields. 

What if we check if there is a hwmodulename in the SAN, and if there is fail if it doesn't match the tpm hwtype ID?
I.E.
* SAN missing: warn and pass
* SAN present but relevant hw field missing: warn and pass
* SAN present with HW info: pass or fail with error based on hw, no warning either way

THS-on:
As this is a mostly a workaround for existing certificates, can you double check what you saw in the wild?

Otherwise your proposed steps is probably the best compromise that we can find.

Isaac-Matthews:
Looks like there is a SAN but you bring up a good point about it being completely absent as the assert will fail. I will force push an update. Well spotted.

keylime/cert_utils.py
```diff
@@ -240,9 +245,10 @@ def iak_idevid_cert_checks(
     logger.debug("IAK cert verified.")
 
     # TPM cert checks
+    # Check if the IDevID and IAK certificates came from a TPM by checking the SAN
+    # These checks now warn on failure but do not fail the verification
```
THS-on, line 249: This message is now not fully correct

THS-on:
Can add a regression test for this with ideally a real certificate to https://github.com/RedHat-SP-Security/keylime-tests?

Isaac-Matthews:
> Can add a regression test for this with ideally a real certificate to https://github.com/RedHat-SP-Security/keylime-tests?

Sure. I have added an issue on that repo to track:
[Testing issue](https://github.com/RedHat-SP-Security/keylime-tests/issues/651)

THS-on:
LGTM, but someone with more knowledge of the IDevID side should also have a look. The linter fail is not related to this change.

THS-on (approved):

ansasaki (approved):
LGTM

ansasaki:
The linting errors are unrelated to the changes

Issue #1611, opened 2024-08-08
`keylime_policy` option `--rootfs` results in policy with a mix of absolute and relative paths


ansasaki:
Is your issue a feature request? If so, please raise it as an [enhancement](https://github.com/keylime/enhancements)

# Environment
 * OS / version: N/A
 * Processor architecture: N/A 
 * TPM Manufacturer: N/A
 * Keylime version: current master (dc118c780a06cb72725f72bc8100665a07a5a12f)

# Description 
When using the `--rootfs` option to create a policy using a directory as the root directory, the resulting policy contains a mix of absolute and relative paths.

## Expected behavior vs. actual behavior 

Running the command `keylime_policy create runtime --rootfs rootfs` where rootfs has the following structure:
```
rootfs/
├── nested
│   └── nested
└── test
```

Actual:
Results in a json similar to:
```
{
  "meta": {
    "version": 1,
    "generator": 3,
    "timestamp": "2024-08-08 08:57:19.017184"
  },
  "release": 0,
  "digests": {
    "rootfs/test": [
      "a948904f2f0f479b8f8197694b30184b0d2ed1c1cd2a1ec0fb85d299a192a447"
    ],
    "/var/tmp/tmt/run-014/plans/upstream-keylime-all-tests/discover/default-0/tests/functional/keylime_policy-commands/rootfs/nested/nested": [
      "414f8e9fd34ff68f66cbdab5ec63a5e738aa107f3454fa7edb51f49528abf9c6"
    ]
  },
  "excludes": [],
  "keyrings": {},
  "ima": {
    "ignored_keyrings": [],
    "log_hash_alg": "sha256",
    "dm_policy": null
  },
  "ima-buf": {},
  "verification-keys": ""
}
```

Expected:

The paths should have only absolute paths and all with the prefix directory `/rootfs` trimmed:
```
{
  "meta": {
    "version": 1,
    "generator": 3,
    "timestamp": "2024-08-08 08:57:19.017184"
  },
  "release": 0,
  "digests": {
    "/test": [
      "a948904f2f0f479b8f8197694b30184b0d2ed1c1cd2a1ec0fb85d299a192a447"
    ],
    "/nested/nested": [
      "414f8e9fd34ff68f66cbdab5ec63a5e738aa107f3454fa7edb51f49528abf9c6"
    ]
  },
  "excludes": [],
  "keyrings": {},
  "ima": {
    "ignored_keyrings": [],
    "log_hash_alg": "sha256",
[...]

PR #1638, opened 2024-09-12
keylime_policy: Use multiple threads to calculate file digests and other improvements

### LINKED ISSUE #1624: keylime_policy allows mixing inputs that use different digest algorithms

# Environment
 * OS / version: N/A
 * Processor architecture: N/A
 * TPM Manufacturer: N/A
 * Keylime version: current master (ad79ae8af1a70d81059a0d0778a4117425d98742)

# Description 
When multiple inputs are provided using a combination of `--use-ima-measurement-list -m`, `--rootfs`, `--ramdisk`, `--allowlist`, and `--exclude-list`, `keylime_policy` does not validate that all inputs are using the same digest algorithm, generating a policy with mixed digest length.

This can lead agents to fail attestation due to its IMA log containing entries calculated with a different digest algorithm than the one present in the policy, even when the file is the same as expected.

Expected: `keylime_policy` should validate that all inputs are using the same digest algorithm to have a consistent digest length in the policy

Actual: `keylime_policy` does not validate that all inputs are using the same digest algorithm and generate policies with incompatible entries.

## Steps to reproduce problem
Example for mixing algorithms using `--use-ima-measurement-list` and `--allowlist`:

1. `$ echo test > test && sha1sum test > allowlist.txt`
2. `$ echo "10 67bbcd56083572b7e3a4c0b648aa5dd0b636fc97 ima-ng sha256:f2d65f36dbae04c794028a89a7581c452ccca727d9cd6d12504e7aacc1ceaf33 /test" > ima_log.txt`
3. `$ keylime_policy create runtime --use-ima-measurement-list -m ima_log.txt --allowlist allowlist.txt`

# Relevant logs
```
$ keylime_policy create runtime --use-ima-measurement-list -m ima_log.txt --allowlist allowlist.txt
{"meta": {"version": 1, "generator": 3, "timestamp": "2024-08-21 10:46:59.471443"}, "release": 0, "digests": {"test": ["4e1243bd22c66e76c2ba9eddc1f91394e57f9f83"], "/test": ["f2d65f36dbae04c794028a89a7581c452ccca727d9cd6d12504e7aacc1ceaf33"]}, "excludes": [], "keyrings": {}, "ima": {"ignored_keyrings": [], "log_hash_alg": "sha256", "dm_policy": null}, "ima-buf": {}, "verification-keys": ""}
```

Note the difference in lengths of the digests in the generated policy

ansasaki:
This patchset introduces the following functional changes:

- When calculating file digests, use multiple threads
  - On my testing VM, this led to a reduction from ~16 seconds to ~6 seconds to measure ~50000 files
- Do not include symbolic links in the policy.
  - Previously, symbolic links were added to the generated runtime policy. Since IMA always resolves the symbolic links to their final target before adding the measurement to the log, symbolic links are never added to the measurement list provided by the agent. Therefore, there is no need to include these in the policy
- Allow the directory provided with `--rootfs` to be in any directory
  - Previously, if the provided directory was inside one of the excluded directories, it would be ignored. For example, if the user wanted to pass `--rootfs /tmp/my-rootfs` as the root directory, the tool would ignore it completely because `/tmp` is excluded by default. The changes makes the checks relative to the provided rootfs, for example `/tmp/my-rootfs/tmp` is ignored because it matches `/tmp` from the passed `/tmp/my-rootfs`, but other directories under `/tmp/my-rootfs` would not be ignored if they don't match any excluded directory.
- Validate the `verification-keys` content against a JSON schema
- Check that the algorithms used for digests from all sources match to avoid mixing digests calculated with different algorithms

Resolves #1624
Related #1623, #1611


ansasaki:
The test failure in Fedora 41 is due to a missing dependency. A fix is proposed via: https://github.com/RedHat-SP-Security/keylime-tests/pull/653

keylime/policy/create_runtime_policy.py
```diff
@@ -304,12 +304,9 @@ def boot_aggregate_parse(line: str) -> Tuple[str, str]:
     :return: tuple with two values, the algorithm used and the digest of the
              boot aggregate
     """
-    def_alg = FALLBACK_HASH_ALGO
-    def_agg = "0" * algorithms.Hash(def_alg).hexdigest_len()
-
     alg, digest, fpath, ok = process_ima_sig_ima_ng_line(line)
     if not ok or fpath != "boot_aggregate":
-        return def_alg, def_agg
+        return "unknown", ""
```
THS-on, line 346: Any reason why to return "unknown" instead of something falsely?

keylime/policy/create_runtime_policy.py
```diff
@@ -43,6 +43,9 @@
 IGNORED_KEYRINGS: List[str] = []
 DEFAULT_FILE_DIGEST_ALGORITHM = algorithms.Hash.SHA256
 
+# This is to follow IMA default
+# See in keylime/ima/ima.py
+DEFAULT_IMA_TEMPLATE_DIGEST_ALGORITHM = algorithms.Hash.SHA1
```
THS-on, line 49: Yes, I think this is still actually hard-coded in the kernel anyway.

THS-on (approved):
Generally LGTM

sergio-correia (approved):
Good job, looks very good to me.

keylime/policy/create_runtime_policy.py
sergio-correia: Or defining a constant or something similar, given it is used a few times throughout the code?

ansasaki: @sergio-correia Yeah, it makes sense to make it a constant.

@THS-on When I was writing it, I felt the need to make it easy to tell apart the case when we tried to guess the algorithm and failed from the case when there was no input.

And I already see some issues coming from the fact that `sha256` and `sm3_256` has the same output length. I will probably address this in a future PR.

If you agree, I'll follow Sergio's suggestion and make it a constant. If not, I'll try to figure a way without using it.

THS-on: Yeah let's make it a constant

keylime/policy/create_runtime_policy.py
maugustosilva: Unfortunately, it is.

maugustosilva:
Ok, I see two approvals here (been out recovering from surgery). I am ok merging now, but do you want to make the `return "unknown", ""` into a constant now, or in a subsequent PR @ansasaki . I see no problem either way.

ansasaki:
> Ok, I see two approvals here (been out recovering from surgery). I am ok merging now, but do you want to make the `return "unknown", ""` into a constant now, or in a subsequent PR @ansasaki . I see no problem either way.

I'll change it here, I think it shouldn't take long for me to put the PR together. I'll try to address the other issue I mentioned (the ambiguity of trying to guess the digest algorithm from the output length) in a separate PR.

ansasaki:
@maugustosilva Now I think this is ready to merge

ansasaki:
Since this was approved by multiple people and @maugustosilva already manifested the willing to merge it, I'll merge it now.

PR #1643, opened 2024-09-18
docs: add distro installation instructions and include IDevID also in sidebar

THS-on:
- **docs: point to installation guide from RHEL and SLE Micro**
- **docs: include IDevID page also in the sidebar**

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -1,8 +1,19 @@
 Installation
 ============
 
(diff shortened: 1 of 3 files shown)


maugustosilva (approved):

maugustosilva:
Thanks for adding RHEL and SL new instructions.

Issue #830, opened 2022-01-12
[Discussion] Drop SPDX-License-Identifier header?


aplanas:
I will like to discuss the drop of the `SPDX-License-Identifier` and the Copyritght holders headers in the Python code.  The `LICENSE` file should cover the common case, and the copyright notice can be in the documentation, or in some more general place.

THS-on:
I like having the licence identifier directly in the file, because then there is no ambiguity when other files in a project might have different licences. 
Moving the copyright into a different file is fine for me because it does not make sense in most cases where multiple organisations now have worked on the same file.

aplanas:
Multiple license can be an issue sometimes (there are some incompatibilities), but when this situation is present would make sense that the file can have its own SPDX header when is not the default one.

I agree with the second argument.

mpeters:
I agree that pulling both into a single file would make sense and only have a separate license section for a file if it's different. But I'm pretty sure we would never want to have multiple licenses as not only are there incompatibilities, but it would make life very difficult for packagers.

I'm in favor of this change, but wanted to give it a couple of days to make sure there weren't any big objections from the community.

aplanas:
I will do it!

PR #1645, opened 2024-09-19
Add missing license note

sarroutbi:
In order to homogeneize all files, it would be
recommended to add license header to all of them

### DIFF
--- a/keylime/policy/rpm_repo.py
+++ b/keylime/policy/rpm_repo.py
@@ -1,6 +1,11 @@
 #!/usr/bin/env python3
 
-"""Analyze local and remote RPM repositories."""
+"""
+Analyze local and remote RPM repositories.
+
+SPDX-License-Identifier: Apache-2.0
+Copyright 2024 Red Hat, Inc.
+"""
 
 import gzip
 import logging


ansasaki (dismissed):

THS-on (changes requested):
I wouldn't reintroduce them. Rather remove the rest leftover ones. See discussion here, that we had a while back: https://github.com/keylime/keylime/issues/830

ansasaki:
> I wouldn't reintroduce them. Rather remove the rest leftover ones. See discussion here, that we had a while back: #830

I didn't have this context, thank you for pointing out to the discussion.

@sarroutbi Thank you for the contribution, but apparently the preference is to not add the license header, but rather drop it from the files that still have it. If you have free cycles and are willing to contribute in this front, maybe we could reverse the purpose of this PR. Otherwise I think we should close this. I'll drop my approval for now, I'll let you decide.

sarroutbi:
> > I wouldn't reintroduce them. Rather remove the rest leftover ones. See discussion here, that we had a while back: #830
> 
> I didn't have this context, thank you for pointing out to the discussion.
> 
> @sarroutbi Thank you for the contribution, but apparently the preference is to not add the license header, but rather drop it from the files that still have it. If you have free cycles and are willing to contribute in this front, maybe we could reverse the purpose of this PR. Otherwise I think we should close this. I'll drop my approval for now, I'll let you decide.

OK. It makes sense to close it. Sorry for the noise.

Issue #1610, opened 2024-08-08
`keylime_policy` options names are inconsistent


ansasaki:
Is your issue a feature request? If so, please raise it as an [enhancement](https://github.com/keylime/enhancements)

# Environment
 * OS / version: N/A
 * Processor architecture: N/A
 * TPM Manufacturer: N/A
 * Keylime version: current master (dc118c780a06cb72725f72bc8100665a07a5a12f)

# Description 
`keylime_policy` option names `--allowlist` and `--exclude-list` are inconsistent. They should both have the `-` or not.

Also some short options are not intuitive, probably better to drop them and keep only the long version. For example `-A` for `--add-ima-signature-verification-key`.

## Expected behavior vs. actual behavior 

Actual:
The options `--allowlist` and `--exclude-list` are inconsistent

Expected:
The options are consistent: either both have the `-` or not:

- `--allow-list` and `--exclude-list` or
- `--allowlist` and `--excludelist`

## Steps to reproduce problem
1. Run `keylime_policy create runtime --help` 

# Relevant logs

Attach any relevant log files that can help to debug your issue.

sergio-correia:
What would be the preferred way here?
`--allowlist/--excludelist` or `--allow-list/--exclude-list`?

PR #1648, opened 2024-09-23
keylime-policy: use consistent names for create runtime

sergio-correia:
This PR has a few changes related to consistency within the policy tool, and has the following changes:

1) Let's keep option names consistent, i.e. use `--allowlist`/`--excludelist`, instead of `--allowlist`/`--exclude-list`
2) drop some confusing "short" shortcuts, like `-i` and `-A`, and keep long versions instead, for those
3) drop superfluous option `--use-ima-measurement-list`, as we already have `--ima-measurement-list`; the behavior is the following: if `--ima-measurement-list` is used without any argument, it will use the default IMA measurement list file, `/sys/kernel/security/ima/ascii_runtime_measurements`; if a parameter is given as argument, that parameter indicates the file to read as a source for the IMA measurement list
4) rename the tool itself from `keylime_policy` to `keylime-policy`

Fixes: #1610

### DIFF
--- a/keylime/cmd/keylime_policy.py
+++ b/keylime/cmd/keylime_policy.py
@@ -20,11 +20,11 @@ except ModuleNotFoundError:
 
 from keylime.policy import create_runtime_policy
 
-logger = logging.getLogger("keylime_policy")
+logger = logging.getLogger("keylime-policy")
 
 
 def main() -> None:
-    """keylime_policy entry point."""
+    """keylime-policy entry point."""
     if os.geteuid() != 0:
         logger.critical("Please, run this program as root")
         sys.exit(1)
--- a/keylime/policy/create_runtime_policy.py
+++ b/keylime/policy/create_runtime_policy.py
@@ -40,7 +40,10 @@ else:
 
 logger = logging.getLogger("policy.create_runtime_policy")
 
+# We use /dev/null to indicate an empty ima measurement list.
+EMPTY_IMA_MEASUREMENT_LIST = "/dev/null"
 IMA_MEASUREMENT_LIST = "/sys/kernel/security/ima/ascii_runtime_measurements"
+
 IGNORED_KEYRINGS: List[str] = []
 DEFAULT_FILE_DIGEST_ALGORITHM = algorithms.Hash.SHA256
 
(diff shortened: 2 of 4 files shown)


THS-on:
> rename the tool itself from keylime_policy to keylime-policy

Any reason for this, as all the other tools we have are keylime_XYZ?

ansasaki:
> > rename the tool itself from keylime_policy to keylime-policy
> 
> Any reason for this, as all the other tools we have are keylime_XYZ?

I'm in favor of `keylime-policy` for ergonomic reasons. We are making all these changes to the tools to make them easier/more comfortable to use. I think that there is no reason to keep the name of the tool cumbersome to use  just because historically all the binaries in the project are cumbersome to use.

Moreover, the other `keylime_XYZ` are mostly run as services, not as command line tools. In my opinion, the `keylime_tenant` also should be renamed at some point as the name is misleading to newcomers. It is always a pain to explain that `keylime_tenant` is not a service, but a command-line tool.

sergio-correia:
> > rename the tool itself from keylime_policy to keylime-policy
> 
> Any reason for this, as all the other tools we have are keylime_XYZ?

It will probably boil down to personal preference, in the end. In my opinion, this would be more in line with other command-line tools, and I actually find a dash even easier to use.

Just to have some anecdotal data to illustrate it from my system as I type this reply:
```
ls  /usr/bin/ | grep _ | wc -l
271

ls  /usr/bin/ | grep - | wc -l
1239
```

So, you see that while underscores are popular, dashes seem to be way more common.

I see your point that the other tools we have are keylime_XYZ, but I also I think we have a last chance to do better this time, as this one tool was not yet present in a releasem, so it technically does not yet exist :)

In practice, as the aim of the tool is to simplify user experience with handling Keylime policies, I think even this dash instead of underscore would play a role here -- it certainly would for me, at least.

THS-on:
Yeah makes sense, then lets rename it and maybe do it with the rest of the commandline tools later.

maugustosilva:
+1 on the suggestion of renaming the **user facing** utilities from `_` to `-`, for "ergonomic reasons" too. I have not thought through all the implications for the package maintainers, but we could even have a "transition phase" where the `keylime-*` utilities are symlinked to `keylime_`.

maugustosilva (approved):

sergio-correia:
I dropped the temporary commit pointing the tests to my personal repo, to prevent it from being merged by accident. Tests will now fail as they need https://github.com/RedHat-SP-Security/keylime-tests/pull/657 merged, which updates the tests to confirm with the changes proposed here.

ansasaki (approved):

PR #1652, opened 2024-09-24
create_runtime_policy: Fix log level for debug messages

### LINKED ISSUE #1613: `keylime_policy` log messages with incorrect level

# Environment
 * OS / version: N/A
 * Processor architecture: N/A
 * TPM Manufacturer: N/A
 * Keylime version: current master (dc118c780a06cb72725f72bc8100665a07a5a12f)

# Description 
`keylime_policy` logs a message for each line of the IMA measurement log, with the `INFO` level. Also, configuration files loading are also logged with `INFO` level.

These messages make the command line output excessively polluted. Moreover, these messages are not actionable from the user point of view and should probably be logged with `DEBUG` level.

Expected: Messages that are not relevant for the user should be logged with `DEBUG` level
Actual: Messages that are not relevant for the user are logged with `INFO` level

## Steps to reproduce problem
1. run `keylime_policy create runtime --use-ima-measurement-list`

# Relevant logs
The reproducer command generates an output similar to:
```
2024-08-08 17:36:22.993 - policy.create_runtime_policy - INFO - Skipping line that was split into 5 pieces, expected 6: 10 0adefe762c149c7cec19da62f0da1297fcfbffff ima-ng sha256:0000000000000000000000000000000000000000000000000000000000000000 boot_aggregate

{"meta": {"version": 1, "generator": 3, "timestamp": "2024-08-08 17:36:22.993441"}, "release": 0, "digests": {"boot_aggregate": ["0000000000000000000000000000000000000000000000000000000000000000"]}, "excludes": [], "keyrings": {}, "ima": {"ignored_keyrings": [], "log_hash_alg": "sha256", "dm_policy": null}, "ima-buf": {}, "verification-keys": ""}
```

ansasaki:
The messages that are not relevant to the user should be logged using DEBUG level instead of INFO.

Fixes #1613

### DIFF
--- a/keylime/policy/create_runtime_policy.py
+++ b/keylime/policy/create_runtime_policy.py
@@ -396,7 +396,7 @@ def process_flat_allowlist(allowlist_file: str, hashes_map: Dict[str, List[str]]
                 pieces = line.split(None, 1)
                 if not len(pieces) == 2:
                     logmsg = f"Skipping line that was split into {len(pieces)} parts, expected 2: {line}"
-                    logger.info(logmsg)
+                    logger.debug(logmsg)
                     continue
 
                 (checksum_hash, path) = pieces
@@ -620,7 +620,7 @@ def get_hashes_from_measurement_list(
                 pieces = line.split(" ")
                 if len(pieces) < 5:
                     errmsg = f"Skipping line that was split into {len(pieces)} pieces, expected at least 5: {line}"
-                    logger.info(errmsg)
+                    logger.debug(errmsg)
                     continue
                 if pieces[2] not in ("ima-sig", "ima-ng"):
                     continue
@@ -746,7 +746,7 @@ def process_ima_buf_in_measurement_list(
                 pieces = line.split(" ")
                 if len(pieces) != 6:
                     errmsg = f"Skipping line that was split into {len(pieces)} pieces, expected 6: {line}"
-                    logger.info(errmsg)
+                    logger.debug(errmsg)
                     continue
                 if pieces[2] not in ("ima-buf"):
                     continue


keylime/policy/create_runtime_policy.py
```diff
@@ -396,7 +396,7 @@ def process_flat_allowlist(allowlist_file: str, hashes_map: Dict[str, List[str]]
                 pieces = line.split(None, 1)
                 if not len(pieces) == 2:
                     logmsg = f"Skipping line that was split into {len(pieces)} parts, expected 2: {line}"
```
THS-on, line 398: Not part of this PR, but why are the log messages separately? I assume this is because of the pylint warning, that one should pass variables as positional arguments instead of using format strings and this works around it. Not that it really matters here, but its kinda weird.

sergio-correia: I don't recall why, to be honest, and looking at it again, I have to agree it looks very odd. I don't think it was to work around pylint, though. We should probably improve that at some point.

THS-on: Thanks!

THS-on:
Looking more closely at those messages, I would expect that we generally never reach those parts with valid logs. So moving this to debug hides actual error cases. The question is why we have the edge case that we run into those conditions and should we not handle it?

ansasaki:
> Looking more closely at those messages, I would expect that we generally never reach those parts with valid logs. So moving this to debug hides actual error cases. The question is why we have the edge case that we run into those conditions and should we not handle it?

That is not the case, these messages are reached for basically every valid log. For example, for every entry using `ima-buf` template, a message will be printed when parsing for `ima-ng` or `ima-sig`. If processing `ima-buf` entries is enabled by passing `--ima-buf`, then a message is printed for every `ima-ng` or `ima-sig` entry.

I understand your concern, but it is not the log level that is incorrect, but the placement. If they were logged for exception cases, then I agree they should have even higher level like `WARNING`. I can create a separate PR to try to improve the entry validation.

THS-on:
Ah makes sense, yeah then change the logging for now and do the validation change in another PR.

THS-on (approved):

PR #1653, opened 2024-09-25
keylime.conf: full removal

THS-on:
We no longer need the legacy config


maugustosilva:
I think the time has come.

sergio-correia (approved):
rip :)

ansasaki (approved):

PR #1657, opened 2024-09-30
keylime-policy: add `sign runtime` and `create measured-boot` subcommands

sergio-correia:
This PR adds the following subcommands to the policy tool:
* `sign runtime` is a port of `cmd/sign_runtime_policy.py`
* `create measured-boot` is a port of `scripts/create_mb_refstate`

It also adds a logger to be used with the tool, that logs to stderr by default and that prevents the logs from propagating to other loggers, which would duplicate the logs and might also write to stdout anyway. That way, we can pipe the output of the tool (when set to write the relevant data to stdout) to tools like `jq`, for instance.


keylime/policy/sign_runtime_policy.py
```diff
@@ -0,0 +1,129 @@
...
+        help="The output file path for the x509 certificate, if using x509 DSSE backend",
+        default="",
+    )
+
+    sign_p.set_defaults(func=sign_sign_policy)
+
+    return sign_p
+
+
+def sign_sign_policy(args: argparse.Namespace) -> Optional[str]:
```
ansasaki: Is there a reason for the duplicate `sign_sign_` prefix?

keylime/policy/sign_runtime_policy.py
```diff
@@ -0,0 +1,129 @@
...
+        required=False,
+        help="The private key to sign the policy with",
+        default="",
+    )
+    sign_p.add_argument(
+        "-p",
+        "--keypath",
+        dest="keypath",
+        required=False,
+        help="The filename to write the created private key, " "if one is not provided via the --keyfile argument",
```
ansasaki, line 63: I think `black` does not like strings split in multiple lines, but when it puts them together is creates weird quoting.

keylime/policy/create_mb_policy.py
```diff
@@ -0,0 +1,228 @@
...
+        default=sys.stdout,
+        help="Output path for the generated measured boot policy",
+    )
+
+    mbref_p.set_defaults(func=create_mb_refstate)
+    return mbref_p
+
+
+def create_mb_refstate(args: argparse.Namespace) -> Optional[Dict[str, object]]:
+    """Create a measured bot reference state."""
```
ansasaki, line 189: s/bot/boot/

test/test_create_runtime_policy.py
```diff
@@ -797,206 +798,23 @@ def test_digest_algorithm_priority(self):
...
 
-            with self.assertLogs("policy.create_runtime_policy", level="DEBUG") as logs:
-                _policy = create_runtime_policy.create_runtime_policy(args)
-
-                self.assertIn(
-                    f"DEBUG:policy.create_runtime_policy:Using digest algorithm '{expected_algo}' obtained from the {expected_source}",
-                    logs.output,
-                )
-
-    def test_digest_algorithm_priority_exceptions(self):
```
ansasaki, line 808: Apparently during rebases some tests were removed by accident. I think the removal of the `test_digest_algorithm_priority_exceptions()` was not intentional

test/test_create_runtime_policy.py
```diff
@@ -797,206 +798,23 @@ def test_digest_algorithm_priority(self):
...
-
-            with self.assertLogs("policy.create_runtime_policy", level="DEBUG") as logs:
-                policy = create_runtime_policy.create_runtime_policy(args)
-                self.assertIn(
-                    f"WARNING:policy.create_runtime_policy:The digest algorithm in the {case['source']} does not match the previously set 'sha1' algorithm",
-                    logs.output,
-                )
-                self.assertEqual(policy, None)
-
-    def test_unknown_algorithm_sources(self):
```
ansasaki, line 943: Again for `test_unknown_algorithm_sources()`, I think it was unintentionally removed.

sergio-correia:
Thanks for looking into this. I will address the raised issues, especially the problem with some tests being removed accidentally.

keylime/policy/create_mb_policy.py
sergio-correia: Fixed.

keylime/policy/sign_runtime_policy.py
ansasaki: TBH, I just follow the linter rules, which means prefer long strings over splitting into multiple lines.

test/test_create_runtime_policy.py
sergio-correia: I believe I have resolved these issues; please double check.

keylime/policy/sign_runtime_policy.py
sergio-correia: Fixed.

keylime/policy/logger.py
```diff
@@ -0,0 +1,81 @@
+"""
+Module to assist with logging in the police tool.
```
ansasaki, line 2: s/police/policy/

test/test_create_runtime_policy.py
```diff
@@ -865,19 +881,31 @@ def test_digest_algorithm_priority_exceptions(self):
...
+                # Let's replace the log stream so that we can capture it.
+                stderr = StringIO()
+                self.logger.setStream(stderr)
+
+                # Now let's run it with the arguments and capture stderr.
                 _policy = create_runtime_policy.create_runtime_policy(args)
+                logs = stderr.getvalue()
+            finally:
+                # Restore the stream.
+                self.logger.setStream(sys.stderr)
```
ansasaki, line 896: @sergio-correia Do you think it would be possible to make this a private method so that we can reuse it without duplicating the code?

Not a blocker, just to make it a bit nicer.

test/test_sign_runtime_policy.py
```diff
@@ -0,0 +1,194 @@
...
+EC_PUBKEY = os.path.join(DATA_DIR, "ec-p521-public.pem")
+RSA_PRIVKEY = os.path.join(DATA_DIR, "rsa-4096-private.pem")
+
+POLICY = os.path.join(DATA_DIR, "runtime-policy.json")
+POLICY_EMPTY = os.path.join(DATA_DIR, "runtime-policy-empty.json")
+POLICY_BOGUS = os.path.join(DATA_DIR, "runtime-policy-bogus.json")
+
+
+class SignRuntimePolicy_Test(unittest.TestCase):
+    def test__get_signer(self):
```
ansasaki, line 31: s/test__get_signer/test_get_signer/

keylime/policy/logger.py
sergio-correia: Fixed.

test/test_create_runtime_policy.py
sergio-correia: Please take another look, now it should be better.

test/test_sign_runtime_policy.py
sergio-correia: This was because the method name is `_get_signer()`, but I can remove one of these underscores, if you feel strongly about it.

ansasaki: No need for the change

ansasaki (approved):

ansasaki:
The failure on Fedora 41 is unrelated, and reported as: https://bugzilla.redhat.com/show_bug.cgi?id=2326964

Issue #1666, opened 2024-10-10
docker build fails as pinned fedora baseimage is no longer available


tuminoid:
# Environment
 * OS / version: ubuntu 24.04
 * Processor architecture:  amd64
 * TPM Manufacturer: na
 * Keylime version:  head of master

# Description 

Docker build does not work.

## Expected behavior vs. actual behavior 

Docker build works.

## Steps to reproduce problem

1. cd docker/ci; docker build .

# Relevant logs

```console
$ docker build .
[+] Building 1.4s (2/2) FINISHED                                                                                                                        docker:default
 => [internal] load build definition from Dockerfile                                                                                                              0.1s
 => => transferring dockerfile: 1.99kB                                                                                                                            0.0s
 => WARN: MaintainerDeprecated: Maintainer instruction is deprecated in favor of using label (line 9)                                                             0.1s
 => ERROR [internal] load metadata for quay.io/fedora/fedora@sha256:1ba88682f9ccc835f87ea8b81b46cafbd0a0214c200d5eb9fdf2808b13cdd070                              1.2s
------
 > [internal] load metadata for quay.io/fedora/fedora@sha256:1ba88682f9ccc835f87ea8b81b46cafbd0a0214c200d5eb9fdf2808b13cdd070:
------

 1 warning found (use docker --debug to expand):
 - MaintainerDeprecated: Maintainer instruction is deprecated in favor of using label (line 9)
Dockerfile:8
--------------------
   6 |     ##############################################################################
   7 |     
   8 | >>> FROM quay.io/fedora/fedora@sha256:1ba88682f9ccc835f87ea8b81b46cafbd0a0214c200d5eb9fdf2808b13cdd070
   9 |     MAINTAINER Luke Hinds <lhinds@redhat.com>
  10 |     LABEL version="1.2.0" description="Keylime - Bootstrapping and Maintaining Trust in the Cloud"
--------------------
ERROR: failed to solve: quay.io/fedora/fedora@sha256:1ba88682f9ccc835f87ea8b81b46cafbd0a0214c200d5eb9fdf2808b13cdd070: failed to resolve source metadata for quay.io/fedora/fedora@sha256:1ba88682f9ccc835f87ea8b81b46cafbd0a0214c200d5eb9fdf2808b13cdd070: quay.io/fedora/fedora@sha256:1ba88682f9ccc835f87ea8b81b46cafbd0a0214c200d5eb9fdf2808b13cdd070: not found
```

[...]

PR #1667, opened 2024-10-10
fix docker build by bumping fedora baseimage

tuminoid:
Dockerbuild does not work as the sha256 pinned fedora image no longer exists in Quay. Bumping it to latest sha256 fixes it.

Fixes: #1666

### DIFF
--- a/docker/ci/Dockerfile
+++ b/docker/ci/Dockerfile
@@ -5,7 +5,7 @@
 # It is not recommended for use beyond testing scenarios.
 ##############################################################################
 
-FROM quay.io/fedora/fedora@sha256:1ba88682f9ccc835f87ea8b81b46cafbd0a0214c200d5eb9fdf2808b13cdd070
+FROM quay.io/fedora/fedora@sha256:fb6b7814deb770c0bd99ad618f6f53d9a12261df061989f1dfd3d90ad1ac193e
 MAINTAINER Luke Hinds <lhinds@redhat.com>
 LABEL version="1.2.0" description="Keylime - Bootstrapping and Maintaining Trust in the Cloud"
 


tuminoid:
Rebased as there was some recent modifications in upstream.

ansasaki:
To be honest, this is not very effective, since tomorrow the images won't build again as the pinned fedora base image will not be available anymore.

We implemented automatic updates of the base image as part of the CI, which runs monthly (see `.github/workflows/update-base-image.yml`). It basically generates a PR updating the pinned base image version like #1658. After we check that the proposed base image is ok and merge it, the images are rebuilt.

For the release images, we have a script to build them locally (`docker/release/build_locally.sh`), which will automatically obtain the current available fedora image and build the images locally on top of it.

@tuminoid I believe what you need is a similar script for the CI image, to build it locally using the currently available base image, correct?

tuminoid:
> To be honest, this is not very effective, since tomorrow the images won't build again as the pinned fedora base image will not be available anymore.

OK, now that I took a deeper look how this works, I can see this PR makes little sense.

> We implemented automatic updates of the base image as part of the CI, which runs monthly (see `.github/workflows/update-base-image.yml`). It basically generates a PR updating the pinned base image version like #1658. After we check that the proposed base image is ok and merge it, the images are rebuilt.

Not being a fedora consumer myself, I'm actually a bit baffled by the rate of rot fedora image seems to have, and keylime building on top of such images. I might miss the point completely here, but wouldn't it make sense to build on top of a bit more stable images?

> For the release images, we have a script to build them locally (`docker/release/build_locally.sh`), which will automatically obtain the current available fedora image and build the images locally on top of it.
> 
> @tuminoid I believe what you need is a similar script for the CI image, to build it locally using the currently available base image, correct?

Yes, that'd be the real solution to my problem.

I'll close this PR as it is useless for the project.

PR #1668, opened 2024-10-18
Add support for EK Certificate Chain, resolves #1552

### LINKED ISSUE #1552: Expand support for EK certificates

# Environment
 * OS / version: NA
 * Processor architecture: NA
 * TPM Manufacturer: 
 * Keylime version: NA

# Description 

1. The latest TCG EK specification permits the TPM to have an internal CA, and to supply intermediate CA certificates. The Rest API should support sending these certificates.
2. Assuming https://keylime-docs.readthedocs.io/en/latest/rest_apis.html#cloud-registrar is current, it says the EK cert is extracted from 01c00002. That's the original RSA-2048 certificate. TPMs are appearing with various ECC curves and RSA-3072.  Keylime should support those as well.

See description.

## Steps to reproduce problem

See description.

# Relevant logs

ematery:
The EK Certificate chain can be stored in the NVRAM Indexes ranging from 0x01c00100 to 0x01c001ff, with the certificates stored concatenated in DER format.

The certificate chain will be provided by the rust-keylime agent, with the 'ek_ca_chain' key, and added to the database. Before the change the tenant checked the EK certificate against the tpm_cert_store. With the changes first the EK certificate will be checked against the ek_ca_chain and on success the EK certificate will be replaced by the most top level certificate from the chain, so the procedure can continue as before.


ematery:
Should I add a new api version (2.4) for the registrar, as the ekcert value changes?

-    :>json string ekcert: base64 encoded EK certificate. Should be in `DER` format. Gets extracted from NV `0x1c00002`.
+    :>json string ekcert: PEM encoded EK certificate chain. Gets extracted from NV `0x1c00002` (ekcert) + `[0x01c00100..0x01c001ff]` (ek ca chain).

keylime/cert_utils.py
```diff
@@ -253,3 +264,24 @@ def iak_idevid_cert_checks(
     if not check_tpm_origin(iak_cert_509, "IAK"):
         return "Error: IAK certificate might not be from a TPM", None, None
     return "", idevid_pub, iak_pub
+
+
+def to_cert_list(ekcert: str) -> List[Certificate]:
+    """Transform single string of multiple certificates in PEM format to list of certificates.
+    :param ekcert: string with certificate chain
+    :returns: List of certificates
+    :raises: ValueErrror, if no valid or an invalid certificate is found
```
sarroutbi, line 273: typo (ValueErrror->ValueError)

ematery:
keylime/registrar.py:

```python
...
    if int(cryptography.__version__.split(".", maxsplit=1)[0]) < 38:
        if tpm_identity == "iak_idevid":
            logger.error(
                "DevID is REQUIRED in config ('tpm_identity = %s') but python-cryptography version < 38", tpm_identity
            )
            sys.exit(1)
...
```

keylime/cert_utils.py:

```python
def x509_pem_cert(pem_cert_data: str) -> Certificate:
    """Load an x509 certificate provided in PEM format
    :param pem_cert_data: the base-64 encoded PEM certificate
    :type pem_cert_data: str
    :returns: cryptography.x509.Certificate
    """
    try:
        return x509.load_pem_x509_certificate(data=pem_cert_data.encode("utf-8"), backend=default_backend())
    except Exception as err:
        logger.warning(
            "Failed to parse PEM data with python-cryptography (might not be strictly conforming to DER ASN.1 encoding): %s",
            err,
        )
        # Let's read the DER bytes from the base-64 PEM.
        der_data = pem.readPemFromFile(io.StringIO(pem_cert_data))
        # Now we can load it as we do in x509_der_cert().
        #
        # This operation will modify the certificate, otherwise reading the DER data again with the same function as
        # before should result in an exception again. This is a problem for signature checks, as the signature will
        # no longer match the new certificate content.
        pyasn1_cert = decoder.decode(der_data, asn1Spec=rfc2459.Certificate())[0]
        return x509.load_der_x509_certificate(data=encoder.encode(pyasn1_cert), backend=default_backend())
```

I'm currently testing my changes with an intel firmware TPM.

In the case of `tpm_identity == "iak_idevid"` the required crypthography version will be at least version 39. The intel ek certificate uses an ASN.1 extension that will result in an exception when using `x509.load_pem_x509_certificate()` and issue a reencoding using `pyasn1` . The reencoding will break the signature, which will result in a chain that can no longer be verified. 

```txt
WARNING - Certificate(s) 'ekcert' may not conform to strict ASN.1 DER encoding rules and were re-encoded before parsing by python-cryptography
```

I initially solved the problem by using a cryptography version <= 3.4.8. Due to the requirement of version > 38 this not possible.

Any suggestions on how to solve the problem?

ansasaki:
> I'm currently testing my changes with an intel firmware TPM.
> 
> In the case of `tpm_identity == "iak_idevid"` the required crypthography version will be at least version 39. The intel ek certificate uses an ASN.1 extension that will result in an exception when using `x509.load_pem_x509_certificate()` and issue a reencoding using `pyasn1` . The reencoding will break the signature, which will result in a chain that can no longer be verified.

I agree that "reencoding" (which means corrupting) the certificate is not the way to fix this issue, but we could not find a better way to workaround broken certificates that are deployed in large scale. Could you please provide an example of the broken certificate from Intel? 

> I initially solved the problem by using a cryptography version <= 3.4.8. Due to the requirement of version > 38 this not possible.
> 
> Any suggestions on how to solve the problem?

Pinning a crypto library to an old version is dangerous in the long term, as it would not be automatically updated with security fixes, etc. It would be less dangerous to simply not support Intel broken certificates (and hope they fix the certificates issued in future).

For me the possibilities are:

- Work with Intel to fix their certificates, if possible (e.g. if it is possible to update the certificates with a firmware update). If python cryptography rejects the certificates, most probably they are broken. This would be the best, but I understand it would probably be very hard or impossible to fix the broken certificates deployed in large scale.
- Work with the python cryptography upstream and try to "relax" the ASN.1 parsing, maybe adding an option to explicitly allow some issues in non-conforming certificates. We already tried this before and they are adamant on keeping the parser strict (see https://github.com/pyca/cryptography/issues/7189).
- Try to find a way to use `pyasn1` to parse the certificate and generate the DER encoding without corrupting the certificate, meaning generating exactly the bytes that were signed, then load the certificate and create a `cryptography.x509.Certificate` object. Maybe the python cryptography certificate verification does not try to parse the certificate again, avoiding the exception during parsing. This would allow us to be flexible with broken certificates, but probably requires a lot of extra work.

ansasaki:
Just to provide all info: there is the possibility to set an alternative certificate verification script to use a more relaxed parser/certificate verification (e.g. openssl), as described in 0ba819e. This is a workaround for the issue of broken certificates found in the past.

Another related change that is still a work in progress is #1670 which will allow adding webhooks for alternative identity verification. But fixing using this also depends on the EK certificate verification to be moved from the tenant to the registrar.

ematery:
> Pinning a crypto library to an old version is dangerous in the long term, as it would not be automatically updated with security fixes, etc. It would be less dangerous to simply not support Intel broken certificates (and hope they fix the certificates issued in future).

I agree.

> 
> For me the possibilities are:
> 
>   * Try to find a way to use `pyasn1` to parse the certificate and generate the DER encoding without corrupting the certificate, meaning generating exactly the bytes that were signed, then load the certificate and create a `cryptography.x509.Certificate` object. Maybe the python cryptography certificate verification does not try to parse the certificate again, avoiding the exception during parsing. This would allow us to be flexible with broken certificates, but probably requires a lot of extra work.

As I understand it, the whole point of using pyasn1 is to modify the certificate in a way that allows it to be loaded by `cryptography.x509.Certificate`. However, this will always result in a broken signature. Additionally, generating the exact same DER output as the original certificate will also lead to an exception, as it will be identical input. I don't believe there is a way to handle "broken" certificates in this manner.

> * Work with Intel to fix their certificates, if possible (e.g. if it is possible to update the certificates with a firmware update). If python cryptography rejects the certificates, most probably they are broken. This would be the best, but I understand it would probably be very hard or impossible to fix the broken certificates deployed in large scale.
> * Work with the python cryptography upstream and try to "relax" the ASN.1 parsing, maybe adding an option to explicitly allow some issues in non-conforming certificates. We already tried this before and they are adamant on keeping the parser strict (see https://github.com/pyca/cryptography/issues/7189).

Upon closer inspection there does not seem to be a problem with the intel certificate. Here you can find the decoded [original intel certificate](https://lapo.it/asn1js/#MIIDfTCCAwOgAwIBAgIQcEsRi8-Fdnm6pDlRYUODODAKBggqhkjOPQQDAzAeMRwwGgYDVQQDDBNDU01FIE1DQyBQVFQgIDAxU1ZOMB4XDTIwMTEyNTAwMDAwMFoXDTQ5MTIzMTIzNTk1OVowADCCASIwDQYJKoZIhvcNAQEBBQADggEPADCCAQoCggEBANVJpV6bONLCJUAuSYY2I9U9w5Iub-rmXZgkz6QfmPyd_1J37N0trb8WZhjJBuaKrih7BDTn38onx4hh_EncTspxeNw4aD4NcKb7DmltO-AaFOC8gVeGz63pjLG6GfmW_OkZPOBU6abkHISZ83tFABy-TbDJ5WnMORX6J5WJkZmHbXOYYWM5l8FObvjdhVJZBdPE3XtjjrfULeUr8rvQdfrSetD-O5wdCWqseVOlT5K20BlQl9_xHaxvNrsjIb2odgyXghCQjO_HqkF6cspVhMTFgPGRNV80Ovxs2DparDu_4oszLX1BBixz1C1qnd2q5_4sehI3OjUNwZbq4NqOr28CAwEAAaOCAXQwggFwMB8GA1UdIwQYMBaAFKdHi_DnvMmrvGdd3uFLuy4cAW07MA8GA1UdEwEB_wQFMAMBAQAwDgYDVR0PAQH_BAQDAgUgMIGgBgNVHR8EgZgwgZUwgZKgSqBIhkZodHRwczovL3RzY2kuaW50ZWwuY29tL2NvbnRlbnQvT25EaWVDQS9jcmxzL09uRGllX0NBX0NTTUVfSW5kaXJlY3QuY3JsokSkQjBAMSYwJAYDVQQLDB1PbkRpZSBDQSBDU01FIEludGVybWVkaWF0ZSBDQTEWMBQGA1UEAwwNd3d3LmludGVsLmNvbTAQBgNVHSUECTAHBgVngQUIATAlBgNVHQkBAQAEGzAZMBcGBWeBBQIQMQ4wDAwDMi4wAgEAAgIAijBQBgNVHREBAf8ERjBEpEIwQDEWMBQGBWeBBQIBDAtpZDo0OTRFNTQ0MzEOMAwGBWeBBQICDANNQ0MxFjAUBgVngQUCAwwLaWQ6MDI1ODAwMEYwCgYIKoZIzj0EAwMDaAAwZQIwRR8zP55haDq3gzY-32uNrYTnNfyIuT-ihoIGmXkF4iq8brbDBl3cVIK5bS9clEgaAjEA-6C7L27nBLG9xIZisXzGYyDXXmAoSPBNwRKZap71T024F5U7ACmGmzc834NE7_L7) . When compared to the [re-encoded and working version of pyasn1](https://lapo.it/asn1js/#MIIDejCCAwCgAwIBAgIQcEsRi8-Fdnm6pDlRYUODODAKBggqhkjOPQQDAzAeMRwwGgYDVQQDDBNDU01FIE1DQyBQVFQgIDAxU1ZOMB4XDTIwMTEyNTAwMDAwMFoXDTQ5MTIzMTIzNTk1OVowADCCASIwDQYJKoZIhvcNAQEBBQADggEPADCCAQoCggEBANVJpV6bONLCJUAuSYY2I9U9w5Iub-rmXZgkz6QfmPyd_1J37N0trb8WZhjJBuaKrih7BDTn38onx4hh_EncTspxeNw4aD4NcKb7DmltO-AaFOC8gVeGz63pjLG6GfmW_OkZPOBU6abkHISZ83tFABy-TbDJ5WnMORX6J5WJkZmHbXOYYWM5l8FObvjdhVJZBdPE3XtjjrfULeUr8rvQdfrSetD-O5wdCWqseVOlT5K20BlQl9_xHaxvNrsjIb2odgyXghCQjO_HqkF6cspVhMTFgPGRNV80Ovxs2DparDu_4oszLX1BBixz1C1qnd2q5_4sehI3OjUNwZbq4NqOr28CAwEAAaOCAXEwggFtMB8GA1UdIwQYMBaAFKdHi_DnvMmrvGdd3uFLuy4cAW07MA8GA1UdEwEB_wQFMAMBAQAwDgYDVR0PAQH_BAQDAgUgMIGgBgNVHR8EgZgwgZUwgZKgSqBIhkZodHRwczovL3RzY2kuaW50ZWwuY29tL2NvbnRlbnQvT25EaWVDQS9jcmxzL09uRGllX0NBX0NTTUVfSW5kaXJlY3QuY3JsokSkQjBAMSYwJAYDVQQLDB1PbkRpZSBDQSBDU01FIEludGVybWVkaWF0ZSBDQTEWMBQGA1UEAwwNd3d3LmludGVsLmNvbTAQBgNVHSUECTAHBgVngQUIATAiBgNVHQkEGzAZMBcGBWeBBQIQMQ4wDAwDMi4wAgEAAgIAijBQBgNVHREBAf8ERjBEpEIwQDEWMBQGBWeBBQIBDAtpZDo0OTRFNTQ0MzEOMAwGBWeBBQICDANNQ0MxFjAUBgVngQUCAwwLaWQ6MDI1ODAwMEYwCgYIKoZIzj0EAwMDaAAwZQIwRR8zP55haDq3gzY-32uNrYTnNfyIuT-ihoIGmXkF4iq8brbDBl3cVIK5bS9clEgaAjEA-6C7L27nBLG9xIZisXzGYyDXXmAoSPBNwRKZap71T024F5U7ACmGmzc834NE7_L7), there is only one difference in the `Extension > extnID OBJECT IDENTIFIER 2.5.29.9 subjectDirectoryAttributes`, where the `critical BOOLEAN false` (3 Bytes) attribute is removed in the pyasn1 version but present in the original intel version. openssl also has no problem with the certificate. According to this [documentation](https://datatracker.ietf.org/doc/html/rfc5280#section-4.1) the presence of the attribute seems to be valid.  This is the resulting exeption when the certificate is loaded by `cryptography`:

```python
>>> x509.load_pem_x509_certificate(data=ek.encode("utf-8"), backend=default_backend())
  File "<python-input-13>", line 1, in <module>
    x509.load_pem_x509_certificate(data=ek.encode("utf-8"), backend=default_backend())
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib64/python3.13/site-packages/cryptography/x509/base.py", line 551, in load_pem_x509_certificate
    return rust_x509.load_pem_x509_certificate(data)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
ValueError: error parsing asn1 value: ParseError { kind: EncodedDefault, location: ["RawCertificate::tbs_cert", "TbsCertificate::extensions", 5, "Extension::critical"] }
```

Without further inspection, I would guess that the problem appears to be related to `EncodedDefault`, as a default value is explicitly set.

It this really a problem, what is your opinion on that?

ansasaki:
> Without further inspection, I would guess that the problem appears to be related to `EncodedDefault`, as a default value (critical = FALSE) is explicitly set?
>
> It this really a problem, what is your opinion on that?

Thank you for investigating this and finding the root cause. But, unfortunately, explicitly encoding a default value inside a `SEQUENCE` is indeed incorrect, making the certificate malformed. The DER encoding specification ([ITU-T X.690](https://www.itu.int/ITU-T/studygroups/com17/languages/X.690-0207.pdf)), section 11.5, says:

> "The encoding of a set value or sequence value shall not include an encoding for any component value which is equal to its default value."

In my opinion, although the certificate content is malformed under the DER encoding rules, the signature can still be verified, which is the most important part. As you mentioned, openssl is more flexible and does not enforce the rule, resulting in a successful signature verification.

But my opinion cannot change the fact that the certificate is malformed, and the python cryptography developers have the right to follow the specification and enforce the rules.

PR #1688, opened 2024-11-21
docs: update TCTI environment variable usage

tuminoid:
Keylime Rust agent does not understand `TPM2TOOLS_` prefixed configuration. For example, `TPM2TOOLS_TCTI` is ignored when setting up swtpm.

`.ci/test_wrapper.sh` is already exporting both formats, but installation documentation is not mentioning the shorter format at all.


tuminoid:
Note: https://github.com/keylime/rust-keylime/issues/871 was created in Rust agent, so I did not replace the long format but added short one in addition.

tuminoid:
How do I rerun a failed test? I don't think attestation-in-localhost test is related to documentation change.

PR #1693, opened 2024-11-27
Add agent-driven (push) attestation protocol

gnurugs:
This PR introduces an agent-driven attestation protocol to Keylime as part of the work to support the [push model](keylime/enhancements#103). This includes:

- A PushAttestation model to manage state across requests
- A PushAttestationController to handle incoming requests

Agent-driven (push) attestation is done in two phases:

1. **Capabilities negotiation**: The agent makes known what capabilities it supports and the verifier uses this to instruct the agent on what attributes to use when preparing the evidence (e.g., nonce, signing algorithm to use, etc.).
2. **Evidence handling**: The agent collects and sends the evidence according to the verifier's specification.

Verification then proceeds out-of-band after issuing the HTTP response.

A number of tasks still need to be completed before this PR is ready to be merged:

- [ ] Sync with master
- [ ] Fix model associations
- [ ] Improve JSON rendering of attestation resources
- [ ] Add config options for `nonce_lifetime` and `verification_timeout`.
- [ ] Improve documentation
- [ ] Linting and formatting fixes
- [ ] Fix test failures


kkaarreell:
> This PR introduces an agent-driven attestation protocol to Keylime as part of the work to support the [push model](keylime/enhancements#103).

Hi. Just FYI, the link is not working.

kkaarreell:
Hi @stringlytyped , do you also plan to implement new (unit) tests for this? Existing tests won't be able to test the added functionality and while I can help with new functional tests it would be great to have it covered on a unit test level as well. I am asking because I can't see such an item on the ToDo list above.

stringlytyped:
Hi @kkaarreell,

> > This PR introduces an agent-driven attestation protocol to Keylime as part of the work to support the [push model](keylime/enhancements#103).
> 
> Hi. Just FYI, the link is not working.

Whoops, that was meant to go to keylime/enhancements#103. That's now fixed.

> Hi @stringlytyped , do you also plan to implement new (unit) tests for this? Existing tests won't be able to test the added functionality and while I can help with new functional tests it would be great to have it covered on a unit test level as well. I am asking because I can't see such an item on the ToDo list above.

I agree that unit tests would be great to have, especially for the library-level code. I am still making changes to the model libraries in `keylime.model.base` with some regularity, but the web-layer code in `keylime.web.base` is fairly stable, so I can try and add some unit tests there.

For functional tests, I am waiting until the agent-side changes are further along before we add those. 

Even without full test coverage, I don't think that should prevent merging, as push mode will be off by default at first. When turned on, the user is [warned](https://github.com/keylime/keylime/blob/2bf9119779a20baa9db7912b4603b574c062987c/keylime/cmd/verifier.py#L16-L21) that the feature is experimental.

keylime/common/algorithms.py
```diff
@@ -71,13 +71,13 @@ def __str__(self) -> str:
         return self.value
 
 
-class Encrypt(str, enum.Enum):
+class Key(str, enum.Enum):
```
kkaarreell, line 85: Hi @stringlytyped , this change was not reflected in the testsuite where import is failing.
```
File "/tmp/tmp.e280gE3el5/test/test_algorithms.py", line 5, in <module>
    from keylime.common.algorithms import Encrypt, Hash, Sign
ImportError: cannot import name 'Encrypt' from 'keylime.common.algorithms' (/usr/local/lib/python3.13/site-packages/keylime-7.12.1-py3.13.egg/keylime/common/algorithms.py)
```

stringlytyped: Hi @kkaarreell, sorry I didn't see this comment last week. This should be fixed now.

PR #1694, opened 2024-11-29
tests: Use Fedora 41 to generate code coverage

ansasaki:
Replace the Fedora 39 usage to generate code coverage with Fedora 41.

This is necessary because Fedora 39 is now EOL and the tests are no longer triggered for it.


sergio-correia (approved):
LGTM. The failures are unrelated to this PR. In f41 it fails due to https://bugzilla.redhat.com/show_bug.cgi?id=2326535 and in c10s it is somehow unable to download the srpm of the agent; looks like some compose issue.

PR #1704, opened 2024-12-17
Docs: expand security/threat model page

stringlytyped:
Building and deploying systems for remote attestation is difficult to do securely, as it requires domain-specific expertise in a wide variety of areas (cryptography, TPMs, UEFI, the Linux kernel, etc.). We have been missing a comprehensive overview of these topics, approachable to non-experts. I have taken the liberty of overhauling the existing threat model page to explain how the security of an attestation system, specifically Keylime, is derived. It is intended as the first port of call for new Keylime users and developers alike who need to understand how to achieve secure attestation.

> [!NOTE]  
> You can see a _Read the Docs_-rendered preview of the page here:
> https://keylime-docs-wip.readthedocs.io/en/latest/design/security.html
> (This will automatically remain in sync with subsequent pushes to this branch)

I have used RATS terms and try to link these back to TCG terminology wherever possible. I've had to expand some RATS notions a bit in order to make some complicated topics approachable (for instance, what is the role of the TPM in relation to other parts of a system?)

Because everything is explained from first principles, it should remain relevant as Keylime grows and attestation technologies in general evolve. I've also tried to be forward-looking where possible. For example, while the page frequently mentions the TPM (of course), the discussion applies even in a confidential computing context. (On that note, I've also included a short section on TEE attestation for future implementers of such functionality.)

At the bottom of the page, you will find a fairly terse adversarial model, inspired by the sort you would find in a formal security analysis. This is important for secure systems design, so that the security characteristics of a system are clear, and to make it easy delineate between vulnerabilities which are in scope and those which are out of scope.

@THS-on and @edwards-n have provided feedback on earlier versions of the page, so thank you both for your contributions.

There are still a few things I wish to fix, namely:

- [x] Add a footnote to reference [SBAT](https://github.com/rhboot/shim/blob/main/SBAT.md) (suggested by Thore)
- [ ] Incorporate anything that's still missing from the [existing threat model page](https://keylime.readthedocs.io/en/latest/design/threat_model.html)
- [x] Restate exclusions so that UEFI bootkits are not universally out of scope.
- [ ] Better tie in the adversarial model at the end with the long-form discussion that precedes it
- [ ] Sync branch with master

I look forward to hearing your thoughts!


ansasaki:
@stringlytyped thank you for working on this! I read the current version and could not spot anything missing or incorrect. Looking forward to the completion of this essential page!

stringlytyped:
I've updated the threat model's exclusions section to fix a mistake I noticed. UEFI bootkits should not be universally excluded, as in fact, measured boot can be used to detect them in many instances. The previous exclusion was written as follows:

>  Attacks which rely on modification of an attesting environment (such as by using a UEFI bootkit) are also excluded.

This has now been updated to:

> Attacks which rely on modification of a system root of trust (such as the BMC or TPM) are also excluded.

Additionally, the discussion around Secure Boot limitations and UEFI bootkits has been updated with references to past Secure Boot bypass vulnerabilities and the use of SBAT to revoke vulnerable bootloaders.

Issue #1614, opened 2024-08-09
CA and logging configuration should be separated per component


ansasaki:
Is your issue a feature request? If so, please raise it as an [enhancement](https://github.com/keylime/enhancements)

# Environment
 * OS / version: N/A
 * Processor architecture: N/A
 * TPM Manufacturer: N/A
 * Keylime version: current master (dc118c780a06cb72725f72bc8100665a07a5a12f)

# Description 
The `ca.conf` and `logging.conf` are shared between all components. Ideally these should be separated per component, maybe integrated as part of each component configuration.

This would allow the CA and logging configuration to be modified separately for components running on the same machine.

## Expected behavior vs. actual behavior 
Expected:
The verifier, registrar, and tenant CA and logging configuration can be modified independently when running on the same machine

Actual:
The verifier, registrar, and tenant share the same CA (`/etc/ca.conf`) and logging (`/etc/logging.conf`) configuration when running on the same machine

## Steps to reproduce problem
1. Run any of the components. Observe that they load the same CA and logging configuration files.

# Relevant logs
A line similar to the following is logged for all components:
```
INFO:keylime.config:Reading configuration from ['/etc/keylime/logging.conf']
```
When the component is configured to generate a local CA, a line similar to the following is logged:
```
2024-08-08 14:50:31.541 - keylime.config - INFO - Reading configuration from ['/etc/keylime/ca.conf']
```

Attach any relevant log files that can help to debug your issue.

marcostork:
@ansasaki Is the goal of this proposal to enable adding or updating the CA and logging configurations directly through each component's configuration after loading the ca.conf and logging.conf, or is it to allow the complete bypass of the shared ca.conf and logging.conf files altogether?

I can see two scenarios:

In the first scenario, the presence of ca.conf and logging.conf would be required, but each component's configuration could override or modify the CA and logging settings.
In the second scenario, the component's configuration file would be checked first for the CA and logging settings, with ca.conf and logging.conf used only as fallback options if the settings are not found.

marcostork:
I'm almost finished with the logging configuration solution (I still need to add tests), and I'm working on the CA configuration. Could you please assign it to me to avoid overlapping work?

ansasaki:
> @ansasaki Is the goal of this proposal to enable adding or updating the CA and logging configurations directly through each component's configuration after loading the `ca.conf` and `logging.conf,` or is it to allow the complete bypass of the shared `ca.conf` and `logging.conf` files altogether?

[...]

PR #1715, opened 2025-01-10
Allow separate CA and logging configurations for components

marcostork:
Hello,

I've worked on issue #1614, and while submitting this PR, I’d like to address a few key points:

**Logging**
Currently, both methods of loading a full configuration reset the root logging setup, overwriting any existing configuration. While this behavior is acceptable for standalone command-line tools, it causes significant issues when the module is used as a library, as it disrupts the application's existing logging configuration.
To address this, I propose dynamically setting the logging configuration whenever the logging library is called. This approach allows incremental adjustments to configurations and avoids losing existing setups.

**CA**
Although changes were required across the codebase, they were relatively straightforward. Using the Red Hat documentation (Chapter 7: Ensuring System Integrity with Keylime) as a reference, I implemented several improvements and enabled component-specific configurations. 

I may have gone slightly beyond the original scope, but here are the highlights:

- Default Logging Configuration: A hardcoded default configuration ensures no setup is required for basic use, simplifying Keylime usage while retaining flexibility for advanced configurations.

- Configuration Templates: Templates now include detailed instructions for customizing logging and CA settings (mostly from ca.conf).

- `web_util`: Removed restrictions requiring specific options to be set to "default" when tls_dir is set to "generate." This change allows different certificates for the server and client allow lists, enabling separate certificates for components. Added support for tenant-specific certificates.

- `ca_util`: Updated command-line options to accommodate new component-specific configurations.

- `test_logging`: Added testing for keylime_logging.

- Templates: Added relevant logging and CA configuration options to each component template.

About Testing: Although I lack a comprehensive testing setup, I validated the following scenarios: Verifier, registrar, and tenant running on the same node and each component running on separate nodes. For both scenarios, I tested cases where components shared the same CA and where each component used its own CA.

Some tasks remain unaddressed, but should they be approved, I would gladly add to this PR or make another one:
1. Remove the current logging and CA templates.
2. Add support for creating tenant-specific certificates with init_mtls.
3. Upgrade the tenant component to use init_mtls for CA creation when required (is keylime_tenant currently being refactored?).
4. Clean up the commented-out code that restricts client and server trusted list names.

Please let me know if this PR needs further adjustments or refinements.


marcostork:
The first failure (basic-attestation-on-localhost-api-version-bump) seems unrelated to the PR, and the second failure (opened-conf-files) is expected because the logging and CA configuration files are no longer read.

ansasaki:
Thank you for working on this! I still have to review it and test what are the consequences, but this is a welcome change.

This is a configuration breaking change (you are removing things), which means you will need to create a new series of templates and bump the version to `3.0`. This will also require testing the scenario where the configuration is updated from a `2.x` version to the new `3.0` version.

You cannot remove templates, as they exist to allow users to safely upgrade from an older configuration version to the new version.

You will need to include some necessary "special" handling of the configuration (implemented in the template `adjust.py` file, check the code in the `2.0` template), since you will have to populate the newly generated files using the content from the old configuration (in this particular case, i think this will be straightforward since it is a matter of populating each components configuration with the content of the old shared config).

I'll add more comments as I review the code and see important things missing.

marcostork:
@ansasaki Thanks for this first review. I'll fix the templates and work on bumping to the new version.

marcostork:
@anderson, the PR mostly rearranges the configuration options; it does not include, remove, or change the current mapping options. The templates I've included in the PR expect to recover the information from the old mapping. 
Since there is no need to change the mapping, I consolidated the last updates into a new version; please let me know if there is a better option.
I have also replicated a fix to handle the subcomponent version inside process_update_mapping and merged this PR to the keylime upstream.

marcostork:
Given the number of changes since this PR was opened, I’m going to close it in favor of a fresh implementation.

PR #1717, opened 2025-01-15
Add /version endpoint to Registrar

ansasaki:
Introduce the `/version` endpoint to the registrar, accessible without authentication.
This endpoint is used by the agent to decide which API version to use when making requests to the registrar.

As part of this PR, the unused file `registrar_common.py` is deleted.

This is part of the implementation of the enhancement proposal 114: https://github.com/keylime/enhancements/pull/115


sergio-correia (approved):
LGTM, thanks

PR #1735, opened 2025-02-05
Remove excessive logging on exception

kaifeng:
log.exception already contains exception info, there is no need to log twice with separate logging levels.


kaifeng:
Also addressed lint issues of typing.

ansasaki (approved):
Thank you, LGTM!

PR #1736, opened 2025-02-06
tests: change test_mba_parsing to not need keylime installed

sergio-correia:
This test needs the verifier configuration file available, and on systems that do not have keylime installed (hence, no config file), it would fail.

This commit changes the test so that it creates a verifier conf file in a temporary directory with default values, so that it can use it.


ansasaki:
The code coverage test failure was fixed by https://github.com/keylime/keylime/pull/1738

PR #1746, opened 2025-03-03
tpm_util: fix quote signature extraction for ECDSA

THS-on:
This fixes the signature extraction for ECDSA signatures and unblocks: https://github.com/keylime/rust-keylime/pull/846

### DIFF
--- a/keylime/tpm/tpm_util.py
+++ b/keylime/tpm/tpm_util.py
@@ -223,9 +223,7 @@ def checkquote(
     pcrblob: The state of the PCRs that were quoted; Intel tpm2-tools specific format
     exp_hash_alg: The hash that was expected to have been used for quoting
     """
-    sig_alg, hash_alg, sig_size = struct.unpack_from(">HHH", sigblob, 0)
-
-    (signature,) = struct.unpack_from(f"{sig_size}s", sigblob, 6)
+    sig_alg, hash_alg = struct.unpack_from(">HH", sigblob, 0)
 
     pubkey = serialization.load_pem_public_key(aikblob, backend=backends.default_backend())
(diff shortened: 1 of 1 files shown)


keylime/tpm/tpm_util.py
```diff
@@ -236,6 +234,14 @@ def checkquote(
     if isinstance(pubkey, EllipticCurvePublicKey) and sig_alg not in [tpm2_objects.TPM_ALG_ECDSA]:
         raise ValueError(f"Unsupported quote signature algorithm '{sig_alg:#x}' for EC keys")
 
+    if sig_alg in [tpm2_objects.TPM_ALG_RSASSA]:
+        (sig_size,) = struct.unpack_from(">H", sigblob, 4)
+        (signature,) = struct.unpack_from(f"{sig_size}s", sigblob, 6)
+    elif sig_alg in [tpm2_objects.TPM_ALG_ECDSA]:
+        signature = ecdsa_der_from_tpm(sigblob)
```
stefanberger, line 241: This is probably the better function to call. The commit text does not say anything, but did this not work (correctly) so far? Presumably `pubkey.verify()` of the ECDSA key would have thrown an error... ?

THS-on: Yes `pubkey.verify()` has thrown an error at least on non P256 curves.

stefanberger: Since swtpm uses P384 I thought that it had been working at least with that one...

THS-on: https://github.com/keylime/rust-keylime/pull/846 actually implements non default key sizes/curves. Before I started working on this the necessary support in the rust binding for generating the correct EKs was missing.

stefanberger (approved):

PR #1758, opened 2025-05-09
Fix pylint invalid name warnings (C0103)

msafarik:
I fixed valid C0103 errors that pylint generated, all other errors that are generated are invalid because pylint thinks that some variables are constants and etc. It is only small change and it also does not change anyting functional and also all tests passed (except that one that should not pass)

### DIFF
--- a/keylime/mba/elparsing/tpm2_tools_elparser.py
+++ b/keylime/mba/elparsing/tpm2_tools_elparser.py
@@ -216,7 +216,6 @@ def __add_boot_aggregate(log: Dict[str, Any]) -> None:
 
 _tpm2_tools_version = None
 
-
 def tpm2_tools_getversion() -> str:
     global _tpm2_tools_version
     if _tpm2_tools_version is not None:
--- a/keylime/mba/elparsing/tpm2_tools_elparser.py
+++ b/keylime/mba/elparsing/tpm2_tools_elparser.py
@@ -216,6 +216,7 @@ def __add_boot_aggregate(log: Dict[str, Any]) -> None:
 
 _tpm2_tools_version = None
 
+
 def tpm2_tools_getversion() -> str:
     global _tpm2_tools_version
     if _tpm2_tools_version is not None:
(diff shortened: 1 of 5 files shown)


keylime/revocation_notifier.py
```diff
@@ -190,10 +190,8 @@ def worker_webhook(tosend: Dict[str, Any], url: str) -> None:
     t = threading.Thread(target=w, daemon=True)
     t.start()
 
-
```
ansasaki, line 193: Please add back this empty line as it is required by the current style format

keylime/revocation_notifier.py
```diff
@@ -190,10 +190,8 @@ def worker_webhook(tosend: Dict[str, Any], url: str) -> None:
...
 cert_key = None
 
-
```
ansasaki, line 196: Please add back this empty line as it is required by the current style format

keylime/mba/elparsing/tpm2_tools_elparser.py
```diff
@@ -216,7 +216,6 @@ def __add_boot_aggregate(log: Dict[str, Any]) -> None:
 
 _tpm2_tools_version = None
 
-
```
ansasaki, line 219: Please add back this empty line as it is required by the current style format

keylime/mba/elparsing/tpm_bootlog_enrich.py
```diff
@@ -23,7 +23,7 @@
 ##################################################################################
 
 
-class hexint(int):
```
ansasaki, line 26: I believe this class name is indeed invalid as it should start with a capital letter.
You could fix this as part of this PR, but I'm OK if you prefer to leave this to a future PR

ansasaki (changes requested):
Please fix the parts I commented in my review.

Also, use a more specific title for you pull request. Something like: "Fix pylint invalid name warnings (C0103)"

Finally, please sign-off each of the commits in your pull request.

sarroutbi (changes requested):
class names are conventionally CamelCase so I would go with HexInt rather

ansasaki (approved):
LGTM!

The linting issue is related with a breaking change in `zmq`: https://pyzmq.readthedocs.io/en/latest/api/zmq.devices.html

The `zmq.device` was already deprecated in favor of `zmq.proxy`.
I reported this in a new issue #1760

sarroutbi (approved):
LGTM

Issue #545, opened 2021-02-12
tenant should validate action_list


lukehinds:
Tenant does not parse out local_actions_* from the action_list to ensure all files are present before sending to the agent.

PR #1759, opened 2025-05-12
Tenant should validate action list

msafarik:
Issue: tenant should validate action_list #545

**Resolves:** #545 

## Change Description
I made function that is called in main function in Tenant class to validate if command is from valid commands list.

*(Check all that apply)*
- [ *] No docs needed (requires maintainer approval)

### DIFF
--- a/docker/ci/Dockerfile
+++ b/docker/ci/Dockerfile
@@ -5,7 +5,7 @@
 # It is not recommended for use beyond testing scenarios.
 ##############################################################################
 
-FROM quay.io/fedora/fedora@sha256:5f262b2b2da44b78eaa96d52b45bd1834d330d3d17e2dbcf3027caef10b12c8c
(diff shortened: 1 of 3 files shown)


ansasaki:
I think you misunderstood the issue reported in #545.

The issue is about the lack of validation of the presence of the files listed in the `action_list` file when constructing the the optional secure payload (see the documentation https://keylime.readthedocs.io/en/latest/user_guide/secure_payload.html#)

Basically, it is possible to include scripts as part of the secure payload zip file, and list them in a file called `action_list` in the zip file. The problem is that when the tenant is creating the secure payload zip it does not check if each file listed in the `action_list` file is present in the directory.

The proper fix would be to open the `action_list` file in the directory (when present), parse its content to obtain the names of the files that should be present in the directory, and then check that the files are present.

Issue #1766, opened 2025-06-03
Base image build fails due to `tpm2-tools` compilation failure


ansasaki:
# Environment
 * OS / version: N/A
 * Processor architecture: N/A
 * TPM Manufacturer: N/A
 * Keylime version: current master (b57781c55308930e1a87cb9ae7f41782764c0522)

# Description 
The base image build fails due to failure compiling `tpm2-tools` (in the last run using the commit c6e182cf690ca0022b77fe7d3c174659917a7e7e)

The compilation failure is apparently caused by the use of the option `-Wformat-security` which is not supported by the compiler.

## Expected behavior vs. actual behavior 
Expected: The base image is built successfully
Actual: The base image fails to build due to the failure on `tpm-tools` compilation

## Steps to reproduce problem
1. The CI runs the following command:
```
/usr/bin/docker buildx build --file docker/release/base/Dockerfile --iidfile /home/runner/work/_temp/docker-actions-toolkit-E2uK22/build-iidfile-691af1b26f.txt --label org.opencontainers.image.created=2025-06-02T15:44:19.523Z --label org.opencontainers.image.description=A CNCF Project to Bootstrap & Maintain Trust on the Edge / Cloud and IoT --label org.opencontainers.image.licenses=Apache-2.0 --label org.opencontainers.image.revision=b57781c55308930e1a87cb9ae7f41782764c0522 --label org.opencontainers.image.source=https://github.com/keylime/keylime --label org.opencontainers.image.title=keylime --label org.opencontainers.image.url=https://github.com/keylime/keylime --label org.opencontainers.image.version=master --tag quay.io/keylime/keylime_base:master --tag quay.io/keylime/keylime_base:monthly --tag quay.io/keylime/keylime_base:sha-b57781c --metadata-file /home/runner/work/_temp/docker-actions-toolkit-E2uK22/build-metadata-3d41ebccf3.json --push .
```

# Relevant logs
See https://github.com/keylime/keylime/actions/runs/15396628697/job/43318932325

This is the relevant excerpt:
```
#6 12.67 checking whether the C compiler accepts -Wformat-security... no
#6 12.69 configure: error: Cannot enable -Wformat-security, consider configuring with --disable-hardening
[...]

PR #1767, opened 2025-06-03
docker: Remove tpm2-tools compilation from base image

ansasaki:
# PR Title 
Use the system provided tpm2-tools instead of building from upstream code.

## Type of Change
*(Select all that apply)*
- [x] CI/CD changes

## Related Issues
Resolves: #1766

### Concise Summary
Install the system-provided tpm2-tools instead of building from the source.

This adds the `tpm2-tools` and `awk` packages to the base image.

### Technical Details
 - This avoids the issue with `tpm2-tools` compilation when building the base image in the CI

*(Check all that apply)*

## Verification Process
1. Run the `docker/release/build_locally.sh and check that it correctly builds the base image

## Checklist

## Additional Context
N/A


tuminoid (approved):
> > Should `installer.sh` be cleaned up while at it?
> 
> Could you please explain what exactly are you proposing? I do not think we can remove it from the script as there is no guarantee that the script will always be executed in an environment where the `tpm2-tools` is already installed. The script could be executed in a bare metal machine without any dependency previously installed (other than those necessary to execute the script itself).

I was thinking of removing the git repo cloning and build tool installation etc, but you're correct, we cannot do it. It might be running in other env than Dockerfile. Docker build logs will look a bit funny now, but that is not major issue.

Thanks for fixing the core issue!

sergio-correia (approved):
Thanks, looks good to me.

PR #1771, opened 2025-06-11
CI: Enable CONTAINER_ENGINE to allow alternative container engines

sarroutbi:
Minimum change to allow CONTAINER_ENGINE
environment variable to be set with a
different container engine (i.e.podman)

# PR Title 
`CI: Enable CONTAINER_ENGINE to allow alternative container engines`

## Type of Change
*(Select all that apply)*
- [x] CI/CD changes

## Related Issues
N/A

### Concise Summary
Test run are normally executed locally before pushing changes to ensure regression tests are OK and nothing is broken. However, tests are run through docker, and can not be executed with other container engine (i.e.`podman`).

### Technical Details
Allow using a different container engine rather than `docker` through `CONTAINER_ENGINE` environment variable. If nothing is provided, `docker` will be used by default (as it is now)

## Verification Process
1 - Configure machine where docker is not installed (only podman is)
2 - Execute ci tests:
```bash
$ .ci/run_local.sh
Fetching Keylime CI container
.ci/run_local.sh: line 18: docker: command not found
Running Keylime's test suite
.ci/run_local.sh: line 23: docker: command not found
.ci/run_local.sh: line 28: docker: command not found
.ci/run_local.sh: line 30: docker: command not found
.ci/run_local.sh: line 31: docker: command not found
```
3 - Provide CONTAINER_ENGINE environment variable to clarify `podman` wants to be used:
```bash
$ CONTAINER_ENGINE=podman .ci/run_local.sh
Fetching Keylime CI container
Trying to pull quay.io/keylime/keylime-ci:latest...
...
OK (expected_failures=1, passes=189, skips=1)
==================================================================================
                        Cleanup
==================================================================================
Removing /tmp/tmp.Khv1SrRsxt
```
## Checklist
- [x] CHANGELOG updated (if applicable)

## Additional Context
N/A

### DIFF
--- a/.ci/run_local.sh
+++ b/.ci/run_local.sh
@@ -5,23 +5,37 @@
 # The default value is one directory above where this script is located.
 REPO=${KEYLIME_REPO_PATH:-${1:-$(realpath "$(dirname "$(readlink -f "$0")")/../")}}
 
+# You can specify the container engine to use with the CONTAINER_ENGINE
+# environment variable. The default to be used will be 'docker',
+# or 'podman' if it exists and 'docker' don't
+if [ -z "$CONTAINER_ENGINE" ]; then
+    if command -v docker &> /dev/null; then
+        CONTAINER_ENGINE="docker"
+    elif command -v podman &> /dev/null; then
+        CONTAINER_ENGINE="podman"
+    else
+        echo "ERROR: No container engine specified, and could not find 'docker' or 'podman' in PATH." >&2
+        exit 1
+    fi
+fi
+
 # keylime images
 tpmimage="quay.io/keylime/keylime-ci"
 tpmtag="latest"
 
 echo -e "Fetching Keylime CI container"
 
-docker pull ${tpmimage}:${tpmtag}
+${CONTAINER_ENGINE} pull ${tpmimage}:${tpmtag}
 
 echo -e "Running Keylime's test suite"
 
 container_id=$(mktemp)
-docker run --detach --user root --env KEYLIME_TEST='true' --env RUST_TEST=1\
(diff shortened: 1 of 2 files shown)


ansasaki (commented):
Could you please extend this approach also for the `docker/release/build_locally.sh` script?

.ci/run_local.sh
```diff
@@ -5,23 +5,27 @@
 # The default value is one directory above where this script is located.
 REPO=${KEYLIME_REPO_PATH:-${1:-$(realpath "$(dirname "$(readlink -f "$0")")/../")}}
 
+# You can specify the container engine to use with the CONTAINER_ENGINE
+# environment variable. The default to be used will be 'docker'.
+CONTAINER_ENGINE=${CONTAINER_ENGINE:-docker}
```
sergio-correia: Perhaps it would make sense to check if either `docker` or `podman`, in that order, are available, if `CONTAINER_ENGINE` is not explicitly specified? That way it would work for most cases without requiring to set `CONTAINER_ENGINE` for the case one has podman but not docker.

sarroutbi: The reason to not do that (docker or podman) is to be `open` to other engines ... but I am OK to change it. Thanks for suggestion

sergio-correia: Having the variable means we are open to other engines -- please keep it that way.

In my view, podman seems to be common enough to warrant making it work without any additional configuration out of the box, but I may be wrong here. It was a just suggestion, in any case :)

ansasaki: I like this idea, to make it to work with `podman` without an extra configuration step, and using `docker` as a fallback.

sarroutbi: It would be the other way round, `docker` as default and only `podman` in case `docker` does not exist

ansasaki (approved):

PR #1783, opened 2025-07-23
Add 2.5 templates including Push Model agent changes

sarroutbi:
# PR Title 
Add 2.5 templates to include Push Model agent changes

## Type of Change
*(Select all that apply)*
- [ ] CI/CD changes

### Concise Summary

This commit introduces the necessary configuration template files for version 2.5. These new templates add the required configuration options to support the new agent operating in push mode.

The update follows the established configuration upgrade process by creating a new version directory and defining a non-breaking, minor version update.

### Technical Details
 
Configuration changes motivated from Push Model implementation:
- Includes configuration to access Verifier.
- Includes configuration related to exponential back-off.

*(Check all that apply)*

## Verification Process
New configuration values can be read from agent.

## Checklist

## Additional Context

### DIFF
--- a/templates/2.5/mapping.json
+++ b/templates/2.5/mapping.json
@@ -1,0 +1,33 @@
+{
+    "version": "2.5",
+    "type": "update",
+    "components": {
+        "agent": {
+            "add": {
+                "agent_data_path": "/var/lib/keylime/agent_data.json",
+                "verifier_url": "https://localhost:8881",
+                "exponential_backoff_max_retries": "5",
+                "exponential_backoff_initial_delay": "10000",
+                "exponential_backoff_max_delay": "360000",
+                "certification_keys_server_identifier": "ak",
+                "disabled_signing_algorithms": "",
+                "ima_logs_appendable": "True",
+                "ima_logs_formats": "text/plain",
+                "ima_logs_supports_partial_access": "True",
+                "ima_ml_count_file": "/tmp/ima_ml_count",
+                "uefi_logs_appendable": "False",
+                "uefi_logs_evidence_version": "1.0",
+                "uefi_logs_formats": "application/octet-stream",
+                "uefi_logs_supports_partial_access": "False",
+                "uefi_logs_binary_file_path": "/sys/kernel/security/tpm0/binary_bios_measurements"
+            }
+        },
+        "verifier": {
+            "add": {
+                "mode": "pull",
+                "challenge_lifetime": "1800",
+                "verification_timeout":  "0"
+            }
+        }
+    }
+}
--- a/templates/2.5/verifier.j2
+++ b/templates/2.5/verifier.j2
@@ -1,0 +1,261 @@
+# Keylime verifier configuration
+[verifier]
+
+# The configuration file version number
+version = {{ verifier.version }}
+
+# Unique identifier for the each verifier instances.
+uuid = {{ verifier.uuid }}
+
+# The binding address and port for the verifier server
+ip = "{{ verifier.ip }}"
+port = {{ verifier.port }}
+
+# The address and port of registrar server that the verifier communicates with
+registrar_ip = {{ verifier.registrar_ip }}
+registrar_port = {{ verifier.registrar_port }}
+
+# Enable mTLS communication between agent, verifier and tenant.
+# Details on why setting it to "False" is generally considered insecure can be found
+# on https://github.com/keylime/keylime/security/advisories/GHSA-2m39-75g9-ff5r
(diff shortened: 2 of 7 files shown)


sarroutbi:
> The following options are not present in the agent configuration:
> 
>     * `verifier_ip`
> 
>     * `verifier_port`

Should not we add them? I understand this information should be available in configuration, and way to connect to Verifier should not be exclusive to command line arguments. Apart from that ... it might be better to use something like `verifier_base_url` to allow setting `http://` or `https://`. WDYT?
> 
> 
> There are many missing options added for the push-attestation agent:
> 
>     * `certification_keys_server_identifier`
> 
>     * `disabled_signing_algorithms`
> 
>     * `ima_logs_appendable`
> 
>     * `ima_logs_formats`
> 
>     * `ima_logs_supports_partial_access`
> 
>     * `ima_ml_count_file`
> 
>     * `registrar_api_versions`
> 
>     * `uefi_logs_appendable`
> 
>     * `uefi_logs_evidence_version`
> 
>     * `uefi_logs_formats`
> 
>     * `uefi_logs_supports_partial_access`
> 
>     * `uefi_logs_binary_file_path`
> 

Sure. I'll add them

> 
> There is also an old option that is missing from all versions and we could add:
> 
>     * `agent_data_path`

I'll add it too

> 
> 
> For the verifier, there are also these that I'm aware of, but we could add later as part of the push-attestation branch:
> 
>     * `mode`
> 
>     * `challenge_lifetime`

ansasaki:
> Should not we add them? I understand this information should be available in configuration, and way to connect to Verifier should not be exclusive to command line arguments. Apart from that ... it might be better to use something like verifier_base_url to allow setting http:// or https://. WDYT?

I'm OK adding these verifier-related options to the templates as long as they are also added to the agent code to be actually used.

sarroutbi:
> > Should not we add them? I understand this information should be available in configuration, and way to connect to Verifier should not be exclusive to command line arguments. Apart from that ... it might be better to use something like verifier_base_url to allow setting http:// or https://. WDYT?
> 
> I'm OK adding these verifier-related options to the templates as long as they are also added to the agent code to be actually used.

I added `verifier_url` to the Agent configuration for this purpose

stringlytyped:
For the agent template:

- If we are adding a `verifier_url` option (which I agree we should), we should have a `registrar_url` option to match.
- If either `verifier_url` or `registrar_url` has a scheme of `http://`, that should be logged as a warning on startup.

For the verifier template:

- `mode` and `verification_timeout` look good
- `nonce_lifetime` should be `challenge_lifetime`
- We should add a `verifier_url` option to mirror the agent configuration so that the verifier can report URLs with the correct hostname/IP rather than bind address

For the registrar template:

- I am not seeing any changes from the previous version. I'm not missing anything, right?
- We should add `mode` here too in order to disable certain features when push is used (e.g., registration over insecure HTTP, payload feature, etc.)
- Again, I think we should add a `registrar_url` option for the same reason as above

@ansasaki A couple questions for you:

- Would it be feasible to rename `quote_interval` to `attestation_interval` so that it is more generic and makes sense for non-TPM uses?
- Once this is merged, can I make edits to the template to add more detailed documentation such as which options are push/pull specific?

templates/2.5/verifier.j2
```diff
@@ -0,0 +1,261 @@
...
+zmq_port = {{ revocations.zmq_port }}
+
+# Webhook url for revocation notifications.
+webhook_url = {{ revocations.webhook_url }}
+
+# Attestation mode. Can be 'pull' (traditional) or 'push' (agent-driven).
+mode = {{ verifier_mode }}
+#
+# Lifetime in seconds for nonces sent to agents in push mode.
+nonce_lifetime = {{ verifier_nonce_lifetime }}
```
ansasaki: This should be `challenge_lifetime`

templates/2.5/mapping.json
```diff
@@ -0,0 +1,33 @@
...
+                "uefi_logs_evidence_version": "1.0",
+                "uefi_logs_formats": "application/octet-stream",
+                "uefi_logs_supports_partial_access": "False",
+                "uefi_logs_binary_file_path": "/sys/kernel/security/tpm0/binary_bios_measurements"
+            }
+        },
+        "verifier": {
+            "add": {
+                "mode": "pull",
+                "nonce_lifetime": "1800",
```
ansasaki: This should be `challenge_lifetime`

ansasaki (changes requested):
Sorry @sarroutbi, could you please replace `nonce_lifetime` with `challenge_lifetime` and push again?

Issue #1775, opened 2025-06-27
EK Certificate Verification Fails Due to Issuer Mismatch in Keylime Agent's Retrieved Certificate


jjmorley:
# Environment (Agent)
 * OS / version: Ubuntu 24.04.2 LTS
 * Processor architecture: x86_64
 * TPM Manufacturer: Nuvoton 2111
 * Keylime version: rust_keylime_0.2.7

# Environment (Keylime Management)
 * OS / version: Fedora Server 42
 * Processor architecture: x86_64
 * TPM Manufacturer: N/A
 * Keylime version: 7.12.1

# Description 
When attempting to add an agent using keylime_tenant -add, I'm consistently encountering the error:
```
2025-06-27 11:38:11.922 - keylime.cert_utils - WARNING - Failed to parse PEM data with python-cryptography (might not be strictly conforming to DER ASN.1 encoding): error parsing asn1 value: ParseError { kind: InvalidSetOrdering, location: ["Certificate::tbs_cert", "TbsCertificate::issuer", 0, 2] }
2025-06-27 11:38:11.928 - keylime.cert_utils - ERROR - No Root CA matched EK Certificate
2025-06-27 11:38:11.928 - keylime.tenant - WARNING - Invalid EK certificate for Agent d432fbb3-d2f1-4a97-9ef7-75bd81c00000 (130.195.10.183:9002)
2025-06-27 11:38:11.928 - keylime.tenant - ERROR - TPM Quote from Agent d432fbb3-d2f1-4a97-9ef7-75bd81c00000 (130.195.10.183:9002) is invalid for nonce: mbDG8TPGN8Qn5XS7GKbg
```
Further investigation with the `ek-openssl-verify` script confirms the issue, returning `ERROR: EK signature did not match certificates from TPM cert store`.

I've narrowed down the problem to a discrepancy in the Issuer field of the EK certificate that the Keylime agent retrieves and stores.

Here's a breakdown of my findings:
1. **Manual EK Certificate Retrieval:** On the agent machine, I manually retrieved the EK certificate using `sudo tpm2_getekcertificate -o ek_cert_rsa.bin -o ek_cert_ecc.bin`.
2. **Successful Manual Verification:** When I passed this manually obtained certificate through `ek-openssl-verify`, it verified successfully against the NUVOTON 2111 pem file in my tpm_cert_store.
[...]

PR #1798, opened 2025-09-08
registrar: Do not store corrupted certificate in the DB

ansasaki:
# fix: registrar: Do not store corrupted certificate in the DB

## Type of Change
*(Select all that apply)*
- [ ] CI/CD changes

## Related Issues
Reverts #1794
Resolves: #1775

### Concise Summary
Create a new class `CertificateWrapper`, which handles the caching of the original bytes when necessary transparently, and replace the object used by the `keylime.models.base.types.Certificate` with the new wrapper.

### Technical Details
 
The previous attempt to fix the issue of the registrar storing possibly corrupted certificates in the DB (#1794) was incorrect: the cached original bytes is placed in the model object, and not the certificate instance object, meaning that it would affect following certificate instances.

In this PR, a new `CertificateWrapper` class that handles the existence of the original bits per object instance transparently is introduced.

*(Check all that apply)*

## Verification Process
1. Start the verifier and registrar
2. Start an agent using a TPM with a malformed certificate (e.g. custom swtpm)
3. Let the agent register
4. Check that the certificate stored in the `reg_data.sqlite` (when using sqlite) contains the unmodified certificate obtained from the TPM.

## Checklist

## Additional Context
Note that this PR reverts #1794

### DIFF
--- a/keylime/models/base/types/certificate.py
+++ b/keylime/models/base/types/certificate.py
@@ -12,6 +12,7 @@ from pyasn1_modules import pem as pyasn1_pem
 from pyasn1_modules import rfc2459 as pyasn1_rfc2459
 from sqlalchemy.types import Text
 
+from keylime.certificate_wrapper import CertificateWrapper, wrap_certificate
 from keylime.models.base.type import ModelType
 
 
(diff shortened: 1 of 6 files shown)


ansasaki:
@jjmorley Could you please check if this works with your Nuvoton TPM? Unfortunately our previous attempt to fix the issue was incorrect, hopefully this time it is correct.

keylime/models/base/types/certificate.py
```diff
@@ -223,23 +225,29 @@ def asn1_compliant(self, value: IncomingValue) -> Optional[bool]:
...
             return None
 
         match self.infer_encoding(value):
             case "decoded":
+                # Handle both raw cryptography certificates and our wrapper
+                if isinstance(value, CertificateWrapper):
+                    return value  # type: ignore[return-value]
+                if isinstance(value, cryptography.x509.Certificate):
+                    # Wrap raw certificate without original bytes
+                    return wrap_certificate(value, None)
```
stringlytyped: These should be handled as cases in `self.infer_encoding`. I suggest adding a `"wrapped"` case in addition to the existing `"decoded"` case.

stringlytyped: These cases should also be handled in the `asn1_compliant` method: the current definition does not take wrapped certificates into account, so it will produce unexpected results

ansasaki: Thanks for the suggestion, I'll make the changes.

ansasaki: TL;DR: I had to add also a `"disabled"` case to handle disabled mtls certificate.

When applying the suggested changes, I got an error in the case where the certificate is the `"disabled"` string, which is possible for the mtls certificate. This didn't happen before because the `isinstance` checks would not let the `public_bytes()` method to be called for a `str` variable. I removed those `isinstance` checks, but had to add a check for the special "disabled" string.

PR #1810, opened 2025-10-03
verify/evidence: Add claims to JSON response

tylerfanelli:
# verify/evidence: Add claims to JSON response

## Type of Change
*(Select all that apply)*
- [x] Breaking change (fix/feature causing existing behavior to change)
- [ ] CI/CD changes

This adds a claims response to the `/verify/evidence` API used for one-shot attestation. The change is principally motivated by TEE attestation. Clients request signed claims responses (likely via JWTs) in order to verify that a Keylime verifier has validated the attestation evidence and can attest to those specific claims.

cc/ @mpeters


keylime/cloud_verifier_tornado.py
```diff
@@ -1511,13 +1517,13 @@ def post(self) -> None:
...
             else:
-                attestation_response["valid"] = 1
+                attestation_response["valid"] = True
             # TODO - should we use different error codes for attestation failures even if we processed correctly?
             web_util.echo_json_response(self, 200, "Success", attestation_response)
         except Exception:
             web_util.echo_json_response(self, 500, "Internal Server Error: Failed to process attestation data")
 
-    def _tpm_verify(self, json_body: dict[str, Any]) -> Failure:
+    def _tpm_verify(self, data: dict[str, Any]) -> Tuple[dict[str, Any], Failure]:
```
sergio-correia, line 1699: `_tpm_verify()` returns the input data dict as claims: `return (data, failure)` -- is this intentional? SEV-SNP extracts specific verified claims. Should TPM also extract specific fields (e.g. nonce, hash_alg, PCR values, etc) rather than echoing the input?

keylime/tee/snp.py
```diff
@@ -107,3 +112,46 @@ def nonce_pubkey_freshness_verify(
...
+    host_data = list(report[0xC0:0xE0])
+
+    platform_info = int.from_bytes(report[0x40:0x41], "little")
+    smt_enabled = bool(platform_info & (1 << 0))
+    tsme_enabled = bool(platform_info & (1 << 1))
+
+    reported_tcb_bootloader = int.from_bytes(report[0x180:0x181], "little")
+    reported_tcb_tee = int.from_bytes(report[0x181:0x182], "little")
+    reported_tcb_snp = int.from_bytes(report[0x186:0x187], "little")
+    reported_tcb_microcode = int.from_bytes(report[0x187:0x188], "little")
```
sergio-correia, line 139: sev_snp_claims() parses fixed offsets up to 0x187 without length check. Consider:
    - Adding constants for readability (these are SNP_REPORT structure offsets)
    - Adding validation, e.g.:
```
    if len(report) < 0x188:
      raise ValueError("SNP report too small")
```

keylime/cloud_verifier_tornado.py
```diff
@@ -1589,61 +1595,99 @@ def _tpm_verify(self, json_body: dict[str, Any]) -> Failure:
...
-            byte = string.encode("ascii")
-            report = base64.b64decode(byte)
+        if "tee-evidence" in data and data["tee-evidence"] != "":
+            tee_evidence = data["tee-evidence"]
         else:
-            failure.add_event("missing_param", {"message": 'missing parameter "attestation_report"'}, False)
-            logger.warning("POST returning 400 response. missing query parameter 'attestation_report'")
-            return failure
+            failure.add_event("missing_param", {"message": 'missing parameter "evidence"'}, False)
+            logger.warning("POST returning 400 response. missing query parameter 'evidence'")
```
sergio-correia: Error message says `evidence` but the parameter is `tee-evidence` - should these match?

keylime/cloud_verifier_tornado.py
tylerfanelli: If they are, it's best served for another PR, as I'm unsure what those exact claims would be in a TPM verification.

keylime/tee/snp.py
tylerfanelli: I eventually want to add a class for SEV-SNP reports instead of just indexing some bytes. For now, a simple length check is sufficient.

Added in latest revision.

keylime/cloud_verifier_tornado.py
tylerfanelli: Yes, thanks. Fixed in latest revision.

keylime/tee/snp.py
```diff
@@ -12,18 +13,30 @@
...
+) -> Tuple[dict[str, Any], Failure]:
     failure = Failure(Component.TEE)
 
+    if len(report) < 0x188:
+        failure.add_event(
+            "invalid_input",
+            {"message": "SEV-SNP attestation report input invalid"},
+            False,
+        )
+        return False
```
sergio-correia: `({}, failure)` instead of `False`

keylime/cloud_verifier_tornado.py
```diff
@@ -1658,19 +1658,25 @@ def post(self) -> None:
...
+
         try:
             if evidence_type == "tpm":
-                attestation_failure = self._tpm_verify(data)
-            elif evidence_type == "snp":
-                attestation_failure = self._sev_snp_verify(data)
+                (claims, attestation_failure) = self._tpm_verify(data)
+                attestation_response["claims"] = claims
+            elif evidence_type == "tee":
+                (claims, attestation_failure) = self._tee_verify(data)
```
sergio-correia, line 1672: Here we set the claims and return it unconditionally before checking if `attestation_failure` exists further on. We should probably return the claims only on successful verification, like we do in `verify_attestation()` (`snp.py`)

keylime/cloud_verifier_tornado.py
sergio-correia: Alright, but what about doing it like in `verify_attestation()` and not return it on failures? Then the full data would only be returned when verification actually proceeds.

keylime/cloud_verifier_tornado.py
```diff
@@ -1762,61 +1768,99 @@ def _tpm_verify(self, json_body: dict[str, Any]) -> Failure:
...
+    def _tee_verify(self, data: dict[str, Any]) -> Tuple[dict[str, Any], Failure]:
+        tee_evidence = None
         nonce = None
-        tee_pubkey_x = None
-        tee_pubkey_y = None
+        x = None
+        y = None
 
+        claims: dict[str, Any] = {}
         failure = Failure(Component.DEFAULT)
```
sergio-correia: Should this be `Component.TEE` instead?

keylime/cloud_verifier_tornado.py
```diff
@@ -1762,61 +1768,99 @@ def _tpm_verify(self, json_body: dict[str, Any]) -> Failure:
...
+
+        return (claims, failure)
+
+    def _sev_snp_verify(
+        self, data: dict[str, Any], nonce: bytes, x_b64: bytes, y_b64: bytes
+    ) -> Tuple[dict[str, Any], Failure]:
+        report = None
+
+        claims: dict[str, Any] = {}
+        failure = Failure(Component.DEFAULT)
```
sergio-correia: Should this be `Component.TEE`?

keylime/cloud_verifier_tornado.py
```diff
@@ -1762,61 +1768,99 @@ def _tpm_verify(self, json_body: dict[str, Any]) -> Failure:
...
+        if "snp-report" in data and data["snp-report"] != "":
+            string = data["snp-report"]
             byte = string.encode("ascii")
-            tee_pubkey_y = web_util.urlsafe_nopad_b64decode(byte)
+            report = base64.b64decode(byte)
         else:
-            failure.add_event("missing_param", {"message": 'missing parameter "tee_pubkey_y_b64"'}, False)
-            logger.warning("POST returning 400 response. missing query parameter 'tee_pubkey_y_b64'")
-            return failure
+            failure.add_event("missing_param", {"message": 'missing parameter "report"'}, False)
```
sergio-correia: Nit: `snp-report`

keylime/cloud_verifier_tornado.py
tylerfanelli: Yes, thanks. Fixed in latest revision.

keylime/tee/snp.py
tylerfanelli: Fixed in latest revision.

keylime/cloud_verifier_tornado.py
tylerfanelli: Thanks. Fixed in latest revision.

keylime/cloud_verifier_tornado.py
tylerfanelli: Fixed in latest revision.

keylime/cloud_verifier_tornado.py
tylerfanelli: I misunderstood what you meant. Yes, we can return the claims only upon successful attestation. See the revision I made to https://github.com/keylime/keylime/pull/1810/commits/f7dfd5630109ff5c0061142e6c213caff481cbd2.

tylerfanelli:
Thanks!

Issue #1820, opened 2025-11-24
Keylime registrar does not reject same UUID from different TPM


4N0body5:
# Environment
 * OS / version: Fedora 44
 * Processor architecture: x86_64
 * TPM Manufacturer: SWTPM
 * Keylime version: 
   * keylime-base-7.13.0-2.fc44.noarch
   * keylime-registrar-7.13.0-2.fc44.noarch
   * keylime-tenant-7.13.0-2.fc44.noarch
   * keylime-tools-7.13.0-2.fc44.noarch
   * keylime-verifier-7.13.0-2.fc44.noarch
   * keylime-agent-rust-0.2.8-8.fc44.x86_64
   * keylime-agent-rust-common-0.2.8-8.fc44.noarch
   * keylime-agent-rust-ima-emulator-0.2.8-8.fc44.x86_64

# Description 
The Keylime registrar allows registration of another agent (different TPM device, different EK certificate) with a duplicate UUID.
This presents a critical security vulnerability that allows an attacker to take over an existing agent's identity by re-registering with the same UUID though a different TPM's EK certificate.

## Expected behavior:
When another agent attempts to register with a UUID that already exists in the registrar, but with a different EK certificate, the registrar should:
- **Reject the registration** with HTTP 403 Forbidden
- **Log a security warning** about the duplicate UUID attempt from different TPM
- **Not update** the existing agent record

## Actual behavior 
Currently the registrar does not validate that the EK certificate on device using duplicate UUID matches the previously registered one. But it allows another TPM device with a different EK but duplicate UUID to proceed with a valid registration, without any error or warnings.

Registration of first agent (/dev/tpm0) with UUID "d432fbb3-d2f1-4a97-9ef7-75bd81c00000":

```
INFO:keylime.config:Reading configuration from ['/etc/keylime/logging.conf']
2025-11-23 22:03:45.683 - keylime.tenant - WARNING - The configuration upgrade templates path /usr/share/keylime/templates does not exist
2025-11-23 22:03:45.683 - keylime.config - INFO - Reading configuration from ['/etc/keylime/tenant.conf']
2025-11-23 22:03:45.684 - keylime.tenant - INFO - Setting up client TLS...
2025-11-23 22:03:45.684 - keylime.tenant - INFO - Using default client_cert option for tenant
2025-11-23 22:03:45.684 - keylime.tenant - INFO - Using default client_key option for tenant
[...]

PR #1825, opened 2025-12-01
Fix registrar duplicate UUID vulnerability

sarroutbi:
# PR Title 
Fix registrar duplicate UUID vulnerability

## Type of Change
*(Select all that apply)*
- [ ] CI/CD changes

## Related Issues
**Resolves:** #1820

### Concise Summary
The registrar was allowing re-registration of the same UUID with a different TPM device (different EK/AIK), creating a critical security vulnerability where an attacker could hijack an existing agent's identity by re-registering with the same UUID but their own TPM's credentials.

This commit adds security validation to prevent UUID spoofing attacks

### Technical Details
**Problem:**
The registrar accepts re-registration of existing UUIDs without validating that the TPM identity (EK, AIK, EK certificate) matches the previously registered values. This allows an attacker to:
1. Discover a legitimate agent's UUID
2. Register their own TPM device with the same UUID
3. Successfully impersonate the legitimate agent
4. Bypass all TPM-based attestation security

**Solution Architecture:**
- Added `_check_tpm_identity_immutable()` method in `RegistrarAgent` model that compares TPM identity fields (ek_tpm, ekcert, aik_tpm) between committed database values and new registration attempts
- Modified `agents_controller.create()` to detect TPM identity violations and return HTTP 403 Forbidden instead of allowing the update

*(Check all that apply)*

## Checklist

## Additional Context
N/A

### DIFF
--- a/keylime/web/registrar/agents_controller.py
+++ b/keylime/web/registrar/agents_controller.py
@@ -73,21 +73,15 @@ class AgentsController(Controller):
             agent.update({"agent_id": agent_id, **params})  # type: ignore[union-attr]
 
             # Step 3: Check for TPM identity change security violation
-            if not agent.changes_valid and "agent_id" in agent.errors:  # type: ignore[union-attr]
-                # Check if this is a TPM identity security violation (vs other validation error)
-                agent_id_errors = agent.errors.get("agent_id", [])  # type: ignore[union-attr]
-                is_tpm_identity_violation = any("different TPM identity" in str(err) for err in agent_id_errors)
+            # Use explicit flag instead of fragile string matching for security check
+            if not agent.changes_valid and agent.has_tpm_identity_violation:  # type: ignore[union-attr]
+                # Log the validation errors (includes security warning)
+                self.log_model_errors(agent, logger)
 
-                if is_tpm_identity_violation:
-                    # Log the validation errors (includes security warning)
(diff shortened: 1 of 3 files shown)


keylime/web/registrar/agents_controller.py
```diff
@@ -28,16 +31,97 @@ def show(self, agent_id, **_params):
...
+            agent = RegistrarAgent.get(agent_id) or RegistrarAgent.empty()  # type: ignore[no-untyped-call]
+
+            # Step 2: Update agent with new data and validate (inside lock)
+            agent.update({"agent_id": agent_id, **params})  # type: ignore[union-attr]
+
+            # Step 3: Check for TPM identity change security violation
+            if not agent.changes_valid and "agent_id" in agent.errors:  # type: ignore[union-attr]
+                # Check if this is a TPM identity security violation (vs other validation error)
+                agent_id_errors = agent.errors.get("agent_id", [])  # type: ignore[union-attr]
+                is_tpm_identity_violation = any("different TPM identity" in str(err) for err in agent_id_errors)
```
sergio-correia: This is kinda fragile. If that string gets accidentally changed/updated in the future, the security model would break here silently. Can't we use a different approach, e.g. a custom exception class, or an error code, for instance? If the string match fails here, for instance, a TPM identity violation would incorrectly return 400 instead of 403.

PR #1836, opened 2026-01-09
workflows: Separate upstream test suite from e2e coverage

kkaarreell:
## List of changes
- The workflow that uploads the coverage would mix the coverages from upstream and end-to-end tests, making it not possible to see the coverage of each test suite separately. This separates the reports by disabling the automatic coverage files search.
- Switch code coverage measurement from Fedora 41 to Fedora 43
- Add new tests to e2e test plan

## Type of Change
- [X] CI/CD changes

## Verification Process
verify that codecov coverage measurements differ per coverage file and newly added tests are run.

### DIFF
--- a/packit-ci.fmf
+++ b/packit-ci.fmf
@@ -71,6 +71,7 @@ adjust:
      - /compatibility/basic-attestation-on-localhost-with-allowlist-excludelist
      - /functional/add-agent-with-malformed-ek-cert
      - /functional/agent_UUID_assignment_options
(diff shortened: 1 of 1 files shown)


kkaarreell:
> We need to switch to Fedora 43 in this repo.
> 
> I think we also need to change the scripts/download_packit_coverage.sh to use Fedora 43 instead of Fedora 41

Added. I have also updated CI e2e plan with missing tests.

PR #1838, opened 2026-01-26
Introduce API version 2.5 with version negotiation support

ansasaki:
# feat: Introduce API version 2.5 with version negotiation support

## Type of Change
*(Select all that apply)*
- [x] Documentation update (standalone)
- [ ] CI/CD changes

## Related Issues
N/A

### Concise Summary
Add API v2.5 with explicit algorithm formats and version negotiation

### Technical Details

This PR introduces API version 2.5 with two main improvements:

**1. Explicit TPM Algorithm Formats**
- The `/v2.5/quotes/integrity` endpoint now uses explicit bit-length formats for TPM algorithms:
  - `rsa2048` instead of `rsa`
  - `ecc256` instead of `ecc`
- Server-side normalization ensures backward compatibility with older agents

**2. API Version Negotiation**
- Enhanced `GET /version` endpoint with `supported_versions` field (array of all supported versions)
- Retained `supported_version` field for backward compatibility
- Tenant and verifier now negotiate to use the highest mutually supported API version
- Added `negotiate_version()` function in `keylime/api_version.py`
- Prevents compatibility issues when newer agents communicate with older tenants/verifiers

**3. Enhanced /verify/evidence Endpoint (v2.5) documentation**
- Changed `valid` response field from integer (1/0) to boolean (true/false)
- Added `claims` field to response containing verified claims
- Added TEE (Trusted Execution Environment) verification support

**Motivation:**
- Previously, the tenant/verifier blindly used whatever version the agent reported
- This caused issues when a newer agent reported a version unsupported by older verifiers
- The negotiation mechanism ensures compatibility across mixed-version deployments

**Files Modified:**
- `keylime/api_version.py`: Added version negotiation function, bumped to v2.5
- `keylime/cloud_verifier_tornado.py`: Implemented negotiation in `update_agent_api_version()`
- `keylime/tenant.py`: Implemented negotiation when querying agent version
- `test/test_api_version.py`: Updated test for new version
- `docs/rest_apis/2_5/`: Added complete v2.5 API documentation

*(Check all that apply)*
- [x] Updated markdown docs (file path: `docs/rest_apis/2_5/`)
- [x] Needs user guide updates (`docs/`)
- [x] Needs ReadTheDocs updates

## Verification Process
1. Deploy a verifier/tenant with this change
2. Test with agent supporting v2.5 (new format with `supported_versions` array)
3. Verify negotiation selects highest compatible version
4. Test with older agent (only `supported_version` string) for backward compatibility
5. Confirm the selected version is within verifier's supported range

## Checklist

## Additional Context

- This is the first step toward better cross-version compatibility in Keylime deployments
- The `supported_versions` field requires corresponding agent changes to fully utilize negotiation
- Backward compatibility is maintained: old agents with only `supported_version` still work

### DIFF
--- a/test/test_api_version.py
+++ b/test/test_api_version.py
@@ -56,6 +56,91 @@ class APIVersion_Test(unittest.TestCase):
             with self.subTest(description):
                 self.assertEqual(api_version.is_supported_version(version), supported, description)
 
+    def test_negotiate_version_with_string(self):
+        """Test negotiate_version with a single version string."""
+        # Single version that is supported
+        result = api_version.negotiate_version("2.0")
+        self.assertEqual(result, "2.0")
+
+        # Single version that is not supported
+        result = api_version.negotiate_version("99.0")
+        self.assertIsNone(result)
+
+    def test_negotiate_version_with_list(self):
+        """Test negotiate_version with a list of versions."""
+        # List with multiple supported versions - should return highest
+        result = api_version.negotiate_version(["1.0", "2.0", "2.1"])
+        self.assertEqual(result, "2.1")
+
+        # List with mixed supported/unsupported - should return highest supported
+        result = api_version.negotiate_version(["1.0", "2.0", "99.0"])
+        self.assertEqual(result, "2.0")
+
+        # List with no supported versions
+        result = api_version.negotiate_version(["98.0", "99.0"])
+        self.assertIsNone(result)
+
+    def test_negotiate_version_returns_highest(self):
+        """Test that negotiate_version returns the highest common version."""
+        # All versions supported by both
+        result = api_version.negotiate_version(["1.0", "2.0", "2.1", "2.2", "2.3", "2.4", "2.5"])
+        self.assertEqual(result, "2.5")
+
+        # Only lower versions in common
+        result = api_version.negotiate_version(["1.0", "2.0"])
+        self.assertEqual(result, "2.0")
+
+    def test_negotiate_version_with_custom_local_versions(self):
+        """Test negotiate_version with custom local_versions."""
+        # Restrict local versions to exclude 3.0 (simulating pull mode)
+        local_versions = ["1.0", "2.0", "2.1", "2.2", "2.3", "2.4", "2.5"]
+        result = api_version.negotiate_version(["2.5", "3.0"], local_versions)
+        self.assertEqual(result, "2.5")
+
+        # Remote only has 3.0, local doesn't support it
+        result = api_version.negotiate_version(["3.0"], local_versions)
+        self.assertIsNone(result)
+
+        # Both support 3.0 when included in local
+        local_versions_with_3 = ["1.0", "2.0", "2.5", "3.0"]
+        result = api_version.negotiate_version(["2.5", "3.0"], local_versions_with_3)
+        self.assertEqual(result, "3.0")
+
+    def test_negotiate_version_raise_on_error(self):
+        """Test negotiate_version with raise_on_error=True."""
+        # No compatible version should raise ValueError
+        with self.assertRaises(ValueError) as context:
+            api_version.negotiate_version(["99.0"], raise_on_error=True)
+        self.assertIn("No compatible API version", str(context.exception))
+
+        # Compatible version should not raise
+        result = api_version.negotiate_version(["2.0"], raise_on_error=True)
+        self.assertEqual(result, "2.0")
+
+    def test_negotiate_version_empty_list(self):
+        """Test negotiate_version with empty list."""
+        result = api_version.negotiate_version([])
+        self.assertIsNone(result)
+
+        # With raise_on_error
+        with self.assertRaises(ValueError):
+            api_version.negotiate_version([], raise_on_error=True)
+
+    def test_negotiate_version_proper_version_comparison(self):
+        """Test that version comparison is numeric, not string-based (2.10 > 2.2)."""
+        # String comparison would incorrectly say "2.2" > "2.10" because "2" > "1"
+        # Proper version comparison should correctly identify 2.10 > 2.2
+        local_versions = ["2.2", "2.10"]
+        remote_versions = ["2.2", "2.10"]
+        result = api_version.negotiate_version(remote_versions, local_versions)
+        self.assertEqual(result, "2.10", "2.10 should be greater than 2.2")
+
+        # Test with more versions to ensure proper ordering
+        local_versions = ["1.0", "2.1", "2.2", "2.9", "2.10", "2.11"]
+        remote_versions = ["2.2", "2.9", "2.10"]
+        result = api_version.negotiate_version(remote_versions, local_versions)
+        self.assertEqual(result, "2.10", "2.10 should be the highest common version")
+
 
 if __name__ == "__main__":
     unittest.main()


packit-ci.fmf
```diff
@@ -3,6 +3,8 @@
 environment:
   RUST_IMA_EMULATOR: 1
   TPM_BINARY_MEASUREMENTS: /var/tmp/binary_bios_measurements
+  RUST_KEYLIME_UPSTREAM_URL: "https://github.com/ansasaki/rust-keylime.git"
+  RUST_KEYLIME_UPSTREAM_BRANCH: "rsa2048"
```
mpeters: Is this just for current testing or will this be permanent?

sergio-correia: For current testing, part of a "DO NOT MERGE" commit.

ansasaki: This is only for the current testing as it depends on changes on that branch (the PRs should be merged together).
Note that this change is in the "DO NOT MERGE" commit that I'll drop before pushing.

EDIT: see https://github.com/keylime/rust-keylime/pull/1184

ansasaki:
@kkaarreell This last version I pushed implements version negotiation for the tenant with all components (verifier, registrar, and agent). I added separate API versions state per component to avoid mix-ups.

keylime/cloud_verifier_tornado.py
```diff
@@ -1895,31 +1917,63 @@ async def update_agent_api_version(
...
+            # Validate negotiated version
+            if not keylime_api_version.validate_version(negotiated):
+                logger.error("Negotiated version %s for agent %s is invalid", negotiated, agent_id)
+                return None
 
-            # Check that the new version is greater than current version
-            if new_version_tuple <= old_version_tuple:
+            # Check that the negotiated version is greater than current version (prevent downgrade)
+            negotiated_tuple = str_to_version(negotiated)
+            assert negotiated_tuple, f"Agent {agent_id} version {negotiated} is invalid"
```
sergio-correia: Let me take the opportunity to question why we use asserts here (and elsewhere in the code). Wouldn't it be better to perform a check and raise an exception instead? I believe asserts can be disabled by giving `python` some parameter (`-O`, if I am not mistaken), which means we lose what they were guarding against?

In this particular case, it would be problematic to lose the validation to prevent downgrade attacks, for instance.

keylime/api_version.py
```diff
@@ -34,6 +34,54 @@ def all_versions() -> List[str]:
     return VERSIONS.copy()
 
 
+def negotiate_version(
```
sergio-correia, line 37: It would be nice to have some unit test for this method -- could be done in a follow-up PR.

keylime/tenant.py
```diff
@@ -180,6 +183,131 @@ def set_full_id_str(self) -> None:
...
+            ) from None
+
+    def negotiate_verifier_version(self) -> None:
+        """Fetch and negotiate API version with the verifier."""
+        server_versions = self._fetch_server_versions(self.verifier_base_url, self.tls_context, "verifier")
+        # API version 3.0 is for push attestation only; exclude it in pull mode
+        if self.push_model:
+            local_versions = None  # Use all versions including 3.0
+        else:
+            local_versions = [v for v in keylime_api_version.all_versions() if keylime_api_version.major(v) < 3]
```
sergio-correia: Perhaps it would make sense to have a helper to get the local versions? We use this same snippet a few lines down the road, in `negotiate_registrar_version()`.

keylime/cloud_verifier_tornado.py
```diff
@@ -668,6 +668,24 @@ def post(self) -> None:
...
+                        if not keylime_api_version.is_supported_version(supported_version):
+                            logger.warning(
+                                "Agent %s requested API version %s which is not supported by verifier. "
+                                "Verifier supports: %s. Will attempt version negotiation on first contact.",
+                                agent_id,
+                                supported_version,
+                                keylime_api_version.all_versions(),
+                            )
+                            # Use verifier's current version as fallback
+                            supported_version = keylime_api_version.current_version()
```
sergio-correia, line 684: Should we fallback to `current_version()` or should we try to  negotiate the highest common version instead?
For instance, if the tenant requested version `"2.5"` but the verifier only supports `["2.0", "2.1", "2.2"]`, it would be better to negotiate the highest common version (`"2.2"`), no?

keylime/cloud_verifier_tornado.py
ansasaki: I'll replace these asserts with proper checks, thanks for the suggestion.

keylime/api_version.py
ansasaki: I'll try to add a test for this

keylime/tenant.py
ansasaki: Yes, I agree, I'll replace this to avoid repetition.

keylime/cloud_verifier_tornado.py
ansasaki: So, this is a bit convoluted, but I'll try to explain:

1. Tenant negotiates API version with the agent, and sets the `supported_version`. If the negotiation could not find a common version it aborts the operation. It will set the `supported_version` with the current tenant version only if the agent does not support the `/version` endpoint, which is very old behavior.
2. The verifier checks if it supports the `supported_version`  set by the tenant (it can be a version common to the tenant and agent, but not with the verifier). If not, it will set supported_version = current version and try to negotiate the version with the agent on first contact. This can only happen if the tenant version > latest verifier supported version
4.  Notice that the fallback you highlighted above is for the case where the tenant did not set the `supported_version` at all, which is a very old behavior for the tenant. In any case, the `supported_version` set in the fallback is kind of a placeholder as the verifier will perform version negotiation with the agent on first contact (if the agent does not support it) 
5. On first contact with the agent, the verifier will try the `supported_version` first. If it works, then fine, otherwise it will negotiate the highest common version. The agent would reject the version only if the verifier version is higher than its own version or if it was restricted to a subset of versions that does not include the current verifier version.

PS.: You made me realize that there is an issue with the downgrade detection. It does not differentiate the first contact (where the verifier may be using the fallback version set by the tenant) from the case that there is an ongoing attestation and the agent version was modified, which can lead to false positives:

1. Tenant can't negotiate with agent (old behavior), uses fallback supported_version = 2.5                                                                                                  
2. Verifier stores supported_version = 2.5                                                                                                                                                  
3. On first contact, agent rejects 2.5 (only supports up to 2.3)                                                                                                                            
4. Verifier calls negotiate_agent_version()                                                                                                                                                 
5. Agent reports supported_versions = [1.0, 2.0, 2.1, 2.2, 2.3]                                                                                                                             
6. Verifier negotiates 2.3                                                                                                                                                                  
7. Downgrade check: 2.3 <= 2.5 → TRUE → Returns None, which is a false positive

When is the fallback actually used?

Looking at the tenant code, the fallback only happens when:                                                                                                                                 
1. The agent doesn't support the /version endpoint (very old agent), OR                                                                                                                     
2. The tenant couldn't reach the agent at all (push model)

Normally, the tenant successfully negotiates, so the supported_version passed to the verifier is accurate. The verifier's first request succeeds, and update_agent_api_version() is never called.

The problematic scenario requires:                                                                                                                                                          
- Old agent without /version support, OR tenant couldn't negotiate                                                                                                                          
- Tenant falls back to its current version (2.5)                                                                                                                                            
- Agent actually only supports up to 2.3                                                                                                                                                    
- Verifier rejects the "downgrade" from 2.5 → 2.3

In any case, I'll add an extra check: do not check for the downgrade before the first attestation (`attestation_count == 0`). Let's see if it won't break the existing API bump test case.

keylime/cloud_verifier_tornado.py
ansasaki: Done

keylime/api_version.py
ansasaki: Done

keylime/tenant.py
ansasaki: Done, moved to the `_negotiate_server_version()`, which was the only place this was needed

keylime/cloud_verifier_tornado.py
sergio-correia: Thanks for the explanation, now things are clearer to me. With the updated changes of the PR, it looks good to me, thanks. I will approve it.

ansasaki: Implemented, and it did not break tests :tada:

Issue #1843, opened 2026-02-03
One shot attestation feature isn't functional


Koncpa:
# Environment
 * OS / version: Fedora-Rawhide
 * Processor architecture: x86_64
 * TPM Manufacturer: swTPM
 * Keylime version: Builded keylime directly from upstream 

# Description 
Keylime one shot attestation doesn't work properly and attestation will fail with HTTP Response Code: 404. There suspicion is that some part's of upstream work isn't finished for one shot attestation feature.

## Expected behavior vs. actual behavior 
One shot attestation will be succesfully completed vs.
One shot attesation failed with HTTP error code.

## Steps to reproduce problem
1. obtain fedora machine and install tmt
2. `git clone --branch pk_one_shot_att https://github.com/RedHat-SP-Security/keylime-tests.git` 
3. Inside cloned dir run `tmt --context distro=Fedora-Rawhide run -vvv prepare provision -h connect --guest IP_MACHINE -u root discover plan -n upstream-fedora execute --how tmt --interactive login finish -vvv`

# Relevant logs
[Test log](https://artifacts.dev.testing-farm.io/06a03864-a040-4193-9d1c-508967031d68/work-upstream-basic-attestation-without-revocationx4u2u5_r/plans/upstream-basic-attestation-without-revocation/execute/data/guest/default-0/functional/one-shot-attestation-6/output.txt )

Attach any relevant log files that can help to debug your issue. 

`:: [ 12:45:12 ] :: [  BEGIN   ] :: Run one-shot attestation without policy :: actually running '/keylime-tests/one-shot-attestation-SInBt/keylime_oneshot_attestation             --tpm-pw keylime             --verifier-host 127.0.0.1             --verifier-port 8881             --verifier-cert /var/lib/keylime/cv_ca/server-cert.crt             --verifier-cacert /var/lib/keylime/cv_ca/cacert.crt             --verifier-key /var/lib/keylime/cv_ca/server-private.pem'
INFO:keylime.config:Reading configuration from ['/etc/keylime/logging.conf']
2026-02-02 12:45:12.502 - keylime_oneshot_attestation - INFO - Executing command "tpm2_startup --version"
2026-02-02 12:45:12.511 - keylime_oneshot_attestation - INFO - TPM2-TOOLS Version: 5.7
2026-02-02 12:45:12.511 - keylime_oneshot_attestation - INFO - Executing command "tpm2_startup -c"
[...]

PR #1844, opened 2026-02-03
fix: Update oneshot attestation script for API v2.5 and fix policy handling

ansasaki:
# fix: Update oneshot attestation script for API v2.5 and fix policy handling

## Type of Change
*(Select all that apply)*
- [ ] CI/CD changes

## Related Issues
#1843

### Concise Summary
Fix oneshot attestation API version and policy handling

The oneshot attestation script was using API v2.4 but building the request
body in v2.5 format. Additionally, several issues with empty policy handling
caused JSON parsing errors and incorrect HTTP response codes.

### Technical Details

**Problem:** The `/verify/evidence` endpoint and `keylime_oneshot_attestation`
script had multiple issues preventing successful one-shot attestation.

**Changes to `scripts/keylime_oneshot_attestation`:**
- Updated `API_VERSION` from 2.4 to 2.5 to match the request body format
- Removed unused imports (`secure_mount`, `keylime_logging`)
- Added `sys.exit(1)` after request failure to prevent `AttributeError` when
  accessing `response.status_code` on `None`
- Moved response body logging inside try block to prevent `UnboundLocalError`
  when JSON decoding fails
- Added validation requiring at least one policy (`--runtime-policy`,
  `--tpm-policy`, or `--mb-policy`) before making the request
- Removed unnecessary parameters and usage of verifier private key and certificate

**Changes to `keylime/ima/ima.py`:**
- Fixed `deserialize_runtime_policy()` to return `{}` for `None` or empty
  string input instead of calling `json.loads("")` which raises
  `JSONDecodeError`

**Changes to `keylime/cloud_verifier_tornado.py`:**
- Added policy requirement validation in `_tpm_verify()` - at least one of
  `tpm_policy`, `runtime_policy`, or `mb_policy` must be provided
- Fixed HTTP response codes: input validation errors (`missing_param`,
  `missing_policy`) now return 400 instead of 200

**Changes to `keylime/tpm/tpm_main.py`:**
- Fixed `check_quote()` to handle empty string `tpm_policy` by changing
  `if tpm_policy is None:` to `if not tpm_policy:`
- Fixed `check_pcrs()` to handle empty string `mb_policy` by changing
  `if mb_policy is not None:` to `if mb_policy:`

**Root causes:**
- API version mismatch: script sent v2.5 format to v2.4 endpoint (404)
- Empty string vs None: empty strings passed validation checks but failed
  JSON parsing downstream
- Missing early exit: script continued after request failures
- Incorrect HTTP codes: all responses returned 200, even for input errors

*(Check all that apply)*

## Verification Process
1. Run `keylime_oneshot_attestation` with a valid runtime policy:
   ```bash
   keylime_oneshot_attestation \
       --tpm-pw keylime \
       --verifier-host 127.0.0.1 \
       --verifier-port 8881 \
       --runtime-policy /path/to/policy.json \
       --verifier-cacert /var/lib/keylime/cv_ca/cacert.crt
   ```
2. Verify the request succeeds with HTTP 200 and `valid: true` (or `valid: false`
   with appropriate attestation failures)
3. Run without any policy and verify it fails early with a clear error message
4. Verify input validation errors return HTTP 400

## Checklist

## Additional Context

Files modified:
- `scripts/keylime_oneshot_attestation`
- `keylime/ima/ima.py`
- `keylime/cloud_verifier_tornado.py`
- `keylime/tpm/tpm_main.py`

### DIFF
--- a/scripts/keylime_oneshot_attestation
+++ b/scripts/keylime_oneshot_attestation
@@ -161,29 +161,13 @@ def start_tpm() -> None:
         raise Exception("Error initializing emulated TPM with TPM2_Startup: %s" + str(rc) + ": " + str(errout))
 
 
-def setup_ek_handle(tpm_pw, ek_handle, encrypt_algo):
+def setup_ek_handle(tpm_endorsement_pw, tpm_owner_pw, ek_handle, encrypt_algo):
     global EK_HANDLE, EK
 
-    # do we need to generatea a new EK or use an existing one
+    # do we need to generate a new EK or use an existing one
     if not ek_handle or ek_handle == "generate":
         logger.info("Removing all saved sessions from TPM")
         cmd_return = tpm_cmd(["flushcontext", "-s"], raiseOnError=False)
-        cmd_return = tpm_cmd(["changeauth", "-c", "o", tpm_pw], raiseOnError=False)
-        cmd_return = tpm_cmd(["changeauth", "-c", "e", tpm_pw], raiseOnError=False)
-        rc = cmd_return["code"]
-
-        # if we fail, see if already owned with this pw
-        if rc != EXIT_SUCESS:
-            cmd_return = tpm_cmd(["changeauth", "-c", "o", "-p", tpm_pw, tpm_pw], raiseOnError=False)
-            cmd_return = tpm_cmd(["changeauth", "-c", "e", "-p", tpm_pw, tpm_pw], raiseOnError=False)
-
-            reterr = cmd_return["reterr"]
-            rc = cmd_return["code"]
-            if rc != EXIT_SUCESS:
-                # ut-oh, already owned but not with provided pw!
-                raise Exception("Owner password unknown, TPM reset required. Code %s" + str(rc) + ": " + str(reterr))
-
-        logger.info("TPM Owner password confirmed: %s", tpm_pw)
 
         # create a new EK handle
         with tempfile.NamedTemporaryFile() as tmppath:
@@ -196,9 +180,9 @@ def setup_ek_handle(tpm_pw, ek_handle, encrypt_algo):
                 "-u",
                 tmppath.name,
                 "-w",
-                str(tpm_pw),
+                str(tpm_owner_pw),
                 "-P",
-                str(tpm_pw),
+                str(tpm_endorsement_pw),
             ]
             cmd_return = tpm_cmd(cmd, raiseOnError=False, outputpaths=tmppath.name)
             output = cmd_return["retout"]
@@ -238,7 +222,7 @@ def setup_ek_handle(tpm_pw, ek_handle, encrypt_algo):
                 raise Exception("tpm2_readpublic failed with code " + str(rc) + ": " + str(reterr))
 
 
-def setup_ak_handle(tpm_pw, tpm_encryption_algorithm, tpm_signing_algorithm, tpm_hash_algorithm):
+def setup_ak_handle(tpm_endorsement_pw, tpm_encryption_algorithm, tpm_signing_algorithm, tpm_hash_algorithm):
     global AK_PW, AK, AK_CONTEXT_PATH
     logger.info("Creating a new AK identity")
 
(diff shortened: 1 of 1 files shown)


scripts/keylime_oneshot_attestation
```diff
@@ -541,11 +546,10 @@ def main():
 
     try:
         response_body = response.json()
+        logger.info("Verifier Response Body: %s", response_body)
     except Exception as e:
         logger.error("Failed to decode JSON payload from verifier: %s", e)
```
sarroutbi, line 527: Should not a sys.exit(1) be included here (in case of JSON decoding error) similar to the one before, in line 537?

keylime/cloud_verifier_tornado.py
```diff
@@ -1772,6 +1780,16 @@ def _tpm_verify(self, data: dict[str, Any]) -> Tuple[dict[str, Any], Failure]:
         if "mb_policy" in data and data["mb_policy"] != "":
             mb_policy = data["mb_policy"]
 
+        # At least one policy must be provided for TPM verification to be meaningful
+        if not tpm_policy and runtime_policy is None and not mb_policy:
```
sergio-correia: Looking the validation here, I had to go back to where these are initialized, some ~50 lines above, to find something like this:

```
        tpm_policy = ""
        runtime_policy = None
        mb_policy = ""
```

I wonder if we should be more consistent with these policies, and e.g. use `""` for all or `None` for all? This will also make the validation simpler here, I expect.

scripts/keylime_oneshot_attestation
sergio-correia, line 520: Should we exit here, or is it expected to continue despite the error?

keylime/cloud_verifier_tornado.py
```diff
@@ -1696,19 +1696,27 @@ def post(self) -> None:
...
+                is_input_error = False
                 for event in attestation_failure.events:
                     failures.append(
                         {
                             "type": event.event_id,
                             "context": json.loads(event.context),
                         }
                     )
+                    # Check if this is an input validation error
+                    if event.event_id in ("missing_param", "missing_policy"):
```
sergio-correia: This looks fragile to me. Could we improve it perhaps by using constants or something along the same lines?

scripts/keylime_oneshot_attestation
ansasaki: Thanks, I will add the early exit

keylime/cloud_verifier_tornado.py
ansasaki: I'll make the initialization uniform, and use the python falsy check to handle both `None` and `""` cases.

scripts/keylime_oneshot_attestation
ansasaki: Thanks, I will add the early exit

keylime/cloud_verifier_tornado.py
ansasaki: This one is tricky... The `event_id` is basically created ad-hoc throughout the code and is actually a composition of various parts. The full event_id is actually `{component}.{subcomponent}.{event_id}`. The `"missing_param"` is the most consistently used `event_id` (12 uses), while the great majority is used only once.

So, the check is actually broken as it would compare the full `event_id` against the strings `"missing_param"` or `"missing_policy"`. A simple way to make this check work for now would be to check if the `event_id` ends with those strings. I agree this is fragile, but I'll leave like this for now.

We need to evaluate in future what would be the best way to improve this `event_id` definition. Maybe simply creating the constants per component would help already.

sergio-correia: Right, your `endswith()` approach should be okay for now, we can try to improve this in the future.

ansasaki:
@mpeters I just made changes to make the script to not take the TPM ownership. This could lock other applications from accessing the TPM (including keylime itself). I replaced the `tpm_pw` with `tpm_endorsement_pw` and `tpm_owner_pw` to allow the user to set the passwords separately, and also made these optional.

PR #1847, opened 2026-02-06
tenant: Exclude API version 3.0 from tenant version negotiation

ansasaki:
# tenant: Exclude API version 3.0 from tenant version negotiation

## Type of Change
*(Select all that apply)*
- [ ] CI/CD changes

## Related Issues
N/A

### Concise Summary
Exclude API v3.0 from tenant version negotiation

The tenant does not implement API version 3.0. However, the version
negotiation logic could negotiate v3.0 with the verifier, registrar,
or agent, resulting in the tenant attempting to use an API it does not
support.

### Technical Details

The tenant negotiates API versions with three components: the verifier,
the registrar, and the agent. In all three cases, the tenant must
exclude API version 3.0 from the set of locally supported versions
since it does not implement the v3 API.

Two code paths were affected:

- `_negotiate_server_version()`: Used for both verifier and registrar
  negotiation. Previously, it only excluded v3.0 in pull mode
  (`self.push_model == False`), but allowed it in push mode by passing
  `local_versions = None` (which defaults to all versions including
  3.0). Fixed to always filter out v3.0+ regardless of mode.

- Agent version negotiation in `init_add()`: Called
  `negotiate_version(agent_versions)` without passing a local version
  filter, so `all_versions()` (including 3.0) was used as the default.
  Fixed to explicitly build a filtered version list excluding v3.0+
  and pass it to the negotiation function. Also updated the related
  error and log messages to report the actual filtered version list
  instead of the unfiltered `all_versions()`.

*(Check all that apply)*

## Verification Process
1. Start a keylime deployment with both verifier and registrar
   advertising API version 3.0 in their supported versions list.
2. Run the tenant in both push mode (`--push-model`) and pull mode
   (default) to add an agent.
3. Verify the tenant log messages show a negotiated version below 3.0
   (e.g., 2.5) for all three components (verifier, registrar, agent).

## Checklist
- [ ] Commit messages follow [Chris Beams' How to Write a Git Commit Message article](https://chris.beams.io/posts/git-commit/)

## Additional Context
The tenant is a management tool that uses the v2.x REST API to interact
with the verifier, registrar, and agent. API version 3.0 introduces
different semantics and endpoints that the tenant does not support. Once
the tenant is updated to support v3.0, the version filter can be
adjusted accordingly.

### DIFF
--- a/keylime/tenant.py
+++ b/keylime/tenant.py
@@ -224,8 +224,8 @@ class Tenant:
     def _negotiate_server_version(self, server_versions: Optional[List[str]], server_name: str) -> str:
         """Negotiate API version with a server.
 
-        In pull mode, API version 3.0+ is excluded since it is for push attestation only.
-        In push mode, all versions including 3.0 are available.
+        The tenant always excludes API version 3.0+ from negotiation since it does not
+        implement the v3 API.
 
         Args:
             server_versions: List of versions supported by the server, or None
@@ -237,11 +237,8 @@ class Tenant:
         Raises:
             UserError: If no compatible version is found
         """
-        # API version 3.0 is for push attestation only; exclude it in pull mode
-        if self.push_model:
-            local_versions = None  # Use all versions including 3.0
-        else:
-            local_versions = [v for v in keylime_api_version.all_versions() if keylime_api_version.major(v) < 3]
+        # The tenant does not implement the v3 API; exclude version 3.0+
+        local_versions = [v for v in keylime_api_version.all_versions() if keylime_api_version.major(v) < 3]
 
         if server_versions is None:
             # Fall back to current version when we can't reach the server
@@ -426,7 +423,11 @@ class Tenant:
 
                         if agent_versions:
                             # Negotiate compatible version
-                            negotiated = keylime_api_version.negotiate_version(agent_versions)
+                            # The tenant does not implement the v3 API; exclude version 3.0+
+                            tenant_versions = [
+                                v for v in keylime_api_version.all_versions() if keylime_api_version.major(v) < 3
+                            ]
+                            negotiated = keylime_api_version.negotiate_version(agent_versions, tenant_versions)
 
                             if negotiated is None:
                                 # No compatible version found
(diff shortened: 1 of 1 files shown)


sergio-correia (approved):
Nice catch, thanks

ansasaki:
I believe the failing tests are due to not enabling PoP authentication on push-attestation agent (`enable_authentication = true`). We should either add that to the tests, or merge https://github.com/keylime/rust-keylime/pull/1185 to make it required.

kkaarreell:
I am prepared tests updated in https://github.com/RedHat-SP-Security/keylime-tests/pull/1006 .

keylime/models/verifier/auth_session.py
```diff
@@ -522,7 +522,9 @@ def delete_active_session_for_agent(cls, agent_id: str) -> None:
         with db_manager.session_context() as session:
             agent_id_col = cls.db_table.columns["agent_id"]
             active_col = cls.db_table.columns["active"]
-            delete_stmt = cls.db_table.delete().where((agent_id_col == agent_id) & (active_col == True))  # noqa: E712  # pylint: disable=singleton-comparison
+            delete_stmt = cls.db_table.delete().where(
+                (agent_id_col == agent_id) & (active_col == True)
```
sarroutbi: Did you try changing:
```active_col == True```
to:
```active_col.is_(True)```
It should allow disabling pylint exceptions

File: keylime/tenant.py [hunk 2]
```diff
@@ -237,11 +237,8 @@ def _negotiate_server_version(self, server_versions: Optional[List[str]], server
         Raises:
             UserError: If no compatible version is found
         """
-        # API version 3.0 is for push attestation only; exclude it in pull mode
-        if self.push_model:
-            local_versions = None  # Use all versions including 3.0
-        else:
-            local_versions = [v for v in keylime_api_version.all_versions() if keylime_api_version.major(v) < 3]
+        # The tenant does not implement the v3 API; exclude version 3.0+
+        local_versions = [v for v in keylime_api_version.all_versions() if keylime_api_version.major(v) < 3]
```
sarroutbi, line 241: 3 is a `magic number`. I prefer defining a constant to identify what it is used for. Something like:
MAX_SUPPORTED_MAJOR_VERSION = 2
```if keylime_api_version.major(v) <= MAX_SUPPORTED_MAJOR_VERSION```
but indeed this is a matter of taste.
Whatever you prefer

keylime/models/verifier/auth_session.py
ansasaki: I'll try

ansasaki: Done! Thank you for the tip!

keylime/tenant.py
ansasaki: Thanks for the suggestion, but in this particular case I'll leave the magic number there. We can re-evaluate later, but either we will implement the v3 support or replace the tenant for the push-attestation use case.

PR #1849, opened 2026-02-09
Fix session token migration failure on MySQL

ansasaki:
# Fix session token migration failure on MySQL

## Type of Change
*(Select all that apply)*
- [ ] CI/CD changes

## Related Issues
N/A

### Concise Summary
Fix session token migration failure on MySQL

The database migration `5a8b2c3d4e6f` (hash session tokens for secure
storage with PBKDF2) fails when the verifier has existing sessions in
the database. On MySQL, the `batch_alter_table` operation generates
`ALTER TABLE sessions CHANGE token token_hash VARCHAR(64) NULL`, which
MySQL rejects because the `token` column is part of the PRIMARY KEY
and primary key columns cannot be made nullable.

### Technical Details

The migration changes the primary key of the `sessions` table from
`token` to a new `session_id` column, and renames `token` to
`token_hash`. The original implementation used Alembic's
`batch_alter_table` in two phases: first to add columns and rename,
then with `recreate="always"` to set constraints. This approach fails
on MySQL because the first phase emits an ALTER TABLE that makes the
primary key column nullable during the rename.

- The root cause is that `batch_alter_table` without
  `recreate="always"` issues individual ALTER statements on MySQL
  (unlike SQLite where it always recreates). The column rename
  produces a CHANGE statement that sets the column to NULL, which
  MySQL rejects for primary key columns (error 1171).

- The fix replaces the batch alter approach with a simple
  `op.drop_table` followed by `op.create_table` for both upgrade and
  downgrade. This is correct because existing sessions are invalidated
  in both directions: plaintext tokens cannot be retroactively hashed
  (upgrade), and hashed tokens cannot be converted back to plaintext
  (downgrade). Agents will need to re-authenticate after migration.

- This approach works consistently across all supported database
  backends (SQLite, MySQL, PostgreSQL) and avoids the complexity of
  managing primary key changes through ALTER operations.

- An alternative considered was deleting all rows and using a single
  `batch_alter_table` with `recreate="always"`, but managing primary
  key changes in batch mode adds unnecessary complexity when the data
  is being discarded anyway.

*(Check all that apply)*

## Verification Process
1. Set up a keylime verifier with a MySQL backend at the migration
   revision prior to this one (`517a2d6b5cd3`), with active sessions
   in the `sessions` table.
2. Run the verifier (which triggers the migration on startup) and
   verify it completes without errors.
3. Confirm the `sessions` table has the new schema: `session_id` as
   primary key, `token_hash` and `token_salt` columns present, old
   `token` column removed, and no leftover rows.
4. Downgrade the migration and verify the table is restored to the
   original schema with `token` as primary key.
5. Repeat steps 1-4 with SQLite and PostgreSQL backends.

## Checklist
- [ ] Commit messages follow [Chris Beams' How to Write a Git Commit Message article](https://chris.beams.io/posts/git-commit/)

## Additional Context
The migration script comment and docstring have been updated to
document that all sessions are deleted during both upgrade and
downgrade. The unused `import uuid` was also removed since the new
approach no longer generates UUIDs for existing rows.


keylime/migrations/versions/5a8b2c3d4e6f_hash_session_tokens.py
```diff
@@ -49,58 +48,51 @@ def downgrade_registrar():
...
-    # These sessions will be invalidated
-    connection = op.get_bind()
-    sessions_table = sa.table(
+    # Drop the sessions table entirely and recreate with the new schema.
+    # Existing sessions are invalidated by this migration since plaintext
+    # tokens cannot be retroactively hashed with PBKDF2.
+    # Dropping and recreating avoids issues with ALTER PRIMARY KEY which
+    # is not supported consistently across database backends (e.g. MySQL
+    # rejects making a PRIMARY KEY column nullable during column rename).
+    op.drop_table("sessions")
```
sarroutbi: The current migration strategy uses a "Drop and Recreate" approach. While effective for resolving schema conflicts across different database backends, it introduces a risk on databases that do not support Transactional DDL (such as MySQL, MariaDB, and SQLite). If the migration is interrupted after the `drop_tabl`e command but before the `create_table` command, the database is left in an inconsistent state. A subsequent attempt to run the migration would fail because the sessions table no longer exists to be dropped.

Action: Incorporate SQLAlchemy’s inspector to check for the existence of the sessions table before attempting to drop it. This makes the migration idempotent, allowing it to be safely re-run in the event of a partial failure or crash during the upgrade or downgrade process:
```
# Check for table existence to ensure idempotency
    connection = op.get_bind()
    inspector = sa.inspect(connection)
    if "sessions" in inspector.get_table_names():
        op.drop_table("sessions")
```

keylime/migrations/versions/5a8b2c3d4e6f_hash_session_tokens.py
```diff
@@ -49,58 +48,51 @@ def downgrade_registrar():
...
-        # Rename token_hash back to token
-        batch_op.alter_column(
-            "token_hash",
-            new_column_name="token",
-            existing_type=sa.String(64),
-        )
+    # Drop the sessions table entirely and recreate with the original schema.
+    # Sessions are invalidated by the downgrade since hashed tokens cannot be
+    # converted back to plaintext.
+    op.drop_table("sessions")
```
sarroutbi: Please, include table existence check similar to the one in `downgrade_cloud_verifier`

ansasaki: I'll add the check

keylime/migrations/versions/5a8b2c3d4e6f_hash_session_tokens.py
ansasaki: Done

PR #1850, opened 2026-02-09
Add sane default for verifier session_lifetime configuration option

ansasaki:
# Add sane default for verifier session_lifetime configuration option

## Type of Change
*(Select all that apply)*
- [ ] CI/CD changes

## Related Issues
N/A

### Concise Summary
Fix session_lifetime fallback to prevent token expiry

When the `session_lifetime` option is missing from the verifier
configuration, `config.getint` returns its default fallback of `-1`.
This causes authentication tokens to be created with an expiration
timestamp one second in the past, making them expire immediately.

### Technical Details

- `config.getint("verifier", "session_lifetime")` was called without
  a `fallback` argument in two locations. The function's default
  fallback is `-1`, which produces `timedelta(seconds=-1)`.
- A `DEFAULT_SESSION_LIFETIME = 180` constant was added to
  `keylime/config.py`, consistent with the existing pattern used by
  `DEFAULT_TIMEOUT` and `DEFAULT_MAX_RETRIES`.
- Both call sites now pass `fallback=config.DEFAULT_SESSION_LIFETIME`:
  - `keylime/models/verifier/auth_session.py` (token creation)
  - `keylime/verification/tpm_engine.py` (token renewal on attestation)
- The value of 180 seconds matches the default already defined in the
  configuration template at `templates/2.5/mapping.json`.

*(Check all that apply)*

## Verification Process
1. Deploy a verifier without `session_lifetime` set in the configuration.
2. Register and attest an agent.
3. Verify that the authentication session token remains valid for 180
   seconds rather than expiring immediately.
4. Confirm that explicitly setting `session_lifetime` in the config
   still overrides the default.

## Checklist

## Additional Context
The configuration template already defines 180 as the default value for
`session_lifetime`. This fix only affects deployments where the option
is absent from the active configuration file (e.g., upgraded
environments that have not regenerated their config from the template).

### DIFF
--- a/packit-ci.fmf
+++ b/packit-ci.fmf
@@ -94,14 +94,17 @@ adjust:
      - /functional/keylime_tenant-ima-signature-sanity
      - /functional/measured-boot-swtpm-sanity/.*
      - /functional/service-logfiles-logging
+     - /functional/tenant_api_version_negotiation
      - /functional/tenant-runtime-policy-sanity
      - /functional/tpm-issuer-cert-using-ecc
      - /functional/tpm_policy-sanity-on-localhost
      - /functional/use-multiple-ima-sign-verification-keys
      - /functional/webhook-certificate-on-localhost
      - /functional/agent-registration-with-non-default-tpm-algorithms/.*
+     - /functional/attestation-mode-mismatch
      - /functional/push-attestation-on-localhost
      - /functional/push-authentication-*
+     - /functional/verifier_registrar_unreachable/.*
      - /regression/cve-2023-38200
      - /regression/cve-2023-38201
      - /regression/CVE-2023-3674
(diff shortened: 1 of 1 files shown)


keylime/models/verifier/auth_session.py
```diff
@@ -522,7 +522,9 @@ def delete_active_session_for_agent(cls, agent_id: str) -> None:
         with db_manager.session_context() as session:
             agent_id_col = cls.db_table.columns["agent_id"]
             active_col = cls.db_table.columns["active"]
-            delete_stmt = cls.db_table.delete().where((agent_id_col == agent_id) & (active_col == True))  # noqa: E712  # pylint: disable=singleton-comparison
+            delete_stmt = cls.db_table.delete().where(
+                (agent_id_col == agent_id) & (active_col == True)  # noqa: E712  # pylint: disable=singleton-comparison
```
sarroutbi: Did you try changing:
```active_col == True```
to:
```active_col.is_(True)```
It should allow disabling pylint exceptions

sarroutbi (commented):
Same comment regarding active_col checking. Rest of changes LGTM

Issue #1826, opened 2025-12-05
keylime should generate certificates with correct hostnames on default configuration


kkaarreell:
# Environment
 * OS / version: Fedora 42
 * Processor architecture: x86_64 
 * TPM Manufacturer: swtpm
 * Keylime version: upstream [5ba6450](https://github.com/keylime/keylime/commit/5ba6450c275919bdb821e71fda106b9b972ac9ea)

# Description 

When using the default configuration with all services running on localhost, I have to configure `tls_accept_invalid_hostnames true` on the agent to make things work. This is because the generated scripts do not use correct hostname. Generated certificates have `CN=server` and `CN=client`. I believe that for
certificates generated with defaults settings (`generate` on the verifier/registrar) the certificate should rather use `verifier.ip` there. Filename of the certificate and key files have to be preserved so that services can find them at expected locations.

## Expected behavior vs. actual behavior 

agent cannot connect to the verifier due to incorrect hostname in the certificates

## Steps to reproduce problem
1. attempt to attest local push agent with the default configuration

# Relevant logs

Attach any relevant log files that can help to debug your issue. 

```
Dec 03 08:12:24 local keylime_push_model_agent[41689]:  INFO  keylime_push_model_agent::attestation > --- Phase 1: Sending Capabilities Negotiation ---
Dec 03 08:12:24 local keylime_push_model_agent[41689]:  INFO  keylime_push_model_agent::attestation > Capabilities negotiation URL (POST): https://host:8881/v3.0/agents/d432fbb3-d2f1-4a97-9ef7-75bd81c00001/attestations
Dec 03 08:12:24 local keylime_push_model_agent[41689]:  ERROR keylime_push_model_agent::struct_filler > Failed to create UEFI log handler: Io(Os { code: 2, kind: NotFound, message: "No such file or directory" })
Dec 03 08:12:24 local keylime_push_model_agent[41689]:  WARN  keylime::resilient_client               > Network error (connection failed): error sending request for url (https://host:8881/v3.0/agents/d432fbb3-d2f1-4a97-9ef7-75bd81c00001/attestations)
Dec 03 08:12:24 local keylime_push_model_agent[41689]:  WARN  keylime::error                          > TLS Error: Server certificate hostname doesn't match the connection URL
Dec 03 08:12:24 local keylime_push_model_agent[41689]:  WARN  keylime::error                          > → Set tls_accept_invalid_hostnames = true to bypass hostname verification
[...]

PR #1852, opened 2026-02-11
ca: Add Subject Alternative Names to auto-generated TLS certificates

ansasaki:
# ca: Add Subject Alternative Names to auto-generated TLS certificates

## Type of Change
*(Select all that apply)*
- [ ] CI/CD changes

## Related Issues
#1826 

### Concise Summary
Add SANs (hostname, IPs) to auto-generated certificates

When keylime auto-generates TLS certificates (via `tls_dir = generate`), the server certificates now include Subject Alternative Names (SANs) that reflect the system's actual identity, enabling proper hostname verification by TLS clients.

### Technical Details

Previously, auto-generated server certificates only contained the literal string "server" as the Subject Alternative Name. This caused TLS hostname verification failures when clients connected using the server's actual hostname or IP address, since the certificate did not match the connection endpoint.

This change addresses the problem by automatically populating SANs with:

- **Localhost entries**: `localhost`, `127.0.0.1`, and `::1` (always included for local testing)
- **System hostname and FQDN**: detected via `socket.gethostname()` and `socket.getfqdn()`
- **Configured bind address**: the IP or hostname from the component's `ip` configuration option (wildcard addresses `0.0.0.0` and `::` are excluded)

Additionally, a new configuration option `cert_subject_alternative_names` is introduced for both the verifier and registrar, allowing operators to specify extra DNS names or IP addresses (comma-separated) to include in the certificate. This supports deployment scenarios such as:

- External DNS names (e.g., `verifier.example.com`)
- Load balancer addresses
- Additional network interfaces

**Files changed:**

- `keylime/ca_impl_openssl.py`: Added `get_san_entries()` function to gather SAN entries; extended `mk_signed_cert()` to accept and apply `san_dns` and `san_ips` parameters
- `keylime/ca_util.py`: Updated `cmd_mkcert()` to pass SAN parameters through to the certificate generation layer
- `keylime/web_util.py`: Updated `init_tls_dir()` to read the bind address and the new `cert_subject_alternative_names` config option, compute SANs via `get_san_entries()`, and pass them to `cmd_mkcert()`
- `templates/2.5/verifier.j2` and `templates/2.5/registrar.j2`: Added the `cert_subject_alternative_names` configuration option with documentation
- `templates/2.5/mapping.json`: Added default (empty) value for `cert_subject_alternative_names` in both verifier and registrar sections
- `test/test_ca_impl_openssl.py`: Added unit tests for `get_san_entries()` (hostname/IP/wildcard/IPv6/additional SANs) and for `mk_signed_cert()` with DNS SANs, IP SANs, and the fallback behavior

**Backward compatibility:** This is a non-breaking change. Certificates generated without explicit `cert_subject_alternative_names` configuration will automatically include the system hostname and localhost entries, which is strictly more permissive than the previous behavior of only including "server". The new config option defaults to empty, requiring no action from existing deployments.

*(Check all that apply)*

## Verification Process
1. Deploy keylime with `tls_dir = generate` for the verifier and/or registrar
2. Start the verifier or registrar and inspect the generated `server-cert.crt`:
   ```bash
   openssl x509 -in /var/lib/keylime/cv_ca/server-cert.crt -text -noout | grep -A 20 "Subject Alternative Name"
   ```
3. Verify that the SAN extension includes `localhost`, `127.0.0.1`, `::1`, the system hostname, the FQDN, and the configured bind IP
4. Optionally set `cert_subject_alternative_names = extra.example.com,10.0.0.5` and confirm those entries appear in the regenerated certificate
5. Run unit tests:
   ```bash
   python -m pytest test/test_ca_impl_openssl.py -v
   ```

## Checklist

## Additional Context
This change lays the groundwork for enabling strict TLS hostname verification in keylime components. With proper SANs in place, clients can verify that they are connecting to the intended server, improving the overall security posture of the trust chain.

### DIFF
--- a/keylime/ca_impl_openssl.py
+++ b/keylime/ca_impl_openssl.py
@@ -1,4 +1,6 @@
 import datetime
+import ipaddress
+import socket
 from typing import List, Optional, Tuple
 
 from cryptography import x509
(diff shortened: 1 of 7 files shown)


keylime/ca_impl_openssl.py
```diff
@@ -18,6 +20,81 @@
...
+    # Add FQDN if different from hostname
+    try:
+        fqdn = socket.getfqdn()
+        if fqdn and fqdn != hostname:
+            dns_names.add(fqdn)
+    except OSError:
+        pass
+
+    # Process the bind address
+    if bind_address and bind_address not in ("0.0.0.0", "::", ""):
```
sergio-correia: This won't match e.g. `0:0:0:0:0:0:0:0`. You should probably use `ipaddress.ip_address().is_unspecified` here.

keylime/web_util.py
```diff
@@ -106,7 +107,38 @@ def init_tls_dir(component: str, logger: Optional[Logger] = None) -> str:
...
+                for entry in additional_san.split(","):
+                    entry = entry.strip()
+                    if entry:
+                        # Try to determine if it's an IP address or hostname
+                        try:
+                            import ipaddress
+
+                            ipaddress.ip_address(entry)
+                            additional_ips.append(entry)
+                        except ValueError:
```
sergio-correia: Should we add logging here (debug level would be fine, I guess), to make it clear this was not a valid IP and we are treatring it as DNS name?

keylime/ca_impl_openssl.py
```diff
@@ -18,6 +20,81 @@
...
+            if ip:
+                try:
+                    # Validate the IP address
+                    ipaddress.ip_address(ip)
+                    ip_addresses.add(ip)
+                except ValueError:
+                    # Invalid IP, skip
+                    pass
+
+    return list(dns_names), list(ip_addresses)
```
sergio-correia: Use `sorted()` here to guarantee consistent ordering of these lists.

keylime/ca_impl_openssl.py
ansasaki: OK, I'll change this

keylime/web_util.py
ansasaki: I'll add a DEBUG log entry to make this clear

keylime/ca_impl_openssl.py
ansasaki: I wonder if this is relevant. AFAIK, the certificate is required (by x509 rules) to sort it, so I would expect the underlying lib (openssl) to sort the SAN entries when creating the certiicate . But I'll sort it anyway.

keylime/ca_impl_openssl.py
ansasaki: done

keylime/ca_impl_openssl.py
ansasaki: Done

keylime/web_util.py
ansasaki: Done

PR #1879, opened 2026-04-07
tpm: use only policy's relevant PCRs for event log verification

phaer:
# PR Title 
Use only policy's relevant PCRs for tpm event log verification

## Type of Change
*(Select all that apply)*
- [X] New feature (non-breaking change)
- [ ] CI/CD changes

## Change Description

Use only policy's relevant PCRs for tpm event log verification

Intersect MEASUREDBOOT_PCRS with the active policy's get_relevant_pcrs() set for the PCR replay check. PCRs outside the policy's set may have runtime extensions not in the event log (e.g. PCR 11 from systemd-pcrphase).

This is backwards compatible: policies that return all PCRs (or an empty set) behave as before.

*(Check all that apply)*
- [X] Updated markdown docs (file path: ______)

## Checklist
- [X] Code follows project style guidelines
- [X] Unit/integration tests added/updated
- [X] Documentation updated (per above section)
- [X] Commit messages follow [Chris Beams' How to Write a Git Commit Message article] (https://chris.beams.io/posts/git-commit/)
- [X] All tests pass (local & CI)

## Additional Context

## Summary by CodeRabbit

* **New Features**
  * Measured-boot verification now respects a configured measured-boot policy name and limits PCR replay/validation to only those PCRs the active policy marks relevant; if the active policy is absent or reports no relevant PCRs, all configured measured-boot PCRs are validated. Verification and attestation flows accept an explicit measured-boot policy name to control this.

* **Documentation**
  * User guide clarifies PCR replay validation behavior and the empty-policy edge case.

### DIFF
--- a/keylime/tpm/tpm_main.py
+++ b/keylime/tpm/tpm_main.py
@@ -16,6 +16,7 @@ from keylime.ima import ima
 from keylime.ima.file_signatures import ImaKeyrings
 from keylime.ima.types import RuntimePolicyType
 from keylime.mba import mba
+from keylime.mba.elchecking import policies as mb_policies
 from keylime.tpm import tpm2_objects, tpm_util
 from keylime.tpm.errors import (
     HashAlgorithmMismatch,
@@ -26,6 +27,28 @@ from keylime.tpm.errors import (
 )
 
 logger = keylime_logging.init_logging("tpm")
+
+
+def mb_pcrs_to_check(policy_name: str) -> Set[int]:
+    """Return the set of measured-boot PCRs to replay against the event log.
+
+    Intersects ``config.MEASUREDBOOT_PCRS`` (PCRs Keylime collects) with the
+    set returned by ``policy.get_relevant_pcrs()`` for *policy_name*.  PCRs
+    outside the policy's relevant set are excluded so that runtime extensions
+    not captured in the event log (e.g. PCR 11 from ``systemd-pcrphase``) do
+    not cause spurious verification failures.
+
+    Policies that return an empty set from ``get_relevant_pcrs()`` (such as
+    the built-in ``accept-all`` and ``reject-all``) apply no restriction, so
+    the full ``MEASUREDBOOT_PCRS`` set is returned unchanged.
+    """
+    pcr_set: Set[int] = set(config.MEASUREDBOOT_PCRS)
+    policy = mb_policies.get_policy(policy_name)
+    if policy is not None:
+        relevant = policy.get_relevant_pcrs()
+        if relevant:
+            pcr_set &= relevant
+    return pcr_set
 
 
 class Tpm:
@@ -366,6 +389,7 @@ class Tpm:
         hash_alg: Hash,
         count: int,
         skip_data_pcr_check: bool = False,
+        mb_policy_name: str = "accept-all",
     ) -> Failure:
         failure = Failure(Component.PCR_VALIDATION)
 
@@ -451,8 +475,8 @@ class Tpm:
             except Exception:
                 pass
 
-        if not mb_failure and mb_policy is not None and mba.policy_is_valid(mb_policy) and mb_policy_is_nonempty:
-            for pcr_num in set(config.MEASUREDBOOT_PCRS) & pcr_nums:
+        if not mb_failure and mb_policy is not None and mba.policy_is_valid(mb_policy):
+            for pcr_num in mb_pcrs_to_check(mb_policy_name) & pcr_nums:
                 if not mb_measurement_list:
                     logger.error(
                         "Measured Boot PCR %d in policy, but no measurement list provided by agent %s",
@@ -564,7 +588,7 @@ class Tpm:
                 logger.error("Invalid measured boot policy for agent '%s'", agent_id)
                 failure.add_event("invalid_mb_policy", "Invalid measured boot policy", True)
 
-        if not mb_failure and mba.policy_is_valid(mb_policy):
+        if not mb_failure and mb_policy_is_nonempty and mba.policy_is_valid(mb_policy):
             mb_evaluate = config.get("verifier", "measured_boot_evaluate", fallback="once")
 
             # Value of measured_boot_evaluate can be only 'once' or 'always'
@@ -598,6 +622,7 @@ class Tpm:
         ima_keyrings: Optional[ImaKeyrings] = None,
         mb_measurement_list: Optional[Union[bytes, str]] = None,
         mb_policy: Optional[str] = None,
+        mb_policy_name: str = "accept-all",
         compressed: bool = False,
         count: int = -1,
         skip_pcr_check: bool = False,
@@ -673,6 +698,7 @@ class Tpm:
                 hash_alg,
                 count,
                 skip_data_pcr_check,
+                mb_policy_name=mb_policy_name,
             )
 
         return failure
--- a/keylime/tpm/tpm_main.py
+++ b/keylime/tpm/tpm_main.py
@@ -16,6 +16,7 @@ from keylime.ima import ima
 from keylime.ima.file_signatures import ImaKeyrings
 from keylime.ima.types import RuntimePolicyType
 from keylime.mba import mba
+from keylime.mba.elchecking import policies as mb_policies
 from keylime.tpm import tpm2_objects, tpm_util
 from keylime.tpm.errors import (
     HashAlgorithmMismatch,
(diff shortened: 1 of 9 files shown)


keylime/tpm/tpm_main.py
```diff
@@ -452,7 +475,12 @@ def check_pcrs(
                 pass
 
         if not mb_failure and mb_policy is not None and mba.policy_is_valid(mb_policy) and mb_policy_is_nonempty:
-            for pcr_num in set(config.MEASUREDBOOT_PCRS) & pcr_nums:
+            mb_policy_name = config.get(
+                "verifier",
+                "measured_boot_policy_name",
+                fallback="accept-all",
+            )
```
ansasaki: Please pass the policy name as a parameter instead of reading the `verifier.conf` inside the `tpm/tpm_main.py` file, which is a shared module.

ansasaki (commented):
Thank you for contributing! I left a review comment

keylime/tpm/tpm_main.py
phaer: Thank you for the quick review! I just force-pushed and hope I didn't miss anything :)

phaer: Tried to address the coderabbitai comments as good as I could. But haven't found it super easy to navigate CI checks as at least some (e.g. the mypy lint) seem to be broken prior to my PRs as well?

Happy to chat about this and the general approach. I tried to join the [chatroom](https://cloud-native.slack.com/archives/C01ARE2QUTZ) linked in [README: Contributing: First Timers Support](https://github.com/keylime/keylime/tree/master?tab=readme-ov-file#contributing-first-timers-support), but that seems to require either an `@linuxfoundation.org` adress or an invite? (there's an email address listed in my github profile)

ansasaki: Sorry, we need to update the documentation with the steps to join the CNCF slack instance. Could you please try to register here: https://communityinviter.com/apps/cloud-native/cncf

Then please join the #keylime channel.

About the mypy lint issue, it should be fixed soon: I added a commit in https://github.com/keylime/keylime/pull/1874 to address it.

phaer:
@ansasaki Sorry, I had missed that actual CI tests - not just the previously buggy mypy check - were failing as well here.

CI is green now :)

ansasaki (commented):
EDIT: The review is incorrect, it confuses two different concepts that we informally call "measured boot policy" (and mislead me). The `measured_boot_policy_name` config option sets the MB policy evaluation engine (HOW the policy is evaluated), while what we call "measured boot policy name" should actually be called "measured boot reference state/values name" (WHAT values to check against).

I'm keeping the unmodified AI review below for context and to remember: **always triple check the AI review**.

## Review: `mb_policy_name` should come from the per-agent DB record, not global config

> **Disclaimer**: This review was generated with the assistance of AI (Claude Code). The analysis has been verified against the current codebase, but please double-check the specific line references.

### The issue

All call sites in this PR read `mb_policy_name` from global verifier config:

```python
config.get("verifier", "measured_boot_policy_name", fallback="accept-all")
```

However, keylime already stores the policy name **per-agent** in the database. When an agent is enrolled via the tenant, both `mb_policy` (the refstate content) and `mb_policy_name` (the symbolic name) are sent and stored together in the `VerifierMbpolicy` table (columns `name` + `mb_policy`). Each agent has a foreign key (`mb_policy_id`) linking to its specific policy record.

This means that with the current PR, **all agents share the same PCR filtering** regardless of which policy they were individually enrolled with. For example, if agent A was enrolled with the `"example"` policy (PCRs 0-9, 14) and agent B with a custom policy that also covers PCR 11, both would get the same PCR filtering based on whatever is in `verifier.conf` — which defeats the purpose of per-agent policy assignment.

Additionally, the config key `measured_boot_policy_name` does not exist in the verifier's configuration schema, so `config.get()` will always return the `"accept-all"` fallback unless the user manually adds a non-standard key to their config file.

### Suggested fix

The per-agent policy name is already available at attestation time. The verifier already loads `stored_agent.mb_policy` (the `VerifierMbpolicy` relationship) from the database — it currently only fetches `.mb_policy` (the content column). It just needs to also load `.name`.

**1. In the `load_only` calls** (e.g. around line 2565 on master), add `VerifierMbpolicy.name`:

```python
joinedload(VerfierMain.mb_policy).load_only(VerifierMbpolicy.mb_policy, VerifierMbpolicy.name)
```

**2. Extract the name alongside the content** (around line 2572 on master):

```python
if stored_agent and stored_agent.mb_policy:
    mb_policy_data = stored_agent.mb_policy.mb_policy
    mb_policy_name = stored_agent.mb_policy.name  # already in the DB
else:
    mb_policy_name = "accept-all"
```

**3. Pass `mb_policy_name` through `invoke_get_quote`** and the other call paths instead of reading from `config.get(...)`.

The same approach applies to the `_tpm_verify` path (`VerifyEvidenceHandler`) and the `da/attest.py` path — each should resolve the policy name from the agent's stored record, not from a global config key.

This keeps the feature working correctly for deployments where different agents use different measured boot policies.

keylime/tpm/tpm_main.py
```diff
@@ -452,7 +476,7 @@ def check_pcrs(
                 pass
 
         if not mb_failure and mb_policy is not None and mba.policy_is_valid(mb_policy) and mb_policy_is_nonempty:
-            for pcr_num in set(config.MEASUREDBOOT_PCRS) & pcr_nums:
+            for pcr_num in Tpm.mb_pcrs_to_check(mb_policy_name) & pcr_nums:
```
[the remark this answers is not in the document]
phaer: This looks like a false-positive to me? Just pushed to update documentation, but the behaviour of not replaying PCRs seems correct to me? 

At least removing the guard as suggested would result in a test failure again

phaer:
@ansasaki Thank you for the quick review! 

> This means that with the current PR, all agents share the same PCR filtering regardless of which policy they were individually enrolled with. 

While this is true, in my understanding, it was true before this PR - and the introduction of PCR filtering - as well: [`elchecker.bootlog_evaluate()` did read `measured_boot_policy_name` from global config](https://github.com/keylime/keylime/blob/1e2e5a2450e806efa891d250ac7bedbf26b22c23/keylime/mba/elchecking/elchecker.py#L76) as well?

> Additionally, the config key measured_boot_policy_name does not exist in the verifier's configuration schema, so config.get() will always return the "accept-all" fallback unless the user manually adds a non-standard key to their config file.

I don't think that's accurate. Besides the usage in elchecker.py linked above, [the key also exists in the verifier.j2 config templates](https://github.com/keylime/keylime/blob/1e2e5a2450e806efa891d250ac7bedbf26b22c23/templates/2.0/verifier.j2#L151).

---

So while per-agent policy decisions may be desirable here, it seems to be a bit bigger in scope then this PR so far, would involve elchecker.py as well, and might be a breaking change depending on how exactly we would implement it. 

If wanted, I can do a separate PR for per-agent policies first, before introducing the PCR filtering. WDYT?

ansasaki:
> @ansasaki Thank you for the quick review!
> 
> > This means that with the current PR, all agents share the same PCR filtering regardless of which policy they were individually enrolled with.
> 
> While this is true, in my understanding, it was true before this PR - and the introduction of PCR filtering - as well: [`elchecker.bootlog_evaluate()` did read `measured_boot_policy_name` from global config](https://github.com/keylime/keylime/blob/1e2e5a2450e806efa891d250ac7bedbf26b22c23/keylime/mba/elchecking/elchecker.py#L76) as well?
> 
> > Additionally, the config key measured_boot_policy_name does not exist in the verifier's configuration schema, so config.get() will always return the "accept-all" fallback unless the user manually adds a non-standard key to their config file.
> 
> I don't think that's accurate. Besides the usage in elchecker.py linked above, [the key also exists in the verifier.j2 config templates](https://github.com/keylime/keylime/blob/1e2e5a2450e806efa891d250ac7bedbf26b22c23/templates/2.0/verifier.j2#L151).
> 
> So while per-agent policy decisions may be desirable here, it seems to be a bit bigger in scope then this PR so far, would involve elchecker.py as well, and might be a breaking change depending on how exactly we would implement it.
> 
> If wanted, I can do a separate PR for per-agent policies first, before introducing the PCR filtering. WDYT?

@phaer TL;DR: You are right, I got confused by two different things that we call "measured boot policy", and even more after the AI review result. I don't think the per-agent policy is needed before this PCR filtering, we could add an RFE issue to make it possible to have it (use different evaluation engines per agent).

I'm really sorry for the confusion, the AI review strongly mislead me: I remember that a while ago we introduced what we call "named measured boot policies", which allowed individual MB policies to be set per agent. The fact is that the naming is confusing: what is actually available per agent are the reference states, which means WHAT values to compare against. We informally call them policies. And the "named policies" are actually "named reference values", which we can share between agents (for example, when the fleet of monitored machines are very homogeneous, we can use the same reference values for multiple agents).

The configuration option `measured_boot_policy_name` that you correctly pointed that exists since forever sets HOW the values are evaluated: it is global, and unique for the whole verifier instance. It basically defines the evaluation engine to be used. I got confused because it is referred by the "name" in the configuration, and we had the "named policies" that I mentioned above.

ansasaki (approved):
LGTM!

PR #1885, opened 2026-04-15
Fix timezone handling in Unix timestamp conversion

kkaarreell:
## Type of Change
*(Select all that apply)*
- [ ] CI/CD changes

### Concise Summary
Fix datetime.fromtimestamp() calls to explicitly use UTC timezone instead of relying on system local timezone. Without the tz parameter, fromtimestamp() returns timestamps in the local timezone, causing incorrect datetime values on systems with non-UTC timezone settings.

Changes:
- Add tz=timezone.utc parameter to fromtimestamp() calls in _load_str() and _load_float() methods
- Add test_load_int_has_utc_timezone test that reliably validates UTC timezone handling by checking both tzinfo and time components

### Technical Details
 
  Analysis of the Test Failure

  The test test_load_int_zero (line 170-178 in test_timestamp.py:176) is failing because it expects Unix timestamp 0 to convert to January 1, 1970
   at 00:00:00 UTC (the Unix epoch), but it's getting year 1969 instead of 1970.

  Root Cause

  The issue is in the _load_float() and _load_int() methods in timestamp.py:51-56:

  def _load_float(self, value: float) -> datetime:
      ts = datetime.fromtimestamp(value)
      return self._load_datetime(ts)

  def _load_int(self, value: int) -> datetime:
      return self._load_float(float(value))

  The problem is that datetime.fromtimestamp() without specifying a timezone uses the system's local timezone, not UTC.

  When the test runs:
  1. _load_int(0) calls _load_float(0.0)
  2. _load_float(0.0) calls datetime.fromtimestamp(0.0) - this interprets 0 as seconds since epoch in the local timezone
  3. If the system is in a timezone behind UTC (like EST/UTC-5), fromtimestamp(0) returns December 31, 1969, 19:00:00 (local time)
  4. Then _load_datetime() converts this to UTC, but since it's a naive datetime, it just adds UTC timezone without converting, resulting in 1969

  The Fix

  The fix is to explicitly specify tz=timezone.utc when calling fromtimestamp():

*(Check all that apply)*

## Verification Process
1. Describe test environment
2. Step-by-step validation procedure
3. Expected vs actual results

New test case added.

## Checklist

### DIFF
--- a/keylime/models/base/types/timestamp.py
+++ b/keylime/models/base/types/timestamp.py
@@ -36,7 +36,7 @@ class Timestamp(ModelType):
 
         if not ts:
             try:
-                ts = datetime.fromtimestamp(float(value))
+                ts = datetime.fromtimestamp(float(value), tz=timezone.utc)
             except ValueError:
                 pass
 
@@ -49,7 +49,7 @@ class Timestamp(ModelType):
         return self._load_datetime(ts)
 
     def _load_float(self, value: float) -> datetime:
-        ts = datetime.fromtimestamp(value)
+        ts = datetime.fromtimestamp(value, tz=timezone.utc)
         return self._load_datetime(ts)
 
     def _load_int(self, value: int) -> datetime:
--- a/test/test_timestamp.py
+++ b/test/test_timestamp.py
@@ -177,6 +177,20 @@ class TestTimestampLoadNumeric(unittest.TestCase):
         self.assertEqual(result.month, 1)
         self.assertEqual(result.day, 1)
 
+    def test_load_int_has_utc_timezone(self):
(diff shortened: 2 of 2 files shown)


sarroutbi (changes requested):
Can you please change PR title to:
"Fix timezone handling in Unix timestamp conversion"

ansasaki (approved):
LGTM! Currently, we store the timestamps using the ISO8601 format. When we read it, we try first to parse as ISO, and only if it fails we try to parse using `fromtimestamp()`, so the added changes don't need a DB migration.

sarroutbi (approved):
LGTM

PR #1889, opened 2026-04-19
Fix PUSH mode agent status tracking across restarts and recovery

### LINKED ISSUE #1888: PUSH mode agent timeout not detected after verifier restart (check_push_agent_timeouts is dead code)
# Environment
   * OS / version: Fedora 43 / Linux 6.19.12-200.fc43.x86_64
   * TPM Manufacturer: TPM2_PT_MANUFACTURER (0x53544D20): "STM "
   * Mode: PUSH mode agent
   * Keylime version: HEAD (fc85da2)

  # Description

  PUSH mode agents that stop sending attestations are never marked as timed out after a verifier restart. The agent continues to report `attestation_status: "PASS"` indefinitely, even when the last attestation was received hours or days ago and `maximum_attestation_interval` is only 10 seconds.

  The root cause is twofold:

  1. **`check_push_agent_timeouts()` is dead code.** The function at `keylime/push_agent_monitor.py:197` implements database-based timeout detection for PUSH agents — it queries all PUSH mode agents and sets `accept_attestations = False` for any whose `last_received_quote` exceeds the timeout threshold. It is fully implemented and tested
  (`test/test_push_agent_monitor.py`) but **never called from production code** — not on startup, not periodically. The event-driven timeouts (`schedule_agent_timeout()`) use in-memory Tornado IOLoop handles that are lost on verifier restart.

  2. **`activate_agents()` doesn't skip PUSH mode agents.** On startup (`cloud_verifier_tornado.py:2776`), all agents in `START` state enter the PULL polling loop via `process_agent()` → `invoke_get_quote()`, which tries to contact the agent at `http://{ip}:{port}/...`. For PUSH agents where `ip=None` and `port=None`, this generates repeated failed connection
  attempts to `http://None:None/...`.

  **Expected:** Agent should show `attestation_status: "FAIL"` within `maximum_attestation_interval` (10s) after the last received quote, regardless of whether the verifier was restarted.

  **Actual:** Agent shows `attestation_status: "PASS"` indefinitely (18+ hours after last attestation in the reproduction below). `accept_attestations` remains `True` in the database because in-memory timeout handles were lost on restart and `check_push_agent_timeouts()` is never called.

  ## Steps to reproduce problem

  1. Register a PUSH mode agent (no ip/port) with `attestation_period=2s`
  2. Let it attest successfully (agent builds up `attestation_count`)
  3. Stop the agent so it stops sending quotes
  4. Restart the verifier
  5. Query agent status: `keylime.cmd.tenant -c cvstatus -u <agent_id>`
  6. Observe `attestation_status` is still `"PASS"` despite `last_received_quote` being far older than `maximum_attestation_interval`

  # Relevant logs

  `cvstatus` output 18+ hours after the agent stopped attesting:

  ```json
  {
    "a1b2c3d4-0000-1111-2222-333344445555": {
      "operational_state": "Start",
      "v": null,
      "ip": null,
      "port": null,
      "attestation_count": 343,
      "last_received_quote": 1776467203,
      "last_successful_attestation": 1776467203,
      "attestation_status": "PASS",
      "attestation_period": "2s",
      "maximum_attestation_interval": "10s"
    }
  }
```
  Affected files:
  - keylime/push_agent_monitor.py — check_push_agent_timeouts() (line 197, dead code)
  - keylime/cloud_verifier_tornado.py — activate_agents() (line 2765), server_process() (line 2862)
  - keylime/agent_util.py — is_push_mode_agent() (already exists, needs to be used in activate_agents)

### PR DESCRIPTION
**Pull request by sarroutbi**
# Distinguish TIMEOUT from FAIL for PUSH mode agent attestation status

## Type of Change
- [ ] CI/CD changes

## Related Issues
Resolves: #1888

### Concise Summary
Fix PUSH agent status after verifier restart and recovery

When a PUSH mode agent stops sending attestations and the verifier is restarted, the agent incorrectly shows `attestation_status: "PASS"` instead of reflecting that it has timed out. Additionally, once marked as timed out, the agent cannot recover to `FAIL` or `PASS` status when it resumes sending attestations.

### Technical Details

Three related problems are fixed:

- **Startup timeout detection missing from actual code path.** The verifier entry point (`keylime/cmd/verifier.py`) uses `VerifierServer` from `keylime.web`, which calls `_prepare_agents_on_startup()` in `keylime/web/verifier_server.py`. This method had no logic to detect PUSH agents that timed out while the verifier was down. On restart,
`accept_attestations` remained `True`, so `process_get_status()` reported `"PASS"` even when the agent hadn't attested in hours. The fix checks each PUSH agent's `last_received_quote` against the timeout threshold (`quote_interval * 5`) and sets `accept_attestations = False` if exceeded.

- **No distinction between TIMEOUT and FAIL.** Previously, `process_get_status()` in `cloud_verifier_common.py` returned `"FAIL"` when `accept_attestations` was `False`. This conflated two different conditions: the agent stopped communicating (timeout) vs. the agent is communicating but attestation verification is failing. The fix introduces a `"TIMEOUT"`
status for the former case, while `"FAIL"` is reserved for `consecutive_attestation_failures > 0` with `accept_attestations = True`.

- **PUSH agents stuck in TIMEOUT after resuming attestations.** When a timed-out agent resumed sending attestations, `accept_attestations` was only reset to `True` on *successful* verification (`tpm_engine.py` line 747). If the agent's attestations failed (e.g., IMA mismatch), `accept_attestations` stayed `False` and the status remained `"TIMEOUT"`
indefinitely. The fix resets `accept_attestations = True`, updates `last_received_quote`, and reschedules the timeout in the failure branch of `_process_results()` for PUSH mode agents, since receiving any attestation — even a failing one — proves the agent is communicating.

Equivalent startup logic was also added to `cloud_verifier_tornado.py:main()` for the legacy code path, along with skipping PULL-mode activation for PUSH agents in `activate_agents()`.

*(Check all that apply)*

## Verification Process
1. **Environment:** Fedora 43 with `python3-keylime-7.14.1`, rust-keylime agent in PUSH mode, swtpm.
2. **Startup timeout detection:**
   - Enroll a PUSH mode agent and let it attest successfully (`attestation_status: "PASS"`).
   - Stop the agent and wait longer than the timeout threshold (default: `quote_interval * 5 = 10s`).
   - Restart the verifier.
   - Run `keylime_tenant -c cvstatus -u <agent_id>` and verify `attestation_status: "TIMEOUT"`.
3. **Recovery from timeout:**
   - With the agent in `"TIMEOUT"` state, restart the agent.
   - Verify that `attestation_status` transitions to `"PASS"` (if attestation succeeds) or `"FAIL"` (if verification fails), but no longer stays stuck at `"TIMEOUT"`.
4. **Unit tests:**
   - `pytest test/test_tpm_engine.py test/test_cloud_verifier_common.py test/test_push_agent_monitor.py` — all 49 tests pass.

## Checklist

## Additional Context
- The `"TIMEOUT"` status is a new value for `attestation_status` in the API response. Consumers that only expect `"PASS"`, `"FAIL"`, or `"PENDING"` should be updated to handle `"TIMEOUT"`.

## Summary by CodeRabbit

* **New Features**
  * Detect and apply timeouts for PUSH-mode agents at startup and via periodic monitoring.
  * PUSH-mode agents can re-enable and refresh last-attestation timestamps on certain failures.
  * New configuration option to control automatic recovery of attestation failures.

* **Bug Fixes**
  * Treat missing or non-positive last-attestation timestamps as “no attestation” to avoid spurious timeouts.
  
* **Chores**
  * Updated test metadata source and branch for CI.

### DIFF
--- a/keylime/cloud_verifier_common.py
+++ b/keylime/cloud_verifier_common.py
@@ -279,8 +279,13 @@ def process_get_status(agent: VerfierMain) -> Dict[str, Any]:
         # PUSH mode: determine status based on accept_attestations flag and attestation history
         # Agent must have successfully attested at least once to show PASS
         if hasattr(agent, "accept_attestations") and agent.accept_attestations is False:
-            # Timeout has fired - agent is not accepting attestations
-            attestation_status = "FAIL"
+            # Timeout has fired - check if there were prior attestation failures
+            # to avoid downgrading from FAIL to TIMEOUT
+            consecutive_failures = getattr(agent, "consecutive_attestation_failures", None)
+            if consecutive_failures is not None and consecutive_failures > 0:
+                attestation_status = "FAIL"
+            else:
+                attestation_status = "TIMEOUT"
         elif hasattr(agent, "accept_attestations") and agent.accept_attestations is True:
             # Check if there are recent failures (consecutive_attestation_failures > 0)
             # This prevents showing PASS when recent attestations failed but timeout hasn't fired yet
--- a/keylime/cloud_verifier_tornado.py
+++ b/keylime/cloud_verifier_tornado.py
@@ -6,6 +6,7 @@ import os
 import signal
 import sys
 import threading
+import time
 import traceback
 from concurrent.futures import ThreadPoolExecutor
 from contextlib import contextmanager
@@ -21,6 +22,7 @@ from sqlalchemy.exc import SQLAlchemyError
 from sqlalchemy.orm import Session, joinedload
 from sqlalchemy.orm.exc import NoResultFound  # pyright: ignore
 
+from keylime import agent_util
 from keylime import api_version as keylime_api_version
 from keylime import (
     cloud_verifier_common,
@@ -2825,11 +2827,50 @@ async def process_agent(
         _exit_operation()
 
 
+def check_push_agent_timeout_on_startup(agent: VerfierMain, current_time: int, timeout_seconds: float) -> bool:
+    """Check if a PUSH mode agent timed out during verifier downtime.
+
+    Returns True if the agent was marked as timed out.
+    """
+    if not agent_util.is_push_mode_agent(agent):
+        return False
+
+    logger.debug(
+        "PUSH mode agent %s: last_received_quote=%s, accept_attestations=%s (type=%s)",
+        agent.agent_id,
+        agent.last_received_quote,
+        agent.accept_attestations,
+        type(agent.accept_attestations).__name__,
+    )
+    if (
+        agent.last_received_quote is not None and agent.last_received_quote > 0 and bool(agent.accept_attestations)
+    ):  # pyright: ignore[reportGeneralTypeIssues]
+        last_quote = int(agent.last_received_quote)  # pyright: ignore[reportArgumentType]
+        time_since_last = current_time - last_quote
+        if time_since_last > timeout_seconds:
+            logger.warning(
+                "PUSH mode agent %s timed out during verifier downtime "
+                "(%.1f seconds since last attestation, threshold: %.1f seconds). "
+                "Setting accept_attestations to False.",
+                agent.agent_id,
+                time_since_last,
+                timeout_seconds,
+            )
+            agent.accept_attestations = False  # pyright: ignore[reportAttributeAccessIssue]
+            return True
+    return False
+
+
 async def activate_agents(agents: List[VerfierMain], verifier_ip: str, verifier_port: int) -> None:
     aas = get_AgentAttestStates()
     for agent in agents:
         agent.verifier_ip = verifier_ip  # pyright: ignore
         agent.verifier_port = verifier_port  # pyright: ignore
+
+        if agent_util.is_push_mode_agent(agent):
+            logger.debug("Skipping PULL activation for PUSH mode agent %s", agent.agent_id)
+            continue
+
         agent_run = _from_db_obj(agent)
         if agent_run["mtls_cert"] and agent_run["mtls_cert"] != "disabled":
             agent_run["ssl_context"] = web_util.generate_agent_tls_context(
@@ -2887,10 +2928,16 @@ def main() -> None:
     VerfierMain.metadata.create_all(engine, checkfirst=True)  # pyright: ignore
     with session_context() as session:
         try:
+            quote_interval = config.getfloat("verifier", "quote_interval", fallback=2.0)
+            timeout_seconds = push_agent_monitor.get_maximum_attestation_interval(quote_interval)
+            current_time = int(time.time())
+
             query_all = session.query(VerfierMain).all()
             for row in query_all:
                 if row.operational_state in states.APPROVED_REACTIVATE_STATES:
                     row.operational_state = states.START  # pyright: ignore
+
+                check_push_agent_timeout_on_startup(row, current_time, timeout_seconds)
             # session.commit() is automatically called by context manager
         except SQLAlchemyError as e:
             logger.error("SQLAlchemy Error: %s", e)
@@ -2947,6 +2994,7 @@ def main() -> None:
             # Cancel all pending attestation timeouts (retries, polls)
             cancel_all_pending_events()
             push_agent_monitor.cancel_all_timeouts()
+            push_timeout_periodic.stop()
 
             # Wait for in-flight operations, then close connections and stop
             async def stop() -> None:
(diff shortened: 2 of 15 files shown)

### CODE REVIEW

**Comment by sarroutbi** [comment]
For CI to work, this needs to be merged:
https://github.com/RedHat-SP-Security/keylime-tests/pull/1060

**Comment by kkaarreell** [comment]
>     * The `"TIMEOUT"` status is a new value for `attestation_status` in the API response. Consumers that only expect `"PASS"`, `"FAIL"`, or `"PENDING"` should be updated to handle `"TIMEOUT"`.

Should we bump the API version then?

**Comment by kkaarreell** [comment]
> * The fix resets `accept_attestations = True`, updates `last_received_quote`, and reschedules the timeout in the failure branch of `_process_results()` for PUSH mode agents, since receiving any attestation — even a failing one — proves the agent is communicating.

But if the agent fails attestation we should stop communicating anyway, right? So what's the point of resuming ccept_attestations = True`? And do we even want allow agents to resume from timeout?

**Comment by sarroutbi** [comment]
> > ```
> > * The `"TIMEOUT"` status is a new value for `attestation_status` in the API response. Consumers that only expect `"PASS"`, `"FAIL"`, or `"PENDING"` should be updated to handle `"TIMEOUT"`.
> > ```
> 
> Should we bump the API version then?

I doubt about this ... this is not API version change (API path) but API contents change (backwards compatible). I need to check it (AP for me)

**Comment by sarroutbi** [comment]
> > * The fix resets `accept_attestations = True`, updates `last_received_quote`, and reschedules the timeout in the failure branch of `_process_results()` for PUSH mode agents, since receiving any attestation — even a failing one — proves the agent is communicating.
> 
> But if the agent fails attestation we should stop communicating anyway, right? So what's the point of resuming ccept_attestations = True`? And do we even want allow agents to resume from timeout?

We could bring this to the rest of the community members. IMHO, it should be recoverable... a Timeout could just mean an issue in network or whatever so, when agent gets alive back again (through the Exponential Backoff for example), it should be possible for it to `PASS` attestation again

**Comment by ansasaki** [comment]
> > > ```
> > > * The `"TIMEOUT"` status is a new value for `attestation_status` in the API response. Consumers that only expect `"PASS"`, `"FAIL"`, or `"PENDING"` should be updated to handle `"TIMEOUT"`.
> > > ```
> > 
> > 
> > Should we bump the API version then?
> 
> I doubt about this ... this is not API version change (API path) but API contents change (backwards compatible). I need to check it (AP for me)

Strictly speaking, this is a change in the interface as it is changing the possible values that are returned. It won't break anything internally, but if someone relied on the `attestation_status` for something, they may break or behave unexpectedly.

Backwards compatible changes means that minor version bump is sufficient. If it was a breaking change, then a major version bump is required.

Honestly speaking, we didn't do a great job in the past regarding API/ABI stability, but we are trying to improve in this area. The question is: do we consider the push model API stable? I have my doubts since we haven't finished yet the verifier v3 routes, which means that the API for push model will change even more.

I would say that we could add this change without bumping the API version, but let's make a plan to make the interfaces stable after the next major release (I mean, 8.0.0 when we move the push attestation to be the default).

**Comment by ansasaki** [comment]
> > > * The fix resets `accept_attestations = True`, updates `last_received_quote`, and reschedules the timeout in the failure branch of `_process_results()` for PUSH mode agents, since receiving any attestation — even a failing one — proves the agent is communicating.
> > 
> > 
> > But if the agent fails attestation we should stop communicating anyway, right? So what's the point of resuming ccept_attestations = True`? And do we even want allow agents to resume from timeout?
> 
> We could bring this to the rest of the community members. IMHO, it should be recoverable... a Timeout could just mean an issue in network or whatever so, when agent gets alive back again (through the Exponential Backoff for example), it should be possible for it to `PASS` attestation again

We have to discuss with a wider audience what should be the default behavior. In my opinion:

- TIMEOUT --> PASS: should be allowed by default to cover for unstable connectivity issues
- FAIL --> PASS: should not be allowed by default as it could make some unexpected change less visible

File: keylime/cloud_verifier_common.py [hunk 1]
```diff
@@ -279,8 +279,8 @@ def process_get_status(agent: VerfierMain) -> Dict[str, Any]:
         # PUSH mode: determine status based on accept_attestations flag and attestation history
         # Agent must have successfully attested at least once to show PASS
         if hasattr(agent, "accept_attestations") and agent.accept_attestations is False:
-            # Timeout has fired - agent is not accepting attestations
-            attestation_status = "FAIL"
+            # Timeout has fired - agent stopped sending attestations
+            attestation_status = "TIMEOUT"
```
ansasaki: This can make the state to move from `FAIL` to `TIMEOUT`, which is downgrading the severity: this will make the operator believe there are connectivity issues when actually the attestation is failing.

sarroutbi: You are completely right. I will fix it.

templates/2.6/mapping.json
```diff
@@ -10,7 +10,8 @@
         },
         "registrar": {
             "add": {
-                "max_workers": "16"
+                "max_workers": "16",
+                "auto_recover_attestation_failures": "False"
```
ansasaki: Hi, this should be moved from the registrar to the verifier

keylime/cloud_verifier_tornado.py
```diff
@@ -2825,11 +2827,50 @@ async def process_agent(
...
+    if not agent_util.is_push_mode_agent(agent):
+        return False
+
+    logger.info(
+        "PUSH mode agent %s: last_received_quote=%s, accept_attestations=%s (type=%s)",
+        agent.agent_id,
+        agent.last_received_quote,
+        agent.accept_attestations,
+        type(agent.accept_attestations).__name__,
+    )
```
ansasaki: Is this logging necessary? It will add a INFO log line for every agent on startup, regardless of the state. If you want to keep it, move it to DEBUG level.

For the agents that timed-out, there is a logging below.

ansasaki (changes requested):

ansasaki (commented):
The changes looks good to me. What we need now is some cleanup in the commits (there are many fixup commits, please squash into the proper commits that you want to keep in git history). Don't forget to drop the testing branch redirection commit.

ansasaki (approved):
LGTM!

PR #1902, opened 2026-04-29
verifier: Prevent process_agent() from overwriting TERMINATED state (TOCTOU race)

### LINKED ISSUE #1900: verifier: process_agent() overwrites TERMINATED state due to TOCTOU race
## Bug Description

When a DELETE request marks a pull-mode agent as `TERMINATED` while an attestation cycle is in-flight, `process_agent()` can overwrite the `TERMINATED` state back to an active state (e.g., `GET_QUOTE`). This causes the agent to continue attesting indefinitely, and the tenant's deletion polling times out.

## Root Cause

`process_agent()` reads the agent from the database, checks for `TERMINATED` (early return if found), and then later unconditionally writes the in-memory agent dict — which still has `operational_state = GET_QUOTE` — back to the database. If the DELETE handler executes between the TERMINATED check and the DB write, the write overwrites `TERMINATED` with the active state.

This is a classic TOCTOU (Time-of-Check-Time-of-Use) race condition.

### Timeline

1. `invoke_get_quote()` completes HTTP call, calls `process_agent(agent, GET_QUOTE)`
2. `process_agent()` reads `stored_agent` from DB — sees `GET_QUOTE` (not yet TERMINATED)
3. TERMINATED check passes — state is still active
4. **DELETE handler executes** — sets `operational_state = TERMINATED` in DB, returns 202
5. `process_agent()` reaches the unconditional state-propagation DB update — writes `operational_state = GET_QUOTE`, overwriting `TERMINATED`
6. Schedules next poll via `call_later` — attestation continues

### Affected DB writes in `process_agent()`

1. **State propagation write** (unconditional, runs every cycle) — the main culprit
2. **Failure state write** (FAILED/INVALID_QUOTE path) — could overwrite TERMINATED with a failure state

Additionally, `update_agent_api_version()` writes the full agent dict during version negotiation, with the same race.

## Observed Behavior

In CI (`functional/tenant-runtime-policy-sanity`), the `-c update` command internally does DELETE + POST. The DELETE returns 202, and the tenant polls for deletion:

```
First poll:  agent shows "Terminated" (attestation_count=2)  ← correct
Second poll: agent shows "Get Quote"  (attestation_count=3)  ← state reverted!
...attestation continues, attestation_count keeps increasing...
After 5 retries: tenant times out with error
```

## Expected Behavior

After DELETE returns 202, the in-flight attestation should complete, `process_agent()` should detect `TERMINATED`, delete the agent from DB, and the tenant's next poll should get 404.

## Proposed Fix

Add a SQL `WHERE operational_state != TERMINATED` filter to each of the three DB update queries that write the agent dict. If the update matches 0 rows, the agent was terminated — stop polling by returning early instead of scheduling the next cycle.

> This issue was generated with the assistance of Claude Opus 4.6 (Anthropic).

ansasaki:
# verifier: Prevent process_agent() from overwriting TERMINATED/TENANT_FAILED state (TOCTOU race)

## Type of Change
*(Select all that apply)*
- [ ] CI/CD changes

## Related Issues
Closes keylime/keylime#1900

### Concise Summary
Fix TOCTOU race where `process_agent()` overwrites TERMINATED or TENANT_FAILED state set by a concurrent DELETE or PUT /stop handler, causing the agent to continue attesting indefinitely.

### Technical Details

**Problem:**

When a DELETE request arrives during an in-flight attestation cycle, the
DELETE handler correctly sets `operational_state = TERMINATED` in the
database and returns 202. However, the in-flight `process_agent()`
coroutine can overwrite TERMINATED back to an active state (e.g.
`GET_QUOTE`) because the DB writes in `process_agent()` are
unconditional.

The same race applies to `TENANT_FAILED`: if `PUT /stop` sets
`TENANT_FAILED` between the initial read and the DB write,
`process_agent()` overwrites it back to an active state, and the
stop is permanently lost.

The race window:

1. `process_agent()` reads `stored_agent` from DB — sees `GET_QUOTE`
2. TERMINATED/TENANT_FAILED check passes (state is still active)
3. DELETE/PUT handler executes — sets terminal state in DB, returns
4. `process_agent()` writes the in-memory agent dict (with
   `operational_state = GET_QUOTE`) back to DB — **overwrites
   the terminal state**
5. Schedules next poll — attestation continues indefinitely

This was observed in CI: after DELETE returned 202, the agent briefly
showed "Terminated" but then reverted to "Get Quote" with
`attestation_count` increasing on each poll. The tenant timed out after
5 retries.

**Fix:**

Add `WHERE operational_state NOT IN (TERMINATED, TENANT_FAILED)` filters
to the three DB update queries that write the agent dict back to the
database:

1. The main state-propagation write (runs every attestation cycle)
2. The failure-state write (FAILED/INVALID_QUOTE path)
3. The version-negotiation write in `update_agent_api_version()`

If the update matches zero rows, the agent was terminated or stopped
concurrently. A new helper `_complete_deletion_if_terminated()` re-reads
the agent to determine the appropriate action:

- **TERMINATED**: completes the deletion immediately (no future
  `process_agent()` cycle will be scheduled to do it)
- **TENANT_FAILED**: leaves the agent for the DELETE handler to clean up
- **Already deleted**: logs and returns
- **Unexpected state**: logs a warning and stops polling

**Pyright fix (separate commit):**

Narrowed `agent_data["mtls_cert"]` with `isinstance(mtls_cert, str)`
before passing to `generate_agent_tls_context()`, fixing a
`reportArgumentType` error. The function's internal `enable_agent_mtls`
check already provides the config-level guard.

*(Check all that apply)*

## Verification Process
1. Start verifier with a single pull-mode agent and software TPM
2. Wait for successful attestation cycles (`operational_state` = `Get Quote`)
3. Issue `keylime_tenant -c update` (which internally does DELETE + POST)
4. Verify the DELETE returns 202, agent transitions to Terminated, and
   the in-flight attestation does NOT revert the state back to Get Quote
5. Verify `keylime_tenant -c cvstatus` returns 404 (agent deleted)
6. Verify no `AssertionError` in verifier logs
7. Run unit tests: `python -m pytest test/test_cloud_verifier_tornado.py`

## Checklist

## Additional Context
- This is a follow-up to the earlier race condition fix (preventing
  re-deletion of TERMINATED agents). That fix addressed the DELETE
  handler's behavior; this fix addresses the complementary problem on
  the `process_agent()` side.
- The `WHERE` filter makes the check-and-write atomic at the SQL level,
  eliminating the TOCTOU window entirely.
- The TENANT_FAILED guard addresses the same structural race for
  `PUT /stop` — without it, a concurrent stop request would be
  silently overwritten by the next attestation cycle.

> This PR description was generated with the assistance of Claude Opus 4.6 (Anthropic).

### DIFF
--- a/keylime/cloud_verifier_tornado.py
+++ b/keylime/cloud_verifier_tornado.py
@@ -369,6 +369,7 @@ def _complete_deletion_if_terminated(agent_id: str) -> None:
 
     If the agent still exists and is TERMINATED, complete the deletion
     now — there will be no future process_agent() cycle to do it.
+    If the agent is TENANT_FAILED, leave it for the DELETE handler.
     If the agent is already gone, just log and return.
     """
     try:
@@ -379,6 +380,8 @@ def _complete_deletion_if_terminated(agent_id: str) -> None:
             elif agent.operational_state == states.TERMINATED:  # pyright: ignore
                 logger.info("Agent %s was terminated during attestation, completing deletion", agent_id)
                 verifier_db_delete_agent(session, agent_id)
+            elif agent.operational_state == states.TENANT_FAILED:  # pyright: ignore
+                logger.info("Agent %s tenant quote check failed, stopping poll cycle", agent_id)
             else:
                 logger.warning(
                     "Agent %s update matched 0 rows, but agent exists in state %s. Stopping poll cycle.",
@@ -2220,12 +2223,13 @@ async def update_agent_api_version(
                     session.query(VerfierMain)
                     .filter_by(agent_id=agent_id)
                     .filter(VerfierMain.operational_state != states.TERMINATED)  # pyright: ignore
+                    .filter(VerfierMain.operational_state != states.TENANT_FAILED)  # pyright: ignore
                     .update(agent_db)  # pyright: ignore
                 )
                 # session.commit() is automatically called by context manager
 
             if rows == 0:
-                logger.info("Agent %s was terminated during version negotiation, stopping", agent_id)
+                logger.info("Agent %s was terminated or stopped during version negotiation, stopping", agent_id)
                 _complete_deletion_if_terminated(agent_id)
                 return None
 
@@ -2624,6 +2628,7 @@ async def process_agent(
                             session.query(VerfierMain)
                             .filter_by(agent_id=agent["agent_id"])
                             .filter(VerfierMain.operational_state != states.TERMINATED)  # pyright: ignore
+                            .filter(VerfierMain.operational_state != states.TENANT_FAILED)  # pyright: ignore
                             .update(agent)  # type: ignore[arg-type]
                         )
                         # session.commit() is automatically called by context manager
@@ -2640,15 +2645,17 @@ async def process_agent(
                     del agent_db[key]
 
             # Fourth database operation - update agent state.
-            # The TERMINATED filter prevents a TOCTOU race: if a DELETE
-            # handler set TERMINATED between our initial read and this
-            # write, the update matches zero rows and we stop polling
-            # instead of reverting the agent back to an active state.
+            # The TERMINATED/TENANT_FAILED filters prevent a TOCTOU race:
+            # if a DELETE or PUT /stop handler set one of these states
+            # between our initial read and this write, the update matches
+            # zero rows and we stop polling instead of reverting the agent
+            # back to an active state.
             with session_context() as session:
                 rows = (
                     session.query(VerfierMain)
                     .filter_by(agent_id=agent_db["agent_id"])
                     .filter(VerfierMain.operational_state != states.TERMINATED)  # pyright: ignore
+                    .filter(VerfierMain.operational_state != states.TENANT_FAILED)  # pyright: ignore
                     .update(agent_db)  # pyright: ignore
                 )
                 # session.commit() is automatically called by context manager
--- a/keylime/cloud_verifier_tornado.py
+++ b/keylime/cloud_verifier_tornado.py
@@ -364,6 +364,31 @@ def verifier_db_delete_agent(session: Session, agent_id: str) -> None:
     session.commit()
 
 
+def _complete_deletion_if_terminated(agent_id: str) -> None:
+    """Re-read the agent after a guarded update matched zero rows.
+
+    If the agent still exists and is TERMINATED, complete the deletion
+    now — there will be no future process_agent() cycle to do it.
+    If the agent is already gone, just log and return.
+    """
+    try:
+        with session_context() as session:
+            agent = session.query(VerfierMain).filter_by(agent_id=agent_id).first()
+            if agent is None:
+                logger.info("Agent %s was deleted during attestation, stopping poll cycle", agent_id)
+            elif agent.operational_state == states.TERMINATED:  # pyright: ignore
+                logger.info("Agent %s was terminated during attestation, completing deletion", agent_id)
+                verifier_db_delete_agent(session, agent_id)
+            else:
+                logger.warning(
+                    "Agent %s update matched 0 rows, but agent exists in state %s. Stopping poll cycle.",
+                    agent_id,
+                    agent.operational_state,
+                )
+    except SQLAlchemyError:
+        logger.exception("SQLAlchemy Error completing deletion for agent %s", agent_id)
+
+
 def store_attestation_state(agentAttestState: AgentAttestState) -> None:
     # Only store if IMA log was evaluated
     if agentAttestState.get_ima_pcrs():
@@ -2190,8 +2215,18 @@ async def update_agent_api_version(
                     if key in agent_db:
                         del agent_db[key]
 
-                session.query(VerfierMain).filter_by(agent_id=agent_id).update(agent_db)  # pyright: ignore
+                rows = (
+                    session.query(VerfierMain)
+                    .filter_by(agent_id=agent_id)
+                    .filter(VerfierMain.operational_state != states.TERMINATED)  # pyright: ignore
+                    .update(agent_db)  # pyright: ignore
+                )
                 # session.commit() is automatically called by context manager
+
+            if rows == 0:
+                logger.info("Agent %s was terminated during version negotiation, stopping", agent_id)
+                _complete_deletion_if_terminated(agent_id)
+                return None
 
         else:
             logger.warning("Agent %s did not provide version information", agent_id)
@@ -2584,10 +2619,17 @@ async def process_agent(
                         for key in exclude_db:
                             if key in agent:
                                 del agent[key]
-                        session.query(VerfierMain).filter_by(agent_id=agent["agent_id"]).update(
-                            agent  # type: ignore[arg-type]
+                        rows = (
+                            session.query(VerfierMain)
+                            .filter_by(agent_id=agent["agent_id"])
+                            .filter(VerfierMain.operational_state != states.TERMINATED)  # pyright: ignore
+                            .update(agent)  # type: ignore[arg-type]
                         )
                         # session.commit() is automatically called by context manager
+
+                    if rows == 0:
+                        _complete_deletion_if_terminated(agent["agent_id"])
+                        return
 
         # propagate all state, but remove none DB keys first (using exclude_db)
         try:
@@ -2596,10 +2638,23 @@ async def process_agent(
                 if key in agent_db:
                     del agent_db[key]
 
-            # Fourth database operation - update agent state
+            # Fourth database operation - update agent state.
+            # The TERMINATED filter prevents a TOCTOU race: if a DELETE
+            # handler set TERMINATED between our initial read and this
+            # write, the update matches zero rows and we stop polling
+            # instead of reverting the agent back to an active state.
             with session_context() as session:
-                session.query(VerfierMain).filter_by(agent_id=agent_db["agent_id"]).update(agent_db)  # pyright: ignore
+                rows = (
+                    session.query(VerfierMain)
+                    .filter_by(agent_id=agent_db["agent_id"])
+                    .filter(VerfierMain.operational_state != states.TERMINATED)  # pyright: ignore
+                    .update(agent_db)  # pyright: ignore
+                )
                 # session.commit() is automatically called by context manager
+
+            if rows == 0:
+                _complete_deletion_if_terminated(agent["agent_id"])
+                return
         except SQLAlchemyError as e:
             logger.error("SQLAlchemy Error for agent ID %s: %s", agent["agent_id"], e)
 
--- a/keylime/cloud_verifier_tornado.py
+++ b/keylime/cloud_verifier_tornado.py
@@ -1151,9 +1151,10 @@ class AgentsHandler(BaseHandler):
                         if mode == "pull":
                             # Prepare SSLContext for mTLS connections
                             agent_data["ssl_context"] = None
-                            if agent_data["mtls_cert"] and agent_data["mtls_cert"] != "disabled":
+                            mtls_cert = agent_data["mtls_cert"]
+                            if agent_mtls_cert_enabled and isinstance(mtls_cert, str) and mtls_cert != "disabled":
                                 agent_data["ssl_context"] = web_util.generate_agent_tls_context(
-                                    "verifier", agent_data["mtls_cert"], logger=logger
+                                    "verifier", mtls_cert, logger=logger
                                 )
 
                             if agent_data["ssl_context"] is None:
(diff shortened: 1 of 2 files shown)


keylime/cloud_verifier_tornado.py
```diff
@@ -364,6 +364,25 @@ def verifier_db_delete_agent(session: Session, agent_id: str) -> None:
...
+        with session_context() as session:
+            agent = session.query(VerfierMain).filter_by(agent_id=agent_id).first()
+            if agent is not None and agent.operational_state == states.TERMINATED:  # pyright: ignore
+                logger.info("Agent %s was terminated during attestation, completing deletion", agent_id)
+                verifier_db_delete_agent(session, agent_id)
+            else:
+                logger.info("Agent %s was deleted during attestation, stopping poll cycle", agent_id)
+    except SQLAlchemyError as e:
+        logger.error("SQLAlchemy Error completing deletion for agent %s: %s", agent_id, e)
+
```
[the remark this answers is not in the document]
ansasaki: Switched to `logger.exception` to preserve the traceback. Not re-raising though — both callers are inside `process_agent()`, which handles `SQLAlchemyError` the same way (log and continue) at every other DB operation. Re-raising would just add a second traceback for the same error with no retry mechanism to benefit from it. If the deletion fails transiently, the agent stays in TERMINATED and the next `process_agent()` cycle detects it at the TERMINATED check and retries the deletion.

> This comment was generated with the assistance of Claude Opus 4.6 (Anthropic).

keylime/cloud_verifier_tornado.py
```diff
@@ -1126,9 +1145,10 @@ def post(self) -> None:
                         if mode == "pull":
                             # Prepare SSLContext for mTLS connections
                             agent_data["ssl_context"] = None
-                            if agent_mtls_cert_enabled:
+                            mtls_cert = agent_data["mtls_cert"]
+                            if agent_mtls_cert_enabled and isinstance(mtls_cert, str) and mtls_cert != "disabled":
                                 agent_data["ssl_context"] = web_util.generate_agent_tls_context(
-                                    "verifier", agent_data["mtls_cert"], logger=logger
+                                    "verifier", mtls_cert, logger=logger
```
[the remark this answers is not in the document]
ansasaki: Investigated this. The inconsistency is **cosmetic, not behavioral**: `generate_agent_tls_context()` itself reads the `enable_agent_mtls` config and returns `None` when the flag is `False` ([web_util.py lines 356-358](https://github.com/keylime/keylime/blob/master/keylime/web_util.py#L356-L358)):

```python
agent_mtls_enabled = config.getboolean(component, "enable_agent_mtls")
if not agent_mtls_enabled:
    return None
```

So the other three call sites (PUT reactivate at line 1218, `notify_error` at line 2511, and `activate_agents` at line 2820) are already guarded by the function's internal check — they'll get `None` back when mTLS is disabled, and the callers handle that correctly.

The `agent_mtls_cert_enabled` check in the POST handler (line 929) serves a **different purpose**: it gates the early-rejection logic (lines 932-941) that refuses to register an agent without an mTLS cert when the config requires one. The guard we added on line 1149 is for pyright type-narrowing (`isinstance(mtls_cert, str)`), not for the mTLS config decision.

Making the call-site checks uniform would be a nice cleanup but it's not a correctness issue, so it's out of scope for this PR.

sarroutbi (commented):
@ansasaki: Thanks for this PR. However, there are a few logical gaps, edge cases, and test coverage omissions that might be addressed before merging this if you consider it appropriate:

---

### 1. Misleading Logic Branch in `_complete_deletion_if_terminated`

In **Patch 1**, the helper function contains a logic gap where an unexpected agent state will be incorrectly logged as a deleted agent.

**The Issue:**
```python
if agent is not None and agent.operational_state == states.TERMINATED:  # pyright: ignore
    logger.info("Agent %s was terminated during attestation, completing deletion", agent_id)
    verifier_db_delete_agent(session, agent_id)
else:
    logger.info("Agent %s was deleted during attestation, stopping poll cycle", agent_id)
```
If `update()` returns `0` rows, but the agent *does* still exist in the database and its state is **not** `TERMINATED` (e.g., a race condition reverted the state, or a database driver quirk returned 0 changed rows instead of matched rows), the code falls into the `else` block. It will falsely log that the agent was deleted.

**The Fix:**
Split the logic into three distinct branches to prevent hiding unexpected states:
```python
if agent is None:
    logger.info("Agent %s was deleted during attestation, stopping poll cycle", agent_id)
elif agent.operational_state == states.TERMINATED:  # pyright: ignore
    logger.info("Agent %s was terminated during attestation, completing deletion", agent_id)
    verifier_db_delete_agent(session, agent_id)
else:
    logger.warning("Agent %s update matched 0 rows, but agent exists in state %s. Stopping poll cycle.", agent_id, agent.operational_state)
```

---

### 2. Risk of Zombie Agents in `update_agent_api_version`

In **Patch 1**, when `update_agent_api_version` detects that the agent was terminated, it stops but does not clean up.

**The Issue:**
```python
if rows == 0:
    logger.info("Agent %s was terminated during version negotiation, stopping", agent_id)
    return None
```
Unlike `process_agent`, this block does not call `_complete_deletion_if_terminated(agent_id)`. If the `DELETE` handler relies on the background polling loop to finalize the deletion of a `TERMINATED` agent, returning `None` here without completing the deletion could leave the agent permanently orphaned in the database as a "zombie."

**The Fix:**
Determine if version negotiation is responsible for cleanup. If it is, call the cleanup helper before returning:
```python
if rows == 0:
    logger.info("Agent %s was terminated during version negotiation, stopping", agent_id)
    _complete_deletion_if_terminated(agent_id)
    return None
```

---

### 3. Inconsistent Dictionary Access

There are minor consistency and safety issues regarding dictionary keys in **Patch 1** and **Patch 2**.

**The Issue (Patch 1):**
In the fourth database operation of `process_agent`, you use `agent_db` for the query but `agent` for the helper call.
```python
.filter_by(agent_id=agent_db["agent_id"]) # Uses agent_db
...
_complete_deletion_if_terminated(agent["agent_id"]) # Uses agent
```
While functionally safe because `agent_db = dict(agent)` occurs right above it, it is slightly inconsistent.

**The Issue (Patch 2):**
```python
mtls_cert = agent_data["mtls_cert"]
if agent_mtls_cert_enabled and isinstance(mtls_cert, str) and mtls_cert != "disabled":
```
You are narrowing the type to fix a pyright issue, but you directly access `agent_data["mtls_cert"]`. If the database ever omits this key (e.g., due to a schema migration or null handling), it will throw a `KeyError`. Using `.get()` is safer:
```python
mtls_cert = agent_data.get("mtls_cert")
```

---

### 4. Test Coverage Gaps

**Patch 3** adds excellent unit tests for the new `_complete_deletion_if_terminated` helper, but it misses the integration points.

**The Issue:**
There are no tests verifying that `process_agent` or `update_agent_api_version` actually break their execution flows when the database update returns `0`. 

**The Fix:**
Please add mock tests for `process_agent` where `session.query().filter_by().filter().update()` is mocked to return `0`. Assert that:
* The rest of the `process_agent` logic does not execute.
* The `_complete_deletion_if_terminated` helper is explicitly called.

sarroutbi:
Thanks for addressing suggested changes. LGTM

sarroutbi (approved):
LGTM

ansasaki:
@sarroutbi Thanks for the thorough review!

### Finding 1 — Misleading logic branch in `_complete_deletion_if_terminated`

**Valid.** Good catch — the two-branch `if/else` conflates "agent is gone" with "agent exists in an unexpected state", which would hide real anomalies behind a misleading log message. I've split it into three branches as you suggested: `agent is None` (info), `TERMINATED` (delete), and an `else` with `logger.warning` that includes the actual state. Fixed.

### Finding 2 — Zombie agents in `update_agent_api_version`

**Valid.** While in practice the caller's `process_agent(agent, FAILED)` would eventually hit the Fourth operation guard and clean up, that's a convoluted and fragile path. Calling `_complete_deletion_if_terminated()` directly is simpler and more defensive. Fixed.

### Finding 3 — Inconsistent dictionary access

**Not fixing — by design.**

- `agent` vs `agent_db`: The line `agent_db = dict(agent)` is immediately above, so `agent["agent_id"]` and `agent_db["agent_id"]` are always the same value. Using `agent` for the helper call is intentional — `agent` is the canonical dict used throughout `process_agent`, while `agent_db` is a throwaway copy with exclude_db keys stripped for the SQL UPDATE. No bug here.

- `.get("mtls_cert")` vs `["mtls_cert"]`: The `mtls_cert` key is always present — it's populated by `_from_db_obj()` which explicitly lists it in the fields array (and it's a database column with a default). Silently defaulting to `None` via `.get()` would actually be worse: it would mask a real schema bug that we'd want to fail loudly on. The `isinstance(mtls_cert, str)` guard on the next line already handles the `None` case safely for pyright's benefit.

### Finding 4 — Test coverage gaps (integration tests for `process_agent`)

**Acknowledged but out of scope.**

`process_agent` is a ~300-line async function with four sequential database operations, HTTP calls, async awaits, and multiple conditional branches. Testing the "Fourth operation returns 0 rows" path alone requires mocking: `shutdown.is_shutting_down()`, two separate `session_context()` calls with different query chain behaviors (First read must return an agent, Fourth update must return 0 rows), plus an agent dict with all required keys and an async test harness. That's ~60-80 lines of dense mock wiring for what amounts to verifying `if rows == 0: call helper and return` — logic that is already directly covered by the `_complete_deletion_if_terminated` unit tests, and exercised end-to-end by the CI functional tests (`functional/tenant-runtime-policy-sanity`).

I've added unit tests covering all four branches of `_complete_deletion_if_terminated()` (TERMINATED → delete, None → no-op, unexpected state → warning, SQLAlchemyError → log). That provides meaningful coverage of the new code without the brittleness of mocking deep into `process_agent`'s internals.

---

Findings 1 and 2 are fixed and squashed into the relevant commits. Force-pushed.

> This comment was written with the help of AI.

keylime/cloud_verifier_tornado.py
```diff
@@ -2584,22 +2620,42 @@ async def process_agent(
...
+            # The TERMINATED filter prevents a TOCTOU race: if a DELETE
+            # handler set TERMINATED between our initial read and this
+            # write, the update matches zero rows and we stop polling
+            # instead of reverting the agent back to an active state.
             with session_context() as session:
-                session.query(VerfierMain).filter_by(agent_id=agent_db["agent_id"]).update(agent_db)  # pyright: ignore
+                rows = (
+                    session.query(VerfierMain)
+                    .filter_by(agent_id=agent_db["agent_id"])
+                    .filter(VerfierMain.operational_state != states.TERMINATED)  # pyright: ignore
```
sergio-correia, line 2657: Don't we have a similar race here with `TENANT_FAILED`? If `PUT /stop` sets `TENANT_FAILED` between the initial read and this write, the update overwrites it back to an active state, same as the `TERMINATED` race. Unlike `TERMINATED`, nothing re-introduces `TENANT_FAILED`, so the stop is permanently lost and the agent polls indefinitely. Same applies to lines 2222 and 2626.

ansasaki: Thank you for the review! True, the same TOCTOU issue exists for the `TENANT_FAILED`. I'll add the fixes.

test/test_cloud_verifier_tornado.py
```diff
@@ -143,11 +143,26 @@ def test_noop_when_agent_already_deleted(self, mock_session_ctx, mock_delete):
...
+
     @patch("keylime.cloud_verifier_tornado.logger")
     @patch("keylime.cloud_verifier_tornado.verifier_db_delete_agent")
     @patch("keylime.cloud_verifier_tornado.session_context")
     def test_warns_when_agent_in_unexpected_state(self, mock_session_ctx, mock_delete, mock_logger):
-        """Logs warning and does not delete when agent exists in a non-TERMINATED state."""
+        """Logs warning and does not delete when agent exists in an unexpected state."""
         mock_session = MagicMock()
         mock_agent = MagicMock()
         mock_agent.operational_state = 3  # states.GET_QUOTE
```
sergio-correia: The fix looks good to me, but I think this test can be improved:

The `else` branch (unexpected state) also doesn't call `verifier_db_delete_agent`, so this test passes even if the new `elif` is removed entirely. The agent would fall through to else and the assertion would still hold.

To actually verify the `TENANT_FAILED` branch is taken, maybe we could assert on the logging too?

sergio-correia: By the way, I mis-clicked here and it messed up the selected lines. The comment above was supposed to be on line 159.

ansasaki: I'll add the log assertion to distinguish when the `TENANT_FAILED` branch is taken.

ansasaki: Done!

sergio-correia (approved):
LGTM, thanks for the updates!

PR #1919, opened 2026-06-02
Fix test compatibility across Python and cryptography library versions

sergio-correia:
## Type of Change

  ## Related Issues
  None.

  ## Change Description

  ### Concise Summary
  Fix tests that fail with older asyncio and cryptography versions

  ### Technical Details

  **asyncio.Event() loop binding (`test_shutdown.py`):**
  `asyncio.Event()` can bind to the current event loop at creation time on older Python releases. The shutdown tests were
  creating events in `setUp` bound to one loop, then running them from a different loop created inside individual test methods,
   causing `RuntimeError`. The fix creates a single explicit event loop in `setUp`/`tearDown` and reuses it across all async
  tests in the class.

  **ECC error messages (`test_tpm2_objects.py`):**
  The `cryptography` library changed its error message for invalid EC points when EC support was migrated from Python/CFFI to
  Rust in version 42.0.0. Older versions raise `"Invalid EC key."` while newer versions raise `"Invalid EC key. Point is not on
   the curve specified."` The tests were asserting only the newer message. The fix accepts both.

  ## Verification Process
  1. Run the affected tests: `pytest test/test_shutdown.py test/test_tpm2_objects.py`
  2. All tests should pass on both cryptography <42.0.0 (CFFI backend) and >=42.0.0 (Rust backend)

  ## Checklist
  - [x] Commit messages follow [Chris Beams' How to Write a Git Commit Message
  article](https://chris.beams.io/posts/git-commit/)

  ## Additional Context
  The `test_coordinate_exceeds_field_prime_rejected` test at `test_tpm2_objects.py:398` also references `"not on the curve"`
  but does **not** need updating -- keylime's own coordinate bounds check catches the input before the cryptography library is
  reached, so the library's error message is never hit in that test.

### DIFF
--- a/test/test_shutdown.py
+++ b/test/test_shutdown.py
@@ -13,8 +13,14 @@ class TestShutdownFlag(unittest.TestCase):
     """Test the process-wide shutdown flag."""
 
     def setUp(self) -> None:
+        self.loop = asyncio.new_event_loop()
+        asyncio.set_event_loop(self.loop)
         # Reset the module-level event before each test
         shutdown._shutdown_event = asyncio.Event()
+
+    def tearDown(self) -> None:
+        asyncio.set_event_loop(None)
+        self.loop.close()
 
     def test_initial_state_not_shutting_down(self) -> None:
         self.assertFalse(shutdown.is_shutting_down())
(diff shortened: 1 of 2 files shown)


ansasaki (approved):
LGTM!

Issue #1921, opened 2026-06-10
Push-Model attestation fails with PostgreSQL database


jon4hz:
# Environment
 * OS / version: container image `quay.io/keylime/keylime_verifier:v7.14.2`, with PostgreSQL 18.4
 * Processor architecture: x86_64
 * TPM Manufacturer: N/A
 * Keylime version: 7.14.2

# Description

In push mode, the verifier fails with an HTTP 500 on every evidence submission when using PostgreSQL as database backend, because the `evidence_items.agent_id` column is too small any real agent ID.

The migration that introduces the push-attestation tables declares:

* `attestations.agent_id` as `sa.String(80)`:
https://github.com/keylime/keylime/blob/940861a2fd32b16f9d115ebd64c691917dc3b3d2/keylime/migrations/versions/870c218abd9a_add_push_attestation_support.py#L42
* `evidence_items.agent_id` as `sa.String(20)`:
https://github.com/keylime/keylime/blob/940861a2fd32b16f9d115ebd64c691917dc3b3d2/keylime/migrations/versions/870c218abd9a_add_push_attestation_support.py#L59

Agent IDs are usually at minimum a 36-character UUID, and 64 characters when the agent uses `uuid = hash_ek`, so `varchar(20)` can never hold a valid agent ID.

The bug doesn't affect SQLite because SQLite does not enforce VARCHAR length limits.

## Expected behavior vs. actual behavior

**Expected:**
After capabilities negotiation, the verifier persists the chosen evidence items and returns the evidence requests to the push-model agent.

**Actual:**
The INSERT into `evidence_items` fails with `psycopg2.errors.StringDataRightTruncation: value too long for type character varying(20)`, the request handler returns 500, and the attestation is left without challenges. 
The agent's subsequent PATCH then fails with `403 challenges_expired` ("expired at None"), so push-mode attestation never completes.

## Steps to reproduce problem
1. Run registrar + verifier 7.14.2 with `[verifier] mode = push` and a PostgreSQL database
2. Register a `keylime_push_model_agent` (any UUID ≥ 21 chars, e.g. the `hash_ek`) and add it to the verifier
3. Start the `keylime_push_model_agent.service`
4. Observe the 500 + traceback in the verifier log; the agent then receives  `403 challenges_expired` on the evidence PATCH.

**Workaround:**
Patch the database schema:
```sql
ALTER TABLE evidence_items ALTER COLUMN agent_id TYPE varchar(80);
```

# Relevant logs

```
[...]

PR #1922, opened 2026-06-16
fix: Suppress false-positive type checker errors in types.py

ansasaki:
## Type of Change
*(Select all that apply)*
- [x] Bug fix (non-breaking change)
- [ ] New feature (non-breaking change)
- [ ] Breaking change (fix/feature causing existing behavior to change)
- [ ] Documentation update (standalone)
- [ ] Code refactor (no functional changes)
- [ ] Test cases (added/modified)
- [ ] CI/CD changes
- [ ] Other (please specify: ______)

## Related Issues

N/A

### Concise Summary

Suppress false-positive type checker errors for compatibility import

The `CERTIFICATE_PRIVATE_KEY_TYPES` fallback import in `keylime/types.py` triggers mypy and pyright errors when checked against newer versions of the cryptography library. This change adds targeted suppression directives.

### Technical Details

The fallback import exists for compatibility with cryptography v37.0.x-v39.x, where `CertificateIssuerPrivateKeyTypes` did not yet exist. In v40.0+, the old name `CERTIFICATE_PRIVATE_KEY_TYPES` was removed. Type checkers analyze against the installed (newer) version and report:

- mypy: `attr-defined` (symbol doesn't exist) and `no-redef` (already defined on line 7)
- pyright: `reportAttributeAccessIssue` (unknown import symbol)

These are false positives because the import is guarded by `try/except ImportError`, so the missing attribute is handled gracefully at runtime.

The three suppression directives are ordered deliberately:
- `# type: ignore[attr-defined,no-redef]` must come first on the `from` line (mypy only reads `type: ignore` if it is the first directive comment)
- `# isort: skip` follows on the same line to prevent isort from reformatting the multi-line import, which would move the `type: ignore` to the item line where mypy cannot see it
- `# pyright: ignore[reportAttributeAccessIssue]` goes on the item line where pyright reports the issue

## Documentation Updates Required
*(Check all that apply)*
- [ ] Updated markdown docs (file path: ______)
- [ ] Updated code comments/docstrings
- [ ] Needs user guide updates (`docs/`)
- [ ] Needs ReadTheDocs updates
- [x] No docs needed (requires maintainer approval)

## Verification Process
1. Run `tox -e mypy,pyright,isort,black,pylint`
2. All five checks pass with no errors

## Checklist
- [x] Code follows project style guidelines
- [ ] Unit/integration tests added/updated
- [ ] Documentation updated (per above section)
[...]

PR #1923, opened 2026-06-29
DB: increase evidence_items.agent_id column length

ansasaki:
## Type of Change
*(Select all that apply)*
- [ ] CI/CD changes

## Related Issues
**Resolves:** #1921

### Concise Summary
Fix `evidence_items.agent_id` column too small for PostgreSQL, MySQL, and MariaDB

The push-attestation migration (`870c218abd9a`) defined `evidence_items.agent_id`
as `varchar(20)`, which is too small for any valid agent ID. Standard UUIDs are
36 characters and `hash_ek` agent IDs are 64 characters. PostgreSQL enforces
`VARCHAR` length limits (unlike SQLite), causing `StringDataRightTruncation` on
every evidence INSERT, breaking push-mode attestation entirely.

### Technical Details

- **Root cause:** `evidence_items.agent_id` was declared as `String(20)` in migration
  `870c218abd9a`, likely a copy-paste error from the adjacent `evidence_class` column.
  The column should be `String(80)` to match `attestations.agent_id` and
  `verifiermain.agent_id`.

- **Why only non-SQLite backends:** SQLite does not enforce `VARCHAR` length limits,
  so the bug was invisible in SQLite-based deployments.

- **MySQL and MariaDB constraints:** Both block inline `ALTER COLUMN` on
  FK-referenced columns (error 1832). Additionally, Alembic's batch table recreate
  fails because FK constraint names must be unique database-wide, so the temp table
  cannot be created with the same FK name as the original (error 121). MariaDB
  additionally does not honor `FOREIGN_KEY_CHECKS=0` for `ALTER TABLE` column
  modifications, unlike MySQL.

- **Fix:** A corrective Alembic migration (`a59cc9366774`) widens the column from
  `varchar(20)` to `varchar(80)`. For MySQL and MariaDB, the migration queries the
  actual FK constraint name from `information_schema`, drops it, widens the column,
  then restores the FK. For PostgreSQL and SQLite the standard `batch_alter_table`
  path is used. The original migration is left unchanged since it has already been
  published and applied by existing deployments.

- **Downgrade:** Intentionally a no-op, consistent with other column-widening
  migrations in this project (e.g. `160f932fde5b`, `8da20383f6e1`).

*(Check all that apply)*

## Verification Process
1. Deploy verifier + registrar with a **PostgreSQL**, **MySQL**, or **MariaDB**
   database backend.
2. Register an agent using a `hash_ek`-style UUID (64 characters) and add it to
   the verifier with push mode enabled.
3. Start the push-model agent and observe verifier logs — before this fix, every
   evidence submission returns HTTP 500 with a `StringDataRightTruncation` (PostgreSQL)
   or `OperationalError` (MySQL/MariaDB) on the `evidence_items` INSERT.
4. After applying this migration (`alembic upgrade head`), repeat steps 2–3 and
   confirm the verifier processes evidence submissions successfully and attestation
   completes without errors.

## Checklist

## Additional Context

> This PR was created with the assistance of Claude Sonnet 4.6 (Anthropic).

### DIFF
--- a/keylime/migrations/versions/a59cc9366774_fix_evidence_items_agent_id_length.py
+++ b/keylime/migrations/versions/a59cc9366774_fix_evidence_items_agent_id_length.py
@@ -1,0 +1,71 @@
+"""Fix evidence_items.agent_id column length for PostgreSQL, MySQL and MariaDB
+
+Revision ID: a59cc9366774
+Revises: 5a8b2c3d4e6f
+Create Date: 2026-06-29 00:00:00.000000
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "a59cc9366774"
+down_revision = "5a8b2c3d4e6f"
+branch_labels = None
+depends_on = None
+
+
+def upgrade(engine_name):
+    globals()[f"upgrade_{engine_name}"]()
+
+
+def downgrade(engine_name):
+    globals()[f"downgrade_{engine_name}"]()
+
+
+def upgrade_registrar():
+    pass
+
+
+def downgrade_registrar():
+    pass
+
+
+def upgrade_cloud_verifier():
+    bind = op.get_bind()
+    if bind.dialect.name == "mysql":
+        # MySQL and MariaDB both block ALTER COLUMN on FK-referenced columns
+        # (error 1832), and batch recreate fails too because FK names must be
+        # unique database-wide (error 121). The portable fix is to drop the FK,
+        # widen the column, then restore the FK. The FK name is looked up from
+        # information_schema rather than hardcoded.
+        result = bind.execute(
+            sa.text(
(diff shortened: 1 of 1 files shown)


ansasaki:
The lint test failure is caused by some false-positives addressed via: https://github.com/keylime/keylime/pull/1922

sergio-correia (dismissed):
LGTM, thanks

kunalworldwide (commented):
Thanks for the quick fix. Adding a follow-up Alembic migration is the right approach — editing the existing migration wouldn't help deployments that already ran `870c218abd9a`.

The `String(80)` length matches `attestations.agent_id`, which covers both UUID and `hash_ek` agent IDs. One small thing to double-check: the migration's `downgrade` should also restore the original `String(20)` length so the migration is reversible. The patch includes both upgrade and downgrade paths, which is good.

This should resolve the PostgreSQL push-mode failure described in #1921.

sergio-correia (approved):
LGTM

Issue #1924, opened 2026-06-30
Race condition in the verifier GET handler


kkaarreell:
# Environment
 * OS / version:  RHEL-9.9
 * Processor architecture: aarch64
 * TPM Manufacturer: swtpm
 * Keylime version: keylime-7.14.2

# Description 

  1. Test that failed

  /Tests with IMA sign template/functional/ek-cert-use-ek_check_script, phase: "Expected fail of adding keylime agent due verifying of script via 
  ek_ceck_script option, which doesn't have a zero exit code."

  2. What was expected

  The test configures ek_check_script to point to a script that always exits with code 1, then runs keylime_tenant -c update expecting it to fail. The
  assertion at test.sh:86 checks:
  rlAssertGrep "ERROR - External check script failed to validate EK" $rlRun_LOG

  3. What actually happened

  The tenant command failed (exit code 1 as expected), but with a completely different error:
  ValueError: Unable to convert response to JSON format
  The tenant crashed in do_cvdelete() → do_cvstatus() before ever reaching the EK check script validation.

  4. Root cause analysis — race condition in the verifier GET handler

  The -c update command first tries to delete the existing agent from the verifier. The sequence:

  ┌──────────────┬──────────────────────────────────────────────────────────────────┐
  │     Time     │                                                        Event                                                         │
  ├──────────────┼──────────────────────────────────────────────────────────────────┤
  │ 13:56:59.590 │ Tenant sends DELETE /v2.5/agents/UUID → verifier returns 202 (sets agent to TERMINATED state)                        │
  ├──────────────┼──────────────────────────────────────────────────────────────────┤
  │ 13:56:59.620 │ Tenant sends GET /v2.5/agents/UUID to poll for deletion                                                              │
  ├──────────────┼──────────────────────────────────────────────────────────────────┤
  │ 13:56:59.629 │ Verifier attestation loop detects TERMINATED → deletes agent from DB                                                 │
  ├──────────────┼──────────────────────────────────────────────────────────────────┤
  │ 13:56:59.707 │ GET handler calls session.refresh(agent) on the now-deleted record → InvalidRequestError: Could not refresh instance │
  ├──────────────┼──────────────────────────────────────────────────────────────────┤
  │ 13:56:59.717 │ Verifier returns 500 with empty body; tenant crashes parsing it                                                      │
  └──────────────┴──────────────────────────────────────────────────────────────────┘

[...]

PR #1926, opened 2026-07-02
fix: handle concurrent agent deletion in verifier GET handler

ansasaki:
# fix: handle concurrent agent deletion in verifier GET handler

## Type of Change
*(Select all that apply)*
- [ ] CI/CD changes

## Related Issues
**Resolves:** #1924

### Concise Summary

Wrap `session.refresh(agent)` in the verifier GET handler with a
`try/except InvalidRequestError` to handle the race condition where
the attestation loop concurrently deletes a TERMINATED agent between
the initial query and the refresh call.

### Technical Details

The verifier GET handler queries for an agent with `one_or_none()`,
finds it, then calls `session.refresh(agent)` to ensure it has the
latest state. However, the attestation loop runs as an interleaved
coroutine on the same Tornado IOLoop and can delete the agent record
(when it transitions to TERMINATED and the attestation cycle completes)
between these two operations. When this happens, SQLAlchemy raises
`InvalidRequestError("Could not refresh instance '...'")`, which
propagates as an unhandled exception, causing the verifier to return
a 500 response with an empty body. The tenant then crashes parsing
the empty response with `ValueError: Unable to convert response to
JSON format`.

Commit 476ea2f previously hardened the DELETE handler against a
related race, but the GET handler was left vulnerable.

This fix:

- Catches `InvalidRequestError` specifically (not the broader
  `SQLAlchemyError`) to avoid masking genuine database errors such as
  connectivity failures or deadlocks, which should still produce 500
- For the single-agent GET: returns 404 — the agent the caller was
  polling is gone, which is the correct semantic
- For the bulk GET: skips the concurrently-deleted agent with
  `continue` and returns the rest of the listing — a 404 for a bulk
  response because one agent disappeared mid-iteration would be wrong

The `InvalidRequestError` exception was verified against SQLAlchemy
2.0 source: `session.refresh()` raises it (and only it) when the
SELECT for the primary key returns no rows. `StaleDataError` is only
raised by the version-ID validator (optimistic locking, not used on
`VerfierMain`), and `NoResultFound` is only raised by `.one()`, not
by `refresh()`.

## Verification Process

1. **Environment:** Python unit tests, no hardware TPM required
2. **Unit tests** added to `test/test_cloud_verifier_tornado.py`:
   - `TestGetHandlerRaceCondition::test_returns_404_when_refresh_raises_invalid_request_error`
     — mocks `session.refresh()` to raise `InvalidRequestError` and
     asserts the handler returns HTTP 404
   - `TestGetHandlerRaceCondition::test_returns_200_when_no_race`
     — verifies the happy path is unaffected
   - `TestGetHandlerRaceCondition::test_returns_404_when_agent_not_found`
     — verifies the pre-existing 404 path is intact and `refresh` is
     not called when the agent was absent at query time
   - `TestBulkGetHandlerRaceCondition::test_bulk_get_skips_deleted_agent_includes_rest`
     — verifies bulk GET omits the deleted agent and returns the rest
3. **Result:** All 23 tests pass (`python -m pytest test/test_cloud_verifier_tornado.py`)

## Checklist

## Additional Context

The functional test `/functional/ek-cert-use-ek_check_script` that
originally exposed this race is not reliably reproducible (timing
dependent), so the fix is covered by unit tests that deterministically
inject the race via mocking.

> This pull request was created with the assistance of Claude Sonnet 4.6 (Anthropic).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved agent listing and lookup behavior when an agent is removed during a request.
  * Single-agent requests now return a not found response instead of failing unexpectedly.
  * Bulk agent results now skip agents that were deleted mid-request, avoiding partial refresh errors.
  * Reduced unnecessary data refreshes for agent status updates, improving request reliability.

### DIFF
--- a/keylime/cloud_verifier_tornado.py
+++ b/keylime/cloud_verifier_tornado.py
@@ -18,7 +18,7 @@ import tornado.netutil
 import tornado.process
 import tornado.web
 from sqlalchemy.engine import Engine
-from sqlalchemy.exc import SQLAlchemyError
+from sqlalchemy.exc import InvalidRequestError, SQLAlchemyError
 from sqlalchemy.orm import Session, joinedload
 from sqlalchemy.orm.exc import NoResultFound  # pyright: ignore
 
@@ -614,7 +614,17 @@ class AgentsHandler(BaseHandler):
                 if agent is not None:
                     # Refresh agent from database to ensure we have the latest consecutive_attestation_failures
                     # This is critical for PUSH mode status detection when failures occur
-                    session.refresh(agent)
+                    try:
+                        session.refresh(agent, attribute_names=["consecutive_attestation_failures"])
+                    except InvalidRequestError:
+                        # The attestation loop concurrently deleted the agent (TERMINATED → deleted)
+                        # between the initial query and the refresh. Return 404 to the caller.
+                        web_util.echo_json_response(self.req_handler, 404, "agent id not found")
+                        logger.info(
+                            "GET returning 404 response. agent %s was deleted during request processing.",
+                            agent_id,
+                        )
+                        return
                     response = cloud_verifier_common.process_get_status(agent)
                     web_util.echo_json_response(self.req_handler, 200, "Success", response)
                 else:
@@ -658,9 +668,18 @@ class AgentsHandler(BaseHandler):
 
                     json_response = {}
                     for agent in agent_list:
+                        aid = agent.agent_id
                         # Refresh agent from database to ensure fresh consecutive_attestation_failures
-                        session.refresh(agent)
-                        json_response[agent.agent_id] = cloud_verifier_common.process_get_status(agent)
+                        try:
+                            session.refresh(agent, attribute_names=["consecutive_attestation_failures"])
+                        except InvalidRequestError:
+                            # Agent was concurrently deleted during iteration; skip it.
+                            logger.debug(
+                                "Agent %s was deleted during bulk GET request processing, skipping.",
+                                aid,
+                            )
+                            continue
+                        json_response[aid] = cloud_verifier_common.process_get_status(agent)
 
                     web_util.echo_json_response(self.req_handler, 200, "Success", json_response)
                 else:
--- a/test/test_cloud_verifier_tornado.py
+++ b/test/test_cloud_verifier_tornado.py
@@ -3,14 +3,17 @@
 Tests cover:
 1. _register_pending_event / _cancel_pending_event helpers
 2. store_attestation_state graceful handling when agent is deleted
+3. AgentsHandler.get() race condition: session.refresh() raises InvalidRequestError
+   when the attestation loop deletes the agent between query and refresh
 """
 
 # pylint: disable=protected-access
 
 import unittest
+from contextlib import contextmanager
 from unittest.mock import MagicMock, patch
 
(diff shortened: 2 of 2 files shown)


keylime/cloud_verifier_tornado.py
```diff
@@ -659,7 +669,15 @@ def get(self) -> None:
...
+                        try:
+                            session.refresh(agent)
+                        except InvalidRequestError:
+                            # Agent was concurrently deleted during iteration; skip it.
+                            logger.debug(
+                                "Agent %s was deleted during bulk GET request processing, skipping.",
+                                agent.agent_id,
+                            )
+                            continue
                         json_response[agent.agent_id] = cloud_verifier_common.process_get_status(agent)
```
sergio-correia: After `session.refresh(agent)` raises `InvalidRequestError`, the instance's attributes have been expired. Accessing `agent.agent_id` here triggers an unnecessary lazy-load query that will also fail (though it would be caught by this same except block since `ObjectDeletedError` inherits from `InvalidRequestError`). Better to cache the ID before `refresh()`, which also matches the single-agent path above where the local `agent_id` variable is used instead of the ORM attribute.

```suggestion
                    json_response = {}
                    for agent in agent_list:
                        # Refresh agent from database to ensure fresh consecutive_attestation_failures
                        aid = agent.agent_id
                        try:
                            session.refresh(agent)
                        except InvalidRequestError:
                            # Agent was concurrently deleted during iteration; skip it.
                            logger.debug(
                                "Agent %s was deleted during bulk GET request processing, skipping.",
                                aid,
                            )
                            continue
                        json_response[aid] = cloud_verifier_common.process_get_status(agent)
```

keylime/cloud_verifier_tornado.py
```diff
@@ -614,7 +614,17 @@ def get(self) -> None:
                 if agent is not None:
                     # Refresh agent from database to ensure we have the latest consecutive_attestation_failures
                     # This is critical for PUSH mode status detection when failures occur
-                    session.refresh(agent)
+                    try:
+                        session.refresh(agent)
```
sergio-correia: `session.refresh(agent)` without `attribute_names` expires all attributes including the eagerly-loaded relationships from the `joinedload(...).load_only(...)` above. When `process_get_status()` then accesses `agent.mb_policy` and `agent.ima_policy`, those trigger lazy-load SELECTs that fetch all columns, including the full `ima_policy` text blob (which can be very large). In the bulk path this turns 1 query into 3N+1.

Since the comment says only `consecutive_attestation_failures` needs refreshing, maybe pass `attribute_names` to preserve the optimized relationship loading:

```python
session.refresh(agent, attribute_names=["consecutive_attestation_failures"])
```

This still raises `InvalidRequestError` when the row is gone, so the race-condition fix is preserved.

keylime/cloud_verifier_tornado.py
ansasaki: Thanks! Suggestions applied!

keylime/cloud_verifier_tornado.py
ansasaki: Thanks! Suggestions applied!

sergio-correia (approved):
LGTM, thanks!

Issue #1929, opened 2026-07-09
Verifier crashes on startup with empty max_workers config value


ansasaki:
## Description

The verifier crashes on startup with a `ValueError` when the `max_workers` configuration option has an empty string value. This can happen when the configuration file is not properly upgraded using the configuration templates during an update.

## Steps to Reproduce

1. Set `max_workers =` (empty value) in the verifier configuration
2. Start the verifier

## Observed Behavior

The verifier crashes with the following traceback:

```
File "/usr/lib/python3.12/site-packages/keylime/web/verifier_server.py", line 191, in _setup
    self._set_max_workers(from_config="max_workers")
  File "/usr/lib/python3.12/site-packages/keylime/web/base/server.py", line 577, in _set_max_workers
    self._set_option("max_workers", **kwargs)
  File "/usr/lib/python3.12/site-packages/keylime/web/base/server.py", line 491, in _set_option
    value = config.getint(self.config_component, config_name, fallback=fallback)
  ...
ValueError: invalid literal for int() with base 10: ''
```

## Expected Behavior

The verifier should start successfully, either by using a sane default value or by logging a clear warning about the invalid configuration value.

## Version

keylime 7.14.3

---
*This issue was written with the assistance of AI (Claude Code)*

PR #1930, opened 2026-07-09
fix: handle invalid config values for typed server options

ansasaki:
## Type of Change
*(Select all that apply)*
- [ ] CI/CD changes

## Related Issues
**Resolves:** #1929

### Concise Summary
Handle invalid typed config values gracefully

When a typed configuration option (int, float, bool) contains an empty or
invalid value, Python's `configparser` raises `ValueError` instead of using
the fallback. This causes the verifier to crash on startup when e.g.
`max_workers` has an empty value after a configuration file upgrade that
did not properly apply the configuration templates.

### Technical Details

The root cause is that `configparser.getint()` only uses the fallback when
the key is *missing* from the config file. When the key exists but has an
empty or non-numeric value, it attempts `int('')` which raises `ValueError`.

This fix:

- Wraps `config.getint()`, `config.getfloat()`, and `config.getboolean()`
  calls in `_set_option()` with try/except `ValueError` blocks, logging a
  warning and falling back to the caller-provided default
- Adds a default fallback of `0` (meaning "use all CPUs") for
  `_set_max_workers()`, following the same pattern already used by
  `_set_max_upload_size()`

*(Check all that apply)*

## Verification Process
1. Set `max_workers =` (empty value) in the verifier configuration
2. Start the verifier
3. Verify it starts successfully with a warning instead of crashing

## Checklist

## Additional Context
*This PR was written with the assistance of AI (Claude Code)*

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of configuration options using typed integer, float, or boolean parsing: invalid values now either raise an error or, if a fallback is provided, emit a warning and use that fallback.
  * Correctly preserves valid falsy values such as `0` and `False` instead of replacing them with fallbacks.
  * Updated max-worker fallback behavior to default to `0` when no fallback is specified, improving consistency when values are missing.

### DIFF
--- a/keylime/web/base/server.py
+++ b/keylime/web/base/server.py
@@ -449,9 +449,34 @@ class Server(ABC):
         """Sets config component (i.e., namespace) used to locate config values when setting server options."""
         self.__config_component = component
 
+    _TYPED_CONFIG_GETTERS: dict[type, Callable[..., Any]] = {
+        int: config.getint,
+        float: config.getfloat,
+        bool: config.getboolean,
+    }
+
+    def _get_typed_config_value(self, config_name: str, data_type: type, fallback: Any) -> Any:
+        """Read a typed config value, falling back gracefully on ValueError."""
+        getter = self._TYPED_CONFIG_GETTERS[data_type]
+        try:
+            return getter(self.config_component, config_name, fallback=fallback)  # type: ignore
+        except ValueError:
+            if fallback is None:
+                raise
+            raw_value = config.get(self.config_component, config_name, fallback="<unknown>")  # type: ignore[arg-type]
+            logger.warning(
+                "Cannot parse '%s' as %s for option '%s' (component: %s), using fallback: %s",
+                raw_value,
+                data_type.__name__,
+                config_name,
+                self.config_component,
+                fallback,
+            )
+            return fallback
+
     def _set_option(self, name: str, **kwargs: Any) -> None:
         """Sets server option by name either from given value or by obtaining it from user config. If the value is
-        falsy, uses the given fallback value instead or ``None`` if no fallback is provided. Examples::
+        ``None``, uses the given fallback value instead. Examples::
 
             # Set option using provided value
             self._set_option("bind_interface", value="0.0.0.0")
@@ -487,19 +512,13 @@ class Server(ABC):
             case {"from_config": (config_name, data_type)} if data_type is str:
                 value = config.get(self.config_component, config_name, fallback=fallback)  # type: ignore
 
-            case {"from_config": (config_name, data_type)} if data_type is int:
-                value = config.getint(self.config_component, config_name, fallback=fallback)  # type: ignore
-
-            case {"from_config": (config_name, data_type)} if data_type is float:
-                value = config.getfloat(self.config_component, config_name, fallback=fallback)  # type: ignore
-
-            case {"from_config": (config_name, data_type)} if data_type is bool:
-                value = config.getboolean(self.config_component, config_name, fallback=fallback)  # type: ignore
+            case {"from_config": (config_name, data_type)} if data_type in self._TYPED_CONFIG_GETTERS:
+                value = self._get_typed_config_value(config_name, data_type, fallback)
 
             case _:
                 raise TypeError(f"invalid arguments given when setting option '{name}' for {self.__class__.__name__}")
 
-        setattr(self, attr_name, value or fallback)
+        setattr(self, attr_name, value if value is not None else fallback)
 
     def _set_operating_mode(self, **kwargs: Any) -> None:
         """Sets operating mode of the server (push or pull)."""
@@ -573,6 +592,9 @@ class Server(ABC):
 
         if "from_config" in kwargs:
             kwargs.update({"from_config": (kwargs["from_config"], int)})
+
+        if "fallback" not in kwargs:
+            kwargs.update({"fallback": 0})
 
         self._set_option("max_workers", **kwargs)
 


sergio-correia (commented):
A couple of things to address before merging.

The core approach is right, but `value or fallback` on line 529 silently discards valid falsy values like `0` and `False`. This needs to be `value if value is not None else fallback`. See inline comments for details on that and the fail-open behavior for port options.

Not in this diff, but directly affected by the change: line 529 uses `value or fallback`, which treats `0`, `False`, and `0.0` as falsy. A config value of `max_upload_size = 0` with `fallback=104857600` would evaluate `0 or 104857600` and silently apply the 100 MiB default. The existing bool path is also affected since `False or True` always evaluates to `True`, making it impossible to set a boolean option to `False` when the fallback is `True`.

Please use `value if value is not None else fallback` instead of `value or fallback`.

Separate from this PR: `__init__` line 221 uses `setattr(self, f"__{opt}", options[opt])` which bypasses Python's name mangling. The properties read `_Server__http_port` but this creates `__http_port`, so the `**options` constructor parameter is silently non-functional. Worth fixing in a follow-up.

keylime/web/base/server.py
```diff
@@ -488,13 +488,40 @@ def _set_option(self, name: str, **kwargs: Any) -> None:
                 value = config.get(self.config_component, config_name, fallback=fallback)  # type: ignore
 
             case {"from_config": (config_name, data_type)} if data_type is int:
-                value = config.getint(self.config_component, config_name, fallback=fallback)  # type: ignore
+                try:
+                    value = config.getint(self.config_component, config_name, fallback=fallback)  # type: ignore
+                except ValueError:
```
sergio-correia: When `_set_https_port(from_config="tls_port")` is called without an explicit fallback, `fallback` defaults to `None`. If the config value is invalid (e.g. a typo like `tls_port = 889l`), this catch sets `value = None`, which means the HTTPS port gets silently disabled. A warning is logged, but it may be overlooked in production logs, and the registrar would start HTTP-only.

Before this change, the `ValueError` would crash the server immediately, which is the safer behavior for security-critical settings. Would it make sense to re-raise when `fallback is None`?

```suggestion
                try:
                    value = config.getint(self.config_component, config_name, fallback=fallback)  # type: ignore
                except ValueError:
                    if fallback is None:
                        raise
                    logger.warning(
                        "Invalid integer value for '%s' in [%s] config, using default: %s",
                        config_name,
                        self.config_component,
                        fallback,
                    )
                    value = fallback
```

keylime/web/base/server.py
```diff
@@ -488,13 +488,40 @@ def _set_option(self, name: str, **kwargs: Any) -> None:
...
+                    logger.warning(
```
sergio-correia: The warning message does not include the actual invalid value, which makes debugging harder for operators. Also, `config.getint()` can raise `ValueError` from an environment variable override (via `_get_env()`), but the message says "in [%s] config", which would send the operator looking at the config file when the problem is actually in an env var.

Maybe include the raw value and use more neutral phrasing:
```python
raw_value = config.get(self.config_component, config_name, fallback="<unknown>")
logger.warning(
    "Cannot parse '%s' as integer for option '%s' (component: %s), using fallback: %s",
    raw_value, config_name, self.config_component, fallback,
)
```

keylime/web/base/server.py
ansasaki: Yes, crashing is better than silently disabling security features. Thank you for the suggestion, code modified

keylime/web/base/server.py
ansasaki: Thank you, done!

keylime/web/base/server.py
```diff
@@ -488,18 +488,57 @@ def _set_option(self, name: str, **kwargs: Any) -> None:
...
+                        self.config_component,
+                        fallback,
+                    )
+                    value = fallback
 
             case _:
                 raise TypeError(f"invalid arguments given when setting option '{name}' for {self.__class__.__name__}")
 
-        setattr(self, attr_name, value or fallback)
+        setattr(self, attr_name, value if value is not None else fallback)
```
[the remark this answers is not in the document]
ansasaki: Good catch — the docstring is stale after changing `value or fallback` to `value if value is not None else fallback`. Will fix.

> This comment was written with the assistance of Claude Opus 4.6 (Anthropic).

keylime/web/base/server.py
```diff
@@ -488,18 +488,57 @@ def _set_option(self, name: str, **kwargs: Any) -> None:
...
+                        raise
+                    raw_value = config.get(self.config_component, config_name, fallback="<unknown>")
+                    logger.warning(
+                        "Cannot parse '%s' as boolean for option '%s' (component: %s), using fallback: %s",
+                        raw_value,
+                        config_name,
+                        self.config_component,
+                        fallback,
+                    )
+                    value = fallback
```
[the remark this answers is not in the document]
ansasaki: Adopted — extracted into `_get_typed_config_value()` with a `_TYPED_CONFIG_GETTERS` class dict that maps `data_type` to the corresponding `config.get*` callable. The type name for warning messages is derived from `data_type.__name__`, so only `data_type` needs to be passed (no separate string parameter). The three match cases also collapsed into a single `case ... if data_type in self._TYPED_CONFIG_GETTERS`.

> This comment was written with the assistance of Claude Opus 4.6 (Anthropic).

sergio-correia (approved):
LGTM, thanks!

PR #1933, opened 2026-07-10
mba: add an ordered PCR replay policy for boot logs

flaviens:
# `feat: add ordered PCR replay measured-boot policy`

## Type of Change
- [ ] CI/CD changes

## Related Issues
None

### Concise Summary
Add ordered PCR replay measured-boot policy

The policy pins the expected per-PCR values of the boot log. The verifier replays the log once in the trusted parser and binds the resulting per-PCR values to the AK-signed quote, then the policy compares each reference value against that quote-bound map. The default stays accept-all and existing policies are unchanged.

### Technical Details

The stock measured-boot path lets an operator supply a reference state, but it does not pin the boot log's replayed PCR values against the values the quote actually bound. This policy fills that gap for operators who want the boot log checked against a known-good per-PCR outcome.

- The reference bank must match the quoted bank and be one the verifier can authenticate (sha1, sha256, sha384, sha512). Every reference PCR must be bound to the log by the quote and belong to `MEASUREDBOOT_PCRS`.

- Replay happens in the trusted parser, so the policy inherits its handling of StartupLocality and HCRTM. A within-PCR reorder that changes a PCR's final value is rejected. Endpoint-preserving reorders and cross-PCR interleaving are out of scope and documented as such.

- The policy pins the reference against each quote, so it runs on every attestation and during one-shot verification regardless of `measured_boot_evaluate`.

- An empty, missing, or malformed reference state no longer passes as accept-all. The first two fail closed for content-requiring policies and a malformed one is reported as an invalid operator policy instead of raising.

- An unregistered configured policy name is an operator error. The verifier refuses to start on one and the quote check fails closed if such a name still reaches it, hence a misspelled name cannot accept an empty reference state and pass.

- Capability hooks (`accepts_empty_refstate`, `requires_evaluation_every_quote`, `requires_log_bound_pcrs`) let a custom policy opt into these behaviors and reach evaluation instead of being short-circuited.

Compatibility: legacy custom implementations keep the original quote-PCR set positionally. The quoted bank and log-bound set are passed only as keywords to implementations that name them or accept keywords, therefore a four-argument implementation is never handed changed arguments.

- [x] Needs user guide updates (`docs/`) - updated `docs/user_guide/use_measured_boot.rst`

## Verification Process
1. Test environment: Python 3.12 virtualenv with project dev dependencies, on the branch head.
2. Procedure:
   - `python -m unittest test.test_ordered_replay_policy`
   - `python -m unittest test.test_tpm_check_pcrs test.test_mba_parsing test.test_tpm_engine test.test_tpm_verify`
   - `black --check`, `isort --check-only`, `mypy`, and `pylint --rcfile=pylintrc` on the changed files.
   - Import `keylime.cmd.verifier` and confirm the three built-in policies register.
3. Expected vs actual: all suites pass (33 in the ordered-replay suite, neighboring suites green); black, isort, and mypy report no issues; pylint scores 10.00/10. Startup validation raises on a misspelled policy name and the built-in policies register as expected.

## Checklist

## Additional Context
Per-bank-width coverage and a fixture generated from a real `tpm2_eventlog` run are worthwhile follow-ups. They need dedicated tooling and captured logs, so they are left out of this change.

## Summary by CodeRabbit

* **New Features**
  * Added the built-in **ordered-replay** measured boot policy for validating PCR values and detecting replay reordering.
  * Improved measured boot verification across one-shot and ongoing quote evaluations.
  * Added validation for configured policy names, reference states, PCR banks, and quote context.

* **Documentation**
  * Expanded the Measured Boot guide with PCR replay behavior, empty-policy handling, and ordered replay requirements and limitations.

### DIFF
--- a/docs/user_guide/use_measured_boot.rst
+++ b/docs/user_guide/use_measured_boot.rst
@@ -62,15 +62,12 @@ an agent node, the following actions will be taken.
 
 .. note::
 
-   The PCR replay in step 3 is restricted to the intersection of ``MEASUREDBOOT_PCRS``
-   (the PCRs Keylime collects) and the set returned by the active policy's
-   ``get_relevant_pcrs()`` method.  PCRs outside that set are excluded from the
-   replay check.  This allows policies to declare that certain PCRs (e.g. PCR 11,
-   which ``systemd-pcrphase`` extends at runtime) are not part of their trust
-   model, preventing spurious verification failures.
-   Policies that return an empty set from ``get_relevant_pcrs()`` (such as the
-   built-in ``accept-all`` and ``reject-all``) apply no restriction, so all
-   ``MEASUREDBOOT_PCRS`` are replayed.
+   The PCR replay in step 3 uses the intersection of ``MEASUREDBOOT_PCRS`` and
+   the active policy's relevant PCR set. PCRs outside that set are excluded.
+   This allows a policy to omit PCR 11 when ``systemd-pcrphase`` extends it at
+   runtime. A policy that returns an empty set applies no restriction, so
+   Keylime uses all ``MEASUREDBOOT_PCRS``. The built-in ``accept-all`` and
+   ``reject-all`` policies use this behavior.
 
 How to use 
 ---------- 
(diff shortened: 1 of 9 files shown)


kunalworldwide (commented):
This is a substantial addition to Keylime's measured-boot support. The ordered PCR replay policy lets operators pin expected per-PCR values from a boot log, with the verifier replaying the log once and binding the resulting values to the AK-signed quote.

A few things that stand out:
- Separating `get_relevant_pcrs()` from `get_relevant_pcrs_from_refstate()` makes the policy logic cleaner: one answers "what can this verifier check?" and the other answers "what does this specific policy care about?".
- The `BANKABLE_PCRS` set is a good guard against attempting to bind PCRs that the quote path can't authenticate.
- The test coverage in `test_ordered_replay_policy.py` is extensive.

One question: the one-shot verification path treats the attestation count as 0 so that a `once` policy is valid. Does this interact safely with the existing `learn` policy, which expects multiple quotes over time? It looks like the dispatch in `mb_pcrs_to_check` keeps them separate, but confirming the intended behavior would help.

LGTM otherwise.

flaviens:
Hi @kunalworldwide,

Thanks for looking into it!

First, as you might have noted, this is quite vibe-coded so it should be carefully checked by sb who knows the repo well indeed.

Regarding your question, I think it is safe because afaiu measured boot does not have a learn mode (it only supports once or always). In one-shot verification, because there i a single quote, passing `count=0` just treats it as the first quote and ensures a once policy is checked. The continuous verifier behavior is unchanged because it still manages it quote count. The IMA/runtime learning flow is separate and I don't intend to affect it with this PR.

Thanks!
Flavien

PR #1935, opened 2026-07-16
Skip sm3 tests when not supported by OpenSSL

seryan-rh:
## Problem

The SM3 hash algorithm (GB/T 32905-2016) is not available in all OpenSSL builds. On systems where OpenSSL is built without SM3 support (e.g. OpenSSL 4.0+), multiple test failures occur:

- `_get_digest_algorithm_from_hex()` iterates all `Hash` enum members and calls `get_hex_size()` without guarding against unavailable algorithms. This crashes with `ValueError` even for non-SM3 inputs, because the iteration reaches the `sm3_256` entry before finding a match.
- `test_is_recognized` expects `Hash.is_recognized('sm3_256') == True`, but on systems without SM3, `is_recognized` correctly returns `False`.
- `test_get_hex_size` directly calls `Hash('sm3_256').get_hex_size()`, which raises `ValueError` when SM3 is unavailable.
- `test_digest_algorithm_priority` and `test_digest_algorithm_priority_exceptions` include `sm3_256` in algorithm loops that require SM3 IMA log entries to be correctly identified.

## Fix

- Catch `ValueError` in `_get_digest_algorithm_from_hex` so it skips algorithms unsupported by the current OpenSSL. Actual hashing still requires OpenSSL to support the algorithm, which is correct.
- Make SM3 test cases conditional on SM3 actually being present at runtime (detected via `hashlib.new("sm3", b"")`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved runtime policy creation when detecting digest algorithms.
  * Unsupported algorithms no longer interrupt detection, allowing compatible algorithms to be identified successfully.

### DIFF
--- a/keylime/policy/create_runtime_policy.py
+++ b/keylime/policy/create_runtime_policy.py
@@ -827,8 +827,12 @@ def process_signature_verification_keys(verification_keys: List[str], policy: Ru
 def _get_digest_algorithm_from_hex(hexstring: str) -> str:
     """Try to identify the algorithm used to generate the provided value by length"""
     for alg in list(algorithms.Hash):
-        if len(hexstring) == algorithms.Hash(alg).get_hex_size():
-            return str(alg)
+        try:
+            if len(hexstring) == algorithms.Hash(alg).get_hex_size():
+                return str(alg)
+        except ValueError:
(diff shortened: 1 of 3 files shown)


seryan-rh:
Closing in favour of a cleaner fix — replacing with a static size table in algorithms.py so size queries never require crypto backend support.

PR #1936, opened 2026-07-16
algorithms: do not require crypto backend support to query digest sizes

seryan-rh:
## Problem

`Hash.get_size()` and `Hash.get_hex_size()` instantiated the algorithm via `hashlib.new()` just to read constant, specification-defined digest sizes. On crypto backends where an algorithm is compiled out (e.g. sm3 in RHEL/ELN OpenSSL builds) this raised `ValueError`, which broke `Hash.is_recognized()` and crashed any caller iterating over the `Hash` enum, such as `create_runtime_policy`'s digest-length detection.

## Fix

Use a static size table instead. Actual hashing (`Hash.hash()`, `Hash.file_digest()`) still requires backend support, as it must.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved hash algorithm size detection so supported hash types report correct digest sizes even when not all algorithms are present in the underlying cryptographic backend.
  * Standardized how hexadecimal digest length is calculated for supported hash algorithms.
  * No changes to actual hash computation results—only metadata about sizes/lengths is corrected.

### DIFF
--- a/keylime/common/algorithms.py
+++ b/keylime/common/algorithms.py
@@ -23,6 +23,20 @@ def is_accepted(algorithm: str, accepted: List[Any]) -> bool:
     return False
 
 
+# Digest sizes in bits, defined by the respective specifications. Kept
+# static so that size queries and algorithm recognition keep working even
+# when an algorithm is not enabled in the crypto backend (e.g. sm3 with
+# OpenSSL builds that disable it); only actual hashing requires backend
+# support.
+_HASH_SIZES = {
+    "sha1": 160,
+    "sha256": 256,
+    "sha384": 384,
+    "sha512": 512,
+    "sm3_256": 256,
+}
+
+
 class Hash(str, enum.Enum):
     # Names compatible with tpm2-tools (man/common/alg.md)
     SHA1 = "sha1"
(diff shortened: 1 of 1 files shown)


keylime/common/algorithms.py
```diff
@@ -23,6 +23,20 @@ def is_accepted(algorithm: str, accepted: List[Any]) -> bool:
...
+# when an algorithm is not enabled in the crypto backend (e.g. sm3 with
+# OpenSSL builds that disable it); only actual hashing requires backend
+# support.
+_HASH_SIZES = {
+    "sha1": 160,
+    "sha256": 256,
+    "sha384": 384,
+    "sha512": 512,
+    "sm3_256": 256,
+}
```
sergio-correia, line 37: `get_size()` does `_HASH_SIZES[self.value]`, which raises `KeyError` on a missing entry. But `is_recognized()` catches only `ValueError` (from the enum constructor). If someone adds a new algorithm to the `Hash` enum but forgets the `_HASH_SIZES` entry, `is_recognized()` would propagate an unhandled `KeyError` instead of returning `False`.

The existing tests would catch this, but an import-time assertion is more robust:

```python
assert set(_HASH_SIZES.keys()) == {h.value for h in Hash}, \
    f"_HASH_SIZES keys must match Hash enum values"
```

sergio-correia (commented):
Looks good overall. One suggestion on future-proofing the dict/enum sync, plus the DCO check is failing (the commit needs a `Signed-off-by` line).

keylime/common/algorithms.py
seryan-rh: I believe I added these changes now. Thanks :)

sergio-correia (approved):
LGTM, thanks!

seryan-rh:
> LGTM, thanks!

Cool, I don't have permissions to merge. Could you do so whenever you're ready. Thanks :)

ansasaki (approved):
LGTM! Thank you for your contribution!

PR #1939, opened 2026-07-21
test: warning on empty max_workers config value

4N0body5:
## Type of Change

- [ ] CI/CD changes

## Change Description

Add unit test for empty max_workers fallback

## Verification Process
Run in the keylime repo root:
   ```
   PYTHONPATH=. python3 -m unittest test.test_server_base.TestWorkerCount.test_empty_max_workers_falls_back_with_warning -v
   ```

## Checklist

### DIFF
--- a/test/test_server_base.py
+++ b/test/test_server_base.py
@@ -1,5 +1,5 @@
 import unittest
-from unittest.mock import patch
+from unittest.mock import MagicMock, patch
 
 from keylime.web.base.server import Server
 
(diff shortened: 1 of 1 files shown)


ansasaki:
@4N0body5 Please add your sign-off in the commit to make the DCO test to pass. Please also fix the linting errors:
```
test/test_server_base.py:63:8: W0212: Access to a protected member _use_config of a client class (protected-access)
test/test_server_base.py:65:24: W0212: Access to a protected member _TYPED_CONFIG_GETTERS of a client class (protected-access)
test/test_server_base.py:68:20: W0212: Access to a protected member _set_max_workers of a client class (protected-access)
```
tip: run `tox -vv -e 'pylint,pylint310,mypy,pyright,black,isort'` to make sure the linters are clean before commit

kunalworldwide (commented):
The test coverage for the empty `max_workers` case is good — catching the warning and asserting fallback to 0 makes the behavior explicit.

One thing I'd flag: the test uses `patch.dict` to replace `Server._TYPED_CONFIG_GETTERS` with a mocked `int` getter that raises. That proves the warning path, but it also means the test isn't exercising the real config parser that users hit in production. Would it be worth adding a second test that actually sets `max_workers = ""` in a config string and verifies the warning fires through the normal `config.get()` path?

Also, should the production code itself log the warning, or is the intent that callers are expected to validate the value before passing it to `Server`? The current change only adds the test, so the warning seems to be a behavior already present.

Otherwise LGTM.

Issue #1891, opened 2026-04-21
Registrar fails to parse EK certificates wrapped in ASN.1 OCTET STRING


ansasaki:
## Summary

The registrar's EK certificate parsing fails when the certificate data is wrapped in an ASN.1 OCTET STRING (tag `0x04`). Both the primary parser (`cryptography.x509.load_der_x509_certificate`) and the pyasn1 fallback reject the data because they expect it to start with a SEQUENCE tag (`0x30`), which is the outer tag of a valid X.509 certificate.

This occurs when a newer Rust agent sends an EK certificate that was stored in TPM NV RAM wrapped in an OCTET STRING — a known convention per the TCG EK Credential Profile.

## Affected code paths

1. **`keylime/cert_utils.py:x509_der_cert()`** (lines 56–67): Used by the old registrar (`registrar_common.py`) and by `verify_ek()`.
2. **`keylime/models/base/types/certificate.py:_load_der_cert()`** (lines 87–121): Used by the new model-based registrar when casting the `ekcert` field.

Both follow the same pattern: try `cryptography` first, fall back to pyasn1 `decoder.decode(..., asn1Spec=rfc2459.Certificate())`. Neither handles an OCTET STRING wrapper.

## Error

From the old registrar path:

```
File "keylime/cert_utils.py", line 56, in x509_der_cert
    return x509.load_der_x509_certificate(data=der_cert_data, backend=default_backend())
ValueError: error parsing asn1 value: ParseError { kind: UnexpectedTag { actual: 4 } }

During handling of the above exception, another exception occurred:

File "keylime/cert_utils.py", line 60, in x509_der_cert
    pyasn1_cert = decoder.decode(der_cert_data, asn1Spec=rfc2459.Certificate())[0]
pyasn1.error.PyAsn1Error: <TagSet object, tags 0:0:4> not in asn1Spec: <Certificate schema object, ...>
```

The registrar returns a 400 response and the agent fails to register.

## Proposed fix

In both `x509_der_cert()` and `_load_der_cert()`, before attempting to parse as a certificate, check whether the DER data starts with the OCTET STRING tag (`0x04`). If so, decode the OCTET STRING to extract the inner bytes, then parse those as the certificate.

For example, in `cert_utils.py`:

```python
def x509_der_cert(der_cert_data: bytes) -> Certificate:
    # Unwrap OCTET STRING if present (some TPMs store EK certs this way in NV RAM)
    if der_cert_data and der_cert_data[0] == 0x04:
[...]

PR #1946, opened 2026-08-17
fix: Unwrap OCTET STRING wrapped EK certificates

msafarik:
# PR Title 
fix: Unwrap OCTET STRING wrapped EK certificates

## Type of Change
*(Select all that apply)*
- [ ] CI/CD changes

## Related Issues
**Resolves:** #1891 

### Concise Summary
Unwrap OCTET STRING wrapped EK certificates

### Technical Details
Some hardware TPMs store the EK certificate in NV RAM wrapped in an ASN.1 OCTET STRING (tag 0x04) instead of raw DER (tag 0x30), causing both `cert_utils.x509_der_cert()` and `certificate.py`'s `_load_der_cert()` to reject it, which makes agent registration fail with a 400 response.

- Added `unwrap_x509_octet_string()` in `cert_utils.py` that detects the  OCTET STRING wrapper, unwraps it, and verifies the inner content starts with a SEQUENCE tag (0x30) before trusting it.
- If the inner content doesn't look like a certificate, the original bytes are returned unchanged, so unrelated/malformed input isn't misinterpreted.
- Called from both affected code paths.
- Alternative considered: unwrapping without the inner 0x30 check — rejected since it could misidentify arbitrary OCTET-STRING-encoded non-cert data.

*(Check all that apply)*

## Verification Process
1. Added `test_x509_der_cert_octet_string_wrapped` unit test reproducing the exact error from the issue against a real EK certificate wrapped in OCTET STRING; confirmed it fails before the fix and passes after.
2. Verified against a real registrar on Testing Farm (CentOS Stream 10): registered a fake agent with a real swtpm-issued EK cert wrapped in OCTET STRING via the registrar's HTTP API — fails pre-fix, succeeds post-fix, matching a non-wrapped control case.
3. Ran full existing test suite (61 tests) — no regressions.

## Checklist

## Additional Context

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of X.509 certificates wrapped in ASN.1 OCTET STRING data.
  * Certificates are now unwrapped and parsed correctly across supported certificate-loading paths.
  * Malformed or unexpected certificate data is preserved safely, with warnings logged instead of causing unintended processing.

### DIFF
--- a/keylime/cert_utils.py
+++ b/keylime/cert_utils.py
@@ -13,6 +13,7 @@ from cryptography.hazmat.primitives.asymmetric.rsa import RSAPublicKey
 from cryptography.x509 import Certificate
 from cryptography.x509.oid import ExtensionOID
 from pyasn1.codec.der import decoder, encoder
+from pyasn1.type import univ
 from pyasn1_modules import pem, rfc2459, rfc4108
 
 from keylime import keylime_logging, tpm_ek_ca
@@ -59,12 +60,40 @@ def x509_der_cert(der_cert_data: bytes) -> Certificate:
     :type der_cert_data: bytes
     :returns: cryptography.x509.Certificate
     """
+    der_cert_data = unwrap_x509_octet_string(der_cert_data)
+
     try:
         return x509.load_der_x509_certificate(data=der_cert_data, backend=default_backend())
     except Exception as err:
         logger.warning("Failed to parse DER data with python-cryptography: %s", err)
         pyasn1_cert = decoder.decode(der_cert_data, asn1Spec=rfc2459.Certificate())[0]
         return x509.load_der_x509_certificate(data=encoder.encode(pyasn1_cert), backend=default_backend())
+
+
+def unwrap_x509_octet_string(der_cert_data: bytes) -> bytes:
+    """If der_cert_data is an ASN.1 OCTET STRING wrapping a DER certificate,
+    unwrap it and return the inner DER bytes. Otherwise, return der_cert_data unchanged.
+    """
+    if not der_cert_data or der_cert_data[0] != 0x04:
+        return der_cert_data
+
+    try:
+        octet_string, _ = decoder.decode(der_cert_data, asn1Spec=univ.OctetString())
+        unwrapped = bytes(octet_string)
+    except Exception as err:
+        logger.warning(
+            "Data starts with OCTET STRING tag (0x04) but unwrapping failed; passing through original data: %s", err
+        )
+        return der_cert_data
+
+    if unwrapped and unwrapped[0] == 0x30:
+        return unwrapped
+
+    logger.warning(
+        "Data starts with OCTET STRING tag (0x04) and was unwrapped, but the inner content "
+        "does not start with a SEQUENCE tag (0x30); passing through original data."
+    )
+    return der_cert_data
 
 
 def x509_pem_cert(pem_cert_data: str) -> Certificate:
--- a/keylime/models/base/types/certificate.py
+++ b/keylime/models/base/types/certificate.py
@@ -12,6 +12,7 @@ from pyasn1_modules import pem as pyasn1_pem
 from pyasn1_modules import rfc2459 as pyasn1_rfc2459
 from sqlalchemy.types import Text
 
+from keylime import cert_utils
 from keylime.certificate_wrapper import CertificateWrapper, wrap_certificate
 from keylime.models.base.type import ModelType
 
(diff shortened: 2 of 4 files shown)


msafarik:
Before fix: https://artifacts.osci.redhat.com/testing-farm/60a06ad8-7d6c-43b1-b127-2def2186399e/
After fix: https://artifacts.osci.redhat.com/testing-farm/f8129300-2b80-411d-968b-9b1aa603e222/

test/test_cert_utils.py
```diff
@@ -311,3 +313,16 @@ def test_is_x509_cert(self):
...
+        # ASN.1 OCTET STRING (tag 0x04) instead of raw DER (tag 0x30 SEQUENCE).
+        der_cert_data = base64.b64decode(st_sha256_with_rsa_der)
+        wrapped_der_cert_data = encoder.encode(univ.OctetString(der_cert_data))
+
+        try:
+            cert_utils.x509_der_cert(wrapped_der_cert_data)
+        except Exception as e:
+            raise self.failureException(
+                f"Failed to load OCTET STRING wrapped certificate: {type(e).__name__}: {str(e)[:200]}"
+            ) from None
```
sergio-correia: This verifies that no exception is raised, but it doesn't check that the returned certificate is the correct one. A subtle corruption in the unwrapping logic could return a different-but-parseable certificate and this test would still pass.

Would it be better to compare the result against the original unwrapped cert?

```python
expected = cert_utils.x509_der_cert(der_cert_data)
result = cert_utils.x509_der_cert(wrapped_der_cert_data)
self.assertEqual(expected.subject, result.subject)
self.assertEqual(expected.serial_number, result.serial_number)
```

This also avoids the try/except pattern and the `from None` which suppresses the original traceback on failures.

keylime/models/base/types/certificate.py
```diff
@@ -110,6 +111,7 @@ def _load_der_cert(self, der_cert_data: bytes) -> CertificateWrapper:
 
         :returns: A ``CertificateWrapper`` object
         """
+        der_cert_data = cert_utils.unwrap_x509_octet_string(der_cert_data)
```
sergio-correia, line 114: This is the primary production path for EK certs during agent registration (via `RegistrarAgent.update()` -> `cast_changes()` -> `Certificate.cast()`), but only the `cert_utils.x509_der_cert()` path has a test.

Would it make sense to add a test through `Certificate().cast()` with OCTET STRING wrapped bytes too? That way a future refactor that accidentally removes this line won't silently regress.

keylime/cert_utils.py
```diff
@@ -67,6 +71,21 @@ def x509_der_cert(der_cert_data: bytes) -> Certificate:
...
+    unwrap it and return the inner DER bytes. Otherwise, return der_cert_data unchanged.
+    """
+    try:
+        if der_cert_data and der_cert_data[0] == 0x04:
+            octet_string, _ = decoder.decode(der_cert_data, asn1Spec=univ.OctetString())
+            unwrapped = bytes(octet_string)
+            if unwrapped and unwrapped[0] == 0x30:
+                return unwrapped
+    except (PyAsn1Error, IndexError):
+        pass
```
sergio-correia: When data starts with `0x04`, that's a strong signal it is an OCTET STRING wrapped cert. If decoding then fails (e.g., truncated NV RAM read), the silent `pass` here means operators will only see a generic "Failed to parse DER data" from the downstream parser, with no indication that unwrapping was attempted and failed.

Maybe log at debug or warning level here so there's a breadcrumb for debugging TPM NV RAM issues?

```suggestion
    except (PyAsn1Error, IndexError) as e:
        logger.warning(
            "Data starts with OCTET STRING tag (0x04) but unwrapping "
            "failed; passing through original data: %s", e
        )
```

msafarik:
https://artifacts.osci.redhat.com/testing-farm/c064f431-17c7-470d-bd72-f85597f34128/

sergio-correia (changes requested):
Thanks, this looks good to me, but would you please squash the commits? Thanks in advance.

sergio-correia (approved):
Thanks, LGTM

ansasaki (approved):
LGTM! Let's wait for the tests to pass and then merge

ansasaki:
Test failure in Fedora 45 are unrelated to the changes: they are caused by a regression in tpm2-tss v4.2.0 (https://github.com/tpm2-software/tpm2-tss/issues/3145)

